PR #34, opened 2026-05-15
Add OCI image support: pull, unpack, run, prune, status, policy

Max042004:
This PR lands the full elfuse OCI image support. It supersedes the
original Phase 1 scope of this PR (CLI scaffold + pull/inspect) and
now covers Phases 1-4 plus the post-Phase-3 improvements plan: image
layout alignment, GC/prune, layer + stack snapshot caches, store
status, parallel pull, registry policy.json, and a heavy-mode compat
matrix.

## Scope

- **Pull / inspect** — content-addressable blob store, HTTPS + bearer
  token, OCI index walk to the linux/arm64 leaf manifest, partial-
  store-aware inspect renderer.
- **Unpack** — tar reader (ustar + PAX `x`/`g` records), gzip + decode-
  only vendored zstd, whiteout-aware layer apply (typeflag '1'/'2'/'5'
  + `.wh.*` markers), per-image sysroot on a case-sensitive APFS
  sparsebundle.
- **Run** — `elfuse oci run` clones the unpacked tree via clonefile(2),
  honors Entrypoint / Cmd / Env / WorkingDir / User, and reuses the
  existing elfuse launch path so a dynamically-linked guest binary
  runs through the same shim + syscall surface as the non-OCI mode.
- **Lifecycle** — `oci prune` with `--older-than` / `--keep-bytes`;
  layer + stack prune sweep; `oci status` (text + `--json`);
  `oci rebuild-cache` for pre-snapshot stores.
- **Performance** — parallel blob fetch with HTTP Range resume;
  per-layer raw snapshot cache; ChainID stack snapshot cache; APFS
  COW clone-rootfs reuse between runs.
- **Policy** — podman / skopeo-style `policy.json` + `registries.d`
  overlay (per-registry insecure / ca_bundle / auth_file). CLI flags
  override; loopback-only `--insecure`.
- **Test coverage** — 25 OCI unit suites (`test-oci-*`), compat-shell
  smoke (`tests/test-oci-compat.sh`), and an opt-in heavy mode
  (`OCI_COMPAT_TEST=1`) that drives three layered fixtures
  (alpine-shaped, busybox-shaped hardlink dispatch, two-layer
  whiteout) end-to-end through a freshly-provisioned scratch
  sparsebundle.

---
## Summary by cubic
Adds full OCI image lifecycle to elfuse with a new `elfuse oci` subcommand and a BusyBox‑style `elfuse-container` alias. Hardened unpack/path containment, verified blobs and sizes, shared VM launch path, and better `/dev`/`/proc` emulation.

- New Features
  - CLI: `oci pull|inspect|unpack|clone|run|prune|rebuild-cache|status`; `status --json`. Supports `elfuse-container` (e.g., `elfuse-container run alpine:3`).
  - Pull: delegates to `skopeo copy` (auth/TLS/mirrors via `containers-policy.json`/`registries.conf`), pins manifests in `index.json`; multi‑arch defaults to linux/arm64 with optional `--arch amd64`.
  - Unpack/caches: libarchive tar (ustar/GNU/PAX; gzip/zstd), whiteouts, absolute‑symlink rewrite, parent walks with `openat(O_NOFOLLOW)`, manifest/config re‑hash and layer size checks; APFS case‑sensitive sysroot; raw layer + ChainID stack caches; `rebuild-cache`.
  - Runtime/launch: shared VM launcher (`core/launch.*`); PATH resolver; shebang expansion; image `User` lookup (name/group or numeric); inject `/etc/{resolv.conf,hosts,hostname}`; improved `/dev` emulation (`/dev/full` ENOSPC on write, `/dev/console`), and basic `/proc`.

- Migration
  - Pins live in OCI `index.json` (store auto‑migrates from `refs/`).
  - Layer cache schema v2; first open drops legacy v1 entries (blobs/images untouched); symlink rewrite bumps unpack sidecar version and transparently rebuilds caches.
  - Dependencies: use system `libarchive` and `cJSON` via `pkg-config` and require `skopeo` at runtime (`brew install libarchive cjson skopeo` or `apt-get install libarchive-dev libcjson-dev skopeo`).

<sup>Written for commit 73d921677ee628c13423879f998daed1c20e34e6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/34?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jserv (changes requested):
Rebase onto the latest `main` branch and squash/rework the commits into fewer, cleaner ones.

jserv (changes requested):
Refine per review messages from cubic.

docs/usage.md
```diff
@@ -99,6 +99,179 @@ and memory access, and per-thread inspection. Implementation details, including
 the snapshot protocol used to keep Hypervisor.framework register access on the
 owning thread, are documented in [internals.md](internals.md).
 
+## Running OCI Images (`elfuse oci run`)
+
+Phase 3 adds a direct-execution path for pulled OCI images:
```
jserv: Never mention milestone labels such as "Phase 3."

docs/usage.md
```diff
@@ -99,6 +99,179 @@ and memory access, and per-thread inspection. Implementation details, including
...
+
+```sh
+elfuse oci run [OPTIONS] IMAGE [ARG...]
```
jserv, line 200: Rather than using subcommands like `elfuse oci run`, I am considering a Docker-style wrapper command such as `elfuse-container run`.

jserv:
Once a Docker-style wrapper is introduced (e.g., `elfuse-container`), we can adopt an approach similar to how BusyBox handles its applets through the "multicall" technique. This would allow a single executable to expose multiple command entry points while sharing the same underlying implementation, reducing duplication and simplifying maintenance. It would also provide a more familiar command-line experience for users, aligning well with established container tooling conventions.

The canonical reference is the BusyBox documentation itself:
> BusyBox is a multi-call binary. A multi-call binary is an executable program that performs the same job as more than one utility program. There is just a single BusyBox binary, but that single binary acts like a large number of utilities. The built-in utility programs (applets) share common code.

The BusyBox manual also explains the key mechanism:
```shell
ln -s /bin/busybox ls
./ls
```

When invoked through a symbolic link, BusyBox examines `argv[0]` (the name used to invoke the executable) and dispatches to the corresponding applet implementation. [1]

For a deeper explanation of how multicall binaries work internally, the IBM technical note is an excellent reference. It describes how BusyBox uses `argv[0]` to determine which command implementation to execute and why this approach reduces code duplication and binary size. [2]

[1]: https://busybox.net/BusyBox.html
[2]: https://www.redbooks.ibm.com/abstracts/tips0092.html

---

Follow-up after further review of this PR:

The pieces this needs are already in place. `oci run` and the plain launch path were unified into `src/core/launch.{c,h}`, so a multicall front end does not duplicate the VM bring-up -- each applet parses its own argv and calls the shared launcher. The dispatch keys on `argv[0]`:

- Take the basename of `argv[0]` only, and reject any invocation name containing a path separator, so a symlink shipped inside an image (or a crafted name) cannot steer applet selection or smuggle a directory component into the lookup.
- Keep the applet table closed: an unrecognized `argv[0]` falls back to the default `elfuse` behavior rather than erroring, matching BusyBox, which treats an unknown link name as "not an applet".
- The `elfuse-container` name maps to the `oci` subcommand group, so `elfuse-container run alpine:3` and `elfuse oci run alpine:3` resolve to the same entry point.

This keeps the shared-code benefit the multicall model provides without a second code path that can drift from `oci run`.

jserv (changes requested):

jserv (changes requested):

tests/lib/oci-fixture-builder.c
```diff
@@ -0,0 +1,368 @@
+/* OCI store fixture builder for Phase 3 compat tests
```
jserv: Clarify this without mentioning Phase 3.

src/oci/layer-apply.c
```diff
@@ -0,0 +1,672 @@
...
+        seg = strtok_r(NULL, "/", &save);
+    }
+    return 0;
+}
+
+/* mkdir -p for the directory containing 'host_path'. Mode 0755 for
+ * implicit parents; the entry's own mode is applied later when the
+ * tar provides it.
+ */
+static int mkdir_parents(const char *host_path, const char **err)
```
jserv: Containment here is purely lexical. oci_path_join_safe (line 59) rejects absolute names and '..', and oci_symlink_target_check (line 95) rejects escaping symlink targets, but the write primitives that follow resolve host_path with plain path-based syscalls that follow symlinks in every parent component: mkdir_parents here is EEXIST-tolerant, plus the apply_regular open (line 216), apply_symlink (line 383), and apply_hardlink link (line 418, with linkname joined at 403). Only apply_dir's open (line 248) passes O_NOFOLLOW. A symlink written by an earlier tar entry then redirects a later entry's parent outside the sysroot, which is arbitrary host write/read/delete. Ordered layer: a/b -> symlink '..', then a/b/c -> symlink '../etc', then a/b/c/x -> regular file lands outside the root; the hardlink variant can pull a host file back into the guest tree for readback. Resolve every component relative to a root dirfd with openat(O_NOFOLLOW) per segment (or re-check realpath containment before each create), refusing any component that is a symlink.

jserv (commented):
Two issues carry the weight; the rest are hardening.

Load-bearing:
- Layer extraction containment is only lexical -- a symlink planted by an earlier tar entry redirects a later entry outside the sysroot (inline on layer-apply.c).
- Manifest / index / config blobs are parsed without re-verifying their digest, while layers are re-hashed -- asymmetric trust (inline on unpack.c read_blob).

Additional items worth a pass, not line-anchored here to avoid guessing shifted numbers:
- Several size_t additions of the form alen+blen+const / klen+vlen+const in the path, env, runspec, and user-lookup helpers add attacker-influenced lengths (image config, passwd/group fields) with no overflow guard. Guard each against SIZE_MAX before the malloc.
- A few fixed PATH_MAX stack buffers are filled by snprintf from image config and the truncation return is accepted as a real length; treat >= sizeof(buf) as an error.
- read_file_bounded in user-lookup accepts EOF before the fstat size and parses the short buffer; require off == len for regular files.

Checked and cleared: the pull path passes the ref as discrete argv via posix_spawnp (no shell), ref.c restricts refs to a safe charset, blob-store put hashes before publish, manifest/index counts use overflow-safe calloc, and prune_candidate_list_append handles its realloc correctly. None of these need changes.

src/oci/unpack.c
```diff
@@ -0,0 +1,1118 @@
...
+    static const char hex[] = "0123456789abcdef";
+    for (size_t i = 0; i < need; i++) {
+        out[i * 2] = hex[buf[i] >> 4];
+        out[i * 2 + 1] = hex[buf[i] & 0xf];
+    }
+    out[n_hex] = '\0';
+    return 0;
+}
+
+static int read_blob(oci_blob_store_t *bs,
```
jserv, line 72: read_blob opens blobs/<algo>/<hex> and returns the bytes without checking that sha256(content) equals hex. Only layer blobs are re-hashed (reverify_layer_digest, line 115). The manifest (line 580), sub-manifest (line 619), and config (line 730) are parsed straight from read_blob, so the documents that decide which layers, config, entrypoint, env, and user to trust are never verified against their own digest. That is asymmetric with the layer re-verify and defeats the same local-tamper guard: editing blobs/sha256/<manifest_hex> in the store redirects config and layers to attacker-chosen already-present digests. Run the streaming digester over the manifest/index/config bytes and compare to the descriptor hex before parsing, the way reverify_layer_digest does for layers.

src/oci/unpack.c
```diff
@@ -0,0 +1,1118 @@
...
+    *out_buf = buf;
+    *out_len = (size_t) got;
+    return 0;
+}
+
+/* Open the on-disk blob path and confirm its sha256 hash matches the
+ * descriptor's expected digest. Phase 1 already verified at write time,
+ * but unpack re-verifies in case a host-side tool modified the blob.
+ */
+static int reverify_layer_digest(oci_blob_store_t *bs,
```
jserv, line 134: reverify_layer_digest checks only the sha256, not desc->size. OCI descriptors carry both. Counting bytes while hashing and requiring the total to equal desc->size costs nothing and catches a truncation or size mismatch that a bare hash comparison on a partial file could otherwise miss. Accumulate a byte count in the digest loop and compare to desc->size before returning success.

src/oci/unpack.c
```diff
@@ -0,0 +1,1118 @@
...
+        if (lstat(stage_child, &dst) == 0) {
+            if (oci_rm_recursive(stage_child) < 0) {
+                rc = set_err(err, "assemble: unlink dst failed", errno);
+                break;
+            }
+        } else if (errno != ENOENT) {
+            rc = set_err(err, "assemble: dst lstat failed", errno);
+            break;
+        }
+        if (copyfile(raw_child, stage_child, NULL,
```
jserv, line 491: This copyfile uses COPYFILE_CLONE | COPYFILE_ALL with no COPYFILE_NOFOLLOW, unlike the stack restore (line 838) and snapshot (line 1018) which both pass COPYFILE_NOFOLLOW. A symlink in a raw layer or cache is followed during assembly, so host target content can be copied in or the final image semantics changed. Add COPYFILE_NOFOLLOW here, or handle S_ISLNK explicitly with readlink plus symlinkat.

src/oci/unpack.c
```diff
@@ -0,0 +1,1118 @@
...
+                    "elfuse oci unpack: stale image sysroot sha256-%s (%s); "
+                    "rebuilding\n",
+                    image_hex, merr ? merr : strerror(errno));
+        errno = 0;
+        force = true; /* fall into the rm + full re-stage path below */
+    }
+    if (force) {
+        /* Remove any prior commit so the staging rename does not race. */
+        char rm[UN_PATH_MAX];
+        snprintf(rm, sizeof(rm), "rm -rf '%s'", final_dir);
```
jserv: This builds a shell command from a path single-quoted into rm -rf and runs it via system() (same at line 1108). A path containing a single quote breaks out of the quoting, and the snprintf truncation return is ignored, so a path near UN_PATH_MAX can drop the closing quote and rm a shorter prefix. Not reachable from image content today (these paths derive from hex-validated ids), but it is reachable from an operator-supplied store or volume root and is needlessly fragile. Replace both with the oci_rm_recursive() already used elsewhere in this file -- no shell.

src/oci/status.c
```diff
@@ -0,0 +1,807 @@
...
+        if (out_status_code)
+            *out_status_code = OCI_STATUS_PIN_CORRUPT_CONFIG;
+        return NULL;
+    }
+    /* Recurse on the picked sub-manifest. Status code on a sub-manifest miss
+     * is MISSING_MANIFEST rather than INDEX_NO_ARM64: the index itself was
+     * fine, the sub-manifest blob just is not on disk.
+     */
+    oci_status_pin_code_t sub_code = OCI_STATUS_PIN_OK;
+    char *cfg = resolve_config_digest(store, sub_digest, &sub_code);
```
jserv: resolve_config_digest recurses on the picked sub-manifest digest with no depth cap and no visited set (same pattern in dedup-metrics.c:258). A corrupt or crafted local store whose index resolves to an index that eventually points back at itself drives unbounded recursion and stack exhaustion during oci status and the dedup metrics walk. The dedup-metrics.c:255 comment claims arbitrary nesting is handled safely, but nothing enforces it. Pass a visited-digest set, or cap nesting to a small depth and return EINVAL past it.

src/oci/tar.c
```diff
@@ -0,0 +1,220 @@
...
+
+    memset(out, 0, sizeof(*out));
+    out->path = archive_entry_pathname(ae);
+    if (!out->path || !out->path[0]) {
+        if (err)
+            *err = "tar: entry with empty name";
+        errno = EPROTO;
+        return -1;
+    }
+    out->size = (uint64_t) archive_entry_size(ae);
```
jserv: out->size = (uint64_t) archive_entry_size(ae) casts a signed value. archive_entry_size returns -1 when the size is unknown, which becomes UINT64_MAX and makes the regular-file copy loop run to EOF before reporting truncation. Read into a signed temporary and reject a negative size for entry types that must carry a known payload.

src/oci/layer-apply.c
Max042004: Fixed in the updated commit. Layer apply now anchors on a dirfd of the unpack root and resolves every parent component with openat(O_RDONLY | O_DIRECTORY | O_NOFOLLOW) (open_parent_at), creating intermediates via mkdirat. All write primitives are *at-relative: openat(O_CREAT | O_EXCL | O_NOFOLLOW) for regular files, symlinkat, linkat with the hardlink target resolved through the same no-follow walk, and a new oci_rm_recursive_at for every removal (whiteouts and opaque markers included), so a symlink planted by an earlier entry can no longer redirect a later entry's parent chain. Two regression tests reproduce the exact scenarios from this comment: the chained a/b -> '..', a/b/c -> '../etc' redirect and the hardlink variant pulling a host file back into the tree — both now fail the apply with nothing landing outside the root.

src/oci/unpack.c
Max042004: Fixed: read_blob now runs the streaming digester over the returned bytes and rejects a mismatch against the descriptor hex before anything parses them, so manifest / index / config get the same local-tamper guard as layers.

src/oci/unpack.c
Max042004: Fixed: the digest loop accumulates a byte count and reverify_layer_digest now fails unless it equals desc->size.

src/oci/unpack.c
Max042004: Fixed: added COPYFILE_NOFOLLOW here so assembly copies a symlink as a symlink, matching the stack restore / snapshot calls.

src/oci/unpack.c
Max042004: Fixed: both call sites (the force re-stage removal and the fail_stage_dir cleanup) now use oci_rm_recursive() — no shell involved.

src/oci/status.c
Max042004: Fixed: resolve_config_digest in status.c and dedup-metrics.c now carries a depth parameter capped at 4 index levels; past the cap it fails with EINVAL (surfaced as CORRUPT_MANIFEST by the status pin walk). The dedup-metrics comment claiming arbitrary nesting was handled safely is corrected as well.

src/oci/tar.c
Max042004: Fixed: the size is read into a signed int64_t first and a negative (unknown) size is rejected with EPROTO before the uint64_t cast.

tests/lib/oci-fixture-builder.c
Max042004: Done — the header now describes the tool on its own terms, no milestone reference.

docs/usage.md
Max042004: Done — milestone labels are gone from the docs; the remaining scope note now reads 'not implemented yet' instead of 'later phases'.

docs/usage.md
Max042004: Implemented as the BusyBox-style multicall: the build ships a build/elfuse-container symlink, dispatch keys on the basename of argv[0] against a closed applet table (an unrecognized name falls back to the default elfuse behavior), and elfuse-container maps to the oci subcommand group — elfuse-container run alpine:3 and elfuse oci run alpine:3 resolve to the same entry point, with usage text branded by the invoked alias. Documented in docs/usage.md.

henrybear327:
Potential superseding work https://github.com/sysprog21/elfuse/pull/191

jserv:
Close in favor of #191

PR #40, opened 2026-05-19
Add HVC 6 embedder extension hook

doanbaotrung:
Allows embedders to register a custom dispatch handler for HVC immediate 6, keeping it separate from the Linux syscall path (HVC 5).

The hook is registered via two fields on guest_t:
  hvc6_handler   - callback fn(call_nr, args[8], userdata) -> result
  hvc6_userdata  - opaque pointer passed to the callback

The hook is zero-cost when unset (NULL check). Intended for use cases such as JNI bridges, device emulation, or other guest↔host ABI extensions that need their own HVC immediate.

---
## Summary by cubic
Adds an HVC 6 hook for custom host calls, separate from the Linux syscall path (HVC 5). Defaults to a no-op when no handler is registered.

- **New Features**
  - Adds `hvc6_handler(call_nr, args[8], userdata)` and `hvc6_userdata` to `guest_t`.
  - On HVC 6, reads X8 (call) and X0–X7 (args), calls the handler if set, and writes the result to X0.
  - Handler must be thread-safe; the hook is process‑local and must be re‑registered after fork/posix_spawn.

<sup>Written for commit cdbc3b26795110b154c85555692c7cca0c37f5e1. Summary will update on new commits. <a href="https://cubic.dev/pr/sysprog21/elfuse/pull/40?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/core/guest.h
+++ b/src/core/guest.h
@@ -275,6 +275,21 @@ typedef struct {
      * 64-bit to avoid wrap-around stale hits over long-running sessions.
      */
     _Atomic uint64_t pt_gen;
+
+    /*
+     * Optional HVC #6 embedder extension hook.
+     *
+     * The handler may be invoked concurrently from multiple host threads once
+     * the guest becomes multi-threaded, so implementations must be thread-safe.
+     *
+     * The hook is process-local and does not survive fork(). elfuse implements
+     * fork via posix_spawn() of a fresh process, so the child must register its
+     * own handler during bootstrap.
+     */
+    uint64_t (*hvc6_handler)(uint64_t call_nr,
+                             const uint64_t args[8],
+                             void *userdata);
+    void *hvc6_userdata;
 } guest_t;
 
 /* Bump the page table generation counter (call after any PT modification).
--- a/src/syscall/proc.c
+++ b/src/syscall/proc.c
@@ -1742,6 +1742,26 @@ int vcpu_run_loop(hv_vcpu_t vcpu,
                     break;
                 }
 
+                case 6: {
+                    /* HVC #6: embedder extension hook.
+                     * X8 = call number, X0-X7 = arguments.
+                     * If a dispatch function is registered via
+                     * elfuse_set_hvc6_handler(), it is called here.
+                     * Otherwise falls through as a no-op. */
+                    if (g->hvc6_handler) {
+                        uint64_t x8 = 0;
+                        hv_vcpu_get_reg(vcpu, HV_REG_X8, &x8);
+                        uint64_t args[8] = {0};
+                        for (int i = 0; i < 8; i++)
+                            hv_vcpu_get_reg(vcpu, HV_REG_X0 + i, &args[i]);
+                        uint64_t result =
+                            g->hvc6_handler(x8, args, g->hvc6_userdata);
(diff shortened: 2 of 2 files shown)


doanbaotrung:
This is fix for https://github.com/sysprog21/elfuse/issues/39

jserv (changes requested):
Run `make indent` for consistent style.

jserv:

Max042004:
`uint64_t args[8] = {};`  at src/syscall/proc.c:1754
The empty initializer is a C23 feature and trips -Wc23-extensions under the
project's CFLAGS (-Wall -Wextra -Wpedantic, default gnu17 on macOS clang). It
is also the only = {} in the tree.

doanbaotrung:
I fixed empty initializer `uint64_t args[8] = {};` by `uint64_t args[8] = {0};`

Max042004:
Two non-obvious things about the hook, please document them in guest.h

  - It can be called from several threads at once. Once the guest spawns
  threads (each guest thread runs on its own host pthread + vCPU), two threads
  hitting HVC 6 at the same time will call hvc6_handler concurrently. An embedder
  who doesn't know this might write a handler that isn't thread-safe.
  - It does not survive fork. elfuse implements fork by posix_spawn a fresh
   elfuse process, so the child is a different process with its own address
  space. The parent's hvc6_handler function pointer is meaningless in the child. 
   the child has to register its own hook during its bootstrap.
  
  Could you add a short comment above hvc6_handler in guest.h spelling out
  these two points? Just a few lines so the next embedder doesn't have to
  discover them the hard way.

Max042004:
Please squash commit

doanbaotrung:
Added comments to guest.h as below and aslo squash commits.
```
/*
     * Optional HVC #6 embedder extension hook.
     *
     * The handler may be invoked concurrently from multiple host threads once
     * the guest becomes multi-threaded, so implementations must be thread-safe.
     *
     * The hook is process-local and does not survive fork(). elfuse implements
     * fork via posix_spawn() of a fresh process, so the child must register its
     * own handler during bootstrap.
     */
```

jserv (changes requested):
Check https://cbea.ms/git-commit/ carefully and enforce the rules.
Append `Close #6` at the end of git commit message, instead of placing in the subject.

src/core/guest.h
```diff
@@ -275,6 +275,21 @@ typedef struct {
      * 64-bit to avoid wrap-around stale hits over long-running sessions.
      */
     _Atomic uint64_t pt_gen;
+
+    /*
+     * Optional HVC #6 embedder extension hook.
```
jserv, line 280: Remove issue tracker number by explaining rationale explicitly instead.

doanbaotrung: Updated

jserv (changes requested):
You MUST do `git rebase` properly.

doanbaotrung:
I re-base and squash commits again

Max042004:
In your commit message, HVC#6 becomes tag pull request #6. Please fix that.

doanbaotrung:
Fixed.

doanbaotrung:
Hi @Max042004 @jserv 
Please help to review again.

jserv:

PR #46, opened 2026-05-26
Reject CLONE_NEW* namespace flags in legacy clone(2)

### LINKED ISSUE #44: clone(2) silently ignores CLONE_NEW* namespace flags while clone3(2) rejects them with EINVAL
### Summary

The legacy `clone(2)` path silently ignores `CLONE_NEW*` namespace flags, while `clone3(2)` rejects the same flags with `EINVAL`. The two entry points therefore disagree on identical inputs.

Found while reviewing the OCI image support work (#31, PR #34): container runtimes and tools request namespaces through whichever `clone` variant their libc picks, so the behavior a guest sees depends on the syscall number rather than on a single, intentional policy.

### Details

`clone3(2)` explicitly rejects unsupported isolation features:

```c
// src/runtime/forkipc.c  (sys_clone3)
if (ca.flags & LINUX_CLONE_INTO_CGROUP)
    return -LINUX_EINVAL; /* cgroups not implemented */   // forkipc.c:1572
if (ca.flags & LINUX_CLONE3_NS_FLAGS)
    return -LINUX_EINVAL; /* namespaces not implemented */ // forkipc.c:1574
if (ca.set_tid_size != 0)
    return -LINUX_EINVAL; /* set_tid not implemented */    // forkipc.c:1576
```

The legacy path does not. `sc_clone` forwards the raw flags unchanged:

```c
// src/syscall/syscall.c:1543
return sys_clone(current_thread->vcpu, g, x0, x1, 0, 0, x2, x3, x4, ...);
```

and `sys_clone` (`src/runtime/forkipc.c:1061`) contains no `CLONE_NEW*` / cgroup / `set_tid` validation anywhere. The `CLONE_NEW*` bits simply fall through to the normal `posix_spawn`-based fork path.

### Impact

- `clone(CLONE_NEWPID | CLONE_NEWNS | ... )` returns a child PID and **appears to succeed**, but no isolation is set up. A caller that assumes the namespace was created proceeds on a false premise instead of getting a clean error it can handle or fall back from.
- `clone3()` with the same flags returns `EINVAL`, so a runtime that probes `clone3` first and falls back to `clone` will behave differently from one that calls `clone` directly.

### Expected

Both entry points should apply the same policy. Given the project's current "namespaces not implemented" stance, the legacy path should also reject these flags with `EINVAL`.

### Proposed fix

Mirror the `clone3` checks in `sys_clone` before the spawn path, reusing the existing mask:

```c
/* LINUX_CLONE3_NS_FLAGS is already defined for the clone3 path */
if (flags & LINUX_CLONE3_NS_FLAGS)
    return -LINUX_EINVAL; /* namespaces not implemented */
if (flags & LINUX_CLONE_INTO_CGROUP)
    return -LINUX_EINVAL; /* cgroups not implemented */
```

A regression test asserting that `clone()` and `clone3()` return the same result for each `CLONE_NEW*` flag would lock the two paths together.

Refs #31, PR #34.

Max042004:
sys_clone forwarded raw flags to the posix_spawn-based fork path without inspecting CLONE_NEW* bits, so clone(CLONE_NEWPID, ...) silently created an unisolated child and returned a PID as if the namespace had been set up. clone3 already rejected the same flags with EINVAL, so the observable result depended on which entry point the caller used.

Mirror the clone3 policy in sys_clone: reject any namespace flag with EINVAL. The exit signal occupies the CSIGNAL low byte in clone(2), so mask it off before testing. CLONE_NEWTIME (0x80) lives in that byte and, like CLONE_INTO_CGROUP (bit 33) and set_tid, cannot be conveyed through clone(2), so only the higher namespace bits are reachable. Move the namespace flag macros to the top of the file so both entry points share them.

Add a regression test asserting clone(2) rejects each reachable CLONE_NEW* flag, matching the existing clone3 coverage.

Closes #44

---
## Summary by cubic
Reject CLONE_NEW* namespace flags in legacy clone(2) to match clone3 behavior. This fixes a bug where clone() would silently fork without isolation and pretend the namespace was created.

- **Bug Fixes**
  - In `sys_clone`, mask off the CSIGNAL byte and return EINVAL when any CLONE_NEW* bit is set (namespaces aren’t implemented).
  - Share namespace flag macros across `sys_clone` and `sys_clone3`, and add a regression test that asserts clone(2) rejects all reachable flags (excluding CLONE_NEWTIME in CSIGNAL).

<sup>Written for commit 64ea4e3772137622be8489922bf71c9170a9662b. Summary will update on new commits. <a href="https://cubic.dev/pr/sysprog21/elfuse/pull/46?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/runtime/forkipc.c
+++ b/src/runtime/forkipc.c
@@ -360,6 +360,23 @@ int fork_child_main(int ipc_fd,
 #define LINUX_CLONE_CHILD_SETTID 0x01000000
 /* LINUX_SIGCHLD defined in syscall_signal.h (included above) */
 
+/* Namespace flags. elfuse implements no namespace isolation. Both
+ * sys_clone and sys_clone3 reject them.
+ */
+#define LINUX_CLONE_NEWTIME 0x00000080
+#define LINUX_CLONE_NEWNS 0x00020000
+#define LINUX_CLONE_NEWCGROUP 0x02000000
+#define LINUX_CLONE_NEWUTS 0x04000000
+#define LINUX_CLONE_NEWIPC 0x08000000
+#define LINUX_CLONE_NEWUSER 0x10000000
+#define LINUX_CLONE_NEWPID 0x20000000
+#define LINUX_CLONE_NEWNET 0x40000000
+
+#define LINUX_CLONE3_NS_FLAGS                                         \
+    (LINUX_CLONE_NEWNS | LINUX_CLONE_NEWCGROUP | LINUX_CLONE_NEWUTS | \
+     LINUX_CLONE_NEWIPC | LINUX_CLONE_NEWUSER | LINUX_CLONE_NEWPID |  \
+     LINUX_CLONE_NEWNET | LINUX_CLONE_NEWTIME)
+
 /* CLONE_THREAD: create a new guest thread in the same VM. */
 
 /* Arguments passed to the worker pthread. Allocated by sys_clone_thread, freed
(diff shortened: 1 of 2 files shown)


jserv:

PR #50, opened 2026-05-29
Emit GNU symbol versioning in the synthetic vDSO

### LINKED ISSUE #49: Go static aarch64 binaries SIGSEGV at startup (runtime.vdsoFindVersion, FAR=0x2)
A statically-linked Go binary built with `GOOS=linux GOARCH=arm64 CGO_ENABLED=0` SIGSEGVs under elfuse during runtime startup, before its first syscall:

```
HVC #11 EL0 fault EC=0x24 ESR=0x92000007 ELR=0x7f748 FAR=0x2 -> signal 11
```

`go tool addr2line` puts `ELR=0x7f748` in `runtime.vdsoFindVersion` (`runtime/vdso_linux.go`).

### Root cause

The Go runtime parses the kernel vDSO at startup (`vdsoInitFromSysinfoEhdr`) and marks any image with a hash table + symtab + strtab as valid, then calls `vdsoFindVersion`, which dereferences `info.verdef` guarded only by that validity flag. The synthetic vDSO in `src/core/vdso.c` emits `DT_HASH`/`DT_SYMTAB`/`DT_STRTAB` but no `DT_VERDEF`, so `verdef` is NULL and Go reads `vd_flags` at struct offset 2 -> guest address `0x2` -> data abort.

Every real Linux vDSO ships a `.gnu.version_d` defining `LINUX_2.6.39`, so the Go runtime never expects `verdef` to be absent. glibc and musl tolerate the omission (they fall back to unversioned name matching), which is why C, glibc, and musl workloads are unaffected.

### Repro

```
printf 'package main\nimport "fmt"\nfunc main() { fmt.Println("hello from go") }\n' > hello.go
GOOS=linux GOARCH=arm64 CGO_ENABLED=0 go build -o hello-go hello.go
./build/elfuse --verbose -- hello-go     # exit 139, data abort at FAR=0x2
```

### Fix

Emit `.gnu.version` (`DT_VERSYM`) + `.gnu.version_d` (`DT_VERDEF`) with a `LINUX_2.6.39` version node, matching the real arm64 vDSO. PR to follow.

chenhunghan:
## Problem

Static Go aarch64 binaries (`GOOS=linux GOARCH=arm64 CGO_ENABLED=0`) SIGSEGV under elfuse during runtime startup, before the first syscall: a data abort at `FAR=0x2`, PC in `runtime.vdsoFindVersion`.

The Go runtime parses the kernel vDSO (`vdsoInitFromSysinfoEhdr`) and treats any image with a hash table, symtab and strtab as valid, then calls `vdsoFindVersion` guarded only by that flag. `vdsoFindVersion` dereferences `info.verdef` unconditionally. The synthetic vDSO advertised `DT_HASH`/`DT_SYMTAB`/`DT_STRTAB` but no `DT_VERDEF`, so `verdef` was NULL and Go read `vd_flags` at struct offset 2 -> address `0x2`. Real Linux vDSOs always ship `.gnu.version_d` (`LINUX_2.6.39`); glibc and musl tolerate the omission, which is why only Go was affected.

## Fix

Emit `.gnu.version` (`DT_VERSYM`) and `.gnu.version_d` (`DT_VERDEF`): a base version node plus `LINUX_2.6.39`, with each exported symbol bound to it. The page layout shifts to make room (dynsym and hash move; versym and verdef are inserted before the dynamic table). This matches the real arm64 vDSO and keeps the vDSO fast path live for glibc/musl, which look up the same version.

## Test

`tests/test-go.sh` (wired into `make check`) builds a tiny linux/arm64 Go binary and runs it under elfuse, asserting it reaches `main` and that `time.Now()` (the resolved vDSO `clock_gettime`) returns a sane value. It skips (exit 77) when the Go toolchain is unavailable, mirroring the other optional checks.

## Verification

- Before: the Go binary exits 139 (SIGSEGV, `FAR=0x2`). After: it prints and exits 0; `make test-go` passes, and skips cleanly with no `go` on `PATH`.
- Ran `tests/driver.sh` against a prebuilt guest-binary corpus before and after the change: identical pass/fail set, no regression. (The local toolchain could not cross-build the in-tree guest tests, so the full `make check` plus `test-go` will exercise on the CI Apple Silicon runner.)
- `clang-format-22`, `cppcheck` (`.ci/check-cppcheck.sh`), `shellcheck`, `gofmt`: clean.

Closes #49

---
## Summary by cubic
Adds GNU symbol versioning to the synthetic vDSO to fix startup SIGSEGVs in static linux/arm64 Go binaries. The vDSO now matches Linux (with LINUX_2.6.39) and keeps the glibc/musl fast path.

- **Bug Fixes**
  - Emit `.gnu.version` (`DT_VERSYM`) and `.gnu.version_d` (`DT_VERDEF`) with a base node and `LINUX_2.6.39`; bind all exported symbols; add `DT_VERDEFNUM`.
  - Adjust vDSO layout and section counts to include the new tables; behavior matches the real arm64 vDSO for glibc/musl.
  - Add regression test `tests/test-go.sh` wired to `make check` (target `test-go`): builds a tiny static linux/arm64 Go binary, runs it under elfuse, and skips when the Go toolchain is absent.

<sup>Written for commit 729d7d805380dbca3f1604d78a28290a6ea70fc5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/50?utm_source=github">Review in cubic</a>

### DIFF
--- a/src/core/vdso.c
+++ b/src/core/vdso.c
@@ -14,6 +14,16 @@
  * HVF's CNTVOFF_EL2 virtualization, so the seqlock interpolation produced bogus
  * times (year 26382). The fast path is gone; SVC is correct and the trap cost
  * is negligible compared to the work clock_gettime callers tend to do anyway.
+ *
+ * GNU symbol versioning: every real Linux vDSO carries a .gnu.version_d
+ * (DT_VERDEF) defining "LINUX_2.6.39" plus a .gnu.version (DT_VERSYM) array
+ * binding each symbol to it. glibc and musl tolerate its absence, but the Go
+ * runtime does not: runtime.vdsoInitFromSysinfoEhdr marks any vDSO with a hash
+ * + symtab + strtab as valid, then runtime.vdsoFindVersion dereferences
+ * info.verdef (guarded only by info.valid). With no DT_VERDEF that pointer is
+ * NULL and Go reads vd_flags at struct offset 2 -> guest address 0x2 -> SIGSEGV
+ * before the first syscall. Emitting the version tables fixes Go and keeps the
+ * vDSO fast path live for glibc/musl, which look up the same "LINUX_2.6.39".
  */
 
 #include <stdint.h>
@@ -44,6 +54,15 @@ typedef struct {
     uint64_t st_value, st_size;
 } elf64_sym_t;
 
+typedef struct {
+    uint16_t vd_version, vd_flags, vd_ndx, vd_cnt;
+    uint32_t vd_hash, vd_aux, vd_next;
+} elf64_verdef_t;
+
+typedef struct {
+    uint32_t vda_name, vda_next;
+} elf64_verdaux_t;
+
 /* ELF constants */
 #define SHT_STRTAB 3
 #define SHT_HASH 5
@@ -59,6 +78,13 @@ typedef struct {
 #define DT_SYMENT 11
 #define STB_GLOBAL 1
 #define STT_FUNC 2
+#define SHT_GNU_VERSYM 0x6fffffffU
+#define SHT_GNU_VERDEF 0x6ffffffdU
+#define DT_VERSYM 0x6ffffff0
+#define DT_VERDEF 0x6ffffffc
+#define DT_VERDEFNUM 0x6ffffffd
+#define VER_DEF_CURRENT 1
+#define VER_FLG_BASE 0x1
 #define ELF_ST_INFO(bind, type) (((bind) << 4) | ((type) & 0xf))
 
 /* Layout.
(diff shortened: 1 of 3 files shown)


Max042004:
Statically linked Go binary run correctly on M4 machine

jserv (changes requested):
Rebase latest `main` branch, resolve conflicts, and resolve issues raised by cubic.

src/core/vdso.c
```diff
@@ -120,6 +160,19 @@ static const uint32_t sym_text_off[VDSO_NUM_SYMS] = {
...
+{
+    uint32_t h = 0, g;
+    for (const unsigned char *p = (const unsigned char *) name; *p; p++) {
+        h = (h << 4) + *p;
+        if ((g = h & 0xf0000000U))
+            h ^= g >> 24;
+        h &= ~g;
+    }
+    return h;
+}
```
jserv, line 174: The newly-introduced `elf_hash` is byte-for-byte identical to main branch's (both came from the SysV ELF reference).

jserv:
The exit-time SIGSEGV in this PR's regression scenario is a host-side teardown race in elfuse on current main, not a guest-side issue. The `--gdb` flag exposes the in-VM GDB Remote Serial Protocol stub for the guest aarch64-linux binary, whose perspective at exit is clean (it calls `exit_group(0)` and returns); the fault lives in the host elfuse process and surfaces under plain lldb on the macOS binary.

`lldb -b -o "run hello-go" ./build/elfuse` looped against current main (commit 7642bee), 5 of 15 runs:

```
* thread #3, stop reason = EXC_BAD_ACCESS (code=1, address=0xeffe00000)
    frame #0: elfuse`shim_globals_recompute_attention + 48
```

Call site is `src/syscall/proc.c:1170`, inside the HVC #5 epilogue:

```c
case 5: {
    int ret = syscall_dispatch(vcpu, g, &exit_code, verbose);
    if (ret == 1)
        running = false;
    ...
    shim_globals_recompute_attention(g);   // crash site
```

Race sequence: main vCPU runs `sys_exit_group(0)` and starts tearing the VM down; worker threads parked in nanosleep/futex_wait wake with EINTR (the `-> -4` returns visible under `--verbose`); each worker returns through the same HVC #5 epilogue and hits `shim_globals_recompute_attention(g)` before its `running = false` check unwinds the loop. By that point `g->host_base + g->shim_data_base` is stale -> EXC_BAD_ACCESS.

The `shim_globals_recompute_attention` call site was introduced by commit a24fc53, which landed after this PR was branched. PR #50 is based on b6ad3cb, where that call site does not exist, so the race window does not exist either. A/B with the same Go binary:

```
main (7642bee):     21/30 clean, 9/30 SIGSEGV 139
go-vdso-fix (PR):   30/30 clean
```

The exit-time crash is independent of vDSO versioning and belongs in a separate issue. For `tests/test-go.sh`, only the `grep -q 'hello from go vdso'` half of the pass gate pins the regression target (vDSO startup cleared). Dropping the `rc -eq 0` requirement would keep the test green on main once the PR rebases over a24fc53.

chenhunghan:
Closing — the fix this PR set out to add already landed on `main`.

The GNU symbol-versioning emission (`.gnu.version` / `.gnu.version_d` with `LINUX_2.6.39`) was independently added in a24fc53 ("Speedup vDSO CNTVCT and amortized urandom"), which landed after this branch was cut. Rebasing onto current `main` reduces the `src/core/vdso.c` change to a no-op — the file becomes byte-identical to upstream — and a freshly built `main` runs a static `linux/arm64` Go binary (`GOOS=linux GOARCH=arm64 CGO_ENABLED=0`) to completion. The startup SIGSEGV from #49 is therefore already resolved on `main`, so there's nothing left to carry here.

Thanks for the reviews.

PR #52, opened 2026-05-29
Add vDSO seqlock refresh and fast paths

jserv:
This extends the synthetic vDSO at AT_SYSINFO_EHDR with four new fast paths and rebuilds the anchor publication protocol so the anchor can be refreshed safely while concurrent guest readers are interpolating.

New trampolines, each ending in its own SVC fallback so the dynamic linker sees a complete __kernel_* symbol:
```
  __kernel_clock_getres   76 B / 19 instructions. Returns {0, 1}
                          inline for REALTIME, MONOTONIC, MONOTONIC_RAW,
                          REALTIME_COARSE, MONOTONIC_COARSE, BOOTTIME.
                          CPU and dynamic per-pid clockids SVC out.

  __kernel_gettimeofday  160 B / 40 instructions. Mirrors clock_gettime
                          using the REALTIME anchor and divides by 1000
                          for tv_usec. tz, if non-NULL, gets one
                          str xzr to clear the obsolete struct timezone.

  __kernel_getcpu         52 B / 13 instructions. Stores zero to both
                          out-pointers (elfuse models one CPU / one
                          node) and returns 0.

  __kernel_clock_gettime grows to 168 B / 42 instructions to hold the
                          anchor-age cap and the seqlock recheck added
                          to the existing CNTVCT trampoline.
```

VDSO_NUM_SYMS goes from 4 to 5; dynstr_data widens to 119 bytes; all post-text section offsets shift but the page still ends inside the 4 KiB at 0x6A0.

The vvar's first uint32 is now a Linux-style seqlock counter:
```
  0           unseeded, no anchor data
  odd  N >= 1 writer has reserved generation (N+1)/2
  even N >= 2 stable generation N/2, anchor fields readable
```

vdso_seed_anchor publishes through one CAS-then-release-store sequence that handles both initial seeding and refresh:
```
  load(seq, ACQUIRE), bail if odd
  CAS(seq, cur, cur+1, ACQUIRE, RELAXED), bail on contention
  thread_fence(RELEASE)           // CAS odd-publish before field stores
  store(field_i, RELAXED) * 5
  store(seq, cur+2, RELEASE)      // publish next even
```

The thread_fence(RELEASE) lowers to DMB ISH on AArch64 and closes the window where another CPU could observe the relaxed field stores before the odd-publish, since ARMv8 is not multi-copy atomic for unsynchronized stores to different locations. Without it a reader whose snapshot LDAR still saw the old even seq could read fields from the new generation and recheck the same old even seq, accepting torn data.

Trampoline readers snapshot the seq with LDAR, fall back on 0 or odd, plain-load anchor fields, DMB ISHLD, LDAR the seq again, and SVC on mismatch. The DMB ISHLD is load-bearing: LDAR provides forward acquire only, so without it the recheck load can be observed before the plain field LDR/LDPs complete and the race goes undetected. The host helpers (vdso_anchor_age_exceeded, vdso_realtime_drift_exceeded) read through a new vvar_snapshot_anchor() that mirrors this protocol in C: relaxed atomic field loads with __atomic_thread_fence(__ATOMIC_ACQUIRE) before the recheck.

Two staleness gates drive the refresh:
  - The trampolines LSR + CBNZ the CNTVCT delta against a 2**31-cycle cap (~89 s at 24 MHz). A stale anchor falls back to SVC so the host can publish a fresh one.
  - sys_clock_gettime and sys_gettimeofday sample both host clocks back-to-back on the vDSO SVC fallback and call vdso_seed_anchor whenever the anchor is unseeded, has aged out, or has drifted past VDSO_ANCHOR_MAX_DRIFT_NS (100 ms) relative to a fresh REALTIME. This catches macOS NTP wall-clock steps without a host timer thread. The drift detector short-circuits on age, guards anchor_sec + delta_sec with __builtin_add_overflow, and saturates the cross-second diff before the * 1e9 multiply so adversarial vvar values cannot trip signed overflow.

The host gates the HVF reads (ELR_EL1, X9) on clockid 0 or 1 every time, not just before the anchor is published. The previous short-circuit on vdso_anchor_is_seeded left stale anchors stranded.

sys_gettimeofday now writes 8 bytes of zero to tz_gva when non-null so SVC and fast-path callers see the same tz semantics; previously the SVC path ignored tz while the fast path zeroed it, so the first unseeded fallback with tz != NULL silently diverged.

Measured under tests/bench-vdso.c at 200000 iterations, with the seqlock and DMB ISHLD overhead included:
```
  clock_gettime(MONOTONIC) :    4.3 ns/op   (SVC: 1324 ns;  311x)
  clock_gettime(REALTIME)  :    4.1 ns/op   (SVC: 1315 ns;  321x)
  clock_getres(MONOTONIC)  :    2.3 ns/op   (SVC: 1349 ns;  591x)
  gettimeofday             :    4.9 ns/op   (SVC: 1331 ns;  271x)
  getcpu                   :    2.0 ns/op   (SVC: 1519 ns;  755x)
```

The ~1.5 ns increase on clock_gettime over the prior one-shot anchor (~2.5 ns baseline) is the cost of the seqlock recheck plus DMB ISHLD plus the age-cap LSR+CBNZ. clock_getres and getcpu have no anchor dependence so they stay at ~2 ns.

---
## Summary by cubic
Adds a seqlock-protected vDSO time anchor with safe refresh and fast paths for `__kernel_clock_getres`, `__kernel_clock_gettime`, `__kernel_gettimeofday`, and `__kernel_getcpu`, plus an `NT_GNU_ABI_TAG` note so dynamic glibc binds to the vDSO. Also collapses the `clock_gettime` carry to remove the final divide, cutting dyn‑glibc down to ~4 ns.

- **New Features**
  - vvar seqlock time anchor with LDAR + DMB ISHLD + recheck and a 2^22‑cycle age cap; fast trampolines with SVC fallbacks for the four `__kernel_*` symbols. Host refresh on SVC (seed/refresh on unseeded, aged, or >100 ms REALTIME drift); `sys_clock_getres` returns fixed resolutions for common clockids; `sys_gettimeofday` zeros `tz` to match fast‑path behavior.
  - `PT_NOTE` with `NT_GNU_ABI_TAG` (Linux 2.6.39); PHDR table relocated to keep vvar/text offsets stable. New tests and benches, plus a hot‑syscall guardrail in `make check` enforcing ceilings: getpid ≤ 200 ns, libc `clock_gettime` ≤ 50 ns, 1‑byte `/dev/urandom` read ≤ 200 ns. Optional dynamic‑glibc bench verifies glibc routes through the vDSO.

- **Refactors**
  - Replaced the hot‑path UDIV-by-1e9 in `clock_gettime` with SUBS+CSEL+CINC by tightening the anchor age shift from 31 to 22; keeps X9 live for the SVC fallback. Dyn‑glibc `clock_gettime` ~4.0 ns (static ~3.4 ns); costs one extra re‑seed per ~0.175 s idle. Host drift detector updated to the same mult+shift so rounding matches the trampoline.

<sup>Written for commit f5b3e21b571e1543fd7fa227364b588b4ec66b64. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/52?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -221,6 +221,25 @@ $(BUILD_DIR)/test-lowbase-mem-300000: tests/test-lowbase-mem.c | $(BUILD_DIR)
 	$(Q)$(CROSS_COMPILE)gcc -D_GNU_SOURCE -static -O2 -no-pie \
 		-Wl,-Ttext-segment=0x300000 -o $@ $<
 
+# bench-hot-guard-glibc is the dynamic-glibc twin of bench-hot-guard.
+# Built only when the cross-glibc toolchain ships its own sysroot
+# (so a host without that toolchain can still run the rest of the
+# suite). Linked without -static so glibc resolves time / urandom
+# syscalls through the vDSO trampoline -- which is exactly what the
+# guardrail script verifies against the 50 ns / 200 ns ceilings.
+ifneq ($(wildcard $(LINUX_TOOLCHAIN)/aarch64-unknown-linux-gnu/sysroot/.),)
+# -DGUARD_USE_LIBC_CG switches the bench's clock_gettime case from a
+# direct vDSO trampoline call to the libc wrapper, so the dynamic-glibc
+# build measures glibc's actual routing decision. A regression in the
+# NT_GNU_ABI_TAG note or LINUX_2.6.39 versioning would push this
+# measurement from ~7 ns up to SVC time (~2000 ns) and fail the
+# guardrail.
+$(BUILD_DIR)/bench-hot-guard-glibc: tests/bench-hot-guard.c | $(BUILD_DIR)
+	@echo "  CROSS   $< (dynamic glibc)"
+	$(Q)$(CROSS_COMPILE)gcc -D_GNU_SOURCE -DGUARD_USE_LIBC_CG=1 -O2 \
+		-o $@ $<
+endif
+
 endif
 
 include mk/tests.mk
--- a/mk/tests.mk
+++ b/mk/tests.mk
@@ -51,6 +51,28 @@ check: $(ELFUSE_BIN) $(TEST_DEPS) check-syscall-coverage
 	@$(MAKE) --no-print-directory test-timeout-disable
 	@printf "\n$(BLUE)━━━ rosetta CLI gating ━━━$(RESET)\n"
 	@$(MAKE) --no-print-directory test-rosetta-cli
+	@printf "\n$(BLUE)━━━ hot-syscall guardrail ━━━$(RESET)\n"
+	@$(MAKE) --no-print-directory test-bench-guardrail
+
+## Hot-syscall performance guardrail: ensure getpid, libc clock_gettime,
+## and 1-byte /dev/urandom reads stay under their TODO ns/op ceilings.
+## Builds the dynamic-glibc variant opportunistically; the script skips
+## that arm when the cross-toolchain sysroot is missing.
+BENCH_GUARDRAIL_DEPS := $(ELFUSE_BIN)
+BENCH_GUARDRAIL_REQUIRE_STATIC := 0
+ifndef GUEST_TEST_BINARIES
+  BENCH_GUARDRAIL_DEPS += $(BUILD_DIR)/bench-hot-guard
+  BENCH_GUARDRAIL_REQUIRE_STATIC := 1
+  ifneq ($(wildcard $(LINUX_TOOLCHAIN)/aarch64-unknown-linux-gnu/sysroot/.),)
+    BENCH_GUARDRAIL_DEPS += $(BUILD_DIR)/bench-hot-guard-glibc
+  endif
+endif
+test-bench-guardrail: $(BENCH_GUARDRAIL_DEPS)
+	@ELFUSE="$(ELFUSE_BIN)" \
+	    BENCH_GUARDRAIL_DIR="$(TEST_DIR)" \
+	    BENCH_GUARDRAIL_REQUIRE_STATIC="$(BENCH_GUARDRAIL_REQUIRE_STATIC)" \
+	    LINUX_TOOLCHAIN="$(LINUX_TOOLCHAIN)" \
+	    bash tests/test-bench-guardrail.sh
 
 test-sysroot-rename: $(ELFUSE_BIN) $(BUILD_DIR)/test-sysroot-rename
 	@set -e; \
--- a/src/core/elf.h
+++ b/src/core/elf.h
@@ -45,6 +45,7 @@
 #define PT_LOAD 1
 #define PT_DYNAMIC 2
 #define PT_INTERP 3
+#define PT_NOTE 4
 
 /* Program header flags */
 #define PF_X 1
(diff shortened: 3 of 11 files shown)


Max042004:
The same `python3.12` ELF binary was run on Apple M-series HVF under both elfuse and orbstack (Ubuntu 24.04 noble arm64, Linux 7.0.5-orbstack) to compare time-related syscall / vDSO behavior.


### Startup time: `python3.12 -c 'pass'` (10 runs)

| | samples (ms) | min | median |
|---|---|---|---|
| **elfuse (PR #52)** | 48 41 35 34 34 34 34 34 34 34 | **34** | **34** |
| **orbstack** | 55 35 35 35 35 34 35 35 35 36 | **34** | **35** |

### vDSO-heavy Python loops (N=500k, 5 rounds, median ns/op)

| Operation | elfuse (PR #52) | orbstack | Gap |
|---|---:|---:|---|
| `time.time()` (glibc → `__kernel_gettimeofday`) | **36.2** | 37.0 | **elfuse 2% faster** |
| `time.monotonic_ns()` (→ `clock_gettime(MONO)`) | 35.2 | **30.1** | orbstack 17% faster |
| `clock_gettime(CLOCK_REALTIME)` | 49.3 | **40.2** | orbstack 23% faster |
| `clock_gettime(CLOCK_MONOTONIC)` | 49.8 | **40.4** | orbstack 23% faster |
| `clock_getres(CLOCK_MONOTONIC)` | 49.3 | **38.9** | orbstack 27% faster |

### CPU-bound control: `fibonacci(50000)` (10 runs)

| | samples (ms) | min | median |
|---|---|---|---|
| **elfuse (PR #52)** | 51 52 51 51 53 52 51 51 51 51 | **51** | **51** |
| **orbstack** | 52 52 52 51 52 51 53 51 52 52 | **51** | **52** |

Max042004 (approved):

Max042004:
The same statically-linked `bench-hot-guard` binary was executed inside orbstack via `/mnt/mac/...`. Its symbol-resolution code (`AT_SYSINFO_EHDR` + `DT_HASH`) walks whatever vDSO the running kernel publishes — elfuse's synthetic one in the elfuse case, the real Linux 7.0.5 vDSO in the orbstack case — so this is an apples-to-apples comparison of two vDSO `__kernel_clock_gettime` implementations on the same CPU.

5 runs each, ns/op:

| | run 1 | run 2 | run 3 | run 4 | run 5 | median |
| **elfuse `f5b3e21`** | 2.6 | 2.2 | 2.0 | 1.8 | 1.9 | **2.0** |
| **orbstack (Linux 7.0.5 vDSO)** | 8.7 | 9.1 | 9.3 | 9.5 | 8.8 | **9.1** |

PR #56, opened 2026-05-29
Dispatch getsockname/sendto/recvfrom on netlink sockets

### LINKED ISSUE #53: AF_NETLINK sockets reject sendto/recvfrom/getsockname (ENOTSOCK), and RTM_GETLINK ignores the ifi_index/IFLA_IFNAME filter
## Summary

AF_NETLINK / NETLINK_ROUTE sockets only support `bind`/`sendmsg`/`recvmsg`/`read`. The flat-buffer socket calls `sendto(2)`, `recvfrom(2)` and `getsockname(2)` are not dispatched to the netlink emulation, so they fall through to the real host socket calls on the netlink fd's pipe backing and fail with **ENOTSOCK** (errno 88, "Socket operation on non-socket"). This breaks the most common interface-enumeration path: glibc `getifaddrs()` issues `sendto(2)` on a NETLINK_ROUTE socket internally and therefore fails outright.

Separately, `RTM_GETLINK` ignores the request filter (`ifi_index` / `IFLA_IFNAME`): a single-link lookup returns **every** interface instead of the one requested.

## Reproduction

Generic, no manual netlink message construction — glibc does it for us:

```c
// repro.c
#define _GNU_SOURCE
#include <stdio.h>
#include <errno.h>
#include <string.h>
#include <ifaddrs.h>
int main(void) {
    struct ifaddrs *ifa, *p;
    if (getifaddrs(&ifa) < 0) {
        printf("getifaddrs: FAILED errno=%d (%s)\n", errno, strerror(errno));
        return 1;
    }
    int n = 0; for (p = ifa; p; p = p->ifa_next) n++;
    printf("getifaddrs: OK, %d address entries\n", n);
    freeifaddrs(ifa); return 0;
}
```

```
$ aarch64-linux-gnu-gcc -D_GNU_SOURCE -static -O2 -o repro repro.c
$ ./elfuse -- ./repro
getifaddrs: FAILED errno=88 (Socket operation on non-socket)
```

Per-call confirmation (direct syscalls on an `AF_NETLINK`/`SOCK_RAW`/`NETLINK_ROUTE` socket):

```
getsockname: ret=-1 errno=88 (Socket operation on non-socket)
sendto:      ret=-1 errno=88 (Socket operation on non-socket)
recvfrom:    ret=-1 errno=88 (Socket operation on non-socket)
```

Filter bug (observed via the `sendmsg` path, which is dispatched): a non-DUMP `RTM_GETLINK` with `ifi_index=1` returns `RTM_NEWLINK count=24` instead of `1`.

## Expected vs Actual

| | Expected | Actual (this build) |
|---|---|---|
| glibc `getifaddrs()` | succeeds, lists addresses | `FAILED errno=88 (ENOTSOCK)` |
| `getsockname` on netlink fd | returns bound `nl_pid`, family `AF_NETLINK` | `-1 ENOTSOCK` |
| `sendto`/`recvfrom` on netlink fd | transfer rtnetlink messages | `-1 ENOTSOCK` |
| `RTM_GETLINK` with `ifi_index=N` (no DUMP) | exactly the matching link | every interface |

## Root cause (file:line)

Netlink sockets are synthetic: `socket(AF_NETLINK,...)` allocates an `FD_NETLINK` descriptor whose host backing is the **read end of a pipe** — `src/syscall/netlink.c:418` (`fd_alloc(FD_NETLINK, pipefd[0], ...)`). `bind`/`sendmsg`/`recvmsg`/`read` branch on `fd_get_type(fd) == FD_NETLINK`, but three handlers do not:

- `sys_getsockname` — `src/syscall/net.c:507` (no `FD_NETLINK` branch; calls `host_fd_ref_open()` then host `getsockname()` on the pipe fd)
- `sys_sendto` — `src/syscall/net.c:634` (same; host `send()`/`sendto()` on the pipe)
- `sys_recvfrom` — `src/syscall/net.c:701` (same; host `recv()`/`recvfrom()` on the pipe)

`host_fd_ref_open()` (`src/syscall/internal.h:245`) returns the pipe's host fd, and a socket op on a pipe fd is ENOTSOCK on macOS.

Filter bug: `nl_build_getlink(ns)` (`src/syscall/netlink.c:192`) takes no filter and emits one `RTM_NEWLINK` per host interface unconditionally; the `RTM_GETLINK` request's `ifi_index`/`IFLA_IFNAME` are never parsed.

## Affected programs

Any program that enumerates or looks up network interfaces over rtnetlink:
- **glibc `getifaddrs()` / `if_nameindex()`** (NETLINK_ROUTE-based, `sendto`-driven) — breaks every C/C++ interface enumeration.
- **Go** `net` package and `vishvananda/netlink` (`sendto`/`recvfrom` + `getsockname` for the auto-assigned `nl_pid`).
- **iproute2 `ip`/`ss`, busybox `ip`** (rtnetlink link/addr/route dumps).
- **systemd / systemd-networkd / udev (sd-netlink)**, **NetworkManager**, **dhcpcd/udhcpc** (rtnetlink monitoring).
- **Rust** std interface enumeration and the `nix`/`rtnetlink` crates.
- Anything doing a `LinkByName`/`LinkByIndex`-style single-interface lookup additionally gets the wrong (all-interfaces) result and may error with "more than one link found".

## Fix

Addressed by adding `netlink_send`/`netlink_recv`/`netlink_getsockname` and `FD_NETLINK` dispatch in `sys_sendto`/`sys_recvfrom`/`sys_getsockname`, plus a shared `nl_parse_link_filter` so the `sendmsg` and `sendto` request paths both honor `ifi_index`/`IFLA_IFNAME`. After the fix, `getifaddrs()` returns 38 entries, the three calls succeed, and the filtered `RTM_GETLINK` returns exactly 1 link.

_First observed while bringing up a Go-based control plane (k0s): its node-IP/SAN detection fell back to 127.0.0.1 because the netlink path returned ENOTSOCK. The bug itself is generic to the netlink emulation and independent of any particular workload._

chenhunghan:
Fixes #53.

Dispatches `getsockname`/`sendto`/`recvfrom` to the AF_NETLINK emulation (they previously fell through to the host socket calls on the netlink pipe fd and returned ENOTSOCK), and makes `RTM_GETLINK` honor the `ifi_index`/`IFLA_IFNAME` request filter so a single-link lookup returns exactly one link. Full rationale in the commit message.

Validation on Apple Silicon: `make elfuse`, `make check`, and `make test-matrix` pass; glibc `getifaddrs()` now succeeds where it previously failed with ENOTSOCK (errno 88). (Hosted CI builds + lints only; the runtime tests require an HVF host, so those results are local.)

---
## Summary by cubic
Dispatches netlink getsockname/sendto/recvfrom and honors RTM_GETLINK name/index filters to return a single link when requested. Fixes ENOTSOCK fallthroughs and restores clients using flat send/recv and port ID via getsockname (e.g., getifaddrs and Go netlink).

- **Bug Fixes**
  - Route FD_NETLINK in sys_getsockname/sys_sendto/sys_recvfrom to netlink_getsockname/netlink_send/netlink_recv, fixing ENOTSOCK on the host pipe fd.
  - Honor ifi_index and IFLA_IFNAME in RTM_GETLINK and return only the matching link; share dispatch via nl_process_request so sendmsg and flat send use the same filters.
  - Add tests/test-netlink.c and register it in tests/manifest.txt to cover getsockname/sendto/recvfrom and the getifaddrs() flow.

<sup>Written for commit fafdb43d70fc35f71cb77c7e99c29dcc6640d1ff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/56?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/test-netlink.c
+++ b/tests/test-netlink.c
@@ -1,0 +1,135 @@
+/* Exercise the AF_NETLINK getsockname/sendto/recvfrom dispatch paths.
+ *
+ * Copyright 2026 elfuse contributors
+ * Copyright 2025 Moritz Angermann, zw3rk pte. ltd.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Regression guard for the netlink socket emulation. Before getsockname,
+ * sendto, and recvfrom were routed to the netlink handlers, these calls fell
+ * through to the host socket syscalls on the underlying pipe fd and failed
+ * with ENOTSOCK (errno 88), which in turn broke glibc getifaddrs(). The test
+ * drives each of the three syscalls directly against a NETLINK_ROUTE socket
+ * and then validates the end-to-end getifaddrs() path that originally
+ * regressed.
+ *
+ * The assertions hold for both the elfuse emulation and a real Linux kernel
+ * (the test matrix runs the same binary under qemu-aarch64), so only
+ * implementation-independent netlink semantics are checked.
+ */
+
+#include <errno.h>
+#include <ifaddrs.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+#include <sys/socket.h>
+#include <linux/netlink.h>
+#include <linux/rtnetlink.h>
+
+static int pass, fail;
+
+#define CHECK(cond, msg)                                                       \
+    do {                                                                       \
+        if (cond) {                                                            \
+            printf("PASS: %s\n", (msg));                                       \
+            pass++;                                                            \
+        } else {                                                               \
+            printf("FAIL: %s (errno=%d %s)\n", (msg), errno, strerror(errno)); \
+            fail++;                                                            \
+        }                                                                      \
+    } while (0)
+
+/* RTM_GETLINK dump request: nlmsghdr immediately followed by ifinfomsg. */
+struct getlink_req {
+    struct nlmsghdr nlh;
+    struct ifinfomsg ifi;
+};
+
+int main(void)
+{
+    int fd = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
+    if (fd < 0) {
+        printf("FAIL: socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE): %s\n",
+               strerror(errno));
+        /* Without a socket none of the dispatch paths can be reached. */
+        printf("\n%d passed, %d failed\n", pass, fail + 1);
+        return 1;
+    }
+    printf("PASS: socket(AF_NETLINK, NETLINK_ROUTE) = %d\n", fd);
+    pass++;
+
+    /* bind() with nl_pid=0 lets the kernel/emulation assign a port id. */
+    struct sockaddr_nl local = {.nl_family = AF_NETLINK};
+    CHECK(bind(fd, (struct sockaddr *) &local, sizeof(local)) == 0,
+          "bind(AF_NETLINK)");
+
+    /* 1. getsockname(): previously ENOTSOCK on the pipe fd. */
+    struct sockaddr_nl got = {0};
+    socklen_t gotlen = sizeof(got);
+    int rc = getsockname(fd, (struct sockaddr *) &got, &gotlen);
+    CHECK(rc == 0 && gotlen >= sizeof(struct sockaddr_nl) &&
+              got.nl_family == AF_NETLINK,
+          "getsockname() returns an AF_NETLINK address");
+    CHECK(rc == 0 && got.nl_pid != 0,
+          "getsockname() reports a non-zero port id");
+
+    /* 2. sendto(): flat request buffer, no msghdr. */
+    struct getlink_req req = {0};
+    req.nlh.nlmsg_len = sizeof(req);
+    req.nlh.nlmsg_type = RTM_GETLINK;
+    req.nlh.nlmsg_flags = NLM_F_REQUEST | NLM_F_DUMP;
+    req.nlh.nlmsg_seq = 1;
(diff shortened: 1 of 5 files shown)


Max042004 (changes requested):
make check not cover new three syscall. It is better to implements a simple test so that in future we can catch whether has regression for this.

tests/test-netlink.c
```diff
@@ -0,0 +1,136 @@
...
+ */
+
+#include <errno.h>
+#include <ifaddrs.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+#include <sys/socket.h>
+#include <net/if.h>
```
Max042004: net/if.h not used. Can delete

Max042004 (changes requested):

jserv:
ping?

tests/test-netlink.c
chenhunghan: Removed the unused `net/if.h` include — folded into the test commit (now `fafdb43`). CI is green and the test passes 9/9 under elfuse. Thanks for the review!

jserv:
Thank @chenhunghan for contributing!

PR #57, opened 2026-05-29
Treat chown EPERM as no-op success for emulated-root guests

### LINKED ISSUE #54: chown(2)/fchown(2) fail with EPERM under emulation when the host process is non-root
## Summary
Under the emulator, `chown(2)`, `fchown(2)`, `lchown(2)`, and `fchownat(2)` fail with `EPERM` ("Operation not permitted") whenever the guest tries to set a file's owner/group to an id its emulated identity should be allowed to use (most commonly uid/gid 0). The emulator presents the guest with a synthetic credential model but executes the chown directly against the host kernel as an unprivileged macOS user, so the host's `EPERM` leaks up to the guest. This breaks any program that sets file ownership (package installers, archive extractors, `chown(1)`/`install`).

## Reproduction
Cross-compile a static aarch64 binary (host runs as a non-root user):

```c
#define _GNU_SOURCE
#include <stdio.h>
#include <fcntl.h>
#include <unistd.h>
#include <errno.h>
#include <string.h>
int main(void){
    const char *p = "/tmp/repro-chown-target";
    int fd = open(p, O_CREAT|O_RDWR|O_TRUNC, 0644);
    if (fd < 0){ printf("open: %s\n", strerror(errno)); return 2; }
    errno=0; int r1 = chown(p, 0, 0);
    printf("chown(%s, 0, 0) = %d errno=%d (%s)\n", p, r1, r1?errno:0, r1?strerror(errno):"OK");
    errno=0; int r2 = fchown(fd, 0, 0);
    printf("fchown(fd, 0, 0)      = %d errno=%d (%s)\n", r2, r2?errno:0, r2?strerror(errno):"OK");
    close(fd); unlink(p);
    return (r1==0 && r2==0) ? 0 : 1;
}
```

```
aarch64-linux-gnu-gcc -D_GNU_SOURCE -static -O2 -o /tmp/repro-chown /tmp/repro-chown.c
./build/elfuse -- /tmp/repro-chown
```

Observed output:
```
chown(/tmp/repro-chown-target, 0, 0) = -1 errno=1 (Operation not permitted)
fchown(fd, 0, 0)      = -1 errno=1 (Operation not permitted)
```
(exit code 1)

## Expected vs Actual
- **Expected:** a guest holding an emulated identity that owns the file (or believes it is root) gets `chown`/`fchown` success (return 0), matching real-Linux behavior where the program is run with sufficient privilege. Programs proceed past the ownership step.
- **Actual:** every such `chown`/`fchown` returns `-EPERM`, aborting the operation.

## Root cause
`src/syscall/fs.c:1722` (`sys_fchownat`) and `src/syscall/fs.c:1740` (`sys_fchown`) call the host `fchownat()`/`fchown()` directly and, on failure, return `linux_errno()` (fs.c:1724, fs.c:1742). The emulator gives the guest a fixed emulated identity (`GUEST_UID`/`GUEST_GID`, see `src/syscall/abi.h:263` and `src/syscall/proc-identity.c`) that is independent of the real macOS credentials of the emulator process. macOS only lets the superuser change a file's owner/group to an arbitrary id, so the unprivileged host process gets `EPERM`, which is passed straight through. There is no mapping from the emulated credential model onto the host's real (unprivileged) credentials. Note glibc implements `chown`/`lchown` via `fchownat(AT_FDCWD, ...)`, so this one path covers the whole family.

## Affected programs
- Package managers that chown extracted files: `dpkg`/`apt`, `rpm`, `apk` (BusyBox/Alpine).
- `tar`/`cpio` restoring archived ownership (`-p`, `--same-owner`).
- GNU coreutils `chown(1)`, `chgrp(1)`, `install -o/-g`.
- `cp -p`, `rsync -a`, `mv` across owners (ownership-preservation step).
- Container image unpack tooling (containerd, runc) laying down a rootfs with explicit ownership.
- Build/install flows and fakeroot-style setups that chown installed artifacts.
- `shutil.chown`/`os.chown` (Python) and `fs.chown` (Node) in bootstrap scripts.

## Suggested fix
Treat host `EPERM` from `chown`/`fchown` as a no-op success for the emulated-root/credential model, while still propagating all other errors (`EBADF`, `ENOENT`, etc.). Verified: with the EPERM->0 mapping applied, the repro returns 0 for both calls, and a negative test confirms `fchown(badfd)=EBADF` and `chown(missing)=ENOENT` still propagate correctly.

<sub>First observed while booting the k0s controller, which chowns the key material it writes during startup and aborts on the host EPERM.</sub>

chenhunghan:
Fixes #54.

Maps host `EPERM` from `chown`/`fchown`/`fchownat` to a no-op success: an emulated-root guest expects ownership changes to succeed, but macOS only permits chown-to-arbitrary-uid for the superuser. All other errors propagate. Full rationale in the commit message.

Validation on Apple Silicon: `make elfuse` and `make check` pass; the invalid-`*at`-flags negative path still returns EINVAL (rejected before the chown call).

---
## Summary by cubic
Treat host EPERM from chown/fchown/fchownat as a no-op success for emulated-root guests on macOS to prevent false failures when changing ownership. All other errors propagate normally; invalid *at flags still return EINVAL.

<sup>Written for commit 8a1eec42a345ed64d6da8e7b4c7a94ba3cf9d4cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/57?utm_source=github">Review in cubic</a>

### DIFF
--- a/src/syscall/fs.c
+++ b/src/syscall/fs.c
@@ -1693,6 +1693,14 @@ int64_t sys_fchmodat(guest_t *g,
     return 0;
 }
 
+/* Fake success on EPERM: emulated-root guests expect chown to succeed. */
+static int64_t chown_result(int rc)
+{
+    if (rc < 0)
+        return (errno == EPERM) ? 0 : linux_errno();
+    return 0;
+}
+
 int64_t sys_fchownat(guest_t *g,
                      int dirfd,
                      uint64_t path_gva,
@@ -1719,13 +1727,10 @@ int64_t sys_fchownat(guest_t *g,
         return -LINUX_EBADF;
 
     int mac_flags = translate_at_flags(flags);
-    if (fchownat(dir_ref.fd, tx.host_path, owner, group, mac_flags) < 0) {
-        host_fd_ref_close(&dir_ref);
-        return linux_errno();
-    }
-
+    int64_t out = chown_result(
+        fchownat(dir_ref.fd, tx.host_path, owner, group, mac_flags));
     host_fd_ref_close(&dir_ref);
-    return 0;
+    return out;
 }
 
 int64_t sys_fchown(int fd, uint32_t owner, uint32_t group)
@@ -1737,12 +1742,9 @@ int64_t sys_fchown(int fd, uint32_t owner, uint32_t group)
     host_fd_ref_t host_ref;
     if (host_fd_ref_open(fd, &host_ref) < 0)
         return -LINUX_EBADF;
-    if (fchown(host_ref.fd, owner, group) < 0) {
-        host_fd_ref_close(&host_ref);
-        return linux_errno();
-    }
+    int64_t out = chown_result(fchown(host_ref.fd, owner, group));
     host_fd_ref_close(&host_ref);
-    return 0;
+    return out;
 }
 
 int64_t sys_utimensat(guest_t *g,


Max042004:
This can be a temporarily fix. But still has problem for program need to check owner. So I open a new issue to track #59

Max042004 (approved):

jserv:
Thank @chenhunghan for contributing!

PR #58, opened 2026-05-29
Emit named IN_CREATE/IN_DELETE for inotify directory watches

### LINKED ISSUE #55: inotify directory watches deliver IN_CREATE/IN_DELETE events with an empty name[] field
## Summary
inotify directory watches deliver `IN_CREATE` / `IN_DELETE` events with an **empty `name[]` field** (`len == 0`). On real Linux, directory-watch events for create/delete/move always name the affected child entry. Because the emulated events are nameless, any consumer that filters or dispatches on the entry name (the entire fsnotify / watchdog / chokidar / inotify-tools class of programs) silently drops them — the event is delivered but carries no usable information about *which* file changed.

## Reproduction
Cross-compile and run the minimal program below under the unpatched and patched binaries:

```c
#define _GNU_SOURCE
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/inotify.h>
#include <unistd.h>

int main(void) {
    char dir[] = "/tmp/repro-ino-XXXXXX";
    if (!mkdtemp(dir)) return 2;
    int fd = inotify_init1(IN_NONBLOCK);
    int wd = inotify_add_watch(fd, dir, IN_CREATE);
    char path[300];
    snprintf(path, sizeof(path), "%s/manifest.yaml", dir);
    close(open(path, O_WRONLY | O_CREAT | O_EXCL, 0644));
    for (int a = 0; a < 50; a++) {
        char buf[1024];
        ssize_t n = read(fd, buf, sizeof(buf));
        for (ssize_t off = 0; n > 0 && off + (ssize_t)sizeof(struct inotify_event) <= n;) {
            struct inotify_event *ev = (struct inotify_event *)(buf + off);
            if (ev->mask & IN_CREATE)
                printf("IN_CREATE len=%u name='%s'\n", ev->len, ev->len ? ev->name : "");
            off += (ssize_t)sizeof(struct inotify_event) + ev->len;
        }
        usleep(20000);
    }
    return 0;
}
```

```
aarch64-linux-gnu-gcc -D_GNU_SOURCE -static -O2 -o /tmp/repro-inotify /tmp/repro-inotify.c
elfuse -- /tmp/repro-inotify
```

## Expected vs Actual
**Expected** (matches real Linux): `IN_CREATE len=16 name='manifest.yaml'`

**Actual (unpatched)**:
```
IN_CREATE len=0 name=''
```
The event fires but `name[]` is empty, so the consumer cannot tell which child was created.

## Root cause
`src/syscall/inotify.c`. kqueue's `EVFILT_VNODE` `NOTE_WRITE` reports that a watched directory changed but not which child. The emulation queues the event directly with a `NULL` name:
- `collect_events()` — `src/syscall/inotify.c:331`: `queue_event(inst, w->wd, in_mask, 0, NULL)`
- `inotify_read()` — `src/syscall/inotify.c:630`: `queue_event(inst, w->wd, in_mask, 0, NULL)`

This is by design per the comments at `notes_to_in_mask` (`src/syscall/inotify.c:206-208`) and `collect_events` (`src/syscall/inotify.c:327-330`), which note the filename is omitted because `EVFILT_VNODE` doesn't report the child. The fix is to recover the name out-of-band: snapshot the directory's entry names at `inotify_add_watch` time and, on each `NOTE_WRITE`, re-list and diff to emit a named `IN_CREATE` per added entry and named `IN_DELETE` per removed entry.

## Affected programs
- **Go fsnotify** (`github.com/fsnotify/fsnotify`) — Linux backend reads `name[]` to build the changed path; nameless dir events produce empty `Event.Name` and are filtered out.
- **inotify-tools** (`inotifywait`/`inotifywatch`) — `%f`/per-name filtering relies on `name[]`.
- **Node.js `fs.watch` / chokidar** (webpack/vite/VS Code watchers) — map `name[]` to the changed file path; without it, recursive/dir watching and hot-reload don't fire for the right file.
- **Python `watchdog`** — builds `src_path` from dir path + `name[]`; empty name breaks file-level dispatch and pattern matching.
- **systemd `.path` units / incron** — match on the created/deleted entry name.
- Any **drop-folder / spool-directory** watcher that acts only on new files of a given name/extension.

## Fix
Resolved by snapshot+diff of directory entry names, sharing one `process_vnode_event()` helper across the blocking-read and non-blocking-collect paths; snapshot allocated on `add_watch`, freed on `rm_watch`/close. Regression test added that watches a fresh dir, creates a child, and asserts a named `IN_CREATE` is delivered (fails before the fix).

<sub>First observed while running a Go file-watching workload (k0s manifest applier) whose fsnotify-based directory watch silently never picked up newly written files.</sub>

chenhunghan:
Fixes #55.

Synthesizes named `IN_CREATE`/`IN_DELETE` for directory watches by snapshotting the directory's entry names and diffing on each `NOTE_WRITE` (kqueue `EVFILT_VNODE` reports only that the directory changed, not which child). Full rationale in the commit message.

Adds a regression test (`test-inotify` Test 6) that watches a fresh directory, creates a child, and asserts a named `IN_CREATE` is delivered.

Validation on Apple Silicon: `make elfuse` and `make check` pass; the new test fails before this change (event arrives with an empty name) and passes after.

---
## Summary by cubic
Emit named IN_CREATE/IN_DELETE for directory watches by diffing per‑watch snapshots, and keep the previous baseline on snapshot failure. Snapshots now run without holding `inotify_lock` to avoid blocking other watches. Fixes nameless directory events so fsnotify tools can match child names (fixes #55).

- **Bug Fixes**
  - Maintain a per-watch directory snapshot; on NOTE_WRITE, diff to emit named IN_CREATE/IN_DELETE; suppress redundant bare directory events.
  - Advance the baseline only on a successful snapshot; keep the old baseline on failure; the initial add-watch snapshot is best-effort.
  - Take snapshots with the lock released and re-validate the instance/watch on return (return EBADF if torn down); use one event translator for blocking and non-blocking paths; on overflow, queue IN_Q_OVERFLOW; add a regression test asserting a named IN_CREATE is delivered.

<sup>Written for commit 9c28c569f01b5b0ed94c96f7db9fa92132166f86. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/58?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/inotify.c
+++ b/src/syscall/inotify.c
@@ -28,6 +28,8 @@
 #include <errno.h>
 #include <limits.h>
 #include <pthread.h>
+#include <dirent.h>
+#include <stdlib.h>
 #include <sys/event.h>
 #include <sys/stat.h>
 #include <sys/uio.h>
@@ -82,6 +84,11 @@ typedef struct {
     bool is_dir;   /* true if watching a directory */
     dev_t dev;     /* Device ID (for re-add lookup by inode) */
     ino_t ino;     /* Inode number (for re-add lookup by inode) */
+    /* Dir watches only: path + entry-name snapshot, diffed on change to
+     * recover the child name kqueue omits. NULL/0 for file watches. */
+    char *path;
+    char **entries;
+    int n_entries;
 } inotify_watch_t;
 
 typedef struct {
@@ -295,12 +302,136 @@ static void pipe_drain(inotify_instance_t *inst)
         ;
 }
 
+/* Snapshot the entry names of a directory (excluding "." and ".."). On return
+ * *out is a malloc'd array of malloc'd strings with *n_out entries (free with
+ * free_dir_snapshot). On any failure the snapshot is left empty.
+ */
+static void dir_snapshot(const char *path, char ***out, int *n_out)
+{
+    *out = NULL;
+    *n_out = 0;
+
+    DIR *d = opendir(path);
+    if (!d)
+        return;
+
+    char **names = NULL;
+    int n = 0, cap = 0;
+    struct dirent *de;
+    while ((de = readdir(d)) != NULL) {
+        if (!strcmp(de->d_name, ".") || !strcmp(de->d_name, ".."))
+            continue;
+        if (n == cap) {
+            int ncap = cap ? cap * 2 : 16;
+            char **tmp = realloc(names, (size_t) ncap * sizeof(char *));
+            if (!tmp)
+                break;
+            names = tmp;
+            cap = ncap;
+        }
+        names[n] = strdup(de->d_name);
+        if (!names[n])
+            break;
+        n++;
+    }
+    closedir(d);
+
+    *out = names;
+    *n_out = n;
+}
+
+static void free_dir_snapshot(char **entries, int n)
+{
+    if (!entries)
+        return;
+    for (int i = 0; i < n; i++)
+        free(entries[i]);
+    free(entries);
+}
+
+static bool snapshot_contains(char *const *entries, int n, const char *name)
+{
+    for (int i = 0; i < n; i++)
+        if (!strcmp(entries[i], name))
+            return true;
+    return false;
+}
+
 /* Collect events from kqueue. */
 
-/* Poll the kqueue for pending vnode events and translate them into inotify
- * events in the instance buffer.
- *
- * Returns the number of events collected.
+/* Translate one EVFILT_VNODE notification into queued inotify events for the
+ * watch on host_fd. Returns the number queued, or -1 on buffer overflow (an
+ * IN_Q_OVERFLOW marker is queued). Caller holds inotify_lock.
+ */
+static int process_vnode_event(inotify_instance_t *inst,
+                               int host_fd,
+                               uint32_t fflags)
+{
+    int widx = watch_find_by_hostfd(inst, host_fd);
+    if (widx < 0)
+        return 0;
+
+    inotify_watch_t *w = &inst->watches[widx];
+    int queued = 0;
+    bool overflow = false;
+
+    if (w->is_dir && (fflags & NOTE_WRITE) && w->path) {
+        char **now = NULL;
+        int now_n = 0;
+        dir_snapshot(w->path, &now, &now_n);
+
+        for (int j = 0; j < now_n && !overflow; j++) {
+            if ((w->mask & IN_CREATE) &&
+                !snapshot_contains(w->entries, w->n_entries, now[j])) {
+                if (queue_event(inst, w->wd, IN_CREATE, 0, now[j]) < 0)
+                    overflow = true;
+                else
+                    queued++;
+            }
+        }
+        for (int j = 0; j < w->n_entries && !overflow; j++) {
+            if ((w->mask & IN_DELETE) &&
+                !snapshot_contains(now, now_n, w->entries[j])) {
+                if (queue_event(inst, w->wd, IN_DELETE, 0, w->entries[j]) < 0)
+                    overflow = true;
+                else
+                    queued++;
+            }
+        }
+
+        /* Advance the snapshot regardless: the directory state has moved on,
+         * and any names dropped under overflow are covered by IN_Q_OVERFLOW.
+         */
+        free_dir_snapshot(w->entries, w->n_entries);
+        w->entries = now;
+        w->n_entries = now_n;
+    }
+
+    if (!overflow) {
+        uint32_t in_mask = notes_to_in_mask(fflags, w->mask, w->is_dir);
+        /* The per-child create/delete is emitted by the diff above; only emit
+         * the bare-mask event for file watches or non-create/delete changes.
+         */
+        if (in_mask != 0 &&
+            !(w->is_dir && (in_mask & (IN_CREATE | IN_DELETE)))) {
+            if (queue_event(inst, w->wd, in_mask, 0, NULL) < 0)
+                overflow = true;
+            else
+                queued++;
+        }
+    }
+
+    if (overflow) {
+        /* IN_Q_OVERFLOW (0x4000) uses wd=-1 per Linux semantics. */
+        queue_event(inst, -1, 0x4000, 0, NULL);
+        return -1;
+    }
+    return queued;
+}
+
+/* Poll the kqueue for pending vnode events and translate them into
+ * inotify events in the instance buffer. Returns the number of
+ * events collected.
  */
 static int collect_events(inotify_instance_t *inst)
 {
@@ -312,35 +443,19 @@ static int collect_events(inotify_instance_t *inst)
         return 0;
 
     int collected = 0;
+    bool overflow = false;
     for (int i = 0; i < nev; i++) {
-        int host_fd = (int) kevs[i].ident;
-        int widx = watch_find_by_hostfd(inst, host_fd);
-        if (widx < 0)
-            continue;
-
-        inotify_watch_t *w = &inst->watches[widx];
-        uint32_t in_mask =
-            notes_to_in_mask((uint32_t) kevs[i].fflags, w->mask, w->is_dir);
-        if (in_mask == 0)
-            continue;
-
-        /* Queue event without a filename for file watches. For directory
-         * watches, inotify emulation also omits the filename since kqueue
-         * EVFILT_VNODE does not report which child changed.
-         */
-        if (queue_event(inst, w->wd, in_mask, 0, NULL) == 0) {
-            collected++;
-        } else {
-            /* Fixed inotify queue is full; queue IN_Q_OVERFLOW and stop.
-             * IN_Q_OVERFLOW (0x4000) uses wd=-1 per Linux semantics.
-             */
-            queue_event(inst, -1, 0x4000, 0, NULL);
+        int r = process_vnode_event(inst, (int) kevs[i].ident,
+                                    (uint32_t) kevs[i].fflags);
+        if (r < 0) {
+            overflow = true;
             break;
         }
+        collected += r;
     }
 
     /* Signal the self-pipe so poll/epoll sees readability */
-    if (collected > 0)
+    if (collected > 0 || overflow)
         pipe_signal(inst);
 
     return collected;
@@ -438,12 +553,27 @@ int64_t sys_inotify_add_watch(guest_t *g,
     /* Strip IN_MASK_ADD control flag before storing */
     uint32_t event_mask = mask & ~(uint32_t) IN_MASK_ADD;
 
+    /* For directory watches, snapshot the path + current entries up-front
+     * (outside the lock) so collect_events can diff on each change to emit
+     * named IN_CREATE/IN_DELETE. Ownership moves to the watch slot on success;
+     * every early-exit path below frees these.
+     */
+    char *wpath = NULL;
+    char **wentries = NULL;
+    int wn = 0;
+    if (is_dir) {
+        wpath = strdup(path);
+        dir_snapshot(path, &wentries, &wn);
+    }
+
     pthread_mutex_lock(&inotify_lock);
 
     int slot = inotify_find(inotify_fd);
     if (slot < 0) {
         pthread_mutex_unlock(&inotify_lock);
         close(host_fd);
+        free_dir_snapshot(wentries, wn);
+        free(wpath);
         return -LINUX_EBADF;
     }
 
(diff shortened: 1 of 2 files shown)


src/syscall/inotify.c
```diff
@@ -379,33 +402,38 @@ static int process_vnode_event(inotify_instance_t *inst,
...
-                !snapshot_contains(w->entries, w->n_entries, now[j])) {
-                if (queue_event(inst, w->wd, IN_CREATE, 0, now[j]) < 0)
-                    overflow = true;
-                else
-                    queued++;
+        /* Only diff against -- and advance to -- a snapshot that succeeded.
+         * On failure keep the previous baseline; the next successful snapshot
+         * reconciles whatever changed in between.
+         */
+        if (dir_snapshot(w->path, &now, &now_n)) {
```
Max042004: Stale `w->path` reads the wrong directory after rename

The snapshot/diff design rests on `w->path` (captured by `strdup(path)` at `inotify_add_watch` time, `src/syscall/inotify.c:594`) being a valid handle to the directory whose entries the diff should compare. The watch's *identity*, however, is `w->host_fd` — an `O_EVTONLY` fd that follows the inode across renames. Once the two diverge, `dir_snapshot(w->path, ...)` at `src/syscall/inotify.c:409` no longer reads the directory the watch is registered for.

I reproduced both failure modes with two small static aarch64 binaries cross-built and run against `build/elfuse` at this PR's tip (`7fd5187`).

### Scenario A — old path no longer exists

```
mkdtemp /tmp/A
inotify_add_watch(fd, /tmp/A, IN_CREATE | IN_DELETE)
rename /tmp/A -> /tmp/A-renamed    /* watched inode moves with the fd */
touch /tmp/A-renamed/child         /* in the watched dir */
```

`mv` raises `NOTE_RENAME` on the watched fd (delivered as `IN_MOVE_SELF`, correct). The subsequent `touch` raises `NOTE_WRITE` on the same fd; `process_vnode_event` calls `dir_snapshot("/tmp/A", ...)`, `opendir` returns `ENOENT`, the function returns `false`, the keep-baseline branch fires — and **no** `IN_CREATE` for `child` is ever emitted.

Actual output:
```
watching /tmp/elfuse-renA-KRUJg3 (wd=1)
renamed /tmp/elfuse-renA-KRUJg3 -> /tmp/elfuse-renA-KRUJg3-renamed; old path is now nonexistent
created /tmp/elfuse-renA-KRUJg3-renamed/child in the renamed (watched) dir

IN_CREATE name="child" count: 0 (expected 1 on Linux)
verdict: BUG REPRODUCED (named event lost after rename)
```

From the caller's perspective the watch silently stops naming children after a rename — exactly the failure mode #55 was opened to fix.

### Scenario B — old path is taken by a different inode

```
mkdtemp /tmp/A
inotify_add_watch(fd, /tmp/A, IN_CREATE | IN_DELETE)
rename /tmp/A -> /tmp/A-renamed    /* watched inode moves with the fd */
mkdir /tmp/A                        /* fresh dir, different inode, unwatched */
touch /tmp/A/decoy                  /* in the new unwatched dir */
touch /tmp/A-renamed/child          /* in the watched dir */
```

Now `opendir("/tmp/A")` succeeds, but it reads the **new** `/tmp/A`, not the directory the watch tracks. The diff compares the new directory's entries (`["decoy"]`) to the baseline (`[]`) and synthesises an `IN_CREATE` for the decoy. The real `child` event in the renamed directory never surfaces because the watched dir is never sampled.

Actual output:
```
step 1: watching /tmp/elfuse-rename-A-nydrPl (wd=1)
step 2: rename /tmp/elfuse-rename-A-nydrPl -> /tmp/elfuse-rename-A-nydrPl-renamed (watched inode moves with the fd)
step 3: mkdir /tmp/elfuse-rename-A-nydrPl again (NEW inode at old path -- unwatched)
step 4a: created /tmp/elfuse-rename-A-nydrPl/decoy in the NEW (unwatched) dir
step 4b: created /tmp/elfuse-rename-A-nydrPl-renamed/child in the WATCHED (renamed) dir
step 5: draining events for ~1s
  event: wd=1 mask=0x00000100 (IN_CREATE) len=8 name=decoy

summary:
  total IN_CREATE: 1
  IN_CREATE name="decoy" (BUG if >0): 1
  IN_CREATE name="child" (expected: 1): 0
verdict: BUG REPRODUCED
```

The emitted event is **objectively wrong on both axes**: it names a file in a directory the caller never watched, and it suppresses the named event for the actual child of the watched directory. This isn't theoretical — build systems (yarn/pnpm/Cargo) and test runners routinely `mv build → build.old; mkdir build; …` for atomic swap; the watch on the old `build/` will start fabricating events for the unrelated replacement directory after the swap.

### Reproducers

Both binaries are small static aarch64 ELFs cross-built with `aarch64-linux-gnu-gcc -static -O2 -D_GNU_SOURCE`. Sources attached as Scenario A / Scenario B above; they exit non-zero on the buggy path. A correct fix should make both exit zero (Scenario A: emit `IN_CREATE name="child"`; Scenario B: emit `IN_CREATE name="child"` and **not** `name="decoy"`).

jserv (changes requested):

jserv:

chenhunghan:
Rebased onto the latest `main` and resolved the `inotify.c` conflict — it was a comment-block overlap right before `collect_events`; the new `process_vnode_event` helper and main's `fd_publish_linux_flags` change are both preserved. `test-inotify` passes 6/6 and CI is green.

While here I also addressed the lock-contention review finding in a separate commit (9c28c56): the directory `opendir`/`readdir` snapshot now runs with `inotify_lock` released, re-acquiring and re-validating the instance (by `guest_fd`) and watch (by `host_fd`+`dev`/`ino`) before applying the diff. No change to emitted events. PTAL — thanks!

Max042004:
cubic-dev-ai flagged on PR #58:

> **P1:** Directory diffs are path-based, so a rename makes named
> `IN_CREATE`/`IN_DELETE` report against the stale pathname instead of the
> watched inode.

This is correct, and it currently has **no test coverage**. The named-event
feature (test-inotify Test 6) only exercises the case where the watched
directory keeps its original path, so the path/inode divergence never shows up
in CI.

**Root cause**

A directory watch has two identities that are captured once at
`inotify_add_watch` time and never reconciled afterwards:

- **Watch identity** — `host_fd` (an `O_EVTONLY` fd) plus `dev`/`ino`. The fd
  follows the inode across renames, so `EVFILT_VNODE`/`NOTE_WRITE` keeps firing
  correctly after a `mv`.
- **Diff source** — `w->path`, a string. `process_vnode_event` recovers the
  child name by re-listing this path:

  ```c
  /* src/syscall/inotify.c */
  char *path = strdup(w->path);
  ...
  snap_ok = dir_snapshot(path, &now, &now_n);   /* opendir by PATH, not vnode */
  ```

  (`O_EVTONLY` fds cannot be used for `readdir`, which is why the code re-opens
  by path rather than via the watched fd.)

Once the watched directory is renamed, `w->path` is stale while the watch fd
still tracks the moved inode, and the two diverge.

### Proposed regression tests

Two tests added to `tests/test-inotify.c` (after the existing Test 6), plus a
shared drain helper. They encode the Linux-correct expectation, so they pass
once the diff source is bound back to the watched inode and fail today.

```c
/* Read inotify events for ~1s, recording whether a named IN_CREATE for the
 * (optional) want and forbid names was observed. Drains the whole window so a
 * "must NOT appear" assertion is meaningful. */
static void watch_creates(int fd,
                          const char *want, bool *saw_want,
                          const char *forbid, bool *saw_forbid)
{
    if (saw_want)   *saw_want = false;
    if (saw_forbid) *saw_forbid = false;
    for (int attempt = 0; attempt < 50; attempt++) {
        char buf[2048];
        ssize_t n = read(fd, buf, sizeof(buf));
        for (ssize_t off = 0;
             n > 0 && off + (ssize_t) sizeof(struct inotify_event) <= n;) {
            struct inotify_event *ev = (struct inotify_event *) (buf + off);
            if ((ev->mask & IN_CREATE) && ev->len > 0) {
                if (want && saw_want && strcmp(ev->name, want) == 0)
                    *saw_want = true;
                if (forbid && saw_forbid && strcmp(ev->name, forbid) == 0)
                    *saw_forbid = true;
            }
            off += (ssize_t) sizeof(struct inotify_event) + ev->len;
        }
        usleep(20000);
    }
}

/* Test 7 : a child created in the renamed (still-watched) directory
 * must still produce a named IN_CREATE; today the stale-path snapshot fails and
 * the event is silently lost. */
static void test_dir_rename_oldpath_gone(void)
{
    TEST("named IN_CREATE survives dir rename (old path gone)");
    char dir[] = "/tmp/elfuse-inotify-renA-XXXXXX";
    if (!mkdtemp(dir)) { FAIL("mkdtemp"); return; }
    char renamed[300];
    snprintf(renamed, sizeof(renamed), "%s-renamed", dir);

    int fd = inotify_init1(IN_NONBLOCK);
    if (fd < 0) { FAIL("inotify_init1"); rmdir(dir); return; }
    int wd = inotify_add_watch(fd, dir, IN_CREATE | IN_DELETE);
    if (wd < 0) { FAIL("inotify_add_watch"); close(fd); rmdir(dir); return; }

    if (rename(dir, renamed) != 0) {
        FAIL("rename watched dir");
        inotify_rm_watch(fd, wd); close(fd); rmdir(dir); return;
    }
    const char *child = "child";
    char path[400];
    snprintf(path, sizeof(path), "%s/%s", renamed, child);
    int tfd = open(path, O_WRONLY | O_CREAT | O_EXCL, 0644);
    if (tfd < 0) {
        FAIL("create child in renamed dir");
        inotify_rm_watch(fd, wd); close(fd); rmdir(renamed); return;
    }
    close(tfd);

    bool saw_child = false;
    watch_creates(fd, child, &saw_child, NULL, NULL);
    EXPECT_TRUE(saw_child,
                "named IN_CREATE for child lost after rename (old path gone)");

    unlink(path); inotify_rm_watch(fd, wd); close(fd); rmdir(renamed);
}

/* Test 8 : after the atomic-swap pattern, the watched dir's child
 * must be named and the replacement dir's file must NOT appear. This is also
 * the case the dev/ino re-validation cannot catch -- the watch identity never
 * changed, only the path -> vnode binding (an ABA). */
static void test_dir_rename_oldpath_reused(void)
{
    TEST("dir rename: no fabricated event from replacement dir");
    char dir[] = "/tmp/elfuse-inotify-renB-XXXXXX";
    if (!mkdtemp(dir)) { FAIL("mkdtemp"); return; }
    char renamed[300];
    snprintf(renamed, sizeof(renamed), "%s-renamed", dir);

    int fd = inotify_init1(IN_NONBLOCK);
    if (fd < 0) { FAIL("inotify_init1"); rmdir(dir); return; }
    int wd = inotify_add_watch(fd, dir, IN_CREATE | IN_DELETE);
    if (wd < 0) { FAIL("inotify_add_watch"); close(fd); rmdir(dir); return; }

    if (rename(dir, renamed) != 0) {
        FAIL("rename watched dir");
        inotify_rm_watch(fd, wd); close(fd); rmdir(dir); return;
    }
    if (mkdir(dir, 0755) != 0) {
        FAIL("mkdir replacement dir");
        inotify_rm_watch(fd, wd); close(fd); rmdir(renamed); return;
    }

    const char *decoy = "decoy", *child = "child";
    char decoy_path[400], child_path[400];
    snprintf(decoy_path, sizeof(decoy_path), "%s/%s", dir, decoy);
    snprintf(child_path, sizeof(child_path), "%s/%s", renamed, child);
    int dfd = open(decoy_path, O_WRONLY | O_CREAT | O_EXCL, 0644);
    if (dfd >= 0) close(dfd);
    int cfd = open(child_path, O_WRONLY | O_CREAT | O_EXCL, 0644);
    if (cfd >= 0) close(cfd);
    if (dfd < 0 || cfd < 0) {
        FAIL("create decoy/child");
        unlink(decoy_path); unlink(child_path);
        inotify_rm_watch(fd, wd); close(fd); rmdir(dir); rmdir(renamed); return;
    }

    bool saw_child = false, saw_decoy = false;
    watch_creates(fd, child, &saw_child, decoy, &saw_decoy);
    if (!(saw_child && !saw_decoy))
        printf("[child=%d decoy=%d] ", saw_child, saw_decoy);
    EXPECT_TRUE(saw_child && !saw_decoy,
                saw_decoy ? "fabricated IN_CREATE for replacement-dir file"
                          : "named IN_CREATE for watched child lost");

    unlink(decoy_path); unlink(child_path);
    inotify_rm_watch(fd, wd); close(fd); rmdir(dir); rmdir(renamed);
}
```

Register both in `main()`:

```c
test_dir_create_named();
test_dir_rename_oldpath_gone(); 
test_dir_rename_oldpath_reused();
```

### Verified current behavior (RED on this PR's tip)

```
  watch dir delivers named IN_CREATE                  OK     <- feature works in the simple case
  named IN_CREATE survives dir rename (old path gone) FAIL 
  dir rename: no fabricated event from replacement dir [child=0 decoy=1] FAIL
test-inotify: 6 passed, 2 failed - FAIL
```

Test 6 passing alongside the two new failures confirms the failures are the
real bug, not an event-delivery timing artifact.

Max042004 (changes requested):

jserv:
Close in favor of #122

PR #61, opened 2026-05-30
Bring HVF round-trip floor band into the shim

jserv:
This closes the per-call shim path: a diagnostic counter array so every fast-path bail is attributed, inline getpgid(0) / getsid(0) using two new identity slots, and a 256-byte urandom / getrandom inline copy with 4 KiB ring-wrap split. The bulk copy uses ldp/stp with a tbz-driven 1..15 byte tail; the ring lock uses LSE swpal; the slow-path refill runs arc4random_buf outside the lock; the second AT probe is skipped when buf and buf+len-1 share a host page. PGSID_PUBLISH reads (pgid, sid) under session_lock via a new proc_snapshot_pgsid so a concurrent setsid on a sibling vCPU cannot publish a torn pair.

Measured (Apple M1, 100k iter):
```
  read(urandom, 1)    133 -> 100 ns  (-25 %)
  read(urandom, 64)   224 -> 163 ns  (-27 %)
  read(urandom, 256)  528 -> 355 ns  (-33 %)
  getrandom(1)        128 ->  95 ns  (-26 %)
  getrandom(256)      534 -> 366 ns  (-31 %)
```

CB_URANDOM_RING_WRAP is retained for ABI stability but stays at zero now that wrap is served inline; a non-zero reading flags a regression.

---
## Summary by cubic
Moves short getrandom/urandom and getpgid/getsid work into the EL1 shim to avoid HVF round-trips. Adds a stats gate and per-call counters, cutting 1–256 byte RNG calls by ~25–33% on Apple M1.

- New Features
  - Inline fast paths for read(/dev/urandom) and getrandom up to 256 bytes with a wrap-aware 4 KiB ring copy, LSE `swpal` lock, slow-path refill outside the lock, and a skipped second AT probe when the buffer stays on one page. `getrandom` accepts flags 0 and `GRND_NONBLOCK`; others fall back to the host.
  - Inline getpgid(0)/getsid(0) from new pgid/sid slots; values are published under `session_lock` via a snapshot path and republished on bootstrap, fork, exec, setsid, and setpgid.
  - Diagnostic counters for all hit/bail reasons; `ELFUSE_SHIM_STATS` now gates counter increments and prints a dump on exit; helper APIs included.

- Bug Fixes
  - Fixed wrapped ring copies in the inline RNG path by splitting at the 4 KiB boundary; added `tests/test-shim-urandom-wrap.c`.
  - Kept `CB_URANDOM_RING_WRAP` for ABI stability; it remains zero and flags regressions if non-zero.

<sup>Written for commit a00cc3dbb4c12050b1f8438ee4c11e31c714b60f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/61?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src/core/shim.S
```diff
@@ -125,6 +125,57 @@
...
+ *
+ * Non-atomic ldr/add/str. Multi-vCPU concurrent bails may race and lose
+ * a small fraction of counts; acceptable for diagnostic ratios. Each
+ * use expands to 5 instructions: mrs + add + ldr + add + str.
+ *
+ * Both scratch registers are clobbered. svc_restore_eret's
+ * RESTORE_GPRS_KEEP_X0 reloads X1..X30 from the saved frame, so x29/x30
+ * are safe choices on any fast path.
+ */
+.macro COUNTER_INC byte_off, tmp_addr, tmp_val
```
Max042004: `ELFUSE_SHIM_STATS` gates only the dump. The `COUNTER_INC` expansion in `shim.S` reads no such flag. So even with `ELFUSE_SHIM_STATS` unset , every getpid/getuid/gettid still incrementing.

Second cost is that every host vCPU thread issues a non-atomic load-add-store to the same counter line causing cache-line ping-pong.

PR #69, opened 2026-06-04
Add translated embedder syscall hook

doanbaotrung:
Translated x86_64 guests cannot issue the AArch64 HVC instruction used by the existing embedder extension ABI. This prevents those guests from using embedder-provided services such as the graphics bridge.

Add a private elfuse pseudo-syscall as the translated guest counterpart to HVC 6. The syscall keeps the existing hvc6_handler ABI instead of adding a second embedder callback path, so native and translated guests can share the same dispatch logic.

Use syscall number 999 as an internal ABI value between the translated guest shim and elfuse. This is not a Linux syscall number. It is chosen outside the current generic Linux syscall range to avoid colliding with normal guest syscalls, but it is not reserved by Linux and must remain private to elfuse.

If no embedder handler is registered, dispatch falls back to normal syscall handling.

---
## Summary by cubic
Adds a private elfuse pseudo-syscall for translated x86_64 guests to reach the embedder HVC 6 handler, sharing the same dispatch path as native AArch64. This enables embedder services from translated guests without colliding with Linux syscalls.

- **New Features**
  - Defines `ELFUSE_NR_EMBEDDER_HVC6` (default 999) via CFLAGS; compile-time asserts it is outside the Linux `syscall_table` or fails the build.
  - Gated to translated guests only (`g->is_rosetta`) and requires a registered `g->hvc6_handler`; native AArch64 cannot reach it via SVC.
  - On match: read X0 (call id) and X1 (GVA of uint64_t args[8]); return -EFAULT on read failure; else invoke the handler; otherwise fall through to normal syscall handling.

<sup>Written for commit b0c9e521a505b5c78bcf228fc0c43999472132fc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/69?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -6,6 +6,10 @@ BUILD_DIR := build
 ELFUSE_BIN := $(BUILD_DIR)/elfuse
 VERSION ?= $(shell git describe --tags --always --dirty 2>/dev/null || echo "unknown")
 
+# Private pseudo-syscall number used by translated guests to invoke the
+# embedder HVC 6 hook. This is not a Linux syscall number.
+ELFUSE_NR_EMBEDDER_HVC6 ?= 999
+
 # Test binary directory: either pre-built via GUEST_TEST_BINARIES,
 # auto-detected from build/bin, or locally cross-compiled via $(CROSS_COMPILE)gcc.
 ifeq ($(origin GUEST_TEST_BINARIES), undefined)
@@ -51,3 +55,7 @@ CFLAGS := -O2 -Wall -Wextra -Wpedantic \
           -Wshadow -Wstrict-prototypes -Wmissing-prototypes \
           -Wformat=2 -Wimplicit-fallthrough -Wundef \
           -Wnull-dereference -Wno-unused-parameter
+
+ifneq ($(strip $(ELFUSE_NR_EMBEDDER_HVC6)),)
+CFLAGS += -DELFUSE_NR_EMBEDDER_HVC6=$(ELFUSE_NR_EMBEDDER_HVC6)
+endif
--- a/src/syscall/syscall.c
+++ b/src/syscall/syscall.c
@@ -1706,6 +1706,12 @@ static int64_t sc_futex_waitv(guest_t *g,
 #error "SC_TABLE_SIZE must exceed SC_MAX_SYSCALL_NUM"
 #endif
 
+#if defined(ELFUSE_NR_EMBEDDER_HVC6) && (ELFUSE_NR_EMBEDDER_HVC6 + 0 > 0)
+_Static_assert(ELFUSE_NR_EMBEDDER_HVC6 > SC_TABLE_SIZE,
+               "ELFUSE_NR_EMBEDDER_HVC6 must be outside syscall_table");
+#define ELFUSE_ROSETTA_EMBEDDER_SYSCALL 1
+#endif
+
 typedef struct {
     syscall_handler_t handler;
     bool needs_extra_regs;
@@ -1887,6 +1893,37 @@ int syscall_dispatch(hv_vcpu_t vcpu, guest_t *g, int *exit_code, bool verbose)
 
 slow_path:
     entry = RANGE_CHECK(nr, 0, SC_TABLE_SIZE) ? &syscall_table[nr] : NULL;
+
+/*
+ * Private embedder pseudo-syscall for translated guests.
+ *
+ * This is not a Linux syscall number. Translated x86_64 guests cannot
+ * issue native AArch64 HVC instructions, so elfuse uses this private
+ * build-selected syscall number as the translated counterpart to HVC 6.
+ *
+ * This path is gated on g->is_rosetta so native AArch64 guests cannot
+ * reach the embedder hook through SVC.
+ *
+ * Layout:
+ *   x8 = ELFUSE_ROSETTA_EMBEDDER_SYSCALL
+ *   x0 = embedder call number
+ *   x1 = guest virtual address of uint64_t args[8]
+ */
+#ifdef ELFUSE_ROSETTA_EMBEDDER_SYSCALL
+    if (g->is_rosetta && nr == ELFUSE_NR_EMBEDDER_HVC6 && g->hvc6_handler) {
+        uint64_t call_nr = 0;
+        uint64_t args_gva = 0;
+        hv_vcpu_get_reg(vcpu, HV_REG_X0, &call_nr);
+        hv_vcpu_get_reg(vcpu, HV_REG_X1, &args_gva);
+        uint64_t guest_args[8] = {0};
+        if (guest_read_small(g, args_gva, guest_args, sizeof(guest_args)) < 0) {
+            result = -LINUX_EFAULT;
+        } else {
+            result = g->hvc6_handler(call_nr, guest_args, g->hvc6_userdata);
+        }
+        goto fast_done;
+    }
+#endif
 
     hv_vcpu_get_reg(vcpu, HV_REG_X0, &x0);
     hv_vcpu_get_reg(vcpu, HV_REG_X1, &x1);
--- a/src/core/guest.h
+++ b/src/core/guest.h
@@ -512,7 +512,14 @@ typedef struct {
     _Atomic uint64_t pt_gen;
 
     /*
-     * Optional HVC #6 embedder extension hook.
+     * Optional HVC 6 embedder extension hook.
+     *
+     * Native AArch64 guests reach this through HVC 6. When the build enables
+     * ELFUSE_NR_EMBEDDER_HVC6, translated Rosetta guests may reach the same
+     * handler through a private pseudo-syscall gated on g->is_rosetta.
+     *
+     * Handler implementations must treat call_nr and args as untrusted guest
+     * input and allowlist supported call numbers.
      *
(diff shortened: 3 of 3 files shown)


doanbaotrung:
Hi @Max042004 
Please help to review this

jserv (changes requested):
Check https://cbea.ms/git-commit/ carefully and enforce the rules.
You MUST address the motivations and considerations.

src/syscall/syscall.c
```diff
@@ -1818,6 +1818,37 @@ int syscall_dispatch(hv_vcpu_t vcpu, guest_t *g, int *exit_code, bool verbose)
...
     entry = RANGE_CHECK(nr, 0, SC_TABLE_SIZE) ? &syscall_table[nr] : NULL;
 
+    /* Custom Embedder Syscall Hook:
+     * Guests running on translated user-space architectures (e.g., x86_64
+     * via Apple Rosetta 2) execute at EL0 and cannot compile or execute
+     * native AArch64 'hvc' instructions.
+     *
+     * To allow such guests to invoke the embedder extension hooks (like the
+     * graphics bridge), we intercept a custom system call number (999)
+     * as the x86_64 guest counterpart to AArch64 'hvc 6'.
```
jserv: How can `999` be selected? Any reference?

doanbaotrung: You're right. This is not a Linux syscall number.  This is private ABI values used by translated guests to reach elfuse embedder hooks.

doanbaotrung: Remove hardcode 999 and added #define ELFUSE_NR_EMBEDDER_HVC6 999 in abi.h

jserv (changes requested):
Check https://cbea.ms/git-commit/ carefully and enforce the rules for informative git commit messages.
You MUST address the motivations and considerations.

src/syscall/abi.h
```diff
@@ -742,3 +742,10 @@ void syscall_init(void);
  * Returns 0 to continue, 1 to exit.
  */
 int syscall_dispatch(hv_vcpu_t vcpu, guest_t *g, int *exit_code, bool verbose);
+
+/* Private elfuse embedder pseudo-syscalls.
+ *
+ * These are not Linux syscall numbers. They are private ABI values used by
+ * translated guests to reach elfuse embedder hooks.
+ */
+#define ELFUSE_NR_EMBEDDER_HVC6 999
```
jserv: This should be specified by build system instead of being hard-coded here.

doanbaotrung: Good point. I removed the hard-coded value from abi.h and made the pseudo-syscall number a build-time definition instead.

The default is provided through the Makefile as ELFUSE_NR_EMBEDDER_HVC6, and syscall.c only uses the hook when that macro is defined.

jserv (commented):
Hook itself is correct and the memory-safety story checks out (`guest_read_small` honors `MEM_PERM_EL1_ONLY` via `gva_translate_perm`, so a guest pointing X1 into `shim_data` or the PT pool gets `-EFAULT`, not data exfiltration). Three P2s should land before this does, and one open question gates the whole premise.

P2 -- the PR's stated use case is Rosetta-translated x86_64 guests, but it's not verified that Apple's Rosetta forwards an unknown syscall number (999) through to elfuse's HVC #5 dispatch instead of returning `ENOSYS` at the translation layer. Without an empirical demonstration (graphics-bridge round-trip from an actual x86_64 guest), the PR may only work as a native-aarch64 hook.

P2 -- `make ELFUSE_NR_EMBEDDER_HVC6=` breaks the build (empty override defeats `?=`, CFLAGS gets `-DELFUSE_NR_EMBEDDER_HVC6=`, then `if (nr == )` won't compile). See inline.

P2 -- native aarch64 guests can also reach the embedder handler via `svc #0; x8=999`. Intentional per the description ("share the same dispatch logic") but worth either gating on `g->is_rosetta` or documenting so embedders validate `call_nr` against an allowlist.

P2 -- no test. Hook lands on the hot path of every syscall dispatch; the success, EFAULT, and absent-handler paths all need coverage.

P3 (procedural) -- branch is 3 commits behind main (`a00cc3d`, `9aa54a3`, `7acfe17` all touch `src/syscall/syscall.c`); rebase before merge so the conflict review is honest.

mk/config.mk
```diff
@@ -6,6 +6,10 @@ BUILD_DIR := build
 ELFUSE_BIN := $(BUILD_DIR)/elfuse
 VERSION ?= $(shell git describe --tags --always --dirty 2>/dev/null || echo "unknown")
 
+# Private pseudo-syscall number used by translated guests to invoke the
+# embedder HVC 6 hook. This is not a Linux syscall number.
+ELFUSE_NR_EMBEDDER_HVC6 ?= 999
```
jserv, line 11: `make ELFUSE_NR_EMBEDDER_HVC6=` lets an empty value override `?=`, then CFLAGS picks up `-DELFUSE_NR_EMBEDDER_HVC6=` and the C hook becomes `if (nr == )` -- syntax error. Either skip the `-D` when the value is empty, or guard the C side with a numeric check:

```c
#if defined(ELFUSE_NR_EMBEDDER_HVC6) && (ELFUSE_NR_EMBEDDER_HVC6 + 0 > 0)
```

src/syscall/syscall.c
```diff
@@ -1818,6 +1818,40 @@ int syscall_dispatch(hv_vcpu_t vcpu, guest_t *g, int *exit_code, bool verbose)
...
+ * (ELFUSE_NR_EMBEDDER_HVC6) as the x86_64 guest counterpart to AArch64
+ * 'hvc 6'.
+ *
+ * Syscall layout:
+ *   nr = ELFUSE_NR_EMBEDDER_HVC6
+ *   X0 = call number
+ *   X1 = guest virtual address of the uint64_t args[8] array
+ */
+#ifdef ELFUSE_NR_EMBEDDER_HVC6
+    if (nr == ELFUSE_NR_EMBEDDER_HVC6) {
```
jserv: Four concerns on this hook:

1. No `g->is_rosetta` gate. A native aarch64 EL0 program issuing `svc #0; x8=999` reaches `g->hvc6_handler`, bypassing the HVC #6 trusted-shim path at `src/syscall/proc.c:1825`. If embedders (e.g., a graphics bridge) implicitly trust their caller, this blurs the privilege boundary. Either gate on `g->is_rosetta`, or document the contract at `src/core/guest.h:502` so embedders allowlist `call_nr`.

2. The Rosetta translation premise is unverified. Apple's Rosetta-for-Linux is closed-source; it may translate via a per-syscall table and ENOSYS an unknown nr like 999 at the translator before the SVC ever reaches elfuse. A working graphics-bridge demo from a real x86_64 guest would close this.

3. Future Linux generic syscall range collision (today ~462). Add `_Static_assert(ELFUSE_NR_EMBEDDER_HVC6 > SC_TABLE_SIZE);` here so a future syscall-table growth fails the build instead of silently shadowing the embedder.

4. No test. The hook sits on the syscall_dispatch hot path; the next refactor can break the embedder ABI silently. A ~50-line native-aarch64 test that registers a stub `hvc6_handler`, issues `syscall(999, 42, &args)`, asserts `call_nr` and `args[]`, and covers EFAULT + NULL-handler fallback would lock it in.

Style nits: leading comment block at column 0 should be 4-space indented to match function body; "we intercept" should be third-person per spec.md ("elfuse intercepts" or imperative).

doanbaotrung: With the Rosetta gate added, a native AArch64 syscall(999, ...) test should
verify that the handler is not called and the syscall falls back to ENOSYS.

The positive-path test should use a Rosetta/x86_64 guest or a native dispatch
harness that explicitly sets g->is_rosetta=true.

mk/config.mk
doanbaotrung: Make sure make ELFUSE_NR_EMBEDDER_HVC6=  will disable this variable

doanbaotrung: Hi @jserv ,

Please check if I added block code below can prevent the issue you said.
```
ifneq ($(strip $(ELFUSE_NR_EMBEDDER_HVC6)),)
CFLAGS += -DELFUSE_NR_EMBEDDER_HVC6=$(ELFUSE_NR_EMBEDDER_HVC6)
endif
```

Max042004:
Wrap the commit message body at 72 characters

doanbaotrung:
Dear @Max042004 ,

I ran `cmake indent` which will follow `.clang-format` already. I think we missed `ColumnLimit: 72` in `.clang-format`, but I cannot update it because it may impact to other files when run `cmake indent`

Thanks,
Trung

Max042004:
You can use git amend in this PR to wrap the commit message body in 72 characters.

doanbaotrung:
Dear @Max042004 ,

Fixed.

Thanks,
Trung

Max042004 (approved):

jserv:

PR #71, opened 2026-06-05
Short-circuit mprotect on matching tracker prot

jserv:
Repeat mprotect with the same prot on RELRO, JIT, and GC-style ranges no longer walks page tables: a pre-check confirms every overlapping region already records the requested prot and is not MAP_NORESERVE, in which case the call returns 0 without touching the tracker or PTEs. The skip is gated on a per-request safety helper that forces the slow path for pure PROT_READ, because sys_mmap installs MEM_PERM_RW PTEs for non-exec mappings (including PROT_READ requests) and only the explicit guest_update_perms call inside mprotect tightens them to MEM_PERM_R. tests/test-negative pins this behavior with a regression test that mprotects PROT_READ onto a PROT_READ mmap and confirms a subsequent write traps SIGSEGV.

The fast path runs before any mutation. When PTE work is required, it runs BEFORE guest_region_set_prot so a -ENOMEM from guest_{update_perms, extend_page_tables,invalidate_ptes} leaves the tracker at the OLD prot; the next retry sees the mismatch and re-attempts instead of silently no-op'ing on stale tracker state. The low-IPA branch was also missing return-value checks on guest_update_perms and guest_invalidate_ptes; both now propagate -LINUX_ENOMEM.

guest_region_set_prot's two table-full split paths now set a sticky guest_t.regions_tracker_stale flag. The fast path checks this flag and falls back to unconditional PTE work for the lifetime of the process once the tracker has lied even once. The flag is not propagated through fork IPC v10, so a child that inherits a stale tracker re-arms the flag the next time it hits the same condition; the only window of incorrect skipping is the very first matching mprotect after such a fork.

guest_region_remove is restructured as a single in/out compaction pass. The interior-split branch snapshots the source region, shifts the suffix to make room, writes left and right halves, and returns. The previous in-place layout aliased the source slot when out == in: it overwrote *r with the left half and then read the right half from that clobbered slot, corrupting both halves on the only growth path. The trim-only paths consolidate to a single survivor block.

guest_region_add_ex_owned_gpa replaces the O(n) bubble-insert with a binary search plus single memmove. The two query helpers guest_region_range_prot_uniform and guest_region_range_has_noreserve use the same binary-search prefix-skip.

find_free_gap_inner stops iterating once a region's start crosses max_addr; later regions cannot affect a candidate gap inside the search window.

---
## Summary by cubic
Short-circuits same-prot mprotect to skip page-table work when safe, and hardens region tracking to avoid stale or incorrect skips. This speeds up RELRO/JIT/GC workloads and fixes a region-split corruption bug.

- **New Features**
  - Add safe fast path for mprotect when all overlapping regions already have the target prot and none are MAP_NORESERVE; skips tracker and PTE work. Forces slow path for pure PROT_READ. Gated by a sticky regions_tracker_stale flag.
  - Propagate regions_tracker_stale across fork (IPC v11); the child also arms it if its received region table is truncated.
  - Do PTE changes before tracker updates and return -LINUX_ENOMEM on failures, preventing silent no-ops on retries.

- **Refactors**
  - Rewrite guest_region_remove as a single in/out compaction pass to fix interior-split corruption and simplify trim-only paths; mark regions_tracker_stale if a table-full split drops a tail.
  - Speed up region ops: binary-search insert in guest_region_add_ex_owned_gpa, new guest_region_first_end_above, guest_region_range_prot_uniform, guest_region_range_has_noreserve, and early-exit in gap search.
  - Add regression test: mprotect(PROT_READ) onto PROT_READ mapping keeps it read-only; a write traps SIGSEGV.

<sup>Written for commit 41d43fd79d967c01118ef561550be68214b39642. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/71?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/core/guest.c
+++ b/src/core/guest.c
@@ -1737,6 +1737,40 @@ static bool regions_mergeable(const guest_region_t *a, const guest_region_t *b)
     return a->offset + (a->end - a->start) == b->offset;
 }
 
+/* First region whose start is >= start. regions[] is sorted by start. */
+static int region_lower_bound_start(const guest_t *g, uint64_t start)
+{
+    int lo = 0;
+    int hi = g->nregions;
+
+    while (lo < hi) {
+        int mid = lo + (hi - lo) / 2;
+        if (g->regions[mid].start < start)
+            lo = mid + 1;
+        else
+            hi = mid;
+    }
+    return lo;
+}
+
+/* First region whose end is > addr. See guest.h for the contract; also used
+ * inside this file to skip the untouched prefix for remove and set_prot.
+ */
+int guest_region_first_end_above(const guest_t *g, uint64_t addr)
+{
+    int lo = 0;
+    int hi = g->nregions;
+
+    while (lo < hi) {
+        int mid = lo + (hi - lo) / 2;
+        if (g->regions[mid].end <= addr)
+            lo = mid + 1;
+        else
+            hi = mid;
+    }
+    return lo;
+}
+
 /* Merge region at index i with its right neighbor (i+1) when their layouts
  * agree. No-op if i is the last region or layouts differ.
  */
@@ -1857,12 +1891,10 @@ int guest_region_add_ex_owned_gpa(guest_t *g,
         return -1;
     }
 
-    /* Find insertion point (keep sorted by start address) */
-    int i = g->nregions;
-    while (i > 0 && g->regions[i - 1].start > start) {
-        g->regions[i] = g->regions[i - 1];
-        i--;
-    }
+    /* Find insertion point (keep sorted by start address). */
+    int i = region_lower_bound_start(g, start);
+    memmove(&g->regions[i + 1], &g->regions[i],
+            (g->nregions - i) * sizeof(guest_region_t));
 
     guest_region_t *r = &g->regions[i];
     r->start = start;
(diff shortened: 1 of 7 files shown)


Max042004:
The `guest_region_remove` rewrite looks correct and is cleaner — one nit on the commit message's description of the **old** behavior, unless it's referring to a local draft not in the history.

The message says:

> The previous in-place layout aliased the source slot when out == in: it overwrote `*r` with the left half and then read the right half from that clobbered slot.

But in the parent (`bde0b37`), the interior-split branch writes the **right** half first into a *separate* slot `i+1`, copied from an intact `*r` at slot `i`, and only shortens the left half afterward:

```c
memmove(&g->regions[i + 2], &g->regions[i + 1], ...); /* shift suffix */
guest_region_t *right = &g->regions[i + 1];
*right = *r;            /* copied from the still-intact slot i */
right->start = end; ...
r->end = start;         /* left half shortened AFTER right is read */
```

So `r` (slot `i`) and `right` (slot `i+1`) are distinct — no aliasing, and the right half is read before `*r` is touched. The `out == in` framing comes from the new cursor-based pass, which the parent didn't have. Is the message describing an earlier draft, or could it be reworded to match the committed parent? Either way the new single-pass version is a clear improvement — just flagging the narrative since it outlives the PR in `git log`.

PR #75, opened 2026-06-05
Grow the page-table pool for multi-isolate guests

### LINKED ISSUE #74: Guest page-table pool exhausts under multiple V8 isolates, Node `worker_threads`/`cluster` hard-abort
## Summary

The guest page-table pool is a fixed **960 KiB** arena carved out of the 4 MiB
infrastructure reserve (`src/core/guest.h:53-54`). It is filled by a **bump
allocator that is never reclaimed** within a process — `g->pt_pool_next` only
advances and is reset solely at guest init / reset / fork
(`src/core/guest.c:358,470,1590`, `src/runtime/forkipc.c`). Because each 2 MiB
guest block that needs 4 KiB granularity costs one 4 KiB L3 page from this pool
(`guest_split_block` → `guest_alloc_pt_page`, `src/core/guest.c:166-194`), the
pool budgets roughly **480 MiB of distinct mapped address space over the whole
process lifetime** (240 pages × 2 MiB), regardless of how much is later
`munmap`'d.

A single-isolate Node process stays under that ceiling and runs fine. But every
additional V8 isolate (`worker_threads`, `cluster`) reserves and commits its own
multi-hundred-MiB regions, and the cumulative, non-reclaimed consumption crosses
the 240-page budget. When `guest_alloc_pt_page` returns 0 the guest mmap fails,
V8's allocator then trips on a follow-up `munmap` that returns non-zero, and the
process **hard-aborts**:

```
src/core/guest.c:178: guest: page table pool exhausted (used 983040 / 983040 bytes)

# Fatal error in , line 0
# Check failed: 0 == munmap(address, size).
```

(`983040 == 0xF0000 == 960 KiB`, i.e. the pool is 100% consumed.)

This surfaced while testing advanced Node workloads on `oci run node:alpine`
(node v26.3.0) — `oci run` is the OCI image support from PR #34 — on top of the
`oci-image-rebase` branch with PR #73 applied. Single-threaded, event-loop web
servers (Express, raw `http`) are unaffected; the failure is specific to
spinning up **multiple live V8 isolates**.

## Reproduction

`node:alpine` already pulled. `--entrypoint node` is required because the image
entrypoint is `docker-entrypoint.sh` (a shell script, not an ELF).

### Crashes — concurrent worker isolates after normal allocator traffic

The advanced built-in battery below (crypto / zlib / fs / streams / net, then
two sequential workers, then **four concurrent workers**) deterministically
exhausts the pool at the concurrent-worker stage:

```sh
./build/elfuse oci run --entrypoint node node:alpine -e '
const {Worker}=require("worker_threads");
const mk=n=>new Promise((res,rej)=>{
  const w=new Worker(`require("worker_threads").parentPort.postMessage(${n}*${n})`,{eval:true});
  w.on("message",m=>{w.terminate();res(m);}); w.on("error",rej);
});
Promise.all([mk(2),mk(3),mk(4),mk(5)]).then(r=>console.log(r)).catch(e=>console.error(e));
'
# -> guest: page table pool exhausted (used 983040 / 983040 bytes)
# -> # Check failed: 0 == munmap(address, size).
```

### Works — single isolate / single worker

```sh
# plain node, no worker .................. OK (well under 80% warn threshold)
./build/elfuse oci run --entrypoint node node:alpine --version          # v26.3.0

# one worker (2 isolates) ................ OK, prints 500500
./build/elfuse oci run --entrypoint node node:alpine -e '
const {Worker}=require("worker_threads");
new Promise((res,rej)=>{const w=new Worker("let s=0;for(let i=0;i<=1000;i++)s+=i;require(\"worker_threads\").parentPort.postMessage(s)",{eval:true});
w.on("message",m=>{w.terminate();res(m)});w.on("error",rej);}).then(s=>console.log(s));'
```

### Threshold

Measured with minimal workers held live until all are up: **1 and 2 concurrent
worker isolates pass cleanly; at 3 the run no longer completes** (consistent with
the abort hanging guest teardown), and the 4-concurrent-worker battery above is
the deterministic, cleanly-observed crash. The boundary drops further once the
main isolate has already done real work (an HTTP server plus the
crypto/zlib/stream/net battery), because the pool is cumulative and never
reclaimed. A `worker_threads` pool of size `os.cpus().length` (the common
default for CPU-bound work) or any `cluster`-based server will not start.

For contrast, everything that does **not** add isolates passes on the same
binary: HTTP server reachable from the host, `crypto` (sha256/aes-256-cbc/
randomBytes), `zlib` gzip, `fs`, `stream` pipeline, timers, raw `net` TCP, and a
full Express 5.2.1 REST app serving 100 concurrent requests.

## Root cause

1. **Fixed, boxed-in pool.** `INFRA_PT_POOL_OFF 0x10000 .. INFRA_PT_POOL_END_OFF
   0x100000` = 960 KiB, sitting inside the 4 MiB `INFRA_RESERVE`. The **shim
   code slot starts immediately above at `+0x100000`**
   (`INFRA_SHIM_OFF`, `src/core/guest.h:55`), so the pool cannot grow in place
   without relocating the shim/shim-data slots and enlarging the reserve.

2. **Bump allocator, no reclaim.** `guest_alloc_pt_page` (`guest.c:166-194`)
   only ever advances `pt_pool_next`; there is no free path. `sys_munmap` tears
   down regions and invalidates PTEs but does **not** return L3 pages to the
   pool. So a `worker_threads` pool that spawns and joins workers *leaks* PT
   pages on every churn, even at steady-state isolate count.

3. **Per-isolate footprint.** Each V8 isolate maps its own committed heap /
   code / cage regions; one L3 page per touched 2 MiB block. Two isolates'
   footprints already approach the 240-page budget; three crosses it.

4. **Failure mode is an abort, not a clean OOM.** On exhaustion the guest mmap
   fails, but V8's subsequent `munmap` returns non-zero and trips a `CHECK`,
   killing the process instead of surfacing a JS `RangeError`/OOM. The
   post-exhaustion mmap-bookkeeping is left inconsistent enough that the paired
   `munmap` cannot succeed.

## Notes

- The pool already warns at 80% (`guest.c:185-194`,
  `guest: page table pool at N%`), so the runway is observable before the abort.
- Reachable through the OCI `oci run` path added by PR #34 (`oci run node:alpine`),
  but the page-table pool is core guest infrastructure (`src/core/guest.*`) and
  the limit is not specific to OCI — any multi-isolate guest hits it.
- Independent of PR #73 (epoll-mt). #73 is what lets multi-threaded Node get far
  enough to *hit* this; the pool limit is a separate, pre-existing constraint.
- `io_uring_setup` (syscall 425) is unimplemented and warns, but Node falls back
  to the libuv threadpool cleanly — unrelated to this issue.

Max042004:
Fixes #74.

The guest page-table pool was a fixed 960 KiB arena (240 x 4 KiB L3 pages) at the bottom of the 4 MiB infra reserve. It is a bump allocator that never reclaims on munmap, and each 2 MiB block that needs mixed permissions (e.g. V8 JIT W^X) draws one L3 page, so the pool budgeted only ~480 MiB of split address space over a process lifetime. A single guest stays well under that, but every extra V8 isolate a Node worker_threads pool or cluster spins up reserves its own committed regions; the third isolate exhausted the pool, after which even munmap (which must split a block to invalidate a sub-range) failed and V8 hard-aborted on CHECK(0 == munmap) instead of getting a clean ENOMEM.

Grow INFRA_RESERVE 4 MiB -> 16 MiB and tighten the shim code slot from a round 1 MiB to 40 KiB (INFRA_SHIM_SLOT, ~6x the ~7 KiB shim blob) so the freed space falls through to the pool, which becomes ~13.9 MiB (3558 pages, ~7 GiB of split address space) -- enough for an os.cpus()-sized worker pool on a 24-core Ultra. The reserve is demand-paged and sits in the ~4 GiB dead zone below interp_base, so the unused pool costs no host RAM and the larger virtual reserve is free.

The layout keeps every invariant: shim data still occupies the top 2 MiB block (shim_data_base + 2 MiB == interp_base), shim code still shares the PT pool's last 2 MiB block, and the pool still ends where the shim slot begins. Only the constants in core/guest.h change; consumers derive their addresses from g->shim_base / pt_pool_base, and fork copies only the used pool, so nothing else moves.

Because the shim slot is now tight, a _Static_assert in main.c and a runtime re-check in bootstrap.c fail loudly if the shim ever outgrows INFRA_SHIM_SLOT instead of letting it silently overlap the shim-data block.

Surfaced while bringing up node:alpine through the OCI image work, where worker_threads pools of 8 and 16 isolates now run to completion; the fix is in the core guest runtime and is independent of that. The full guest test suite passes, including test-mremap-infra (infra-reserve boundary) and test-mprotect-mt (W^X L3 splitting).

---
## Summary by cubic
Expand the infra reserve to 16 MiB and tighten the shim code slot to 40 KiB to grow the guest page‑table pool to ~13.9 MiB and avoid PT pool exhaustion in multi‑isolate V8 workloads. Adds build‑time invariants and runtime checks for the infra layout and shim size.

- **Bug Fixes**
  - Increased `INFRA_RESERVE` 4 MiB → 16 MiB; PT pool now ~13.9 MiB (3558 pages, ~7 GiB split space).
  - Tightened shim code slot to 40 KiB (`INFRA_SHIM_SLOT`); shim data stays in the top 2 MiB; layout invariants preserved.
  - Added build‑time `_Static_assert`s for layout invariants (pool end == shim base, shim_data in top 2 MiB, 2 MiB/page alignment) and a runtime size check with slot size in logs.
  - Reserve is demand‑paged; unused pool costs no RAM. Tests pass; Node pools with 8–16 isolates run to completion.

<sup>Written for commit 51c9b0ac1885e9a40ef9ea725f9eaf4a253b0c7a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/75?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src/main.c
```diff
@@ -123,6 +123,14 @@ static void cleanup_main_resources(guest_t *g,
 /* Embedded shim binary (generated by xxd -i from shim.bin) */
 #include "shim_blob.h"
 
+/* The shim code slot in the infra reserve is sized tight (INFRA_SHIM_SLOT, a
+ * few x the current blob) so the rest of the reserve goes to the page-table
+ * pool. If the shim ever outgrows the slot it would overlap the shim-data
+ * block; fail the build loudly rather than corrupt memory at boot. Enlarge
+ * INFRA_SHIM_SLOT (and shrink the pool to match) if this fires. */
+_Static_assert(sizeof(shim_bin) <= INFRA_SHIM_SLOT,
```
jserv, line 131: The shim-blob size is asserted, but the layout invariants documented in `guest.h:62-66` are only guarded by comment. A future edit that grows the pool by shifting `INFRA_PT_POOL_END_OFF` without touching `INFRA_SHIM_OFF` would silently overlap the pool and the shim slot. Consider adding next to this one:

```c
_Static_assert(INFRA_PT_POOL_END_OFF == INFRA_SHIM_OFF,
               "PT pool must end exactly where the shim slot begins");
_Static_assert(INFRA_SHIM_DATA_OFF + BLOCK_2MIB == INFRA_RESERVE,
               "shim_data must occupy the top 2 MiB block of the reserve");
_Static_assert((INFRA_SHIM_DATA_OFF & (BLOCK_2MIB - 1)) == 0,
               "shim_data must be 2 MiB-aligned");
_Static_assert((INFRA_PT_POOL_OFF & 0xFFF) == 0 &&
               (INFRA_PT_POOL_END_OFF & 0xFFF) == 0,
               "PT pool offsets must be page-aligned");
```

jserv (commented):
Layout math, fork-IPC compatibility, `finalize_block_perms` behavior, and the icache-invalidate cache-line edge all check out. Two small follow-ups before this lands, both P3.

One: the layout invariants documented in `src/core/guest.h:62-66` (`INFRA_PT_POOL_END_OFF == INFRA_SHIM_OFF`, `INFRA_SHIM_DATA_OFF + BLOCK_2MIB == INFRA_RESERVE`, alignments) aren't compile-time enforced -- see inline comment on `src/main.c:131`.

Two: `src/runtime/forkipc.c:132` still says "plus the 4 MiB infra reserve below it." Functional check (8 GiB lower bound) still passes, but the comment will mislead anyone debugging guest-size limits. One-line update.

Separately (not this PR's job): `docs/internals.md:95-113` carries a pre-existing memory-map table showing a legacy layout with the infra reserve at LOW addresses; the actual code places it at the top of IPA. Worth a separate doc-cleanup PR rather than folding into this change.

jserv:

PR #76, opened 2026-06-06
Fix cross-process signal delivery to EL0-preempted guests

### LINKED ISSUE #67: test-fork intermittently fails with HVF exit reason 0x3 inside make check
`make check` intermittently fails on `test-fork` with this error:

```
test-fork                                     [ FAIL ] (exit 1)
    ERROR src/syscall/proc.c:2041: elfuse: unexpected exit reason 0x3
```

Running it alone via `bash tests/driver.sh -f test-fork` always passes.
Putting `test-exec` ahead of it (`bash tests/driver.sh -f 'test-exec\|test-fork'`) reproduces the failure. The immediately following tests (`test-session`, `test-pidfd`, the other `test-fork-*` ones) all pass. So whatever state HVF doesn't like, it's inherited from the previous test and only the very next VM is affected.

I did 7 runs of `make clean && make check`, controlled for cross-compiler:

| toolchain                            | runs | failures |
|--------------------------------------|------|----------|
| `aarch64-linux-gnu-` (default)       | 3    | 3        |
| `aarch64-unknown-linux-gnu-` (msnse) | 4    | 2        |

Both are Homebrew GCC 15.2.0, so this isn't toolchain-specific.

0x3 is `HV_EXIT_REASON_UNKNOWN`. The Apple SDK header
`<Hypervisor/hv_vcpu_types.h>` is pretty direct about it:

> Unable to determine exit reason: this should not happen under normal
> operation.

So HVF itself is the one saying the VM is in a state it can't classify, this isn't a missing case in elfuse's switch (`proc.c:2041`), it's
HVF refusing to characterize what happened. Which fits the "post-execve teardown left something behind" hypothesis better than
the obvious "add another case to the switch".

Relevant code paths based on `docs/internals.md`:

- `sys_execve` (`src/syscall/exec.c`) reloads the ELF, rebuilds page tables and restarts the vCPU; `signal_reset_for_exec()` runs after `guest_reset` (`docs/internals.md`).
- The CoW fork path notes that `guest_init_from_shm` zeroes the guest struct and `ttbr0` has to be restored explicitly from the IPC header (`docs/internals.md`). If anything else has the same "must be restored" property and a previous execve didn't clear it, the new VM would enter inconsistent.

One concrete improvement regardless of the root cause: on the `UNKNOWN` path at `proc.c:2041`, dump VCPU register state. HVF won't tell you why it bailed, so any context has to come from elfuse side.

Environment: Apple M3 Max MacBook Pro, macOS 26.5.1, elfuse b1ce739, bash 5.x from Homebrew. Cross-compilers both at
`/opt/homebrew/bin/aarch64-{linux-gnu,unknown-linux-gnu}-gcc`.

Max042004:
To fix #67. test-fork's phase-2 signal child spins in `while (!got_usr1) usleep()` waiting for a SIGUSR1 sent cross-process by its parent. The signal was delivered only ~35% of the time and lost the rest, so `make check` hung at test-fork until the 60s per-test timeout -- often longer, as leaked `elfuse --fork-child` orphans kept the driver's stdout pipe open.

Two complementary defects:

1. vcpu_run_loop treated HV_EXIT_REASON_UNKNOWN as fatal. A host SIGUSR2 (the cross-process guest-signal transport) that interrupts hv_vcpu_run mid-execution aborts the run with UNKNOWN rather than the clean CANCELED that hv_vcpus_exit() produces for a vCPU caught between runs. Route UNKNOWN through the same cancellation handling so the already-queued guest signal is delivered instead of crashing the child.

2. signal_deliver redirected to the handler only via ELR_EL1, which takes effect solely on an ERET from EL1 (the syscall-return path, gated by the shim's X8==2 exec_drop_frame marker). When the signal is delivered from the cancellation branch -- i.e. the vCPU was preempted while running EL0 code (cross-process SIGUSR2, or SIGALRM in a tight loop) -- there is no pending ERET, the resume uses HV_REG_PC, and the ELR_EL1 write is a no-op: the handler never runs and only the X0=signum clobber lands, re-running the interrupted nanosleep with a bogus arg and spinning forever. Detect EL0 preemption from the live PSTATE (CPSR M[3:0]==0), save the interrupted PC from HV_REG_PC instead of the stale ELR_EL1, and redirect HV_REG_PC/CPSR directly; skip the X8==2 marker since there is no shim frame to drop.

test-fork now passes 20/20 (was ~7/20); `make check` is green with no hang.

### DIFF
--- a/src/syscall/proc.c
+++ b/src/syscall/proc.c
@@ -1936,11 +1936,31 @@ int vcpu_run_loop(hv_vcpu_t vcpu,
                 exit_code = 128;
                 running = false;
             }
-        } else if (vexit->reason == HV_EXIT_REASON_CANCELED) {
-            /* Canceled by hv_vcpus_exit(). Can be: alarm timeout, exit_group
-             * from another thread, or signal preemption (signal_queue called
-             * hv_vcpus_exit to deliver a signal while the guest was in a tight
-             * loop).
+        } else if (vexit->reason == HV_EXIT_REASON_CANCELED ||
+                   (vexit->reason == HV_EXIT_REASON_UNKNOWN &&
+                    (signal_pending() || proc_exit_group_requested() ||
+                     (is_main && g_timed_out)))) {
+            /* Canceled by hv_vcpus_exit(). Can be: alarm timeout,
+             * exit_group from another thread, or signal preemption
+             * (signal_queue called hv_vcpus_exit to deliver a signal
+             * while the guest was in a tight loop).
+             *
+             * HV_EXIT_REASON_UNKNOWN is the same event seen from the other
+             * side of a race: when a host signal (e.g. the SIGUSR2 used by the
+             * cross-process guest-signal transport) is delivered to this thread
+             * while it is actively executing guest code inside hv_vcpu_run, the
+             * run aborts with UNKNOWN instead of the clean CANCELED that
+             * hv_vcpus_exit() produces for a vCPU caught between runs. The
+             * pending guest signal has already been drained and queued, so it
+             * is fully deliverable -- fall through to the same handling and
+             * resume rather than treating it as a fatal unexpected exit.
+             *
+             * Gate UNKNOWN on an actionable event actually being present (a
+             * queued signal, a pending exit_group, or a fired timeout) so it is
+             * only absorbed when there is real work to do. If HVF ever reports
+             * UNKNOWN for a genuine fault with nothing pending, fall through to
+             * the "unexpected exit reason" crash path at the end of the switch
+             * rather than silently retrying the run.
              */
             if (is_main && g_timed_out) {
                 /* Timeout already handled above the exception switch -- loop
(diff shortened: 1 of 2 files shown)


src/syscall/proc.c
```diff
@@ -1941,11 +1941,22 @@ int vcpu_run_loop(hv_vcpu_t vcpu,
                 exit_code = 128;
                 running = false;
             }
-        } else if (vexit->reason == HV_EXIT_REASON_CANCELED) {
+        } else if (vexit->reason == HV_EXIT_REASON_CANCELED ||
+                   vexit->reason == HV_EXIT_REASON_UNKNOWN) {
```
jserv: Add proper comments.

jserv: Routing UNKNOWN through the same fall-through as CANCELED is broad: if HVF ever returns UNKNOWN for a genuine fault (not the SIGUSR2 race), the loop will silently retry instead of taking the `unexpected exit reason` crash path at the end of this switch. Consider gating it on something actionable being present after the drain, e.g.:

```c
} else if (vexit->reason == HV_EXIT_REASON_CANCELED ||
           (vexit->reason == HV_EXIT_REASON_UNKNOWN &&
            (signal_pending() ||
             proc_exit_group_requested() ||
             (is_main && g_timed_out)))) {
```

Per the PR description the queued guest signal is already drained before we reach the switch, so `signal_pending()` is true for the cross-process SIGUSR2 race this is targeting -- but a genuine HVF error with no signal queued would still crash visibly instead of looping.

jserv (commented):
Core fix looks right -- the UNKNOWN/CANCELED unification and the EL0-preempt redirect via HV_REG_PC/CPSR both match the bug description. One follow-up worth folding in:

**`src/syscall/proc.c:1995` -- rseq IP fixup has the same root cause this PR fixes in signal.c.** The `rseq_try_abort` call in the CANCELED branch reads `ELR_EL1` unconditionally, so for an EL0-preempted vCPU it sees the stale syscall-return PC and won't fire the IP fixup for a critical section actually interrupted at EL0. Partial cover: when `signal_pending()` is also true on the same exit, `signal_deliver` re-runs `rseq_try_abort` with the el0_preempt-corrected PC, so the abort still happens. The gap is exits with no queued signal (fork-barrier wakeups, future ptrace wake paths). Mirror the CPSR-based PC selection here for symmetry:

```c
uint64_t cur_pc, cur_cpsr = 0;
hv_vcpu_get_reg(vcpu, HV_REG_CPSR, &cur_cpsr);
bool el0 = (cur_cpsr & 0xfULL) == 0;
if (el0)
    hv_vcpu_get_reg(vcpu, HV_REG_PC, &cur_pc);
else
    hv_vcpu_get_sys_reg(vcpu, HV_SYS_REG_ELR_EL1, &cur_pc);
int rseq_rc = rseq_try_abort(g, current_thread->rseq_gva,
                             current_thread->rseq_signature, &cur_pc);
if (rseq_rc == 1) {
    if (el0)
        hv_vcpu_set_reg(vcpu, HV_REG_PC, cur_pc);
    else
        hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_ELR_EL1, cur_pc);
}
```

See inline note on the UNKNOWN routing.

jserv (changes requested):

jserv:

PR #79, opened 2026-06-06
docs: fix memory-map table to show infra reserve at top of IPA

Max042004:
## Summary

The Memory Layout table in `docs/internals.md` showed the page-table pool, shim code, and shim data at **low** addresses (`0x10000`–`0x3FFFFF`) — a legacy layout. The code (`compute_infra_layout` in `src/core/guest.c`) anchors the infrastructure reserve at `[interp_base - INFRA_RESERVE, interp_base)`, in the dead zone above `mmap_limit`, so EL0 binaries are free to load at low addresses without colliding with the runtime.

This is the doc-cleanup follow-up @jserv flagged on #75 ("carries a pre-existing memory-map table showing a legacy layout with the infra reserve at LOW addresses; the actual code places it at the top of IPA. Worth a separate doc-cleanup PR"), kept out of #75 as requested.

## Changes

- Split the table into **low fixed addresses** (ELF, brk, stack, mmap) and the **`interp_base`-anchored high reserve**.
- Describe the reserve's internal layout (null guard, page-table pool, shim code sharing the pool's tail 2 MiB L2 block, shim data) to match `src/core/guest.h`.
- Note the reserve is demand-paged (`MAP_ANON`) so the reservation costs no host RAM.

## Ordering

Numbers reflect the **grown 16 MiB reserve / ~13.9 MiB pool** from #75, not current `main` (4 MiB / 960 KiB). This PR should land **together with or after #75**; the two touch different files (`#75` → `guest.h`/`main.c`, this → `internals.md`), so no merge conflict.

Docs-only; no code change.

### DIFF
--- a/docs/internals.md
+++ b/docs/internals.md
@@ -98,10 +98,9 @@ Guest memory is identity-mapped (guest VA == guest IPA). Large areas use 2 MiB
 block descriptors by default; mappings that need mixed permissions are split
 to 4 KiB L3 pages (see [Page Table Splitting](#page-table-splitting)).
 
+Low, fixed addresses:
+
 ```
-0x000010000  - 0x0000FFFFF:  Page table pool (960 KiB)
-0x000100000  - 0x0001FFFFF:  Shim code (2 MiB block, RX)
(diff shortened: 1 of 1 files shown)


jserv:

PR #80, opened 2026-06-06
Translate fcntl flock l_type between Linux and macOS

Max042004:
## Problem

The `F_GETLK`/`F_SETLK`/`F_SETLKW` handler in `src/syscall/fs.c` copied the
guest's `l_type` field straight into the host `struct flock`, on the
assumption (stated in a comment) that the lock-type constants match. They do
not:

| constant | Linux aarch64 | macOS / BSD |
|----------|:---:|:---:|
| `F_RDLCK` | 0 | 1 |
| `F_WRLCK` | 1 | 3 |
| `F_UNLCK` | 2 | 2 |

A Linux `F_RDLCK` (0) is not a valid lock type on macOS, so the host
`fcntl()` rejects it with `EINVAL`. POSIX-locking databases take a shared
read lock first, so this breaks them. Concretely, **SQLite fails to open any
on-disk database** with `disk I/O error (10)`: its initial `F_RDLCK` on the
1 GiB "pending byte" never succeeds. In-memory databases are unaffected,
which is what made this easy to miss.

Trace from a failing `sqlite3 file.db "CREATE TABLE ..."` under `--verbose`:

```
syscall 25@... (0x3, 0x6, 0x7ffc490, ...)   # fcntl(fd=3, F_SETLK=6, &flock)
  -> -22 (EINVAL)
```

## Fix

Map `l_type` Linux→macOS on input for all three commands, and map it back
macOS→Linux when writing the `F_GETLK` result. Unknown types are rejected
with `EINVAL` rather than passed through.

## Test

Adds `tests/test-flock.c` (registered in `manifest.txt` under "Syscall
coverage tests"), which exercises `F_SETLK`/`F_SETLKW`/`F_GETLK` around the
same 1 GiB pending-byte offsets SQLite uses. It **fails with `errno 22`
before this change and passes after** — verified by reverting the fix.

## Validation

- `make check`: all 97 tests pass (including the new `test-flock`).
- Local CI replication: `check-newline`, `check-format` (clang-format 22),
  `check-security`, `cppcheck`, syscall-dispatch consistency, and
  `make lint` all clean; `make elfuse` builds and codesigns.
- End-to-end with a statically cross-compiled aarch64 SQLite 3.47.2: on-disk
  CRUD, transactions, JSON1 function indexes, FTS5, WAL + `wal_checkpoint`,
  a 50k-row transaction, and `.dump`/`.read` round-trips all succeed; the
  resulting file is readable by the host `sqlite3`.

---
## Summary by cubic
Fixes `fcntl` lock-type translation between Linux guests and macOS hosts to prevent EINVAL on valid read locks. Restores on-disk SQLite and other POSIX-locking apps by correctly mapping `l_type` for `F_GETLK`, `F_SETLK`, and `F_SETLKW`.

- **Bug Fixes**
  - Map `l_type` Linux→macOS on input and macOS→Linux on `F_GETLK` output; reject unknown types with `EINVAL`.
  - Add `tests/test-flock.c` (registered in `tests/manifest.txt`) to cover SQLite’s pending-byte lock pattern; failed before, passes now.

<sup>Written for commit 63e7a9331263a0ebdb422bea784164c29ff5511f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/80?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/fs.c
+++ b/src/syscall/fs.c
@@ -655,11 +655,33 @@ int64_t sys_fcntl(guest_t *g, int fd, int cmd, uint64_t arg)
         memcpy(&l_start, lflock + 8, 8); /* offset 8 due to padding */
         memcpy(&l_len, lflock + 16, 8);
 
+        /* l_type constants differ between Linux and macOS/BSD:
+         *   Linux: F_RDLCK=0, F_WRLCK=1, F_UNLCK=2
+         *   macOS: F_RDLCK=1, F_UNLCK=2, F_WRLCK=3
+         * Passing the Linux value straight through makes a Linux F_RDLCK (0)
+         * an invalid type on macOS, which fcntl() rejects with EINVAL. This is
+         * the lock POSIX databases (e.g. SQLite) take first, so it must map. */
+        short mac_type;
+        switch (l_type) {
+        case 0: /* LINUX_F_RDLCK */
+            mac_type = F_RDLCK;
+            break;
+        case 1: /* LINUX_F_WRLCK */
+            mac_type = F_WRLCK;
+            break;
+        case 2: /* LINUX_F_UNLCK */
+            mac_type = F_UNLCK;
+            break;
+        default:
+            host_fd_ref_close(&host_ref);
+            return -LINUX_EINVAL;
+        }
+
         struct flock mac_fl = {
             .l_start = l_start,
             .l_len = l_len,
             .l_pid = 0,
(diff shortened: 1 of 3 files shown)


jserv:

PR #83, opened 2026-06-06
Support madvise, msync, and mremap on high-VA mmap regions

### LINKED ISSUE #108: `msync()` returns `ENOMEM` (errno 12) for high-VA guest mappings created by Rosetta
### Environment

- **Host**: macOS on Apple Silicon (M-series)
- **Guest arch**: x86_64 (run via Rosetta translation)
- **Trigger**: Running `apt update` or `apt upgrade` inside an x86_64 Linux guest prefix, or any guest program that `msync`s a large `mmap`-backed cache

### Problem Description

`sys_msync` in `src/syscall/mem.c` validates the guest virtual address by comparing it against `g->guest_size` (the primary memory region size) and `g->ipa_base` (the IPA base of the primary region). Addresses that fall outside that primary region are rejected with `-LINUX_ENOMEM`:

```c
if (addr < g->ipa_base)
    return -LINUX_ENOMEM;

uint64_t off = addr - g->ipa_base;
/* sys_msync stays primary-only here for the same reason as madvise: msync
 * iterates regions and reaches into host_base+off to read pages back from
 * the file overlay. Widening to extra-region ranges needs a region-aware
 * iterator landing alongside the data-movement refactor.
 */
if (off > g->guest_size || length > g->guest_size - off)
    return -LINUX_ENOMEM;
uint64_t end = off + length;
```

Under Rosetta translation, the Rosetta runtime allocates its JIT code cache and internal slabs at **high virtual addresses** — typically above `0x00007FFFFFFFFFFF` (the normal 48-bit user-space boundary on Linux x86_64). These mappings are added to `g->regions` via `guest_add_mapping`, so they are known to elfuse, but they have `start` values far beyond `g->guest_size`. The `addr - g->ipa_base` subtraction either underflows (if `addr < g->ipa_base`) or produces an offset that exceeds `g->guest_size`, causing both branches to return `-LINUX_ENOMEM`.

Furthermore, even when the bounds check passes for a legitimate high-VA region, the host-pointer computation in the region-iterating helpers was wrong:

**`sync_shared_aliases_range`** computed the guest host pointer as:
```c
const uint8_t *guest = (const uint8_t *) g->host_base + src->start +
                       (wfile_start - src->offset);
```
This adds `src->start` (a raw GVA for high-VA mappings) to `g->host_base`, producing a completely wrong host address.

**`refresh_shared_region_range`** had the same problem:
```c
uint64_t guest_off = r->start + (rfile_start - r->offset);
uint8_t *buf = (uint8_t *) g->host_base + guest_off;
```

Both helpers assumed all regions are in the primary buffer (`host_base + offset`), which is only correct for regions whose `start` is a primary-buffer-relative offset (i.e., `start < guest_size`).

### Observed Failure

`apt` memory-maps its package-list cache files (the files under `/var/lib/apt/lists/`) with `mmap(MAP_SHARED)` and periodically flushes them with `msync`. On an x86_64 guest under Rosetta, the `apt` process virtual address space contains Rosetta high-VA mappings. When `apt` calls `msync` on its cache, the guest VA falls in the Rosetta high-VA range and `sys_msync` returns `-LINUX_ENOMEM`. `apt` surfaces this as:

```
E: Unable to synchronize mmap - msync (12: Cannot allocate memory)
```

`apt update` and `apt upgrade` both fail at package list parsing before any packages can be installed.

Max042004:
## Problem

Three memory syscalls -- `sys_madvise`, `sys_msync`, and `sys_mremap` -- plus
the file-backed `MAP_SHARED` path in `sys_mmap_high_va` were all
**primary-window-only**. They computed `off = addr - ipa_base` and reached into
`host_base + off`, which only resolves for identity regions (`gpa_base ==
start`). High-VA mmap regions -- Rosetta's own slab/JIT and guest `mmap(NULL)`
placements -- back their pages at `gpa_base` (a named mapping or overflow
segment), so these paths rejected the range with `-ENOMEM`/`-EFAULT` or wrote to
the wrong host address.

Two concrete failures motivate this:

- **x86_64 Node.js / V8.** V8's page allocator decommits guard/code pages with
  `mprotect(PROT_NONE) + madvise(MADV_DONTNEED)` and `CHECK_EQ(0, ret)`s the
  madvise return, so the spurious `-ENOMEM` aborted Node the moment its JIT
  initialized:
  ```
  # Fatal error in , line 0
  # Check failed: 0 == ret.
  ```
  `--verbose` pins it to `madvise(advice=4 MADV_DONTNEED) -> -12 (ENOMEM)` on a
  page in the high-VA window (`mmap(NULL)` under Rosetta lands at e.g.
  `0x7fffff7fd000`).

- **apt (issue #108).** apt memory-maps its package-list cache with `MAP_SHARED`
  and `msync`s it; under an x86_64 guest that mapping is high-VA, so `sys_msync`
  returned `-ENOMEM` and apt failed with
  `E: Unable to synchronize mmap - msync (12: Cannot allocate memory)`. The
  mapping could not even be created, because `sys_mmap_high_va` refused
  file-backed `MAP_SHARED` with `-ENODEV`.

## Fix

Route every admission and data-movement path through the region tracker and
`host_ptr_for_gpa(gpa_base + ...)` so primary and high-VA regions act on their
real backing. Identity regions have `gpa_base == start`, so this collapses to
`host_base + off` and **primary behaviour is byte-for-byte unchanged**.

- **madvise** `MADV_DONTNEED` accepts high-VA ranges the tracker records as
  mapped and zeroes / restores them on their `gpa_base` backing (drops the
  previous zero-fill-only scope-out for high-VA file-backed pages).
- **msync** admission drops the `guest_size` bound (the existing coverage loop
  validates both windows), and `sync_shared_aliases_range` /
  `refresh_shared_region_range` resolve the guest bytes through `gpa_base`.
- **`sys_mmap_high_va`** accepts file-backed `MAP_SHARED` as a snapshot-style
  shared region (pread on map, `msync` writes dirty bytes back), so the high-VA
  shared mappings `msync` operates on can be created.
- **mremap** admits high-VA sources and resolves the source read/zero through
  `gpa_base` in the shrink, `MREMAP_FIXED`, and `MREMAP_MAYMOVE` paths. The
  destination stays in the primary window (`find_free_gap` /
  `mremap_extend_range`), so an `mremap(MAYMOVE)` of a high-VA region relocates
  it there with its contents intact -- no high-VA destination backing or new
  Stage-2 machinery is needed.

## Tests

Vendored x86_64 static ELFs run through elfuse + Rosetta (rebuild recipe in
`tests/fixtures/rosetta/README.md`):

- `x86_64-rosetta-madvise` -- `MADV_DONTNEED` on writable / `PROT_NONE` guard /
  multi-page high-VA ranges (the V8 decommit pattern).
- `x86_64-rosetta-msync` -- high-VA `MAP_SHARED` write-back
  (single/multi-page/`MS_ASYNC`), verified against the backing file.
- `x86_64-rosetta-mremap` -- `MREMAP_MAYMOVE` grow (contents preserved) and
  in-place shrink.

Wired as `make test-rosetta-{madvise,msync,mremap}` (and `test-rosetta-all`)
plus the matching suites in `tests/test-matrix.sh`.

The aarch64 unit tests `test-madvise` / `test-msync` / `test-mremap` /
`test-mremap-infra` cover the unchanged primary-window path. An end-to-end
`MAP_SHARED` store scenario (write + `msync` + cross-mapping refresh + `mremap`
grow + `madvise`) passes with this change and fails (`-ENODEV`/`-ENOMEM`) on the
pre-fix baseline.

## Notes

- macOS CI runners cannot run guest tests (no HVF entitlement), so the Rosetta
  suites were verified locally on Apple silicon; they self-skip (exit 77) where
  the translator or `timeout(1)` is unavailable.
- The vendored x86_64 fixtures are not built in-tree (the Makefile targets
  aarch64 hosts); rebuild them out of tree per the README when the sources
  change.
- For high-VA file-backed `MAP_SHARED` served through the FUSE sysroot layer,
  the backing is materialized read-only, so `msync` write-back to the original
  sysroot file is best-effort; real host-fd backings (e.g. `/dev/shm`) write
  back exactly. This matches how the primary window already treats such
  mappings.

Fixes #108.


jserv (changes requested):

jserv (changes requested):

jserv (changes requested):
Rebase latest `main` branch and resolve conflicts and refine per review.

jserv (changes requested):
Enforce rules described in https://cbea.ms/git-commit/ carefully.

jserv:

PR #84, opened 2026-06-06
Back read-only MAP_SHARED file mappings with MAP_PRIVATE

Max042004:
A MAP_SHARED, PROT_READ mapping of a file opened O_RDONLY could never be
installed. hvf_apply_file_overlay_quiesced() always mmap'd the host page
PROT_READ|PROT_WRITE and mapped the HVF segment RWX. On a read-only fd the
host mmap fails with EACCES (writable mapping of an O_RDONLY fd); forcing
PROT_READ then trips hv_vm_map(), because a MAP_SHARED mapping of an
O_RDONLY fd has macOS max_protection=READ and HVF cannot grant stage-2
rights (RWX) beyond the host region's max_protection (HV_ERROR).

This blocked every workload that maps a read-only file MAP_SHARED -- most
visibly the JVM, which maps its ~135 MiB lib/modules image exactly this
way and crashed on startup.

Choose the host backing from what the fd and the guest actually need:
  - guest wants PROT_WRITE: MAP_SHARED PROT_READ|PROT_WRITE (writes reach
    the file; an O_RDONLY fd still yields EACCES, matching Linux).
  - guest read-only on a writable fd: MAP_SHARED PROT_READ (max_protection
    is RWX, so the segment maps and cross-mapping coherence is preserved).
  - guest read-only on an O_RDONLY fd: MAP_PRIVATE PROT_READ. Its
    max_protection is RWX so the segment maps; the pages still show file
    content, and the guest's stage-1 tables keep the region read-only so
    the private copy is never dirtied -- no observable MAP_SHARED
    divergence for a read-only mapping.

The guest-requested prot is threaded through hvf_apply_file_overlay(),
hvf_apply_file_overlay_quiesced(), and restore_file_overlay_range() so
every overlay install/restore site picks the correct backing.

Add test-mmap-shared-ro covering the O_RDONLY read path, a second
concurrent read-only mapping, EACCES on a writable request, and the
read-only-mapping-on-O_RDWR-fd branch.

(cherry picked from commit 337d39a4313109884112a86a0c4147bddfe18fa1)

---
## Summary by cubic
Fixes read-only MAP_SHARED mappings of O_RDONLY files by using MAP_PRIVATE when needed, returns EACCES for MAP_SHARED|PROT_WRITE on read-only fds, and blocks mprotect(PROT_WRITE) upgrades to preserve Linux max_prot semantics. This unblocks JVM lib/modules and matches Linux behavior.

- **Bug Fixes**
  - Select backing by requested prot and fd mode:
    - Read-only on O_RDONLY: MAP_PRIVATE | PROT_READ.
    - Read-only on writable fd: MAP_SHARED | PROT_READ.
    - Writable: MAP_SHARED | PROT_READ|PROT_WRITE (EACCES on O_RDONLY).
  - In sys_mmap (non-fixed), reject MAP_SHARED | PROT_WRITE on read-only fds early with EACCES (no silent snapshot).
  - Track read-only shared backing with backing_ro and reject mprotect(PROT_WRITE) with EACCES; keep this across snapshots and all sys_mremap paths; prevent region merges when backing_ro differs.

- **Tests**
  - Add `test-mmap-shared-ro` covering: read-only MAP_SHARED on O_RDONLY, concurrent read-only mappings, EACCES on writable request, read-only mapping on an O_RDWR fd, and EACCES on mprotect(PROT_WRITE).

<sup>Written for commit bdde27651a339f80f9d60d28f27ea5c3985cf9cb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/84?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/mem.c
+++ b/src/syscall/mem.c
@@ -81,6 +81,7 @@ typedef struct {
     bool overlay_active;
     uint64_t overlay_start;
     uint64_t overlay_end;
+    bool backing_ro;
     char name[sizeof(((guest_region_t *) 0)->name)];
 } region_snapshot_t;
 
@@ -202,6 +203,21 @@ static void mark_overlay_metadata_range(guest_t *g,
         r->overlay_start = overlay_start;
         r->overlay_end = overlay_end;
         region_clip_overlay(r);
+    }
+}
+
+/* Mark the region spanning exactly [start, end) as backed by a fd that lost
+ * write access, so sys_mprotect rejects a later PROT_WRITE upgrade. Exact
+ * match (not overlap) because callers use this right after installing a
+ * single freshly-added region.
+ */
+static void mark_region_backing_ro(guest_t *g, uint64_t start, uint64_t end)
+{
+    for (int i = 0; i < g->nregions; i++) {
+        if (g->regions[i].start == start && g->regions[i].end == end) {
+            g->regions[i].backing_ro = true;
+            break;
+        }
     }
 }
 
@@ -1099,6 +1115,7 @@ static int capture_region_snapshots(guest_t *g,
         snap->overlay_active = r->overlay_active;
         snap->overlay_start = r->overlay_start;
         snap->overlay_end = r->overlay_end;
+        snap->backing_ro = r->backing_ro;
         str_copy_trunc(snap->name, r->name, sizeof(snap->name));
     }
 
@@ -1211,6 +1228,8 @@ static int restore_region_snapshots(guest_t *g, region_snapshot_t *snaps, int n)
             return -LINUX_ENOMEM;
         }
         snap->backing_fd = -1;
+        if (snap->backing_ro)
+            mark_region_backing_ro(g, snap->start, snap->end);
     }
 
     for (int i = 0; i < n; i++) {
@@ -2426,6 +2445,26 @@ int64_t sys_mmap(guest_t *g,
          * never reachable by the guest because the gap-finder advances the hint
          * to the next host-page boundary after each allocation.
          */
+        /* MAP_SHARED | PROT_WRITE against a backing fd opened without write
+         * access must fail EACCES, matching Linux. The alignment-mismatch and
+         * read-only-fd cases below both fall through to the pread snapshot
+         * path, which always succeeds -- without this check a writable shared
+         * mapping request on a read-only fd would be silently downgraded to a
+         * private snapshot instead of being rejected.
+         */
+        if ((flags & LINUX_MAP_SHARED) && (prot & LINUX_PROT_WRITE) &&
+            !overlay_fd_writable(host_backing_fd)) {
+            int rollback_err = rollback_fresh_mmap_allocation(
+                g, result_off, length, false, 0, 0, saved_mmap_next,
+                saved_mmap_end, saved_mmap_rx_next, saved_mmap_rx_end,
+                saved_rw_gap_hint, saved_rx_gap_hint);
+            if (track_backing_fd >= 0)
+                close(track_backing_fd);
+            host_fd_ref_close(&backing_ref);
+            if (rollback_err < 0)
+                return rollback_err;
+            return -LINUX_EACCES;
+        }
         /* overlay_fd_writable rejects read-only backing fds inside
          * hvf_apply_file_overlay; mirror the check here so a read-only mmap
          * takes the snapshot pread path directly, skipping the thread_quiesce /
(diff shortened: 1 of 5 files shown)


src/syscall/mem.c
```diff
@@ -1459,8 +1464,32 @@ static int hvf_apply_file_overlay_quiesced(guest_t *g,
...
+     *     content, and the guest's stage-1 tables keep the region read-only so
+     *     the private copy is never dirtied (no observable MAP_SHARED
+     *     divergence for a read-only mapping).
+     */
+    bool want_write = (prot & LINUX_PROT_WRITE) != 0;
+    int acc = fcntl(fd, F_GETFL);
+    bool fd_writable = acc >= 0 && ((acc & O_ACCMODE) == O_RDWR ||
+                                    (acc & O_ACCMODE) == O_WRONLY);
+    int host_prot = want_write ? (PROT_READ | PROT_WRITE) : PROT_READ;
+    int share = (want_write || fd_writable) ? MAP_SHARED : MAP_PRIVATE;
```
jserv: The MAP_PRIVATE substitution is sound while the mapping stays read-only, but `sys_mprotect` at mem.c:3275 doesn't know about the backing decision -- it just calls `guest_region_set_prot` + `guest_update_perms(prot_to_perms(prot))`. A guest that does:

```c
int fd = open(path, O_RDONLY);
char *p = mmap(NULL, len, PROT_READ, MAP_SHARED, fd, 0);
mprotect(p, len, PROT_READ | PROT_WRITE);   // Linux: EACCES; here: succeeds
*p = 0xff;                                   // Writes to COW copy, not the file
```

will silently upgrade stage-1 to RW and write into the COW copy. Linux returns `EACCES` because the mapping remembers `max_prot=READ` from the O_RDONLY fd. Before this PR the upgrade was unreachable (the initial mmap failed); the MAP_PRIVATE fallback exposes it.

The cleanest fix is to track a `host_backing_kind` (or `max_prot`) on `guest_region_t`, set it when this branch is taken, carry it through snapshots/splits/merges/mremap, and have `sys_mprotect` return `-LINUX_EACCES` when `PROT_WRITE` would exceed it. That also closes the downstream gap where `sys_msync` at mem.c:3560 skips its pwrite-refresh path for `overlay_active=true` regions on the assumption "the page cache already keeps them coherent with the file" -- false for a MAP_PRIVATE backing.

src/syscall/mem.c
```diff
@@ -1459,8 +1464,32 @@ static int hvf_apply_file_overlay_quiesced(guest_t *g,
...
+     *     cross-mapping coherence; max_protection is RWX because the fd is
+     *     writable, so the segment maps.
+     *   - guest read-only on an O_RDONLY fd: MAP_PRIVATE PROT_READ. Its
+     *     max_protection is RWX so the segment maps; the pages still show file
+     *     content, and the guest's stage-1 tables keep the region read-only so
+     *     the private copy is never dirtied (no observable MAP_SHARED
+     *     divergence for a read-only mapping).
+     */
+    bool want_write = (prot & LINUX_PROT_WRITE) != 0;
+    int acc = fcntl(fd, F_GETFL);
```
jserv: Treating `fcntl` failure as "not writable" silently picks `MAP_PRIVATE` for a valid writable fd whose `F_GETFL` transiently failed. Vanishingly rare on a host fd elfuse already holds via `host_fd_ref_open`, but the failure mode (losing MAP_SHARED coherence) is silent rather than surfaced.

Two options: return `-linux_errno()` on `acc < 0`, or hoist fd-writability detection up to where `host_backing_fd` is resolved and thread a plain `bool fd_writable` through. The latter also eliminates the per-install `fcntl` on the hot mmap path.

tests/test-mmap-shared-ro.c
```diff
@@ -0,0 +1,212 @@
...
+#include <sys/mman.h>
+#include <unistd.h>
+
+#include "test-harness.h"
+
+int passes = 0, fails = 0;
+
+/* Several guest pages so the overlay spans more than one host page and the
+ * containing 2 MiB segment is split and remapped over a realistic range. */
+#define NPAGES 64
```
jserv: 64 x 4 KiB = 256 KiB stays entirely within one 2 MiB segment, so `hvf_segment_split`'s multi-block path isn't exercised. JVM lib/modules is ~135 MiB and crosses many. Bump to at least `NPAGES 768` (3 MiB, two segments) so the segment-split + per-page-marker check catches a misaligned split.

A further test would lock in the corner this PR introduces:

```c
// Linux returns EACCES; with the MAP_PRIVATE fallback in place but no
// backing-kind tracking, elfuse currently lets this through silently.
static void test_rdonly_mprotect_write_rejected(const char *path) {
    int fd = open(path, O_RDONLY);
    char *p = mmap(NULL, FILE_LEN, PROT_READ, MAP_SHARED, fd, 0);
    EXPECT_EQ(mprotect(p, FILE_LEN, PROT_READ | PROT_WRITE), -1, "must reject");
    EXPECT_EQ(errno, EACCES, "errno must be EACCES");
    munmap(p, FILE_LEN); close(fd);
}
```

jserv (changes requested):

jserv (changes requested):

jserv:

PR #85, opened 2026-06-06
Join worker vCPU threads before tearing down guest memory

Max042004:
On exit_group the main thread left vcpu_run_loop and unmapped the guest slab
via guest_destroy() while sibling vCPU threads could still be mid-iteration
touching guest memory: the host process died with SIGSEGV and a guest
exit_group(0) was reported as exit 139. Masked until the fault-delivery fix
let multi-threaded JVM workloads (javac) reach the exit_group teardown.

- main() joins workers after vcpu_run_loop() returns and before any teardown;
  gdb_stub_shutdown() runs first so a worker parked in a GDB stop can
  deactivate instead of timing out the join.
- sc_exit_group no longer joins: it ran on the requesting thread, snapshotted
  never-deactivating slot 0, burned the poll cap, detached the host main
  pthread, and raced main()'s join into a double pthread_join. It now only
  requests + interrupts; main() holds the single authoritative join.
- Timed futex waits slept to the guest deadline, unreachable by teardown for
  the wait's whole duration (the FUTEX_WAIT_BITSET condvar path that glibc
  timed waits and JVM parkNanos issue never routes to os_sync; same in
  futex_lock_pi and futex_waitv). They now sleep in 100 ms quanta and re-check
  exit_group/futex_interrupt each quantum; guest-visible timeout accuracy is
  unchanged.
- thread_join_workers() polls under one shared 500 ms deadline — the old
  100 ms per-worker cap sat below the io wait's 200 ms flag re-check bound —
  and marks timed-out workers join_abandoned so guest_destroy()'s later pass
  cannot join/detach the same pthread twice (UB for a detached thread).
- Every thread_entry_t.active write is now a release store: the joiner polls
  lock-free, so the store/load pair on active is its only happens-before edge
  (flagged by ThreadSanitizer on the teardown test; clean under TSan with the
  fix).
- test-exit-group-worker drives a worker-issued exit_group with siblings
  pinned in each blocking state the join relies on (memory spinners, a timed
  futex wait 600 s out, a blocking pipe read) through 10 fork-child teardowns
  plus the top-level path.

A correctness argument — teardown paths, per-state wake bounds vs the join
cap, and residual risk — follows in a comment.

---
## Summary by cubic
Join worker vCPU threads in main before tearing down guest memory to prevent host SIGSEGVs and keep correct exit codes. Add a pre-join exit/interrupt kick and bound long waits (futex, FUSE, netlink, wait4) so workers always notice teardown; shut down the debugger first to avoid join timeouts.

- **Bug Fixes**
  - Call gdb_stub_shutdown() before thread_join_workers().
  - Join workers in main after vcpu_run_loop(); if no exit was requested, request exit_group(0) then call futex_interrupt_request(), wakeup_pipe_signal(), and thread_interrupt_all(); sc_exit_group now only requests and interrupts to avoid double-join and main-thread detach.
  - In guest_destroy(), run request/interrupt/join when destroy is reached directly so workers can’t touch freed memory.
  - Futex waits: sleep in 100 ms quanta, re-check exit/interrupt each quantum, and also re-check expired itimers and queued signals; futex_waitv quantum reduced to 100 ms; timeout accuracy unchanged.
  - thread_join_workers() polls under one shared 500 ms cap and marks timed-out workers join_abandoned so later passes skip them.
  - All writes to thread_entry_t.active are release stores so the lock-free join path observes a happens-before edge.
  - Make FUSE daemon reads, netlink waits, and specific-pid wait4 blocking paths bounded and interrupt-aware so teardown can reach them and avoid post-unmap memory access.

- **Tests**
  - Add test-exit-group-worker (and build target) to verify worker-issued exit_group exits cleanly in forked children and in the top-level path.

<sup>Written for commit 1fbbebb06dff3d4f87bf095789f4fa38e71fd09c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/85?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -35,10 +35,13 @@
 #include "core/sysroot.h"
 
 #include "runtime/forkipc.h"
+#include "runtime/futex.h" /* futex_interrupt_request */
 #include "runtime/proctitle.h"
+#include "runtime/thread.h"
 
 #include "syscall/fuse.h"
 #include "syscall/path.h"
+#include "syscall/poll.h" /* wakeup_pipe_signal */
 #include "syscall/proc.h"
 
 #include "debug/gdbstub.h"
@@ -673,8 +676,37 @@ int main(int argc, char **argv)
      */
     int exit_code = vcpu_run_loop(vcpu, vexit, &g, verbose, timeout_sec);
 
-    /* Tear down debugger state before freeing guest/vCPU resources. */
+    /* Tear down debugger state before joining workers: a worker parked in
+     * gdb_stub_handle_stop() stays active (not deactivated) until this
+     * broadcasts resume_cond, so joining first would just time out and
+     * detach it while it is still paused. */
     gdb_stub_shutdown();
+
+    /* Wait for worker vCPU threads to stop before tearing down guest memory.
+     * The main thread leaves the run loop as soon as it observes the
+     * exit_group flag, but sibling vCPU threads may still be mid-iteration in
+     * their own run loops (e.g. touching shim_globals). cleanup_main_resources
+     * unmaps the guest slab via guest_destroy, so a still-running worker would
+     * fault on freed guest memory and crash the host with SIGSEGV, masking the
+     * real exit code. thread_join_workers() is a no-op once the workers have
+     * already wound down (the common single-threaded case).
+     *
+     * vcpu_run_loop can also return here without anyone having requested
+     * exit_group or kicked the siblings out of hv_vcpu_run: the alarm timeout
+     * (exit_code 124), a fatal default-disposition signal, or ELR_EL1==0 all
+     * bail out with a bare break. On those paths siblings are still spinning
+     * in the guest, so mirror guest_destroy's request-interrupt prefix before
+     * joining -- otherwise this call burns its full poll cap, detaches every
+     * worker, and guest_destroy's own request-interrupt-join (which honors
+     * join_abandoned) skips them, leaving live pthreads to fault on the
+     * imminent unmap.
+     */
+    if (!proc_exit_group_requested())
+        proc_request_exit_group(0);
+    futex_interrupt_request();
+    wakeup_pipe_signal();
+    thread_interrupt_all();
+    thread_join_workers();
 
     /* Diagnostic counter dump runs before guest_destroy so the shim_data
      * mapping is still valid. ELFUSE_SHIM_STATS is the gate; an unset variable
--- a/src/runtime/futex.c
+++ b/src/runtime/futex.c
@@ -370,6 +370,27 @@ static uint64_t futex_remaining_ns(const struct timespec *deadline,
     return rem < cap_ns ? rem : cap_ns;
 }
 
+/* Absolute CLOCK_REALTIME wait target one bounded quantum from now: the guest
+ * deadline capped at FUTEX_OS_SYNC_POLL_CAP_NS. Every condvar sleep in the
+ * timed-wait paths goes through this so a waiter parked on a distant guest
+ * deadline still re-checks the process-wide teardown flags
+ * (proc_exit_group_requested / futex_interrupt) each quantum; an uncapped
+ * sleep would outlive thread_join_workers' poll cap and race
+ * guest_destroy's unmap of the memory the waiter touches on wake.
+ * Returns false when the guest deadline has already passed.
+ */
+static bool futex_quantum_deadline(const struct timespec *deadline,
+                                   struct timespec *out)
+{
+    uint64_t rem_ns = futex_remaining_ns(deadline, FUTEX_OS_SYNC_POLL_CAP_NS);
+    if (rem_ns == 0)
+        return false;
+    clock_gettime(CLOCK_REALTIME, out);
+    out->tv_nsec += (long) rem_ns;
+    timespec_normalize(out);
+    return true;
+}
+
 #if ELFUSE_HAVE_OS_SYNC_WAIT_ON_ADDRESS
 
 /* Wake up to budget kernel-side waiters at uaddr via
@@ -629,10 +650,45 @@ static int64_t futex_wait(guest_t *g,
 
     while (!__atomic_load_n(&waiter.woken, __ATOMIC_ACQUIRE)) {
         if (has_timeout) {
-            int rc = pthread_cond_timedwait(&waiter.cond, &b->lock, &deadline);
-            if (rc != 0) {
-                /* Timeout (ETIMEDOUT) or error; stop waiting */
+            /* Sleep in bounded quanta rather than to the guest deadline: a
+             * worker parked here for a long guest timeout (JVM parkNanos,
+             * sem_timedwait) would otherwise be unreachable by exit_group /
+             * futex_interrupt, outlive thread_join_workers' poll cap, and
+             * race guest_destroy's unmap. The interrupt checks mirror the
+             * no-timeout branch below.
+             */
+            struct timespec quantum = {0};
+            if (!futex_quantum_deadline(&deadline, &quantum)) {
                 ret = -LINUX_ETIMEDOUT;
+                break;
+            }
+            pthread_cond_timedwait(&waiter.cond, &b->lock, &quantum);
+            if (proc_exit_group_requested() || futex_interrupt_consume()) {
+                ret = -LINUX_EINTR;
+                break;
+            }
+
+            if (__atomic_load_n(&waiter.woken, __ATOMIC_ACQUIRE))
+                break;
+
+            /* Mirror the no-timeout branch's itimer/queued-signal re-check
+             * below: without it, a thread parked in a timed FUTEX_WAIT_BITSET
+             * (glibc sem_timedwait, pthread_cond_timedwait, JVM parkNanos)
+             * only observes an expired guest itimer or a deliverable queued
+             * signal once the futex wakes or the full guest deadline elapses.
+             * Lock order requires dropping the bucket lock before touching
+             * sig_lock (4).
+             */
+            pthread_mutex_unlock(&b->lock);
+            signal_check_timer();
+            bool sig_ready = signal_pending() != 0;
+            pthread_mutex_lock(&b->lock);
+
+            if (__atomic_load_n(&waiter.woken, __ATOMIC_ACQUIRE))
+                break;
+
+            if (sig_ready) {
+                ret = -LINUX_EINTR;
                 break;
             }
             continue;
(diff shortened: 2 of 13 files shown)


src/main.c
```diff
@@ -510,6 +511,16 @@ int main(int argc, char **argv)
...
 
+    /* Wait for worker vCPU threads to stop before tearing down guest memory.
+     * The main thread leaves the run loop as soon as it observes the
+     * exit_group flag, but sibling vCPU threads may still be mid-iteration in
+     * their own run loops (e.g. touching shim_globals). cleanup_main_resources
+     * unmaps the guest slab via guest_destroy, so a still-running worker would
+     * fault on freed guest memory and crash the host with SIGSEGV, masking the
+     * real exit code. thread_join_workers() is a no-op once the workers have
+     * already wound down (the common single-threaded case). */
+    thread_join_workers();
```
jserv, line 709: Ordering vs `gdb_stub_shutdown()` at the next line leaves a residual race. `gdb_stub_shutdown()` (gdbstub.c:1435-1440) broadcasts `resume_cond` to release any GDB-paused workers so they can exit. A worker stuck in `gdb_stub_handle_stop` has `active=1` (paused in ptrace-stop, not deactivated), so the new join here polls 100ms, sees no deactivation, and detaches. Then `gdb_stub_shutdown` releases the worker, it resumes execution, and main proceeds to `shim_globals_counters_dump` / `cleanup_main_resources` -- same SIGSEGV pattern the PR is fixing, just gated on a GDB session.

Move the join one line down so the GDB-paused case is also covered:

```c
int exit_code = vcpu_run_loop(...);

/* Tear down debugger state before freeing guest/vCPU resources. */
gdb_stub_shutdown();

thread_join_workers();

if (shim_globals_stats_enabled())
    shim_globals_counters_dump(&g);
```

jserv, line 709: `sc_exit_group` at syscall.c:719 already calls `thread_join_workers()`, so this is a second join. Two consequences worth tracking:

1. If a worker calls `exit_group`, that worker's `thread_join_workers` snapshots all others including main (slot 0, which never `thread_deactivate`s itself) -- so it polls 100ms then detaches main. Detaching the host process's main pthread is legal but architecturally odd.
2. After that first call returns and main eventually runs this new one, any worker that wound down in between has `active=0` -- main's call then tries `pthread_join` on a thread that was already joined or detached. POSIX says undefined; macOS libpthread returns `EINVAL` in practice (return is unchecked at thread.c:354), so not catastrophic but contract drift.

Cleaner: drop the `thread_join_workers()` from `sc_exit_group`. The exit_group caller still does `proc_request_exit_group` + `thread_for_each(thread_force_exit_cb)` which forces every sibling out of `hv_vcpu_run`; main observes the flag, exits the loop, and runs the single authoritative join here. Removes the detach-main wart and the double-join. Can be a follow-up -- not a blocker for this PR.

jserv: ?

jserv (changes requested):
Despite the lengthy comments, show me the proof.

Max042004:
## (A) Paths: every teardown path establishes the precondition before unmapping

`main()` orders shutdown → join → diagnostics → destroy (`src/main.c:675-702`):

```c
    gdb_stub_shutdown();          /* must precede the join: a worker parked in
                                     gdb_stub_handle_stop() stays active until
                                     resume_cond broadcasts */
    thread_join_workers();
    ...
    if (shim_globals_stats_enabled())
        shim_globals_counters_dump(&g);   /* reads guest memory — must sit
                                             between join and destroy */
    syscall_hist_dump();
    cleanup_main_resources(...);          /* → guest_destroy() */
```

`fork_child_main()` has nine pre-vCPU failure exits (no workers can exist yet)
plus the normal exit; it depends on nothing the caller did because
`guest_destroy()` re-runs the full sequence itself (`src/core/guest.c:621-626`):

```c
    if (!proc_exit_group_requested())
        proc_request_exit_group(0);
    futex_interrupt_request();
    wakeup_pipe_signal();
    thread_interrupt_all();
    thread_join_workers();
```

On the `main()` path this internal join sees an empty snapshot (workers
already reaped, `active` clear).

## (B) the join terminates — every blocking state has a wake-up bounded
below the join cap

| Worker state | Wake mechanism | Bound |
|---|---|---|
| inside `hv_vcpu_run` (incl. shim EL1 fast path) | `hv_vcpus_exit` → CANCELED → flag check | µs |
| host handler, not blocked | flag check at loop top | handler runtime |
| `poll`/`select`/`epoll_pwait` | wakeup pipe; 200 ms re-arm re-check | ≤ 200 ms |
| interruptible io wait (pipe/socket/tty rw) | wakeup pipe + 200 ms flag re-check | ≤ 200 ms |
| futex, untimed (condvar) / os_sync | 100 ms quanta + flag check | ≤ 100 ms |
| **futex, timed (condvar)** — `FUTEX_WAIT_BITSET`, i.e. glibc timed waits, never routed to os_sync | **was: slept to guest deadline, unbounded** → 100 ms quanta; same fix in `futex_lock_pi`; `futex_waitv` 500→100 ms | ≤ 100 ms |
| `gdb_stub_handle_stop()` | `resume_cond` broadcast before join | immediate |

The 200 ms io bound (`src/syscall/io.c:169-178`) — the pipe is
single-consumer, a sibling can steal the wake byte, hence the bounded re-check:

```c
    for (;;) {
        if (proc_exit_group_requested() || futex_interrupt_consume() ||
            signal_pending_interruption(NULL))
            return -LINUX_EINTR;
        int ret = poll(fds, 2, 200);
        ...
```

The unbounded timed-futex state and its fix (`futex_wait`, identical pattern
in `futex_lock_pi` at `src/runtime/futex.c:1205`):

```diff
     while (!__atomic_load_n(&waiter.woken, __ATOMIC_ACQUIRE)) {
         if (has_timeout) {
-            int rc = pthread_cond_timedwait(&waiter.cond, &b->lock, &deadline);
-            if (rc != 0) {
-                /* Timeout (ETIMEDOUT) or error; stop waiting */
+            struct timespec quantum;
+            if (!futex_quantum_deadline(&deadline, &quantum)) {
                 ret = -LINUX_ETIMEDOUT;
                 break;
             }
+            pthread_cond_timedwait(&waiter.cond, &b->lock, &quantum);
+            if (proc_exit_group_requested() || futex_interrupt_consume()) {
+                ret = -LINUX_EINTR;
+                break;
+            }
             continue;
         }
```

`futex_quantum_deadline()` (`src/runtime/futex.c:382`) caps each sleep at
`FUTEX_OS_SYNC_POLL_CAP_NS` (100 ms) or the remaining guest deadline,
whichever is sooner — so guest-visible timeout accuracy is unchanged.

Join cap was 100 ms per worker — below the 200 ms io bound and vacuous against
the unbounded state. Now one shared 500 ms deadline
(`src/runtime/thread.c:366-374`), 2.5× the worst bound, not multiplied by
worker count; the poll exits on deactivation, so the common case stays µs:

```c
    for (int i = 0; i < 100; i++) {
        bool any_active = false;
        for (int w = 0; w < nworkers; w++)
            if (__atomic_load_n(&workers[w].t->active, __ATOMIC_ACQUIRE)) {
                any_active = true;
                break;
            }
        if (!any_active)
            break;
        usleep(5000);
    }
```

## (C) the join itself is UB-free

Double join/detach: a worker detached by a timed-out pass stayed `active`, so
`guest_destroy()`'s second pass would join/detach the same pthread — UB for a
detached thread. Now marked and skipped
(`src/runtime/thread.c:343`, `:379-388`):

```c
        if (t == current_thread || t->join_abandoned)   /* snapshot skip */
            continue;
    ...
        if (!__atomic_load_n(&workers[w].t->active, __ATOMIC_ACQUIRE)) {
            pthread_join(workers[w].thr, NULL);
        } else {
            pthread_detach(workers[w].thr);
            pthread_mutex_lock(&thread_lock);
            workers[w].t->join_abandoned = true;
            pthread_mutex_unlock(&thread_lock);
        }
```

`active` was a plain store under `thread_lock` vs the joiner's lock-free
acquire-load — the mutex gives the joiner no HB edge; without release the
compiler may sink the worker's final guest stores past the flag write. Found
by TSan on the teardown test; every write is now a release store
(`src/runtime/thread.c:226`, in `thread_deactivate`):

```c
    __atomic_store_n(&t->active, 0, __ATOMIC_RELEASE);
```

so: every guest access happens-before the `active` release-store,
happens-before the joiner's acquire-load, happens-before the unmap. vCPUs are
destroyed only after the join (`thread_destroy_all_vcpus`), never cross-thread
while the owner may be in `hv_vcpu_run`.

## Residual

Host condvar parks outside the syscall wait set (`fork_cond` mid-fork-barrier,
`ptrace_cond` mid-ptrace-stop) are not in the table; a worker there can exceed
the cap, be detached, and I then holds only because process exit follows
`guest_destroy()` almost immediately — mitigation, not proof. Same trade-off
as the pre-existing cross-thread `hv_vcpu_destroy` for abandoned workers.

## Evidence (necessary, not sufficient)

- `test-exit-group-worker`: worker-issued exit_group with siblings pinned in
  the load-bearing states — 3 spinners, one timed `FUTEX_WAIT_BITSET` 600 s
  out (sleeps through any cap on the old code), one blocking pipe read —
  through 10 fork-child teardowns + the top-level path. 11/11, ~1.6 s total
  (one uncapped timed-futex sleep would exceed that by 100×). `make check`
  green.
- TSan (host side only — stage-2 accesses inside `hv_vcpu_run` are invisible
  to it): flagged the `active` plain-store/atomic-load pair, fixed above.
  With the fix in and the known race suppressed:
  `test-exit-group-worker`, `test-futex-pi`, `test-futex-waitv`,
  `test-robust-futex`, `test-pthread` — zero reports.

src/runtime/futex.c
```diff
@@ -629,12 +650,23 @@ static int64_t futex_wait(guest_t *g,
...
+             * race guest_destroy's unmap. The interrupt checks mirror the
+             * no-timeout branch below.
+             */
+            struct timespec quantum = {0};
+            if (!futex_quantum_deadline(&deadline, &quantum)) {
                 ret = -LINUX_ETIMEDOUT;
                 break;
             }
+            pthread_cond_timedwait(&waiter.cond, &b->lock, &quantum);
+            if (proc_exit_group_requested() || futex_interrupt_consume()) {
```
jserv, line 666: The untimed branch below drops the bucket lock each poll to run signal_check_timer() and signal_pending(), returning EINTR when a real signal is queued for this thread. This timed branch only checks proc_exit_group_requested()/futex_interrupt_consume(). A thread parked in a timed FUTEX_WAIT_BITSET (glibc sem_timedwait, pthread_cond_timedwait, JVM parkNanos) now will not observe an expired guest itimer or a queued deliverable signal until the futex is woken or the full guest deadline elapses, so signal and timer delivery to that thread is deferred. futex_lock_pi's timed branch has the same omission. Mirror the untimed branch: after each quantum, unlock, run signal_check_timer()/signal_pending(), relock, re-check waiter.woken, and return -EINTR on a pending deliverable signal.

src/runtime/thread.c
```diff
@@ -327,33 +340,51 @@ void thread_join_workers(void)
...
-     * hv_vcpus_exit typically finish within microseconds. Threads stuck in
-     * uninterruptible host calls (blocking read/poll) are given 100ms to
-     * finish. If still alive, detach and let process exit clean up. The vCPU is
-     * NOT destroyed here because HVF vCPUs are thread-affine, so cross-thread
-     * hv_vcpu_destroy while the owning thread may still be inside hv_vcpu_run
-     * is unsafe.
+    /* Poll under one shared 500ms deadline OUTSIDE the lock, then join or
+     * detach each worker. Workers that responded to hv_vcpus_exit typically
+     * finish within microseconds; the cap only matters for workers parked in
+     * host blocking calls, and it must exceed the worst-case teardown-flag
```
jserv, line 498: The cap is defended here as exceeding the worst-case teardown-flag re-check latency of every blocking state, but three worker states never re-check the flags at all. A specific-child wait4 blocks in the host wait4() with no WNOHANG loop or exit-group check (src/syscall/proc.c:755); an AF_NETLINK read blocks in poll(fd, -1) on its private pipe, not the shared wakeup pipe (src/syscall/netlink.c:714); and a FUSE /dev/fuse daemon read blocks in pthread_cond_wait on session->queue_cond (src/syscall/fuse.c:2220). None watch the shared wakeup pipe or poll proc_exit_group_requested(), so neither hv_vcpus_exit nor wakeup_pipe_signal reaches them. A worker in any of these is detached after the cap, then touches guest memory once it returns -- the exact post-unmap SIGSEGV this change targets. This also contradicts the guest_destroy comment that "any thread parked in epoll or poll wakes off the shared pipe." Route these waits through io_wait_fd_or_interrupted / a bounded exit-group re-check, or close and broadcast those sessions during exit-group teardown before the join.

src/main.c
```diff
@@ -667,9 +668,22 @@ int main(int argc, char **argv)
...
```
jserv, line 709: thread_join_workers() only polls active and joins/detaches; it does not request exit_group, interrupt vCPUs, or signal the wakeup pipe. On the exit_group path sc_exit_group already did that, but vcpu_run_loop also returns here on the alarm timeout (exit_code 124, set with a bare break and no exit-group request), on a fatal default-disposition signal, and on ELR_EL1==0 -- none of which requested exit_group or kicked the siblings. On those paths the workers are still active, so this join burns the full 500ms, detaches every one, and sets join_abandoned; guest_destroy's later request-interrupt-join then skips them (it honors join_abandoned), and thread_destroy_all_vcpus destroys their vCPUs cross-thread while they still run guest code -- the same post-unmap crash this change targets, now on the timeout and fatal-signal paths. Before this change guest_destroy did the request-interrupt-join and reaped them. Give this join the same prefix guest_destroy uses (request exit_group if not already requested, futex_interrupt_request, wakeup_pipe_signal, thread_interrupt_all) before calling it, or gate the join on proc_exit_group_requested().

src/runtime/thread.c
```diff
@@ -327,33 +340,51 @@ void thread_join_workers(void)
 
     pthread_mutex_lock(&thread_lock);
     THREAD_FOR_EACH_ACTIVE (t) {
-        if (t == current_thread)
+        if (t == current_thread || t->join_abandoned)
```
jserv, line 475: Two lifecycle hazards in this snapshot. First, vm-clone (CLONE_VM ptrace-inferior) threads are created PTHREAD_CREATE_DETACHED (forkipc.c clone_vm path) but are not excluded here, so main()'s new join reaches an is_vm_clone worker and calls pthread_join/pthread_detach on an already-detached thread -- undefined. Skip is_vm_clone entries; they are reaped via wait4, never joined. Second, host_thread is published outside thread_lock (forkipc.c, after pthread_create returns) while thread_alloc already set active=1 before the create, so if exit_group races a cloner still inside pthread_create, this loop can snapshot the slot with a zeroed host_thread and later pthread_detach(0). Publish host_thread under thread_lock (or have the worker store pthread_self() as its first action), and skip entries whose handle is still zero.

jserv (changes requested):
Rebase latest `main` branch, resolve conflicts, and refine per review messages.

PR #86, opened 2026-06-06
Deliver synchronous fault signals to the faulting thread

Max042004:
> [!NOTE]
> **Stacked on #76.** This branch is based on `fix-cross-process-signal-el0` (#76),
> whose commit appears here until #76 merges; afterwards this PR auto-reduces to the
> single fault-delivery commit. Review #76 first.

A guest exception (SIGSEGV/SIGBUS/SIGILL/SIGFPE/SIGTRAP) was delivered via
`signal_set_fault_info()` (which records si_code/si_addr in a `_Thread_local` slot)
followed by `signal_queue()` + `signal_deliver()`. But `signal_queue()` sets a bit in
the process-wide pending mask, and every vCPU thread runs `signal_deliver()` at its own
poll points, so a fault raised by thread A could be dequeued and delivered by thread B —
whose thread-local fault info is empty. The siginfo then degraded to `si_code=SI_USER`
with no `si_addr` instead of `SEGV_MAPERR`/si_addr. A JVM treats such a SIGSEGV as a
fatal external signal rather than a recoverable implicit null check, so `javac` crashed.
Two threads faulting on the same signal also collapsed into one bitmask bit, dropping
one fault.

Add `signal_deliver_fault()`, which delivers a given signal directly to the current
thread using its thread-local fault info and never touches the process-wide pending set.
`signal_deliver()`'s frame-building tail is factored into `deliver_signal_locked()` and
shared. All four synchronous-fault sites in the vCPU loop (SEGV_ACCERR, BRK→SIGTRAP,
SIGILL, data-abort SIGSEGV) route through it.

Adds `test-fault-signal-mt`: N threads each take recoverable SIGSEGVs and assert every
delivery is `SEGV_MAPERR` with the right `si_addr` on the faulting thread. It fails
deterministically (rc=2) on the old queue-based path and passes on the fix.

---
## Summary by cubic
Deliver synchronous fault signals to the faulting vCPU thread to preserve correct si_code/si_addr and prevent cross‑thread delivery or dropped faults. Fixes JVM crashes where SIGSEGV arrived as SI_USER and makes multi‑threaded faults reliable.

- **Bug Fixes**
  - Added `signal_deliver_fault()` to deliver SIGSEGV/SIGBUS/SIGILL/SIGFPE/SIGTRAP directly to the current thread, bypassing the process‑wide pending set. Follows Linux force_sig semantics: if disposition is `SIG_IGN` or the signal is blocked, reset to `SIG_DFL` and unblock before delivery.
  - Factored handler frame build into `deliver_signal_locked()`; shared by `signal_deliver()` and `signal_deliver_fault()` and preserves EL0 PC/CPSR state.
  - Routed all synchronous guest exceptions in `vcpu_run_loop` (SEGV_ACCERR, BRK→SIGTRAP, SIGILL, data‑abort SIGSEGV) through the new path. Added `test-fault-signal-mt` to validate per‑thread SIGSEGV delivery (correct si_code/si_addr) and to lock in forced‑fault semantics (SIG_IGN/blocked SIGSEGV must terminate). Makefile and manifest updated.

<sup>Written for commit 2d998edd2816fee79f529cd98d571b739e8caeba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/86?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/signal.c
+++ b/src/syscall/signal.c
@@ -1311,37 +1311,27 @@ static void build_sigcontext_reserved(uint8_t *reserved,
     memset(reserved + off, 0, 8);
 }
 
-int signal_deliver(hv_vcpu_t vcpu, guest_t *g, int *exit_code)
+/* Build and install the rt_sigframe for `signum` on the current thread, with
+ * sig_lock held on entry and released on every return path. Shared by
+ * signal_deliver() (signal selected from the process-wide pending set) and
+ * signal_deliver_fault() (synchronous fault forced onto the faulting thread).
+ * rt_info supplies si_code/si_pid/sigval when no thread-local pending_fault is
+ * set; the pending_fault is consumed (one-shot) when valid. Returns 1 if a
+ * handler frame was installed, 0 if the signal was ignored, and -1 (with
+ * *exit_code set) when the default disposition terminates the guest.
+ */
+static int deliver_signal_locked(hv_vcpu_t vcpu,
+                                 guest_t *g,
+                                 int signum,
+                                 signal_rt_info_t rt_info,
+                                 int *exit_code)
 {
-    pthread_mutex_lock(&sig_lock);
     uint64_t *blocked = thread_blocked_ptr();
     uint64_t *saved_ptr = thread_saved_blocked_ptr();
     bool *valid_ptr = thread_saved_valid_ptr();
-    uint64_t deliverable = sig_state.pending & ~*blocked;
-    if (deliverable == 0) {
-        pthread_mutex_unlock(&sig_lock);
-        return 0;
-    }
 
-    /* Find lowest pending unblocked signal */
-    int signum = bit_ctz64(deliverable) + 1;
-    signal_rt_info_t rt_info = signal_default_info(signum);
-
-    /* Dequeue: for RT signals, decrement count and only clear the pending bit
-     * when the queue is empty. Standard signals are always cleared (single
-     * instance, bitmask semantics).
-     */
-    if (signum >= LINUX_SIGRTMIN) {
-        signal_rt_dequeue_locked(signum, &rt_info);
-    } else {
-        rt_info = signal_standard_peek_locked(signum);
-        sig_state.std_info_valid[signum - 1] = false;
-        sig_state.pending &= ~sig_bit(signum);
-    }
-
-    /* signum is bit_ctz64(deliverable) + 1, bounded 1..64 by the 64-bit pending
-     * mask. The static analyzer cannot see the bound, so gate the array access
-     * defensively.
+    /* signum is 1..64 from the caller; the static analyzer cannot see the
+     * bound, so gate the array access defensively.
      */
     int idx = signum - 1;
     if (!RANGE_CHECK(idx, 0, LINUX_NSIG)) {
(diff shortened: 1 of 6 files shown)


src/syscall/signal.c
```diff
@@ -1590,13 +1611,63 @@ int signal_deliver(hv_vcpu_t vcpu, guest_t *g, int *exit_code)
...
+     * is racy: another vCPU thread can dequeue the bit and deliver it with no
+     * fault info (si_code becomes SI_USER, which makes a JVM treat a recoverable
+     * implicit null-check as a fatal external signal), and two threads faulting
+     * on the same signal collapse into one bit so one fault is lost. Deliver
+     * directly here, never touching sig_state.pending. The blocked mask is
+     * intentionally ignored: a synchronous fault cannot be postponed.
+     */
+    pthread_mutex_lock(&sig_lock);
+    signal_rt_info_t rt_info = signal_default_info(signum);
+    return deliver_signal_locked(vcpu, g, signum, rt_info, exit_code);
```
jserv, line 1679: The comment two lines up calls this "intentionally ignored: a synchronous fault cannot be postponed" -- that's only half the Linux contract. `force_sig_info_to_task` in kernel/signal.c resets the disposition to SIG_DFL and unblocks the signum before applying default, so a SIGSEGV/SIGBUS/SIGILL/SIGFPE/SIGTRAP that's blocked OR set to SIG_IGN terminates the process. This PR's behavior diverges in two ways:

1. **SIG_IGN on synchronous fault (preexisting)**: `deliver_signal_locked` at signal.c:1352 returns 0 for SIG_IGN. The vCPU loop resumes, PC unchanged, the same instruction re-faults forever. Linux would terminate.
2. **Blocked synchronous fault (new in this PR)**: pre-PR, `signal_deliver`'s `deliverable = sig_state.pending & ~*blocked` short-circuited and the thread re-faulted infinitely. Post-PR, `signal_deliver_fault` bypasses the blocked mask entirely and runs the user handler despite the block. Linux would reset to SIG_DFL + unblock + terminate.

JVM works either way because it never blocks SIGSEGV, but the contract drift is real. A five-line precheck closes both cases with one shape:

```c
int signal_deliver_fault(hv_vcpu_t vcpu, guest_t *g, int signum, int *exit_code)
{
    pthread_mutex_lock(&sig_lock);
    uint64_t *blocked = thread_blocked_ptr();
    linux_sigaction_t *act = &sig_state.actions[signum - 1];
    if (act->sa_handler == LINUX_SIG_IGN ||
        (*blocked & sig_bit(signum))) {
        /* Linux force_sig_info_to_task: forced synchronous faults reset
         * disposition to SIG_DFL, unblock the signum, then apply default. */
        act->sa_handler = LINUX_SIG_DFL;
        *blocked &= ~sig_bit(signum);
    }
    signal_rt_info_t rt_info = signal_default_info(signum);
    return deliver_signal_locked(vcpu, g, signum, rt_info, exit_code);
}
```

Update the "intentionally ignored" comment accordingly.

jserv (changes requested):

jserv (changes requested):

jserv:

PR #96, opened 2026-06-12
Improve Linux distro rootfs compatibility on Apple Silicon

doanbaotrung:
Improve compatibility with real Linux distro rootfs environments on
Apple Silicon hosts. Package-manager and shell workflows need behavior
closer to Linux for credentials, script execution, fork/clone state,
wait handling, pipes, /proc, and shared mappings.

Preserve dynamic guest UID/GID state in auxv instead of always reporting
fixed guest IDs, and allow the initial guest identity to be configured
with ELFUSE_GUEST_UID and ELFUSE_GUEST_GID. This lets distro workflows
such as apt post-install scripts run with root-like guest credentials
when needed.

Probe ELF binaries quietly before falling back to shebang handling, so
script execution does not emit misleading "not an ELF" diagnostics.

Extend fork IPC state and child restore handling to carry more complete
CPU state, including TLS-related registers, PAC keys, clone flags,
child TID handling, TPIDRRO_EL0, TPIDR2_EL0, and the original SPSR. Add
child process monitoring so host child exit can wake Linux-style wait
and signal behavior.

Align non-fixed file-backed MAP_SHARED mappings to 2 MiB stage-2
boundaries to avoid HVF mapping issues on Apple Silicon.

Improve sysroot symlink creation for absolute guest symlink targets, and
add small Linux compatibility behavior for sync_file_range and pipe
F_SETNOSIGPIPE.

These changes were tested with an Ubuntu arm64 rootfs using shell
pipelines, /proc checks, and apt-get update smoke testing.

---
## Summary by cubic
Improves Linux distro rootfs behavior on Apple Silicon so package managers and shell pipelines run cleanly. Tightens identity, exec, fork/clone, memory, pipes, /proc, and symlink handling to better match real Linux.

- New Features
  - Configurable guest identity: AT_UID/EUID and AT_GID/EGID reflect runtime IDs; set initial IDs via `ELFUSE_GUEST_UID`/`ELFUSE_GUEST_GID`. /proc and getgroups return these values. set*id syscalls follow Linux privileged semantics.
  - Exec: quietly probe ELF before shebang so scripts run without “not an ELF” noise.
  - Memory: non-fixed, file-backed MAP_SHARED mmaps align to 2 MiB stage-2 blocks to avoid HVF issues.
  - FS: `symlinkat` rewrites absolute guest targets to a relative path inside the sysroot; `pipe2` sets F_SETNOSIGPIPE when available; `sync_file_range` is stubbed.
  - Diagnostics: expand syscall debug logging to more path-bearing calls.

- Bug Fixes
  - Fork/clone correctness: carry TLS regs (TPIDRRO_EL0, TPIDR2_EL0), pointer-auth keys, and original SPSR; support CLONE_CHILD_{SETTID,CLEARTID} and write child_tid; smaller SCM_RIGHTS chunks.
  - Child lifecycle: kqueue-based child monitor raises SIGCHLD on host child exit (for all clones); also signal on proc_mark_child_exited for Linux-style wait behavior.
  - Security: include the dynamic linker’s high mapping in forked children so libc’s stack canary stays valid.
  - PTY keepalive: safer dup/snapshot for live and stale entries; avoid FD leaks and ensure CLOEXEC.
  - Signals/pipes: no manual SIGPIPE on write; rely on EPIPE and F_SETNOSIGPIPE to match Linux behavior.
  - Sysroot paths: validate parent directories before create operations to keep writes inside the sysroot.

<sup>Written for commit 22d8532213457a3158132cbe31f53aacb5ac7528. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/96?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/core/guest.c
+++ b/src/core/guest.c
@@ -1661,6 +1661,19 @@ int guest_get_used_regions(const guest_t *g,
         n++;
     }
 
+    /* Interpreter high block. The dynamic linker stores process-global state
+     * such as __stack_chk_guard in its high mapping just above interp_base.
+     * Fork children that take the region-copy path must inherit those bytes;
+     * otherwise libc's post-fork canary check observes zeroed guard storage
+     * and aborts before the child can exec.
+     */
+    if (n < max && g->interp_base > 0 &&
+        g->interp_base <= g->guest_size - BLOCK_2MIB) {
+        out[n].offset = g->interp_base;
+        out[n].size = BLOCK_2MIB;
+        n++;
+    }
+
     /* ELF + brk region: from elf_load_min (set by ELF loader) to brk_current.
      * The lower bound is the actual ELF load address, not ELF_DEFAULT_BASE:
      * ET_EXECs linked below 0x400000 (e.g. at 0x200000) have segments below the
--- a/src/runtime/fork-state.h
+++ b/src/runtime/fork-state.h
@@ -18,14 +18,30 @@
 /* Magic values for IPC frame delimiters */
 #define IPC_MAGIC_HEADER 0x454C464BU   /* "ELFK" */
 #define IPC_MAGIC_SENTINEL 0x454C4F4BU /* "ELOK" */
-/* Bumped to 11 when regions_tracker_stale was added to process state so forked
- * children preserve mprotect fast-path correctness.
+/* Bumped to 13 when pointer-authentication key registers and the remaining
+ * EL0 TLS registers were added so forked children and clone-created vCPUs
+ * resume with the same userspace CPU context as the parent. New Ubuntu arm64
+ * userspace can use PAC in libc and TLS-adjacent state during fork return.
+ *
+ * Bumped to 12 when clone_flags/child_tid_gva were added so fork-process
+ * children can apply CLONE_CHILD_SETTID/CLEARTID inside their own snapshot.
+ *
+ * Bumped to 11 when regions_tracker_stale was added to process state so
+ * forked children preserve mprotect fast-path correctness.
  *
  * Bumped to 10 when the rosetta placement / kbuf / ttbr1 tuple was added so a
  * rosetta-aware child rejects an older parent's header instead of trying to
  * interpret unknown trailing fields.
  */
-#define IPC_VERSION 11
+#define IPC_VERSION 13
+
+typedef struct {
+    uint64_t apiakeylo_el1, apiakeyhi_el1;
+    uint64_t apibkeylo_el1, apibkeyhi_el1;
+    uint64_t apdakeylo_el1, apdakeyhi_el1;
+    uint64_t apdbkeylo_el1, apdbkeyhi_el1;
+    uint64_t apgakeylo_el1, apgakeyhi_el1;
+} ipc_pauth_keys_t;
 
 typedef struct {
     uint32_t magic;
@@ -60,6 +76,8 @@ typedef struct {
     uint64_t rosetta_entry;
     uint64_t kbuf_gpa;
     uint64_t ttbr1;
+    uint64_t clone_flags;
+    uint64_t child_tid_gva;
 } ipc_header_t;
 
 typedef struct {
@@ -74,8 +92,10 @@ typedef struct {
      * access faults.
      */
     uint64_t ttbr1_el1;
-    uint64_t sctlr_el1, tcr_el1, mair_el1, cpacr_el1, tpidr_el0, sp_el1;
+    uint64_t sctlr_el1, tcr_el1, mair_el1, cpacr_el1;
+    uint64_t tpidr_el0, tpidrro_el0, tpidr2_el0, sp_el1;
     uint64_t x[31];
+    ipc_pauth_keys_t pauth_keys;
     vcpu_simd_state_t simd_state;
 } ipc_registers_t;
 
--- a/src/runtime/forkipc.c
+++ b/src/runtime/forkipc.c
@@ -22,6 +22,7 @@
 #include <sys/mman.h>
 #include <sys/stat.h>
 #include <sys/socket.h>
+#include <sys/event.h>
 #include <dirent.h> /* fdopendir, for DIR* reconstruction in child */
 #include <sys/wait.h>
 #include <sys/clonefile.h> /* fclonefileat for CoW shm snapshots */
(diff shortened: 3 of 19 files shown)


src/runtime/forkipc.c
```diff
@@ -48,10 +49,143 @@
...
+#undef SET_PAUTH_KEY
+
+    return HV_SUCCESS;
+}
+
 /* fork_child_main. */
 
 static int fork_child_vfork_notify_fd = -1;
 
+/* Linux clone flags */
```
Max042004, line 179: completely duplicated to forkipc.c:574–582

Max042004 (changes requested):

src/runtime/forkipc.c
doanbaotrung: Removed

src/syscall/io.c
```diff
@@ -833,10 +833,6 @@ static int64_t io_write_result(ssize_t ret)
     if (ret >= 0)
         return ret;
 
-    int saved_errno = errno;
-    if (saved_errno == EPIPE)
-        signal_queue(LINUX_SIGPIPE);
-    errno = saved_errno;
     return linux_errno();
```
jserv, line 836: SIGPIPE is no longer queued on EPIPE for sys_write. F_SETNOSIGPIPE in pipe2 only suppresses the host signal; the guest stops seeing SIGPIPE on broken-pipe write. Restore signal_queue(LINUX_SIGPIPE) on EPIPE here.

src/syscall/syscall.c
```diff
@@ -1952,8 +1953,6 @@ int syscall_dispatch(hv_vcpu_t vcpu, guest_t *g, int *exit_code, bool verbose)
                 result = ret;
                 goto fast_done;
             }
-            if (nr == SYS_write && errno == EPIPE)
-                signal_queue(LINUX_SIGPIPE);
             result = linux_errno();
```
jserv, line 1956: Same SIGPIPE regression on the dispatch fast path. Re-queue LINUX_SIGPIPE when errno == EPIPE before returning linux_errno().

src/syscall/mem.c
```diff
@@ -1887,6 +1873,19 @@ int64_t sys_mmap(guest_t *g,
...
 
+    size_t hps = host_page_size_cached();
+    uint64_t align = hps;
+    if (!is_fixed && !is_anon && fd >= 0 && (flags & LINUX_MAP_SHARED)) {
+        align = BLOCK_2MIB;
+    }
+
+    /* Round length up to align size (overflow-safe) */
+    if (length > UINT64_MAX - (align - 1))
+        return -LINUX_ENOMEM;
```
jserv, line 1884: Rounding length up to BLOCK_2MIB for file-backed MAP_SHARED turns a 4 KiB shm into a 2 MiB VMA. Tail access past EOF SIGBUSes and the Linux-visible length is wrong. Only 2 MiB-align placement (search start); keep length at PAGE_ALIGN_UP(length, 4 KiB).

src/syscall/fs.c
```diff
@@ -288,6 +288,136 @@ static int64_t reject_unsupported_fuse_path_op(const path_translation_t *tx)
...
+    char link_host[LINUX_PATH_MAX];
+    if (tx->host_path[0] == '/') {
+        if (str_copy_trunc(link_host, tx->host_path, sizeof(link_host)) >=
+            sizeof(link_host)) {
+            errno = ENAMETOOLONG;
+            return NULL;
+        }
+    } else {
+        char dir_host[LINUX_PATH_MAX];
+        if (fcntl(dir_ref->fd, F_GETPATH, dir_host) < 0)
```
jserv, line 403: fcntl(dir_ref->fd, F_GETPATH, ...) returns EBADF when dirfd is AT_FDCWD, so symlinkat(target="/abs", AT_FDCWD, rel-linkpath) fails with -EBADF. Branch on dir_ref->fd == AT_FDCWD and use getcwd(); fall back to the original guest target if path recovery still fails.

src/runtime/forkipc.c
```diff
@@ -762,6 +913,9 @@ static void *thread_create_and_run(void *arg)
     } else {
         WORKER_HV(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_TPIDR_EL0, tca->tpidr));
     }
```
jserv, line 915: Two lines below this added pauth restore, line 942 still has hv_vcpu_set_reg(vcpu, HV_REG_CPSR, 0) /* EL0t */. Same pattern at vm_clone_thread_run line 1252. fork_child_main was fixed to use regs.spsr_el1; these two in-process worker paths should also set HV_REG_CPSR to tca->spsr so parent NZCV/PSTATE survives the clone return.

src/syscall/proc-state.c
```diff
@@ -615,14 +624,6 @@ const char *proc_resolve_sysroot_create_path(const char *path,
...
 
-    /* Parent doesn't exist in sysroot. Only /tmp, /var/tmp, and ccache get
-     * forcefully redirected to the sysroot to avoid host case-collisions;
-     * everything else falls back to the host literal.
-     */
-    if (strncmp(path, "/tmp/", 5) && strncmp(path, "/var/tmp/", 9) &&
-        !strstr(path, "/.ccache/"))
-        return path;
-
     if (!create_parents) {
```
jserv, line 627: Removing the /tmp, /var/tmp, /.ccache host-literal fallback means O_CREAT under those paths now silently writes inside the sysroot (or fails) when the sysroot lacks the dir. Either restore the fallback or have sysroot bootstrap pre-create those directories.

src/syscall/proc-state.c
```diff
@@ -553,7 +553,16 @@ static const char *proc_resolve_sysroot_path_flags(const char *path,
...
-    return path;
+    char parent[LINUX_PATH_MAX];
+    str_copy_trunc(parent, buf, sizeof(parent));
+    char *slash = strrchr(parent, '/');
+    if (!slash || slash == parent)
+        return buf;
+    *slash = '\0';
+
+    if (sysroot_validate_dir_prefix(parent) < 0)
+        return NULL;
```
jserv, line 564: Returning NULL on parent prefix validation failure defeats the prior host-literal fallback for read paths. Return path (host-literal) on validation failure to preserve the old contract.

src/runtime/forkipc.c
```diff
@@ -48,10 +49,143 @@
...
+                     (int) host_pid, strerror(errno));
+        close(kq);
+        return NULL;
+    }
+
+    do {
+        errno = 0;
+    } while (kevent(kq, NULL, 0, &kev, 1, NULL) < 0 && errno == EINTR);
+    close(kq);
+    signal_queue(LINUX_SIGCHLD);
```
jserv, line 84: The kqueue loop sets errno=0 then waits; on non-EINTR kevent failure or zero events SIGCHLD is queued anyway. Also: no waitpid/status capture, no pidfd notify, no shutdown hook on exit_group. Gate signal_queue on (ret == 1 && (kev.fflags & NOTE_EXIT)); add pidfd notification; tie monitor lifetime to a shutdown flag.

src/runtime/procemu.c
```diff
@@ -1676,7 +1676,13 @@ static int pty_keepalive_register_locked(int master_host_fd,
     if (pty_keepalive_table[slot].slave_host_fd >= 0 &&
         pty_keepalive_table[slot].slave_host_fd != slave_host_fd)
         close(pty_keepalive_table[slot].slave_host_fd);
-    pty_keepalive_table[slot].slave_host_fd = slave_host_fd;
+    if (stale_open_once) {
+        pty_keepalive_table[slot].slave_host_fd = slave_host_fd;
+    } else {
+        if (slave_host_fd >= 0)
+            close(slave_host_fd);
```
jserv, line 1683: Closing slave_host_fd when stale_open_once is false defeats the keepalive's HUP suppression for live entries. If the intent is to let HUP propagate on real child close, split that into an explicit flag distinct from stale_open_once and add a regression test for master HUP behavior.

src/syscall/sys.c
```diff
@@ -487,17 +459,15 @@ int64_t sys_sched_rr_get_interval(guest_t *g, int pid, uint64_t ts_gva)
 
 int64_t sys_getgroups(guest_t *g, int size, uint64_t list_gva)
 {
-    int ngroups = get_cached_linux_groups();
-    if (ngroups < 0)
-        return linux_errno();
+    const int ngroups = 1;
```
jserv, line 462: Returning [proc_get_gid()] fabricates membership in the primary gid (Linux supplementary groups are independent of primary gid). Return ngroups=0 until elfuse implements setgroups.

src/core/guest.c
```diff
@@ -1661,6 +1661,19 @@ int guest_get_used_regions(const guest_t *g,
         n++;
     }
 
+    /* Interpreter high block. The dynamic linker stores process-global state
+     * such as __stack_chk_guard in its high mapping just above interp_base.
+     * Fork children that take the region-copy path must inherit those bytes;
+     * otherwise libc's post-fork canary check observes zeroed guard storage
+     * and aborts before the child can exec.
+     */
+    if (n < max && g->interp_base > 0 &&
```
jserv, line 1670: Hardcoded BLOCK_2MIB at interp_base may over- or under-copy a future dynamic linker. Track the interpreter's actual load_min..load_max in elf_resolve_interp and emit a region rounded from that. The comment about __stack_chk_guard is also misleading: that symbol normally lives in libc, not ld.so.

src/runtime/fork-state.h
```diff
@@ -18,14 +18,30 @@
...
+ *
+ * Bumped to 11 when regions_tracker_stale was added to process state so
+ * forked children preserve mprotect fast-path correctness.
  *
  * Bumped to 10 when the rosetta placement / kbuf / ttbr1 tuple was added so a
  * rosetta-aware child rejects an older parent's header instead of trying to
  * interpret unknown trailing fields.
  */
-#define IPC_VERSION 11
+#define IPC_VERSION 13
```
jserv, line 36: IPC_VERSION jumps 11 -> 13, skipping wire value 12. The magic-mismatch check still catches old children, but the gap reads like a rebase artifact. Either renumber to 12 or note in the comment that version 12 was rolled into the same release.

jserv (commented):
Flagging real Linux semantics divergences and a few behavioral regressions; the pauth/TLS reg plumbing and identity-driven AT_UID changes are good and worth keeping.

src/runtime/fork-state.h
jserv: `IPC_VERSION` was removed in #93
Don't touch this portion.

PR #98, opened 2026-06-15
Align non-fixed MAP_SHARED file mmap to 2 MiB

### LINKED ISSUE #94: Align guest MAP_SHARED allocations to stage-2 block boundaries (2 MiB)
## Issue Description
When a guest application makes consecutive non-fixed `MAP_SHARED` file-backed mappings (such as SHM pools created via `memfd_create`), subsequent `mmap` calls can fail with `-EFAULT` (Bad Address). 

This is not a failure of the host `mmap` itself, but rather a stage-2 page table boundary restriction inside `elfuse`'s memory manager.

---

## Root Cause
`elfuse` uses `2 MiB` blocks (`BLOCK_2MIB`) for hypervisor mapping segments in stage-2 translation. When a guest requests a file-backed `MAP_SHARED` mapping, `elfuse` installs a host file overlay over the guest anonymous mapping slab via `hvf_apply_file_overlay_quiesced`. This requires splitting the containing stage-2 segments using `hvf_segment_split`.

By default, the address space allocator `find_free_gap` aligns non-fixed allocations to the host page size (`16 KiB` on Apple Silicon). If consecutive mappings are allocated at arbitrary `16 KiB` boundaries, they can span across the boundary of an existing stage-2 segment that was split by a previous mapping.

For example, when `aligned_start` of a new mapping lands inside an existing split segment (`segments[1]`), but `aligned_end` extends beyond it into another segment (`segments[2]`), `hvf_segment_split` fails because it is not designed to split ranges crossing multiple segments:

```c
static int hvf_segment_split(guest_t *g,
                              uint64_t aligned_start,
                              uint64_t aligned_end)
{
    int idx = hvf_segment_find(g, aligned_start);
    if (idx < 0)
        return -LINUX_EFAULT;
    hvf_segment_t orig = g->segments[idx];
    if (aligned_end > orig.ipa + orig.len)
        return -LINUX_EFAULT; // <-- Fails here if aligned_end crosses segment boundary
```

---

## Evidence / Syscall Trace

A trace of consecutive `16 MiB` shared memory allocations illustrates the issue:

1. **First `16 MiB` mapping (Succeeds)**:
   ```
   syscall 222(0x0, 0x1000000, 0x3, 0x1, 0xe, 0x0) -> 0x21120c000
   ```
   - Target range: `[0x21120c000, 0x21220c000)`
   - Segment boundaries aligned to 2 MiB: `aligned_start = 0x210000000`, `aligned_end = 0x212400000`.
   - `hvf_segment_split` splits segment `[0, size)` into `[0, 0x210000000)`, `[0x210000000, 0x212400000)` (36MB), and `[0x212400000, size)`.

2. **Second `16 MiB` mapping (Fails)**:
   ```
   syscall 222(0x0, 0x1000000, 0x3, 0x1, 0xf, 0x0) -> -14 (EFAULT)
   ```
   - Due to `16 KiB` alignment, the allocator places the mapping at `0x21220c000` (right after the first mapping).
   - Target range: `[0x21220c000, 0x21320c000)`
   - Segment boundaries aligned to 2 MiB: `aligned_start = 0x212200000`, `aligned_end = 0x213400000`.
   - `aligned_start` (`0x212200000`) lies inside `[0x210000000, 0x212400000)` (`segments[1]`), but `aligned_end` (`0x213400000`) lies outside of it.
   - `hvf_segment_split` rejects this with `-LINUX_EFAULT` since it crosses segment boundaries.

---

## Proposed Fix
To prevent mappings from crossing hypervisor stage-2 block boundaries, we should align both the allocation address (`result_off`) and allocation `length` to `2 MiB` (`BLOCK_2MIB`) for all **non-fixed, shared, file-backed** `mmap` requests:

1. **Add an alignment parameter** to `find_free_gap` and `find_free_gap_inner`.
2. **Compute the alignment requirement** in `sys_mmap`:
   ```c
   size_t hps = host_page_size_cached();
   uint64_t align = hps;
   if (!is_fixed && !is_anon && fd >= 0 && (flags & LINUX_MAP_SHARED)) {
       align = BLOCK_2MIB;
   }
   ```
3. **Round up `length` and call `find_free_gap`** using the computed `align`.

jserv:
Non-fixed MAP_SHARED file-backed mmap allocations whose result landed mid 2 MiB block forced hvf_apply_file_overlay_quiesced to split the containing HVF stage-2 segment at both ends. Back-to-back memfd-style allocations burned segments at roughly two per mmap and could approach GUEST_MAX_HVF_SEGMENTS (256). The earlier multi-segment overlay fix handles correctness; this change reduces the segment-table churn.

find_free_gap and find_free_gap_inner gain a per-call alignment argument. sys_mmap passes BLOCK_2MIB when the allocation actually qualifies for the host overlay fast path (MAP_SHARED, host-page-aligned offset, writable backer) and host_page_size_cached() otherwise. The backing fd is now opened before the gap finder so overlay_fd_writable can gate the alignment decision; every failure path closes the ref.

Alignment is a best-effort placement preference, not a Linux-visible constraint:
  - The cached gap hint advances by host-page granularity so a small MAP_PRIVATE 4 KiB allocation can still occupy the trailing slack of the 2 MiB block.
  - When no 2 MiB-aligned gap is available, sys_mmap retries with host-page alignment instead of returning -ENOMEM.
  - For non-zero addr hints the path probes the exact host-page-aligned hint window first, falls back to 2 MiB alignment in the wider hint range, then to host-page alignment. A guest that pinpoints a free address still gets the address it asked for.

mremap stays at host-page alignment because the destination does not reinstall a file overlay.

Close #94

---
## Summary by cubic
Aligns non-fixed MAP_SHARED file-backed mmap to 2 MiB when eligible to reduce HVF stage-2 segment fragmentation and avoid segment exhaustion. Falls back to host-page alignment when needed and preserves exact hints; resolves #94.

- **Refactors**
  - Gap finder now accepts an alignment; sys_mmap uses 2 MiB for eligible MAP_SHARED allocations and retries at host-page if no 2 MiB gap exists; mremap stays host-page aligned.
  - Open the backing fd before placement to check writability and gate 2 MiB alignment; close on all failure paths.
  - Hint handling: probe the exact host-page-aligned window first, then try 2 MiB within the hint range, then host-page; gap hints still advance by host page so small MAP_PRIVATE mappings can use trailing slack. Added tests for hint fallback and back-to-back MAP_SHARED memfd mappings.

<sup>Written for commit d8b89522c002a197c506001fc7edb9badea4a641. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/98?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/mem.c
+++ b/src/syscall/mem.c
@@ -266,17 +266,21 @@ static void split_regions_at_boundary(guest_t *g, uint64_t boundary)
 static uint64_t find_free_gap_inner(const guest_t *g,
                                     uint64_t length,
                                     uint64_t min_addr,
-                                    uint64_t max_addr)
+                                    uint64_t max_addr,
+                                    uint64_t align)
 {
-    /* Round the search start up to the next host-page boundary so an unaligned
-     * addr hint cannot return a result that lands inside a host page already
-     * covered by a preceding region's overlay tail (the overlay extends to
+    /* Round the search start up to the requested alignment so an unaligned addr
+     * hint cannot return a result that lands inside a host page already covered
+     * by a preceding region's overlay tail (the overlay extends to
      * ALIGN_UP(r->end, hps)). Apple Silicon enforces 16 KiB host pages;
      * aligning to the guest 4 KiB page is not enough. Advance past each walked
-     * region to the same boundary for the same reason.
+     * region to the same boundary for the same reason. MAP_SHARED file-backed
+     * allocations may request 2 MiB alignment as a best-effort placement
+     * preference so consecutive mappings usually avoid sharing an HVF stage-2
+     * segment, which reduces segment-table fragmentation for memfd-style
+     * allocation patterns.
      */
-    size_t hps = host_page_size_cached();
-    uint64_t gap_start = ALIGN_UP(min_addr, hps);
+    uint64_t gap_start = ALIGN_UP(min_addr, align);
 
     /* Skip the prefix of regions entirely below gap_start in O(log n). After a
      * successful allocation the gap hint advances near or past the existing
@@ -307,8 +311,10 @@ static uint64_t find_free_gap_inner(const guest_t *g,
             g->regions[i].start >= gap_start + length)
             return gap_start;
 
-        /* Region overlaps; advance past it and round to the next host page */
-        gap_start = ALIGN_UP(g->regions[i].end, hps);
+        /* Region overlaps; advance past it and round to the next aligned
+         * boundary so the caller's alignment promise holds across allocations.
+         */
+        gap_start = ALIGN_UP(g->regions[i].end, align);
     }
 
     /* Check trailing space after all regions */
@@ -321,12 +327,16 @@ static uint64_t find_free_gap_inner(const guest_t *g,
  * The hint tracks the first address after the last successful mapping in each
  * region, which avoids rescanning the same prefix on sequential mmap activity.
  * A miss falls back to the region base so holes reopened by munmap are still
- * reusable.
+ * reusable. The align argument is the per-call start boundary the result must
+ * satisfy; some sys_mmap callers first pass BLOCK_2MIB as a best-effort
+ * placement preference for MAP_SHARED file-backed allocations, then retry with
+ * host-page alignment when no 2 MiB-aligned gap is available.
  */
 static uint64_t find_free_gap(guest_t *g,
                               uint64_t length,
                               uint64_t min_addr,
-                              uint64_t max_addr)
+                              uint64_t max_addr,
+                              uint64_t align)
 {
     /* RX and RW mappings advance independently, so keep separate hints. */
     uint64_t *hint =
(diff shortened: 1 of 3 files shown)


jserv:
@doanbaotrung , Please validate this PR.

doanbaotrung:
Look like it passed but I got another error when running command `ls -ail` in foot terminal
```
*** stack smashing detected *** terminated
```

jserv:
> Look like it passed but I got another error when running command `ls -ail` in foot terminal
> ```
> *** stack smashing detected *** terminated
> ```

elfuse + wilco

I am using [Rust rewrite of the GNU coreutils](https://github.com/uutils/coreutils), which works well under foot + elfuse.

PR #102, opened 2026-06-17
Fix trailing-slash mkdir under sysroot

### LINKED ISSUE #100: Bug: `mkdir`/`mkdirat` fails with ENOENT when directory path has a trailing slash
### Description
When running binaries under `elfuse` with a configured `--sysroot`, creating a directory where the path ends with a trailing slash (e.g. `mkdir("/var/lib/pacman/local/pambase-20260614-1/", 0755)`) fails with `ENOENT` ("No such file or directory"), even though the parent directory `/var/lib/pacman/local/` exists in the sysroot. 

Creating the same directory without the trailing slash succeeds.

### Steps to Reproduce
1. Set up a sysroot with an existing directory `/var/lib/pacman/local`.
2. Run a guest binary that invokes `mkdir` or `mkdirat` with a trailing slash:
   ```c
   mkdir("/var/lib/pacman/local/test-dir/", 0755);
   ```
3. The call fails with `ENOENT` instead of creating the directory in the sysroot.

### Root Cause Analysis
The issue resides in `proc_resolve_sysroot_create_path()` within `src/syscall/proc-state.c`. 

When checking if the parent directory exists, `elfuse` copies the resolved path into a `parent` buffer and finds the final component's separator slash using `strrchr()`:

```c
    char parent[LINUX_PATH_MAX];
    str_copy_trunc(parent, buf, sizeof(parent));
    char *slash = strrchr(parent, '/');
    if (!slash || slash == parent)
        return buf;

    *slash = '\0';
    if (access(parent, F_OK) == 0) { ... }
```

If the path ends with a trailing slash, `strrchr` matches the trailing slash itself. After null-terminating it, the `parent` path remains the target directory itself (which doesn't exist yet) rather than its actual parent. 

As a result, `access(parent, F_OK)` fails, and the function mistakenly falls back to returning the guest literal path (without the sysroot prefix). The host then attempts to run `mkdirat` on the literal host path `/var/lib/pacman/local/...` which fails with `ENOENT`.

jserv:
proc_resolve_sysroot_create_path located the leaf separator with strrchr on the resolved sysroot path. A guest path ending in a slash made strrchr match that trailing slash, so the parent buffer collapsed to the target itself. The parent-existence access() then failed and the resolver wrongly fell back to the host literal path (or auto-created the target as if it were its own parent), so mkdir("/dir/") reported ENOENT or EEXIST instead of creating the directory in the sysroot.

Drop trailing slashes before splitting off the leaf so the check sees the real parent. Guard all-slash guest paths ("/", "///") up front: they name the root, which always exists and has no parent, and trimming would otherwise walk strrchr into the sysroot prefix and trip the containment check (ELOOP) on a multi-component sysroot.

Close #100

---
## Summary by cubic
Fixes trailing-slash handling in sysroot path resolution so mkdir("/dir/") creates the directory under the sysroot instead of returning ENOENT/EEXIST or falling back to the host. Also correctly treats "/", "///" as the root.

- **Bug Fixes**
  - Trim trailing slashes before splitting the leaf in `proc_resolve_sysroot_create_path` so the parent check uses the real parent.
  - Handle all-slash guest paths as root and add tests for root mkdir and trailing-slash mkdir under "/tmp" to verify correct routing inside the sysroot.

<sup>Written for commit fb7bb95209e7c835ce7f8f64684df695a4a43184. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/102?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/proc-state.c
+++ b/src/syscall/proc-state.c
@@ -592,8 +592,24 @@ const char *proc_resolve_sysroot_create_path(const char *path,
         return NULL;
     }
 
+    /* An all-slash guest path ("/", "///") names the root, which always exists
+     * and has no parent to check; trimming it would walk strrchr into the
+     * sysroot prefix and trip the containment guard.
+     *
+     * Return buf as-is.
+     */
+    if (path[strspn(path, "/")] == '\0')
+        return buf;
+
     char parent[LINUX_PATH_MAX];
     str_copy_trunc(parent, buf, sizeof(parent));
+    /* Trailing slashes name the same directory (POSIX); drop them so the final
+     * separator splits off the leaf component rather than matching the trailing
+     * slash itself, which would leave the target as its own parent.
+     */
+    size_t plen = strlen(parent);
+    while (plen > 1 && parent[plen - 1] == '/')
+        parent[--plen] = '\0';
     char *slash = strrchr(parent, '/');
     if (!slash || slash == parent)
         return buf;
--- a/tests/test-sysroot-create-paths.c
+++ b/tests/test-sysroot-create-paths.c
@@ -6,8 +6,10 @@
 
 #include <errno.h>
 #include <fcntl.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <string.h>
+#include <sys/stat.h>
 #include <sys/xattr.h>
 #include <unistd.h>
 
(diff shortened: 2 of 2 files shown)


jserv:
@doanbaotrung , Please validate this PR.

doanbaotrung:
It worked!

PR #109, opened 2026-06-23
Support dynamic guest identity (UID/GID) configuration via environment variables

### LINKED ISSUE #103: Support dynamic guest identity (UID/GID) configuration via environment variables
#### **Problem Description**
Currently, `elfuse` hardcodes the guest environment's simulated user identity (UID and GID) to a static non-root user (e.g., `1000` / `1000`). 

This hardcoded configuration prevents running package managers (like `pacman`, `apt`, `apk`, or `dnf`) or other system administration utilities inside the simulated guest. When executed, these applications perform internal root checks (expecting UID `0`) and immediately abort with errors such as:
```
error: you cannot perform this operation unless you are root.
```

To run these tools successfully, the guest environment needs to simulate a root user identity (UID `0`, GID `0`) dynamically.

---

#### **Proposed Solution**
Allow users to dynamically configure the simulated guest UID and GID at runtime via environment variables, for example: `ELFUSE_GUEST_UID` and `ELFUSE_GUEST_GID`.

If these variables are present in the host process environment, `elfuse` should parse them and override the default guest UID/GID.

**Override parsed UID/GID at startup (`src/syscall/proc-identity.c`)**
   In `proc_identity_init()`, check the host environment for `ELFUSE_GUEST_UID` and `ELFUSE_GUEST_GID`. If defined, parse them (using `strtoul`) and override the default values of `emu_uid` and `emu_gid` instead of falling back to the hardcoded constants.

**Propagate dynamic identity to the guest process stack (`src/core/stack.c`)**
   In `build_linux_stack()`, the auxiliary vectors (`AT_UID`, `AT_EUID`, `AT_GID`, `AT_EGID`) are currently populated with hardcoded macros (e.g., `GUEST_UID` and `GUEST_GID`). 
   These should be replaced with calls to dynamic lookup helpers (such as `proc_get_uid()`, `proc_get_euid()`, etc.) so that the guest dynamic linker and standard library (`glibc`/`musl`) correctly read the dynamically configured simulated identity.

doanbaotrung:
Allow users to dynamically configure the simulated guest
UID and GID at runtime via environment variables, for
example: ELFUSE_GUEST_UID and ELFUSE_GUEST_GID.

If these variables are present in the host process
environment, elfuse should parse them and override
the default guest UID/GID.

Fixes #103

---
## Summary by cubic
Add runtime overrides for the simulated guest UID/GID via `ELFUSE_GUEST_UID` and `ELFUSE_GUEST_GID`. The values propagate to auxv and procfs so programs see the correct identity.

- **New Features**
  - Parse env vars in `proc_identity_init()` with strict uint32 parsing; ignore invalid/empty/out‑of‑range; fall back to `GUEST_UID`/`GUEST_GID` when unset.
  - Auxv `AT_UID/AT_EUID/AT_GID/AT_EGID` and `/proc/self/{status,task/*/status}` now use `proc_get_*` so real/effective/saved‑set and fs IDs reflect overrides.
  - Added `test-identity-override-host`; built and run by `make check`.

<sup>Written for commit da2339d0763508f4283ab9112b7b7e5d8187fdc5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/109?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/proc-identity.c
+++ b/src/syscall/proc-identity.c
@@ -6,6 +6,8 @@
 
 #include <stdatomic.h>
 #include <stdbool.h>
+#include <stdlib.h>
+#include <errno.h>
 #include <pthread.h>
 
 #include "syscall/abi.h"
@@ -24,16 +26,45 @@ static _Atomic int64_t guest_sid = 1, guest_pgid = 1;
 static _Atomic int64_t guest_fg_pgrp = 1;
 static _Atomic int32_t guest_has_ctty = 1;
 
+static bool parse_env_identity(const char *env_name, uint32_t *out_val)
+{
+    const char *env_str = getenv(env_name);
+    if (!env_str || *env_str == '\0') {
+        return false;
+    }
+    char *endptr = NULL;
+    errno = 0;
+    unsigned long val = strtoul(env_str, &endptr, 10);
+    const char *p = env_str;
+    while (*p == ' ' || *p == '\t' || *p == '\n' || *p == '\v' || *p == '\f' ||
+           *p == '\r') {
+        p++;
+    }
+    if (p != endptr && errno != ERANGE && *endptr == '\0' &&
+        val <= UINT32_MAX && *p >= '0' && *p <= '9') {
+        *out_val = (uint32_t) val;
+        return true;
+    }
+    return false;
+}
+
 void proc_identity_init(void)
 {
     guest_pid = 1;
     parent_pid = 0;
-    emu_uid = GUEST_UID;
-    emu_euid = GUEST_UID;
-    emu_suid = GUEST_UID;
-    emu_gid = GUEST_GID;
-    emu_egid = GUEST_GID;
-    emu_sgid = GUEST_GID;
+
+    uint32_t uid = GUEST_UID;
+    parse_env_identity("ELFUSE_GUEST_UID", &uid);
+
+    uint32_t gid = GUEST_GID;
+    parse_env_identity("ELFUSE_GUEST_GID", &gid);
+
+    emu_uid = uid;
+    emu_euid = uid;
+    emu_suid = uid;
+    emu_gid = gid;
+    emu_egid = gid;
+    emu_sgid = gid;
     emu_nice = 0;
     guest_sid = 1;
     guest_pgid = 1;
--- a/Makefile
+++ b/Makefile
@@ -151,6 +151,14 @@ $(BUILD_DIR)/test-fork-ipc-protocol-host: \
 	@echo "  LD      $@"
 	$(Q)$(CC) $(CFLAGS) -o $@ $^
 
+## Build the identity override host test (native macOS binary)
+$(BUILD_DIR)/test-identity-override-host: \
+		$(BUILD_DIR)/test-identity-override-host.o \
+		$(BUILD_DIR)/syscall/proc-identity.o | $(BUILD_DIR)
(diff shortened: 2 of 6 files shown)


src/syscall/proc-identity.c
```diff
@@ -28,12 +30,47 @@ void proc_identity_init(void)
...
+        const char *p = env_gid;
+        while (*p == ' ' || *p == '\t' || *p == '\n' || *p == '\v' ||
+               *p == '\f' || *p == '\r') {
+            p++;
+        }
+        if (p != endptr && errno != ERANGE && *endptr == '\0' &&
+            val <= UINT32_MAX && *p >= '0' && *p <= '9') {
+            gid = (uint32_t) val;
+        }
+    }
```
jserv: Eliminate code duplication.

doanbaotrung: Updated!

jserv (changes requested):
Use C-style comments (`/* ... */`),

doanbaotrung:
Fixed

jserv:
I am not sure whether `getenv` is an appropriate mechanism for specifying the desired UID/GID. Instead, I am considering adopting the same approach as [fakeroot](https://manpages.ubuntu.com/manpages/trusty/man1/fakeroot-tcp.1.html).

Max042004:
fakeroot does two things at once:

1. **Process identity** — it intercepts `getuid()/geteuid()` so the program
   believes it is root.
2. **File ownership** — it intercepts `chown()/stat()` so ownership-sensitive
   operations stay self-consistent without real privilege.

We already implement (2). `src/syscall/chown-overlay.c` is literally headed
*"Virtual chown overlay (fakeroot-style)"*:

So I prefer reframing it as the identity half of the existing overlay:
switch the interface to a fakeroot mode defaulting to root, and complete
`capget`/`getgroups` for the emulated-root case.

- **Interface.** fakeroot's model is *"run this and you are root"* — a mode, not a
  free-form numeric knob. I'd rather expose a single switch (e.g. `--fakeroot` /
  `ELFUSE_FAKEROOT=1`) that sets `uid=euid=suid=0, gid=0` in one go, aligned with
  the `chown-overlay.c` naming, instead of asking users to hand-set two arbitrary
  integers.
- **Consistency.** To truly behave like root the whole identity surface has to
  agree. One real gap today: `sc_capget()` (`syscall.c:1293`) returns an empty
  capability set regardless of uid, so a tool that reads `/proc/self/status`
  `CapEff` or checks a specific capability still sees "not root". A named fakeroot
  mode makes it natural to flip `capget` (full caps) and `getgroups` (gid 0)
  together; a raw UID env var leaves those half-done.

doanbaotrung:
@Max042004 You mean this changes right?
https://github.com/sysprog21/elfuse/pull/114

We can close this PR and continue for fakeroot on my new PR https://github.com/sysprog21/elfuse/pull/114

Max042004 (approved):

jserv:
Close in favor of #114

PR #110, opened 2026-06-25
Dangling Symlink Resolution and Missing sync_file_range and syncfs System Calls

doanbaotrung:
the `sys_renameat2` and `sys_linkat` system calls
were using the resolving PATH_TR_NONE translation
mode for the source paths.
This caused elfuse to follow the symlink
leaf names to their targets, causing
dangling symlinks during package extraction
to fail with ENOENT (No such file or directory).
We updated these to use PATH_TR_NOFOLLOW for
the leaf components, preventing path resolution
from walking past the symlink itself.

Missing sync_file_range and syncfs System
Calls: We registered and defined
SYS_sync_file_range (84) and
SYS_syncfs (267) in abi.h and dispatch.tbl,
and implemented them in syscall.c as
aliases of sc_fsync_common.

---
## Summary by cubic
Fixes dangling symlink handling in `renameat2`/`linkat` to prevent ENOENT during package extraction. Adds `SYS_sync_file_range` (84) and `SYS_syncfs` (267) with minimal Linux-compatible behavior and smoke tests.

- **Bug Fixes**
  - `renameat2` and `linkat` stop following leaf symlinks (`PATH_TR_NOFOLLOW`); `linkat` honors `AT_SYMLINK_FOLLOW` for the source. New targets use `PATH_TR_CREATE | PATH_TR_NOFOLLOW`.

- **New Features**
  - Registered and implemented `SYS_sync_file_range`/`SYS_syncfs`. `sync_file_range` validates fd/offset/len, allows only flag bits 1/2/4, returns immediately for WRITE-only, calls `fsync(fd)` when WAIT flags are set, and returns `ESPIPE` on non-regular fds (e.g., pipes). `syncfs` validates the fd then calls `sync()`. Added smoke tests covering success and errors (`EINVAL`, `EBADF`, `ESPIPE`).

<sup>Written for commit bc965a309df83f0234bfb81b4965104defdf5919. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/110?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/dispatch.tbl
+++ b/src/syscall/dispatch.tbl
@@ -100,6 +100,8 @@ SYS_fremovexattr sc_fremovexattr 1
 SYS_sync sc_sync 1
 SYS_fsync sc_fsync 1
 SYS_fdatasync sc_fdatasync 1
+SYS_sync_file_range sc_sync_file_range 1
+SYS_syncfs sc_syncfs 0
 SYS_msync sc_msync 0
 SYS_membarrier sc_membarrier 0
 
--- a/src/syscall/syscall.c
+++ b/src/syscall/syscall.c
@@ -1236,6 +1236,99 @@ static int64_t sc_fsync_common(guest_t *g,
 #define sc_fsync sc_fsync_common
 #define sc_fdatasync sc_fsync_common
 
+static int64_t sc_sync_file_range(guest_t *g,
+                                  uint64_t x0,
+                                  uint64_t x1,
+                                  uint64_t x2,
+                                  uint64_t x3,
+                                  uint64_t x4,
+                                  uint64_t x5,
+                                  bool verbose)
+{
+    (void) g;
+    (void) x4;
+    (void) x5;
+    (void) verbose;
+
+    int fd = (int) x0;
+    int64_t offset = (int64_t) x1;
+    int64_t nbytes = (int64_t) x2;
+    unsigned int flags = (unsigned int) x3;
+
+    if (offset < 0 || nbytes < 0 || INT64_MAX - offset < nbytes)
+        return -LINUX_EINVAL;
+
+    host_fd_ref_t host_ref;
+    int64_t err = host_fd_ref_open_io(fd, &host_ref);
+    if (err < 0)
+        return err;
+
+    struct stat st;
+    if (fstat(host_ref.fd, &st) == 0 && !S_ISREG(st.st_mode) &&
+        !S_ISBLK(st.st_mode) && !S_ISDIR(st.st_mode)) {
+        host_fd_ref_close(&host_ref);
+        return -LINUX_ESPIPE;
+    }
+
+    /* Validate flags: only bits 1, 2, 4 are valid */
+    if (flags & ~7u) {
+        host_fd_ref_close(&host_ref);
+        return -LINUX_EINVAL;
+    }
+
+    /*
+     * If the flags only ask to initiate asynchronous write-out without waiting
+     * (i.e. SYNC_FILE_RANGE_WRITE (2)), we return 0 immediately to avoid
+     * blocking. The host OS's background page-out daemon will handle flushing
+     * dirty buffers. WAIT_BEFORE (1) and WAIT_AFTER (4) require blocking until
+     * writes complete.
+     *
+     * Note on macOS/Darwin: macOS does not provide a system call equivalent to
+     * Linux's sync_file_range(2) that can synchronize file data without writing
+     * back metadata. Therefore, we use fsync() to accomplish the
+     * synchronization, which will also write back metadata, unlike native Linux
+     * sync_file_range(2).
+     */
+    int64_t ret = 0;
+    if (flags & (1u | 4u)) {
+        ret = (fsync(host_ref.fd) < 0) ? linux_errno() : 0;
+    }
+
+    host_fd_ref_close(&host_ref);
+    return ret;
+}
+
+static int64_t sc_syncfs(guest_t *g,
+                         uint64_t x0,
+                         uint64_t x1,
+                         uint64_t x2,
+                         uint64_t x3,
+                         uint64_t x4,
+                         uint64_t x5,
+                         bool verbose)
+{
+    (void) g;
+    (void) x1;
+    (void) x2;
+    (void) x3;
+    (void) x4;
+    (void) x5;
+    (void) verbose;
+
+    int fd = (int) x0;
+    host_fd_ref_t host_ref;
+    int64_t err = host_fd_ref_open_io(fd, &host_ref);
+    if (err < 0)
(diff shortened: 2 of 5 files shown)


Max042004 (changes requested):
Please add smoke test for two new syscall.

src/syscall/syscall.c
```diff
@@ -1236,6 +1236,83 @@ static int64_t sc_fsync_common(guest_t *g,
...
+    /*
+     * If the flags only ask to initiate asynchronous write-out without waiting
+     * (i.e. SYNC_FILE_RANGE_WRITE (2)), we return 0 immediately to avoid
+     * blocking. The host OS's background page-out daemon will handle flushing
+     * dirty buffers. WAIT_BEFORE (1) and WAIT_AFTER (4) require blocking until
+     * writes complete.
+     */
+    int64_t ret = 0;
+    if (flags & (1u | 4u)) {
+        ret = (fsync(host_ref.fd) < 0) ? linux_errno() : 0;
```
Max042004, line 1294: sync_file_range(2) on native Linux don't write back metadata. But MacOS don't provide syscall to accomplish such feature. fsync() will write back metadata.
So please add comment for mentioning different behavior comparing to Linux sync_file_range(2)

src/syscall/syscall.c
```diff
@@ -1236,6 +1236,83 @@ static int64_t sc_fsync_common(guest_t *g,
...
+    int fd = (int) x0;
+    unsigned int flags = (unsigned int) x3;
+
+    host_fd_ref_t host_ref;
+    int64_t err = host_fd_ref_open_io(fd, &host_ref);
+    if (err < 0)
+        return err;
+
+    /* Validate flags: only bits 1, 2, 4 are valid */
+    if (flags & ~7u) {
```
Max042004, line 1274: Add offset, nbytes invalid check 
```
if (offset < 0 || nbytes < 0 || offset + nbytes < offset)
    return -LINUX_EINVAL;
```

Max042004 (changes requested):

src/syscall/syscall.c
```diff
@@ -1236,6 +1236,92 @@ static int64_t sc_fsync_common(guest_t *g,
...
+    int fd = (int) x0;
+    int64_t offset = (int64_t) x1;
+    int64_t nbytes = (int64_t) x2;
+    unsigned int flags = (unsigned int) x3;
+
+    if (offset < 0 || nbytes < 0 || INT64_MAX - offset < nbytes)
+        return -LINUX_EINVAL;
+
+    host_fd_ref_t host_ref;
+    int64_t err = host_fd_ref_open_io(fd, &host_ref);
```
Max042004, line 1262: Mainline Linux rejects `sync_file_range()` on a descriptor that is not a
regular file, block device, or directory, returning **`-ESPIPE`**.
The current shim does not
classify the fd, so calling it on a pipe / socket / fifo / char device. Linux returns `-ESPIPE` in both cases.

**Suggested fix** (after `host_fd_ref_open_io` succeeds):

```c
struct stat st;
if (fstat(host_ref.fd, &st) == 0 &&
    !S_ISREG(st.st_mode) && !S_ISBLK(st.st_mode) && !S_ISDIR(st.st_mode)) {
    host_fd_ref_close(&host_ref);
    return -LINUX_ESPIPE;
}
```

src/syscall/dispatch.tbl
```diff
@@ -100,6 +100,8 @@ SYS_fremovexattr sc_fremovexattr 1
 SYS_sync sc_sync 1
 SYS_fsync sc_fsync 1
 SYS_fdatasync sc_fdatasync 1
+SYS_sync_file_range sc_sync_file_range 1
+SYS_syncfs sc_syncfs 1
```
Max042004: `scripts/gen-syscall-dispatch.py:141-142`:

```
 * E=0 means only x0-x2 are required by the wrapper.
 * E=1 means x3-x5 may also be consumed by the wrapper.
 ```

`sc_syncfs` (`src/syscall/syscall.c:1294`) only uses `x0` and explicitly voids
the rest.

**Suggested fix:**

```
SYS_syncfs sc_syncfs 0
```

**Review by Max042004 — CHANGES_REQUESTED** [review]

**Comment by Max042004** [comment]
The added tests pass (`test-syscall-smoke`: 21 passed, 0 failed under elfuse)
and `scripts/check-syscall-coverage.py` reports `PASS` — both syscalls are
recognized as covered. At the *branch* level, however, `test_sync_file_range`
only ever passes `flags` values of `7` and `8`, so two branches of
`sc_sync_file_range` are never hit:

| Branch in `src/syscall/syscall.c` | Exercised? | By |
| --- | --- | --- |
| `offset/nbytes/overflow → -EINVAL` (1258) | yes | neg-offset / neg-nbytes / overflow cases |
| `host_fd_ref_open_io → -EBADF` (1262) | **no** | — |
| `flags & ~7u → -EINVAL` (1267) | yes | `flags = 8` case |
| `flags & (1\|4)` **true** → `fsync()` (1286) | yes | `flags = 7` case |
| `flags & (1\|4)` **false** → `return 0` (1285) | **no** | — |

The untested `return 0` branch is the most notable gap: it is the
`SYNC_FILE_RANGE_WRITE`-only (async-hint) path, and it is exactly the branch
where this shim **intentionally diverges** from Linux (returns success without
initiating any write-out). It currently has zero coverage. The `-EBADF` path for
`sync_file_range` is also untested (`syncfs` does test its `-EBADF`).

Both gaps close with two extra cases that also pass on native Linux (so no
qemu-mode divergence):

```c
/* SYNC_FILE_RANGE_WRITE only (async hint) returns 0 without blocking.
 * Covers the deliberate-divergence branch. */
if (syscall(SYS_sync_file_range, fd, (off64_t) 0, (off64_t) 0, 2) != 0) {
    FAIL("sync_file_range write-only");
    close(fd);
    return;
}

/* Bad fd → EBADF. */
errno = 0;
if (syscall(SYS_sync_file_range, -1, (off64_t) 0, (off64_t) 0, 7) != -1 ||
    errno != EBADF) {
    FAIL("sync_file_range bad fd");
    close(fd);
    return;
}
```

`test_syncfs` is fine — both of its branches (success and `-EBADF`) are covered.

doanbaotrung:
Updated

Max042004 (approved):

jserv:

PR #112, opened 2026-06-25
Emulated Netlink sockets cause infinite busy loops under event-loop

### LINKED ISSUE #105: Emulated Netlink sockets cause infinite busy loops under event-loop pollers (e.g., avahi-daemon)
### Problem Description
On macOS hosts, `elfuse` emulates Linux netlink route sockets (`AF_NETLINK`, `NETLINK_ROUTE`) using a host pipe. 

Currently, the write end of this pipe (`pipefd[1]`) is closed immediately at socket creation in `netlink_socket()`. This makes the read end of the pipe (`pipefd[0]`) permanently readable, returning EOF (`0` bytes) when read.

When guest applications (such as `avahi-daemon`) run event loops that wait for netlink messages using `ppoll()` or `select()`, the emulated descriptor immediately returns `POLLIN` because the write end is closed. The application's subsequent `recvmsg` or `read` call returns `0` bytes. Because no sender credentials can be resolved from a `0`-byte read, the application ignores the read and immediately queries `ppoll()` again. This causes the guest application to spin in an infinite busy-loop consuming 100% CPU.

### Proposed Fix
Implement a non-blocking self-pipe signaling mechanism inside `src/syscall/netlink.c`:

1. **Retain the write descriptor**: Keep the write descriptor of the pipe (`pipe_wr`) open in the netlink socket state (`netlink_state_t`) instead of closing it at socket creation. Ensure it is closed during `netlink_close()` to prevent file descriptor leaks.
2. **Signal readability**: Write a dummy byte to `pipe_wr` when populating synthetic responses in the netlink buffer.
3. **Clear readability**: Read the dummy byte back from the host descriptor once the guest has fully drained the buffer.
4. **Return `-EAGAIN`**: Change `netlink_recvmsg` and `netlink_read` to return `-EAGAIN` (instead of `0`) when the buffer is empty, indicating standard non-blocking socket behavior.

**Location**: `src/syscall/netlink.c`.

doanbaotrung:
When guest applications (such as avahi-daemon) run event loops that
wait for netlink messages using ppoll() or select(), the emulated
descriptor immediately returns POLLIN because the write end is
closed. The application's subsequent recvmsg or read call returns 0
bytes.

Because no sender credentials can be resolved from a 0-byte read,
the application ignores the read and immediately queries ppoll()
again. This causes the guest application to spin in an infinite
busy-loop consuming 100% CPU.

To address this, implement a non-blocking self-pipe signaling
mechanism inside src/syscall/netlink.c:

1. Store and use pipe_rd in netlink_state_t directly to avoid
   lockless global state reads in netlink_clear_readable.
2. Honor MSG_DONTWAIT and O_NONBLOCK flags and implement proper
   blocking semantics by polling the pipe outside nl_lock.
3. Signal readability only on empty-to-nonempty transition to
   prevent pipe buffer drift.
4. Handle zero-length reads/recvs before empty-buffer check.

Close #105

---
## Summary by cubic
Fixes infinite busy loops in emulated Netlink sockets by signaling readiness only when data exists and enforcing correct blocking. Stops 100% CPU spin in ppoll/select loops (e.g., avahi-daemon) and honors MSG_DONTWAIT/O_NONBLOCK.

- **Bug Fixes**
  - Add a per-socket non-blocking self-pipe; store rd/wr in state (avoid lockless reads), set both O_NONBLOCK, keep write end open, and clean up on close.
  - Signal only on empty→non-empty; drain the pipe when buffers empty; handle 0-length read/recv/recvmsg first to avoid spurious POLLIN.
  - Enforce blocking correctly: return -EAGAIN for non-blocking; otherwise poll the pipe outside `nl_lock`; pass flags from `sys_recvfrom` to `netlink_recv`.

<sup>Written for commit 035e92a3fc949e48931a3e2cb05baf7b98aa4915. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/112?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/netlink.c
+++ b/src/syscall/netlink.c
@@ -33,6 +33,12 @@
 #include "syscall/abi.h"
 #include "syscall/internal.h"
 #include "syscall/net.h"
+#include "utils.h"
+#include <poll.h>
+
+#ifndef LINUX_MSG_DONTWAIT
+#define LINUX_MSG_DONTWAIT 0x40
+#endif
 
 static void netlink_close(int guest_fd);
 
@@ -130,6 +136,8 @@ typedef struct {
     size_t buf_pos; /* Current read position */
     uint32_t seq;   /* Sequence number from last request */
     uint32_t pid;   /* Bound PID (from bind or auto-assigned) */
+    int pipe_wr;    /* Host pipe write descriptor */
+    int pipe_rd;    /* Host pipe read descriptor */
 } netlink_state_t;
 
 static netlink_state_t nl_state[MAX_NETLINK_FDS];
@@ -156,6 +164,8 @@ static netlink_state_t *nl_alloc(int guest_fd)
         memset(s, 0, sizeof(*s));
         s->in_use = true;
         s->guest_fd = guest_fd;
+        s->pipe_wr = -1;
+        s->pipe_rd = -1;
         s->buf = malloc(NETLINK_BUF_SIZE);
         if (!s->buf) {
             s->in_use = false;
@@ -165,6 +175,26 @@ static netlink_state_t *nl_alloc(int guest_fd)
         return s;
     }
     return NULL;
+}
+
+static void netlink_signal_readable(netlink_state_t *ns)
+{
+    if (ns->pipe_wr != -1) {
+        uint8_t dummy = 1;
+        (void) write(ns->pipe_wr, &dummy, 1);
+    }
+}
+
+static void netlink_clear_readable(netlink_state_t *ns)
+{
+    int host_fd = ns->pipe_rd;
+    if (host_fd < 0)
+        return;
+
+    uint8_t dummy[128];
+    while (read(host_fd, dummy, sizeof(dummy)) > 0) {
+        /* Drain non-blocking pipe */
+    }
 }
 
 /* Append a netlink attribute to the buffer. Returns bytes written. */
@@ -423,6 +453,12 @@ int64_t netlink_socket(int protocol, int type)
     if (pipe(pipefd) < 0)
         return -LINUX_EMFILE;
 
+    if (fd_set_nonblock(pipefd[0]) < 0 || fd_set_nonblock(pipefd[1]) < 0) {
+        close(pipefd[0]);
+        close(pipefd[1]);
+        return -LINUX_EMFILE;
+    }
+
     int gfd = fd_alloc(FD_NETLINK, pipefd[0], netlink_close);
     if (gfd < 0) {
         close(pipefd[0]);
@@ -438,10 +474,8 @@ int64_t netlink_socket(int protocol, int type)
         return -LINUX_ENOMEM;
     }
 
-    /* No poll wakeup fd is needed because recvmsg drains the buffered response
-     * directly.
-     */
-    close(pipefd[1]); /* The write end is unused */
+    ns->pipe_wr = pipefd[1];
+    ns->pipe_rd = pipefd[0];
 
     return gfd;
 }
@@ -604,7 +638,12 @@ int64_t netlink_sendmsg(int guest_fd, guest_t *g, uint64_t msg_gva, int flags)
         goto out;
     }
 
+    bool was_empty = ns->buf_pos >= ns->buf_len;
     int ret = nl_process_request(ns, req, rlen);
+    if (ret == 0) {
+        if (was_empty && ns->buf_pos < ns->buf_len)
+            netlink_signal_readable(ns);
+    }
     result = (ret < 0) ? ret : (int64_t) iov.iov_len;
 
 out:
@@ -635,7 +674,12 @@ int64_t netlink_send(int guest_fd, guest_t *g, uint64_t buf_gva, uint64_t len)
         goto out;
     }
 
+    bool was_empty = ns->buf_pos >= ns->buf_len;
     int ret = nl_process_request(ns, req, rlen);
+    if (ret == 0) {
+        if (was_empty && ns->buf_pos < ns->buf_len)
+            netlink_signal_readable(ns);
+    }
     result = (ret < 0) ? ret : (int64_t) len;
 
 out:
@@ -650,6 +694,7 @@ int64_t netlink_recv(int guest_fd,
                      guest_t *g,
                      uint64_t buf_gva,
                      uint64_t len,
+                     int flags,
                      uint64_t src_gva,
                      uint64_t addrlen_gva)
 {
@@ -660,9 +705,42 @@ int64_t netlink_recv(int guest_fd,
         return -LINUX_EBADF;
     }
 
-    if (ns->buf_pos >= ns->buf_len) {
+    if (len == 0) {
         pthread_mutex_unlock(&nl_lock);
         return 0;
+    }
+
+    /* Wait for data to become available. If the buffer is empty, block
+     * on the host pipe read end. Honor MSG_DONTWAIT and O_NONBLOCK flags.
+     */
+    while (ns->buf_pos >= ns->buf_len) {
+        bool nonblock = (flags & LINUX_MSG_DONTWAIT) ||
+                        (fd_table[guest_fd].linux_flags & LINUX_O_NONBLOCK);
+        if (nonblock) {
+            pthread_mutex_unlock(&nl_lock);
+            return -LINUX_EAGAIN;
+        }
+
+        int rd_fd = ns->pipe_rd;
+        pthread_mutex_unlock(&nl_lock);
+
+        struct pollfd pfd = {
+            .fd = rd_fd,
+            .events = POLLIN,
+        };
+        int ret = poll(&pfd, 1, -1);
+        if (ret < 0) {
+            if (errno == EINTR)
+                return -LINUX_EINTR;
+            return -LINUX_EIO;
+        }
+
+        pthread_mutex_lock(&nl_lock);
+        netlink_state_t *current_ns = nl_find(guest_fd);
+        if (!current_ns || current_ns != ns) {
+            pthread_mutex_unlock(&nl_lock);
+            return -LINUX_EBADF;
+        }
     }
 
     size_t avail = ns->buf_len - ns->buf_pos;
@@ -688,6 +766,9 @@ int64_t netlink_recv(int guest_fd,
         return -LINUX_EFAULT;
     }
     ns->buf_pos += msg_end;
+
+    if (ns->buf_pos >= ns->buf_len)
+        netlink_clear_readable(ns);
 
     if (src_gva && addrlen_gva) {
         sockaddr_nl_t snl = {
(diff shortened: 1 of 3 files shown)


doanbaotrung:
Fix #105

Max042004 (changes requested):
Add a SEQPACKET test.

src/syscall/netlink.c
```diff
@@ -167,6 +170,26 @@ static netlink_state_t *nl_alloc(int guest_fd)
...
+{
+    if (ns->pipe_wr != -1) {
+        uint8_t dummy = 1;
+        (void) write(ns->pipe_wr, &dummy, 1);
+    }
+}
+
+static void netlink_clear_readable(int guest_fd)
+{
+    int host_fd = fd_to_host(guest_fd);
```
jserv: Draining via `fd_to_host(guest_fd)` reaches back into the fd table through a lockless read while only `nl_lock` is held. If another thread closes this guest fd and the slot gets reused for a different host fd between `nl_find()` and here, this drains the wrong fd (or misses the real pipe). The read end is already known at socket() time -- store it in `netlink_state_t` (e.g. `pipe_rd`) next to `pipe_wr`, and drain that exact fd. That also drops the dependency on lockless global state entirely.

src/syscall/netlink.c
```diff
@@ -662,7 +694,7 @@ int64_t netlink_recv(int guest_fd,
 
     if (ns->buf_pos >= ns->buf_len) {
         pthread_mutex_unlock(&nl_lock);
-        return 0;
+        return -LINUX_EAGAIN;
```
jserv: Empty buffer now returns `-EAGAIN` unconditionally, regardless of `O_NONBLOCK` / `MSG_DONTWAIT`. This fixes the reported EOF busy-loop, but it is not Linux blocking semantics: a blocking recv with no data should block, not error. avahi polls first so it is fine in practice, but a blocking netlink client can end up in its own retry loop. Consider honoring `MSG_DONTWAIT`/`O_NONBLOCK`: for blocking mode, wait on the pipe outside `nl_lock`, then recheck `ns->buf`. Same pattern repeats at lines 787 (recvmsg) and 882 (read).

src/syscall/netlink.c
```diff
@@ -605,6 +631,9 @@ int64_t netlink_sendmsg(int guest_fd, guest_t *g, uint64_t msg_gva, int flags)
     }
 
     int ret = nl_process_request(ns, req, rlen);
```
jserv, line 642: `netlink_signal_readable()` writes a token on every successful send. Since each request overwrites `ns->buf` (buf_pos reset to 0), repeated sends without an intervening recv keep pushing bytes into the pipe until it fills and the writes silently drop on EAGAIN. Harmless today, but it makes the pipe level drift from `ns->buf` state and can mask a real missed-wakeup later. Signal only on the empty-to-nonempty transition: capture `bool was_empty = ns->buf_pos >= ns->buf_len` before `nl_process_request()` and write only when it goes non-empty. Same at line 667 (netlink_send).

src/syscall/netlink.c
```diff
@@ -423,6 +446,12 @@ int64_t netlink_socket(int protocol, int type)
     if (pipe(pipefd) < 0)
         return -LINUX_EMFILE;
 
+    if (fd_set_nonblock(pipefd[0]) < 0 || fd_set_nonblock(pipefd[1]) < 0) {
```
jserv, line 456: `fd_set_nonblock()` failure returns `-LINUX_EMFILE`, which is misleading -- this is an fcntl failure, not fd exhaustion. Preserve the real errno (return `-linux_errno()` from the failing call) after closing both pipe fds.

src/syscall/netlink.c
```diff
@@ -839,7 +879,7 @@ int64_t netlink_read(int guest_fd, guest_t *g, uint64_t buf_gva, uint64_t count)
 
     if (ns->buf_pos >= ns->buf_len) {
         pthread_mutex_unlock(&nl_lock);
-        return 0;
+        return -LINUX_EAGAIN;
```
jserv: Minor: with an empty buffer this returns `-EAGAIN` even for a zero-length read, but a POSIX zero-length `read()` should return 0. Handle `count == 0` (and the recv `len == 0` case) before the empty-buffer check.

jserv (commented):
Netlink busy-loop fix review. The core approach is sound: keeping the pipe write end open and signaling readiness only when a response is buffered fixes the spurious peer-closed POLLIN that spun avahi-daemon. A few correctness and robustness points below.

Append `Close #105` at the end of git commit messages.

doanbaotrung:
Address netlink review feedback on busy-loop fix
1. Store and use pipe_rd in netlink_state_t directly to avoid
   lockless global state reads in netlink_clear_readable.
2. Honor MSG_DONTWAIT and O_NONBLOCK flags and implement proper
   blocking semantics by polling the pipe outside nl_lock.
3. Signal readability only on empty-to-nonempty transition to
   prevent pipe buffer drift.
4. Handle zero-length reads/recvs before empty-buffer check.

jserv (changes requested):
Squash commits.

src/syscall/netlink.c
```diff
@@ -660,11 +707,42 @@ int64_t netlink_recv(int guest_fd,
...
 
-    if (ns->buf_pos >= ns->buf_len) {
+    if (len == 0) {
         pthread_mutex_unlock(&nl_lock);
         return 0;
     }
 
+    while (ns->buf_pos >= ns->buf_len) {
+        bool nonblock = (flags & LINUX_MSG_DONTWAIT) ||
+                        (fd_table[guest_fd].linux_flags & LINUX_O_NONBLOCK);
```
jserv, line 718: Add proper comments.

src/syscall/netlink.c
```diff
@@ -771,6 +847,42 @@ int64_t netlink_recvmsg(int guest_fd, guest_t *g, uint64_t msg_gva, int flags)
...
+
+        struct pollfd pfd = {
+            .fd = rd_fd,
+            .events = POLLIN,
+        };
+        int ret = poll(&pfd, 1, -1);
+        if (ret < 0) {
+            if (errno == EINTR) {
+                return -LINUX_EINTR;
+            }
```
jserv: Simply rewrite as following:
```c
            if (errno == EINTR)
                return -LINUX_EINTR;
```

src/syscall/netlink.c
```diff
@@ -804,6 +916,10 @@ int64_t netlink_recvmsg(int guest_fd, guest_t *g, uint64_t msg_gva, int flags)
 
     ns->buf_pos += msg_end;
 
+    if (ns->buf_pos >= ns->buf_len) {
```
jserv: Ditto. Use brackets only when necessary.

src/syscall/netlink.c
doanbaotrung: oh my bad. Fixed.

src/syscall/netlink.c
doanbaotrung: Fixed.

src/syscall/netlink.c
doanbaotrung: Added

src/syscall/netlink.c
```diff
@@ -604,7 +638,13 @@ int64_t netlink_sendmsg(int guest_fd, guest_t *g, uint64_t msg_gva, int flags)
         goto out;
     }
 
+    bool was_empty = ns->buf_pos >= ns->buf_len;
     int ret = nl_process_request(ns, req, rlen);
+    if (ret == 0) {
+        if (was_empty && ns->buf_pos < ns->buf_len) {
+            netlink_signal_readable(ns);
+        }
```
jserv: Use brackets only when necessary.

src/syscall/netlink.c
```diff
@@ -635,7 +675,13 @@ int64_t netlink_send(int guest_fd, guest_t *g, uint64_t buf_gva, uint64_t len)
...
```
jserv: Use brackets only when necessary.

src/syscall/netlink.c
```diff
@@ -660,11 +707,45 @@ int64_t netlink_recv(int guest_fd,
...
+
+        struct pollfd pfd = {
+            .fd = rd_fd,
+            .events = POLLIN,
+        };
+        int ret = poll(&pfd, 1, -1);
+        if (ret < 0) {
+            if (errno == EINTR) {
+                return -LINUX_EINTR;
+            }
```
jserv: Use brackets only when necessary.

src/syscall/netlink.c
```diff
@@ -689,6 +770,10 @@ int64_t netlink_recv(int guest_fd,
     }
     ns->buf_pos += msg_end;
 
+    if (ns->buf_pos >= ns->buf_len) {
+        netlink_clear_readable(ns);
+    }
```
jserv: Use brackets only when necessary.

src/syscall/netlink.c
```diff
@@ -837,11 +957,45 @@ int64_t netlink_read(int guest_fd, guest_t *g, uint64_t buf_gva, uint64_t count)
...
+
+        struct pollfd pfd = {
+            .fd = rd_fd,
+            .events = POLLIN,
+        };
+        int ret = poll(&pfd, 1, -1);
+        if (ret < 0) {
+            if (errno == EINTR) {
+                return -LINUX_EINTR;
+            }
```
jserv: Use brackets only when necessary.

src/syscall/netlink.c
```diff
@@ -851,6 +1005,11 @@ int64_t netlink_read(int guest_fd, guest_t *g, uint64_t buf_gva, uint64_t count)
     }
 
     ns->buf_pos += to_copy;
+
+    if (ns->buf_pos >= ns->buf_len) {
+        netlink_clear_readable(ns);
+    }
```
jserv: Use brackets only when necessary.

jserv (changes requested):
Be aware of preferable code style.

src/syscall/netlink.c
doanbaotrung: fixed

src/syscall/netlink.c
doanbaotrung: fixed

src/syscall/netlink.c
doanbaotrung: fixed

src/syscall/netlink.c
doanbaotrung: fixed

src/syscall/netlink.c
doanbaotrung: fixed

src/syscall/netlink.c
doanbaotrung: fixed

doanbaotrung:
Noted. I'll aware of using brackets only when necessary.

jserv:

PR #113, opened 2026-06-28
tests: resolve Alpine fixture versions from the live APKINDEX

Max042004:
The fixture builder pinned an exact version for every Alpine package and for the minirootfs tarball.  Alpine's mirror keeps only the latest build of each package, so a pinned version 404s as soon as a newer build lands upstream — forcing a manual version bump on every refresh just to keep the test fixtures fetchable.

Resolve versions dynamically instead:

  - PKGS becomes a plain "repo:name" list; resolve_versions() downloads each repo's APKINDEX and fills PKG_VERSION[repo:name] with whatever the CDN currently serves.
  - resolve_minirootfs() picks the newest minirootfs the releases listing advertises (ALPINE_PATCH still pins one explicitly when set).
  - A versions.lock manifest records the resolved set; when it changes the staged rootfs/kernel/initramfs are rebuilt so a mirror bump actually takes effect instead of reusing stale apks.
  - Both resolvers fall back to the cache when the mirror is unreachable, preserving warm offline re-runs, and error cleanly when neither the mirror nor a cache is available.

---
## Summary by cubic
Resolve Alpine test fixture versions from the live APKINDEX to prevent 404s from stale pins and remove manual bumps. Auto-pick the newest minirootfs and rebuild the staged tree when versions change, with offline cache fallback.

- **New Features**
  - `PKGS` now "repo:name"; `resolve_versions` reads live APKINDEX, caches "name version", and fills `PKG_RESOLVED` as "repo:name:version".
  - `resolve_minirootfs` selects the newest release and writes it to `ALPINE_PATCH` (still honors an explicit pin).
  - `versions.lock` records packages + minirootfs; any diff sets `REBUILD=1` to rebuild the whole staged tree (including x86_64).
  - Offline fallback reuses cached indexes/tarballs and errors cleanly if no cache is available.

<sup>Written for commit 34bf78e21f4db7e43f9c8131967b10cf9b59cb59. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/113?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/fetch-fixtures.sh
+++ b/tests/fetch-fixtures.sh
@@ -36,7 +36,10 @@ set -euo pipefail
 . "$(dirname "$0")/lib/bash-compat.sh"
 
 ALPINE_VERSION="${ALPINE_VERSION:-3.21}"
-ALPINE_PATCH="${ALPINE_PATCH:-3.21.0}"
+# Empty by default -- the exact point release is resolved from the live
+# releases listing at fetch time (see resolve_minirootfs), then written back
+# here so the x86_64 path reuses the same patch. Set explicitly to pin one.
+ALPINE_PATCH="${ALPINE_PATCH:-}"
 ALPINE_ARCH="${ALPINE_ARCH:-aarch64}"
 
 CDN_BASE="https://dl-cdn.alpinelinux.org/alpine/v${ALPINE_VERSION}"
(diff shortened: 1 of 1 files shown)


jserv:

PR #91, opened 2026-06-09
Add self-hosted macOS HVF runtime job to CI

Max042004:
GitHub-hosted macOS runners do not expose Apple's Hypervisor.framework (HVF):
the hosted runner OS itself runs inside a virtualization layer that withholds
HVF, so `hv_vm_create()` returns `HV_UNSUPPORTED`. Because of this, the existing
`build-macos` job can only compile elfuse and verify the HVF entitlement — it
can never actually boot a guest.

This PR adds a `runtime-macos` job that runs on a **self-hosted Apple Silicon
runner** (an M4 machine) where HVF *is* available, so the real runtime path —
booting a Linux guest under HVF + Rosetta — is exercised in CI for the first
time. The hosted `build-macos` job is unchanged and keeps gating builds as
before; this job runs *after* it (`needs: build-macos`).

Closes #51.

> [!NOTE]
> **Known failure:** the `runtime-macos` job currently fails in the
> **`make check`**. Once #76 lands (or is merged into / rebased onto this branch), 
> it is expected to go green.

## What the `runtime-macos` job does

- `runs-on: [self-hosted, macOS, arm64]`, `needs: build-macos`, 20-minute
  timeout.
- Per-ref concurrency: cancels in-progress runs for the same PR, but lets
  `main` runs finish.

### Where it runs (trigger gating)

- **PR targeting `main` — runs.** A `pull_request` from any fork is evaluated
  in the upstream (`sysprog21/elfuse`) Actions context, so a collaborator who
  pushes a branch to their own fork and opens a PR against upstream gets the
  full runtime suite. This is the normal contributor flow and must work.
- **Push to `main` — runs.** Post-merge validation on the mainline.
- **Activity in a fork's own context — skipped.

### Failing fast on superseded re-runs

There is only one self-hosted runner, so a wasted run on it blocks every other
PR. Per-ref `concurrency` already cancels an in-progress run when a newer commit
is pushed to the same PR, but it does **not** cover a manual *Re-run jobs* on an
old run: a re-run replays the original (now stale) commit and would otherwise
occupy the runner for the full timeout re-testing code that is no longer HEAD.

The job's first step compares the run's frozen `head.sha` against the PR's live
HEAD (via the REST API) and, when they differ, prints a `::error::` saying *the
commit is no longer the latest* and `exit 1`, end up failed.

PR #118, opened 2026-07-02
Defer sidecar walks to the sysroot host fallback

Max042004:
## Problem

On a case-insensitive sysroot (macOS default APFS, where the sidecar
case-folding layer is active), absolute guest paths that live outside the
sysroot became unreachable. `proc_resolve_sysroot_path_flags` documents the
contract: resolve inside the sysroot when the path exists there, otherwise
fall back to the literal host path so guests can still reach host resources
such as mktemp directories or `/etc/resolv.conf`. The sidecar walk vetoed
that fallback in two ways:

- `sidecar_translate_lookup_at` anchored every absolute path at the sysroot
  root and returned ENOENT as soon as an intermediate component was missing
  there; when only the final component was missing it still rewrote the path
  to `${sysroot}${path}`. Either way the resolver's host-literal decision was
  overridden.
- Mutating syscalls (`openat` with `O_CREAT`, `mkdirat`, `unlinkat`,
  `linkat`, `renameat`) hit the same wall through `sidecar_walk_parent_at`,
  which hard-anchored the parent directory at the sysroot.

### Symptom

All 27 failures in the test-matrix **"musl dyn"** coreutils suite on the
self-hosted runtime job (PR #91) and on any local run where the fixtures
live on a case-insensitive volume:

```
cat   [ FAIL ] (exit 1)   ... cat: /var/folders/.../hello.txt: No such file or directory
touch [ FAIL ] (got 1)    ... cannot touch '/var/folders/.../touched-1'
env   [ FAIL ] (got 127)  ... execve could not reach the host binary
```

Dynamic binaries run under `--sysroot`; macOS `mktemp -d` lives in
`/var/folders/...` and the sysroot only contains `/var`, so the walk
descended into `${sysroot}/var`, failed at `folders`, and turned the
documented fallback into ENOENT. Static coreutils run without `--sysroot`
and passed, which is why only the dynamic suite failed.

## Fix

Track whether the walk consulted an index mapping:

- A walk that leaves the sysroot tree **without** touching a mapping returns
  "no translation" (0), so the resolver's fallback decision stands.
- Beneath a **mapped** prefix the sysroot view stays authoritative: missing
  suffixes append literally and surface as ENOENT against the translated
  path (no index can exist under a missing directory).
- An unmapped final component is claimed only if it actually exists in the
  sysroot, restoring the fallback for paths like `/etc/resolv.conf` whose
  parent exists in the sysroot but whose leaf does not.

`sidecar_walk_parent_at` now reports a parent that resolves outside the
sysroot as a distinct verdict, and all mutation entry points hand such paths
back as `SIDECAR_NOT_HANDLED` to the regular translation flow.

## Validation

- `tests/test-matrix.sh elfuse-aarch64`: **168 passed, 0 failed** (was
  141/27); the remaining 4 skips are missing optional fixtures
  (base32/b2sum/numfmt applets, glibc dyn suite), unrelated to this bug.
- New regression test `test-sysroot-host-fallback` (wired into
  `make check`): 7 cases covering reads through missing intermediates,
  create/write/unlink, mkdir/rmdir, rename on host paths, the
  missing-final-component fallback, and sysroot-over-host precedence.
  Without the fix six of the seven cases fail with ENOENT.
- `make check`: all green (104 driver tests + host unit tests, 0 failed).
- `test-case-collision` and `test-case-collision-fallback` (12/12 both):
  the sidecar's collision compensation is unaffected.

---
## Summary by cubic
Defers sidecar path walks to the resolver’s host-literal fallback on case-insensitive sysroots so absolute guest paths outside the sysroot remain reachable. Restores access to host resources (e.g., mktemp dirs, `/etc/resolv.conf`) and fixes the “musl dyn” coreutils failures.

- **Bug Fixes**
  - Lookup returns “no translation” if a walk leaves the sysroot without an index mapping; mapped prefixes stay authoritative and missing dirs surface ENOENT; unmapped final components are only claimed if they exist.
  - `sidecar_walk_parent_at` reports parents outside the sysroot; mutation syscalls (`openat`, `mkdirat`, `unlinkat`, `linkat`, `renameat`) return `SIDECAR_NOT_HANDLED` for the regular resolver.
  - Adds `test-sysroot-host-fallback` and wires it into `make check`; matrix now 168 passed, 0 failed (was 141/27).

<sup>Written for commit 1304c68a9d36317307bf172dced3b0725739cf85. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/118?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/mk/tests.mk
+++ b/mk/tests.mk
@@ -10,6 +10,7 @@
         test-matrix test-matrix-elfuse-aarch64 test-matrix-qemu-aarch64 \
         test-full test-multi-vcpu test-rwx test-sysroot-rename \
         test-case-collision test-case-collision-fallback test-getdents64-overlong \
+        test-sysroot-host-fallback \
         test-sysroot-create-paths test-fork-ipc-protocol-host test-identity-override-host \
         test-proctitle-host test-proctitle-low-stack \
         test-sysroot-procfs-exec test-timeout-disable test-fuse-alpine \
@@ -57,6 +58,8 @@ check: $(ELFUSE_BIN) $(TEST_DEPS) check-syscall-coverage \
 	@$(MAKE) --no-print-directory test-sysroot-procfs-exec
 	@printf "\n$(BLUE)━━━ getdents64 overlong-UTF-8 dirent skip ━━━$(RESET)\n"
 	@$(MAKE) --no-print-directory test-getdents64-overlong
+	@printf "\n$(BLUE)━━━ sysroot host-fallback validation ━━━$(RESET)\n"
+	@$(MAKE) --no-print-directory test-sysroot-host-fallback
 	@printf "\n$(BLUE)━━━ Alpine sysroot FUSE validation ━━━$(RESET)\n"
 	@$(MAKE) --no-print-directory test-fuse-alpine
 	@printf "\n$(BLUE)━━━ timeout=0 validation ━━━$(RESET)\n"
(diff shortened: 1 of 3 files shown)


jserv:

PR #124, opened 2026-07-03
Verify on-disk spelling in sidecar lookups

Max042004:
Linux path resolution treats file names as byte strings: a lookup whose spelling differs from an existing entry only by case (or Unicode normalization form) must fail with `ENOENT`. On a case-insensitive sysroot the sidecar walk probed existence with plain `openat`/`fstatat`, which APFS resolves case- and normalization-insensitively, so read-path syscalls leaked the folded match. With only `Makefile` staged in the sysroot:

```
stat("/data/MAKEFILE")   -> succeeds   (Linux: ENOENT)
open("/data/makefile")   -> opens Makefile
access("/data/mAkEfIlE") -> succeeds
open("/data/SUB/f.txt")  -> folds the intermediate directory
```

Mutating syscalls were already byte-exact through the sidecar's readdir probe, leaving the semantics asymmetric: `unlink` reported `ENOENT` on the very spelling `stat` had just confirmed. Real-world impact: any unpacked rootfs (literal names on disk) gives false positives to case-probing workloads — `make` finding `makefile` when only `Makefile` exists, configure-style existence probes, tools distinguishing `README`/`readme`.

### Fix

Probe every unmapped component of the walk with `getattrlistat(ATTR_CMN_NAME, FSOPT_NOFOLLOW)`: it resolves through the same folding lookup but returns the on-disk spelling for a byte comparison. `FSOPT_NOFOLLOW` pins the probe to the entry itself, so a symlink component is verified by its own name, not its target's.

- exact match — proceed as before;
- genuinely absent — keep the existing deferral flows (resolver host-literal fallback without a mapping, translated-path `ENOENT` beneath a mapped prefix, #118 semantics unchanged);
- folded match — `ENOENT` outright. The veto must bypass the resolver too, whose own `access()` existence probe folds the same way.

Index-mapped components stay unprobed (tokens resolve byte-exactly by construction), filesystems that do not return `ATTR_CMN_NAME` fall back to the readdir scan. Cost: the probe replaces the final-component `fstatat` one-for-one and adds one `getattrlistat` per unmapped intermediate component, only on case-insensitive sysroots; the hot-syscall guardrail stays green.

The byte comparison also rejects normalization-folded matches (NFD lookup of an NFC entry), which APFS folds even on case-*sensitive* volumes, so with the sidecar active those now fail with `ENOENT` and `O_CREAT|O_EXCL` of the other form creates a distinct entry.

### Scope

Lookups that leave the sysroot for the host-literal fallback keep host semantics — the sidecar cannot speak for entries it does not index, so case folding on the host tree itself remains reachable through that hatch (it is a host-resource escape by design). Case-sensitive sysroots (`--create-sysroot`) are unaffected: the sidecar is inactive there and APFS enforces case exactness natively; its normalization folding remains a known limitation of that mode.

### Testing

`test-sysroot-case-exact` stages exact-case fixtures host-side and asserts folded spellings do not resolve: wrong-case `stat`/`open`/`access`/`unlink`/`rename` on final and intermediate components, collision-pair creation via wrong-case `O_CREAT|O_EXCL`, and NFC/NFD probes; wired into `make check`. Without the fix six of its thirteen cases fail. Full `make check` passes.

---
## Summary by cubic
Make sidecar lookups byte-exact on case-insensitive sysroots by verifying the on-disk spelling and returning ENOENT for case- or normalization-folded matches. This aligns stat/open/access with Linux and removes asymmetric behavior vs mutating syscalls.

- **Bug Fixes**
  - Probe each unmapped component with getattrlistat(ATTR_CMN_NAME, FSOPT_NOFOLLOW) and compare the returned name byte-for-byte; exact proceeds, folded returns ENOENT (vetoes resolver fallback), genuine absence keeps existing fallback flows.
  - Do not probe index-mapped components; fall back to a readdir scan when ATTR_CMN_NAME is unavailable; symlink components are validated by their own name.
  - Reject normalization folding (e.g., NFD vs NFC); O_CREAT|O_EXCL with the other form creates a distinct entry.
  - Add test-sysroot-case-exact and wire it into make check; update docs to note byte-exact lookup semantics.

<sup>Written for commit 800395657fb0aba4b5d4b12b77bf73ec122071e4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/124?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/sidecar.c
+++ b/src/syscall/sidecar.c
@@ -16,6 +16,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/attr.h>
 #include <sys/stat.h>
 #include <sys/uio.h>
 #include <unistd.h>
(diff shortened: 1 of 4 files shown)


src/syscall/sidecar.c
```diff
@@ -565,6 +566,69 @@ static int sidecar_open_base(guest_fd_t dirfd,
...
+ * byte comparison. FSOPT_NOFOLLOW keeps the probe on the entry itself so a
+ * symlink component is verified by its own name, not its target's. The name
+ * buffer covers the APFS maximum (255 UTF-16 units, up to 765 UTF-8 bytes),
+ * so a returned name never truncates into a false mismatch.
+ *
+ * Returns a SIDECAR_NAME_* verdict, or -1 with errno on probe failure.
+ * Filesystems that do not return ATTR_CMN_NAME fall back to the readdir
+ * scan, with a folded fstatat to separate ABSENT from CASEFOLD.
+ */
+static int sidecar_probe_exact_name(int dirfd, const char *name)
```
jserv: Specify `SIDECAR_` enumeration for return type rather than plain `int.`

Max042004: Done in 8003956: the verdicts are now a named `sidecar_name_verdict_t` typedef returned by `sidecar_probe_exact_name`, with the error case promoted to an explicit `SIDECAR_NAME_ERROR = -1` member (errno still set) so every return value belongs to the enumeration; call sites compare against `SIDECAR_NAME_ERROR` instead of `< 0`.

jserv:

PR #127, opened 2026-07-03
Increase sys_execve limits MAX_ARGS and STR_BUF_SIZE

### LINKED ISSUE #125: Increase `sys_execve` limits (`MAX_ARGS` and `STR_BUF_SIZE`) to support modern guest workloads
Currently, `elfuse` imposes a limit of `256` arguments (`MAX_ARGS`) and a `256 KB` buffer size (`STR_BUF_SIZE`) for command-line arguments and environment variables inside `sys_execve`.

While these limits are sufficient for basic workloads, they are easily exceeded by modern developer tools, compilation pipelines, package managers, and large projects. For example:
- Deeply nested package manager calls (e.g., `npm`, `yarn`, `cargo`)
- Compilers or test runners compiling large codebases (e.g., `javac` or clang with many source paths)
- Shell scripts using tools like `find` combined with `xargs`

When these limits are hit, the guest process exits with `E2BIG` (or fails with `EFAULT` when the buffer is exhausted), blocking successful emulation of these tools.

To match standard Unix/Linux environments more closely and support rich CLI developer tools, we propose increasing these limits:
- **`MAX_ARGS`**: Increase from `256` to `4096` arguments.
- **`STR_BUF_SIZE`**: Increase from `256 KB` to `2 MB` (`2048 * 1024` bytes), matching typical modern Linux defaults for `ARG_MAX`.

---

#### 1. Stack Usage
In `src/syscall/exec.c`, the pointer arrays `argv` and `envp` are allocated on the stack inside `sys_execve`:
```c
char *argv[MAX_ARGS + 1];
char *envp[MAX_ENVS + 1];
```
- With `MAX_ARGS` and `MAX_ENVS` set to `4096`, each array occupies `4097 * 8 bytes ≈ 32 KB` of stack space.
- In the shebang (interpreter) execution path, an additional stack array `new_argv[MAX_ARGS + 3]` is allocated, occupying another `32 KB`.
- This leads to a peak stack usage in `sys_execve` of approximately `64 KB` (or `96 KB` on shebang paths), which is extremely safe on macOS (default thread stack size is 512 KB for spawned threads, and 8 MB for the main thread).

#### 2. Heap Usage
The string buffers themselves (`argv_buf` and `envp_buf`) are heap-allocated via `malloc`:
```c
char *argv_buf = malloc(STR_BUF_SIZE);
char *envp_buf = malloc(STR_BUF_SIZE);
```
Since they are on the heap, increasing their size from `256 KB` to `2 MB` does not carry any stack overflow risks. These allocations are also cleaned up correctly on all exit paths via `exec_cleanup_inputs`.

#### 3. Bounds-Checking Integrity
All existing bounds checks in `sys_execve` remain correct:
- The shebang argument reconstruction check:
  ```c
  if (argc > MAX_ARGS - prefix + 1) {
      err = -LINUX_E2BIG;
      goto fail;
  }
  ```
  handles the new limits correctly and prevents out-of-bounds writes on `new_argv[MAX_ARGS + 3]`.
- `read_string_array` correctly respects the passed limits (`MAX_ARGS` / `MAX_ENVS` and `STR_BUF_SIZE`) ensuring memory safety.

---

### Proposed Changes

Below is the diff of the changes I think I can work in `src/syscall/exec.c`:

```diff
diff --git a/src/syscall/exec.c b/src/syscall/exec.c
index 9c21052..60afb3e 100644
--- a/src/syscall/exec.c
+++ b/src/syscall/exec.c
@@ -255,9 +255,9 @@ int64_t sys_execve(hv_vcpu_t vcpu,
     char interp_host_buf[LINUX_PATH_MAX];
     bool interp_host_temp = false;
 
-#define MAX_ARGS 256
+#define MAX_ARGS 4096
 #define MAX_ENVS 4096
-#define STR_BUF_SIZE ((size_t) 256 * 1024)
+#define STR_BUF_SIZE ((size_t) 2048 * 1024)
 
     int64_t err = 0;
     char *argv[MAX_ARGS + 1];
```

doanbaotrung:
MAX_ARGS: Increased from 256 to 4096 arguments.

STR_BUF_SIZE: Increased from 256 KB to 2 MB
(2048 * 1024 bytes).

fix #125

---
## Summary by cubic
Use heap allocations for sys_execve argv/envp to avoid stack overflows and align limits with Linux, fixing #125. Enforces 128 KB per-arg (MAX_ARG_STRLEN) and 2 MB total (ARG_MAX) across argv/envp with correct E2BIG errors, and adds boundary tests.

<sup>Written for commit ffd68837c55f83dacc43284b78271e532c3677b4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/127?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/exec.c
+++ b/src/syscall/exec.c
@@ -100,7 +100,9 @@ static void exec_republish_shim_globals_or_die(hv_vcpu_t vcpu,
  * before crossing the point of no return. Used by both the Rosetta and the
  * aarch64 success paths.
  */
-static void exec_cleanup_inputs(char *argv_buf,
+static void exec_cleanup_inputs(char **argv,
+                                char **envp,
+                                char *argv_buf,
                                 char *envp_buf,
                                 const char *path_host_buf,
                                 bool path_host_temp,
@@ -111,6 +113,8 @@ static void exec_cleanup_inputs(char *argv_buf,
         unlink(path_host_buf);
     if (interp_host_temp)
         unlink(interp_host_buf);
+    free(argv);
+    free(envp);
     free(argv_buf);
     free(envp_buf);
 }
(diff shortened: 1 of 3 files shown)


src/syscall/exec.c
```diff
@@ -255,9 +255,9 @@ int64_t sys_execve(hv_vcpu_t vcpu,
...
     bool interp_host_temp = false;
 
-#define MAX_ARGS 256
+#define MAX_ARGS 4096
 #define MAX_ENVS 4096
-#define STR_BUF_SIZE ((size_t) 256 * 1024)
+#define STR_BUF_SIZE ((size_t) 2048 * 1024)
 
     int64_t err = 0;
     char *argv[MAX_ARGS + 1];
```
jserv: Stack budget: argv[4097] + envp[4097] are now ~64 KB in this single frame (only argv[] grew here; MAX_ENVS was already 4096). The shebang retry path adds new_argv[MAX_ARGS+3] (~32 KB) on top, and sys_execve runs on worker/vCPU pthreads created with no pthread_attr_setstacksize -- i.e. the macOS default (~512 KB) secondary-thread stack (no setstacksize exists in src/runtime/thread.c or forkipc.c). ~96 KB of pointer arrays plus the deep execve -> elf_load -> path_translate call chain risks overflowing that stack, worst on nested shebangs. Note this file already heap-moves the cloexec array for exactly this reason. Fix: malloc argv[]/envp[]/new_argv[] alongside argv_buf/envp_buf, or raise the worker stack to >=1 MB.

src/syscall/exec.c
```diff
@@ -255,9 +255,9 @@ int64_t sys_execve(hv_vcpu_t vcpu,
     char interp_host_buf[LINUX_PATH_MAX];
     bool interp_host_temp = false;
 
-#define MAX_ARGS 256
+#define MAX_ARGS 4096
```
jserv: A fixed 4096-arg count cap diverges from Linux, which imposes no argv-count limit -- it bounds total argv+envp bytes (ARG_MAX, ~2 MB) and per-string length (MAX_ARG_STRLEN, 128 KB). build_linux_stack() already carries a much higher 131072-entry guard (src/core/stack.c), so this becomes the effective lower limit and rejects arg counts Linux would accept within the byte budget. Low practical impact, but if parity matters, drop the count cap and gate E2BIG on total bytes instead.

src/syscall/exec.c
```diff
@@ -255,9 +255,9 @@ int64_t sys_execve(hv_vcpu_t vcpu,
...
 #define MAX_ENVS 4096
-#define STR_BUF_SIZE ((size_t) 256 * 1024)
+#define STR_BUF_SIZE ((size_t) 2048 * 1024)
```
jserv: Three things on this buffer size:

1. 2 MB is applied independently to argv and envp, so a process can pass ~4 MB of strings vs Linux ARG_MAX of ~2 MB combined; there is also no per-string 128 KB cap. More permissive than Linux (not a crash).

2. E2BIG vs EFAULT: when the strings overflow this buffer, read_string_array hits `remaining < 2` (or guest_read_str fails on an oversized string) and returns -1 -> EFAULT (exec.c:200-203). Linux returns E2BIG for that case, and this is exactly the overflow this larger buffer makes reachable. Fix: return -2 on the buffer-full / oversized-string branch so the caller maps it to E2BIG.

3. Every execve now unconditionally mallocs 2x2 MB = 4 MB even for a tiny argv. Optional: keep a smaller default and grow on demand.

jserv (commented):
Concerns:
- Stack: the enlarged pointer arrays live on a small worker pthread stack (see inline on argv[]).
- Semantics: the argv/env byte budget and the E2BIG vs EFAULT handling now diverge from Linux (see inline on the two #defines).
- Tests: no test exercises the raised limits. tests/test-exec.c and tests/test-fork-exec.c only pass a handful of args. Worth adding a case that execs near MAX_ARGS (expect success) and just past it / past the byte budget (expect E2BIG) to lock in the boundary.

doanbaotrung:
Align `sys_execve` with Linux Limits (Heap Allocation & Proper `E2BIG`/`EFAULT` handling)

### Problem Description
Previously, `elfuse` set fixed compilation limits in `sys_execve` (`MAX_ARGS = 256`, `STR_BUF_SIZE = 256 KB`), which easily broke modern CLI tools (like compilers, package managers, and test runners). Simply bumping these values on the stack introduced resource issues (stack frame usage up to ~96 KB on workers with a macOS default stack of ~512 KB), and unconditionally allocating large static buffers (2x2 MB) caused memory overhead even for tiny argument lists.

Moreover, the behavior differed from Linux because:
1. Linux does not enforce a strict hard-coded count on `argv` (only a byte budget).
2. Buffer overflow in string copy incorrectly returned `EFAULT` (bad page address) instead of `E2BIG` (argument list too long).

---

### Solution Description

We redesigned the argument list preparation in `src/syscall/exec.c` to resolve these stack, memory, and semantic mismatches:

1. **Heap Allocation of `argv[]`/`envp[]`/`new_argv[]`**:
   - `sys_execve` no longer allocates large arrays on the stack.
   - Pointers are allocated on the heap dynamically according to the actual number of arguments present in guest memory.
   - For shebang resolution, `new_argv` is dynamically allocated.

2. **On-Demand Buffer Growth**:
   - Instead of unconditionally allocating 4 MB on every `execve`, the string buffers (`argv_buf` / `envp_buf`) are initially allocated to a small size (4 KB) and grown dynamically via `realloc` on demand.

3. **Linux Semantic Alignment**:
   - **`ARG_MAX` (2 MB)**: Enforced as a combined byte limit across all `argv` and `envp` strings.
   - **`MAX_ARG_STRLEN` (128 KB)**: Enforced as a maximum size of a single argument string.
   - **`E2BIG` vs `EFAULT`**: If a string exceeds 128 KB or the combined buffers exceed 2 MB, the emulator returns `-LINUX_E2BIG`. If it encounters a bad memory address, it returns `-LINUX_EFAULT`.

4. **Integration Test Case**:
   - A new test `tests/test-exec-limits.c` has been added to verify:
     - Executing under the limits succeeds.
     - Single string length exceeding 128 KB fails with `E2BIG`.
     - Combined byte list exceeding 2 MB fails with `E2BIG`.

jserv:

PR #143, opened 2026-07-03
Return 0 for zero-iovcnt vector I/O

Max042004:
readv/writev/preadv/pwritev/pwritev2 with `iovcnt == 0` returned EINVAL because the request was funneled into the host syscall path and macOS rejects a zero count. Linux returns 0 and leaves the file offset untouched -- even for pwritev2(RWF_APPEND), which must not seek to EOF. Verified against a Linux reference under qemu-system-aarch64: `pwritev2(fd, iov, 0, -1, RWF_APPEND)` returns 0 with the offset unmoved.

Short-circuit `iovcnt == 0` in the five vector entry points before any host call: validate the fd (closed/O_PATH report EBADF, matching Linux), return 0, and leave the offset alone. test-readv-writev test 7 now encodes the Linux behavior, so the same binary passes under both elfuse and the qemu reference lane of tests/test-matrix.sh.

Testing: `make check` 105/105; test-readv-writev 7/7 under elfuse and under qemu.

---
## Summary by cubic
Return 0 for zero-iovcnt vector I/O and match Linux’s validation order while keeping file offsets unchanged. Applies to readv, writev, preadv, pwritev, preadv2, and pwritev2, including pwritev2(RWF_APPEND).

- **Bug Fixes**
  - Short-circuit `iovcnt == 0` in all six syscalls to avoid macOS `EINVAL`.
  - Follow Linux’s ordering: negative offsets return `EINVAL` (`*v2` allows `-1` and treats it as non-positional), positional on non-seekable fds return `ESPIPE`, wrong-direction or `O_PATH`/closed fds return `EBADF`.
  - Skip `RWF_*` validation for empty vectors and keep the file offset unchanged (incl. `pwritev2(RWF_APPEND)`).
  - Tests: update append case and add a table covering closed/O_PATH/mode/pipe/offset/flag paths; passing on elfuse and the qemu Linux reference.

<sup>Written for commit 1e12fc530b4cca1a3ae590e720223d5b4d1855ff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/143?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/io.c
+++ b/src/syscall/io.c
@@ -1250,8 +1250,62 @@ static int64_t single_guest_iov(guest_t *g,
     return 0;
 }
 
+/* Linux returns 0 for zero-iovcnt vector I/O once the fd validates:
+ * import_iovec() yields an empty iterator and do_iter_read/do_iter_write
+ * return before the file offset is touched, so even pwritev2(RWF_APPEND)
+ * leaves the position alone. macOS readv/writev instead reject iovcnt == 0
+ * with EINVAL, so short-circuit before any host call -- the append path's
+ * SEEK_END would otherwise move the shared offset. The iov pointer is not
+ * dereferenced (Linux ignores it for an empty vector), and negative counts
+ * keep flowing into host_iov_prepare's EINVAL.
+ *
+ * The checks Linux runs before its empty-vector return still apply, in this
+ * order: the positional variants fail non-seekable files with ESPIPE
+ * (FMODE_PREAD/FMODE_PWRITE in do_preadv/do_pwritev), and all variants fail
+ * wrong-direction fds with EBADF (FMODE_READ/FMODE_WRITE in
+ * do_iter_read/do_iter_write). Both are probed on the host fd, whose open
+ * mode mirrors the guest's for host-backed types; fd_table linux_flags
+ * cannot be used because pipe/socket entries only track CLOEXEC there.
+ * Virtual multiplexed fds (eventfd/timerfd/signalfd/...) sit on host pipes
+ * or kqueues whose mode says nothing about the guest-visible fd (the Linux
+ * anon-inode equivalents are O_RDWR), so they validate existence only. No
+ * F_SEAL_WRITE check: Linux never reaches the write path for an empty
+ * vector, so a write-sealed memfd returns 0 here too.
+ */
+static int64_t vec_zero_iovcnt(int fd, bool op_is_write, bool positional)
+{
+    fd_entry_t snap;
+    if (!fd_snapshot(fd, &snap) || snap.type == FD_PATH)
+        return -LINUX_EBADF;
+
+    bool host_mode_mirrors_guest =
+        snap.type == FD_REGULAR || snap.type == FD_DIR ||
+        snap.type == FD_PIPE || snap.type == FD_SOCKET ||
+        snap.type == FD_STDIO || snap.type == FD_URANDOM;
+    if (!host_mode_mirrors_guest)
+        return 0;
+
+    host_fd_ref_t host_ref;
+    if (host_fd_ref_open(fd, &host_ref) < 0)
+        return -LINUX_EBADF;
+
+    int64_t ret = 0;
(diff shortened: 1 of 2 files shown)


src/syscall/io.c
```diff
@@ -1150,8 +1150,27 @@ static int64_t single_guest_iov(guest_t *g,
...
+ * leaves the position alone. macOS readv/writev instead reject iovcnt == 0
+ * with EINVAL, so short-circuit before any host call -- the append path's
+ * SEEK_END would otherwise move the shared offset. The iov pointer is not
+ * dereferenced (Linux ignores it for an empty vector), and negative counts
+ * keep flowing into host_iov_prepare's EINVAL.
+ */
+static int64_t vec_zero_iovcnt(int fd)
+{
+    fd_entry_t snap;
+    if (!fd_snapshot(fd, &snap) || snap.type == FD_PATH)
```
jserv, line 1278: The helper validates existence and O_PATH but not access mode, so it diverges from Linux for empty vectors. Linux checks FMODE_READ/FMODE_WRITE in do_iter_read/do_iter_write and returns EBADF before the `tot_len == 0` short-circuit. So readv/preadv/preadv2 with iovcnt==0 on a write-only fd, and writev/pwritev/pwritev2 with iovcnt==0 on a read-only fd, return EBADF on Linux but 0 here. This is also looser than the iovcnt>=1 path, which surfaces EBADF via the host read/write on the mode-restricted fd.

If you fix it, pass a direction into the helper and reject the wrong access mode with EBADF, but do not source the mode from snap.linux_flags: pipe/socket ends store only CLOEXEC there, so a pipe write-end would read as O_RDONLY and get wrongly rejected. Use the host fd's real mode (fcntl(host_fd, F_GETFL) & O_ACCMODE).

Do not add an F_SEAL_WRITE EPERM here: Linux never invokes the write op for an empty vector, so writev on a write-sealed memfd with iovcnt==0 returns 0, which this helper already matches.

This is an obscure conformance corner (empty-vector I/O on a wrong-direction fd); real programs essentially never hit it, but LTP-style suites do. Reasonable to leave as-is if scoped out -- just noting the gap.

src/syscall/io.c
```diff
@@ -1380,6 +1405,9 @@ static int64_t sys_pwritev_append(guest_t *g,
                                   int iovcnt,
                                   bool update_file_offset)
 {
+    if (iovcnt == 0)
+        return vec_zero_iovcnt(fd);
```
jserv: RWF flag validation in sys_preadv2 and sys_pwritev2 runs before the iovcnt==0 short-circuit, but Linux skips flag validation entirely for empty vectors (do_iter_read/do_iter_write jump past kiocb_set_rw_flags via `if (!tot_len) goto out`). So preadv2(fd, iov, 0, off, RWF_APPEND) returns EINVAL here but 0 on Linux, and pwritev2(fd, iov, 0, off, <unsupported flag>) returns EOPNOTSUPP here but 0 on Linux. Low impact and optional; if you want parity, do the fd/access validation and return 0 before the flag checks.

tests/test-readv-writev.c
```diff
@@ -276,7 +280,7 @@ static void test_pwritev2_append_zero_iovcnt(void)
     close(fd);
     unlink(path);
 
-    EXPECT_TRUE(ret == -EINVAL && pos == 1, "zero iovcnt changed offset");
+    EXPECT_TRUE(ret == 0 && pos == 1, "zero iovcnt not a no-op");
```
jserv, line 320: Coverage only pins the pwritev2(RWF_APPEND) offset invariant. The cases this change actually owns are untested: bad fd and O_PATH (EBADF), wrong access mode, and the non-append entry points (readv, writev, preadv, pwritev, preadv2). A small table over those would guard the validation ordering.

jserv:

PR #145, opened 2026-07-03
Route vCPU preemption signals via sigwait thread

### LINKED ISSUE #77: Route vCPU preemption signals through a dedicated sigwait thread to eliminate `HV_EXIT_REASON_UNKNOWN`
## Background

elfuse interrupts a running vCPU (for the cross-process guest-signal transport
and for the per-iteration safety timeout) by sending a host signal to the vCPU
thread, whose handler calls `hv_vcpus_exit()`:

- **SIGUSR2** — cross-process guest-signal doorbell. `proc_send_guest_signal()`
  (`src/syscall/proc.c:358`) writes the guest signum to `/tmp/elfuse-sig-<pid>`
  and sends host SIGUSR2; the receiver's handler
  `guest_signal_transport_handler()` (`proc.c:965`) sets `g_external_guest_signal`
  and calls `hv_vcpus_exit(&g_timeout_vcpu, 1)`.
- **SIGALRM** — per-iteration timeout. `alarm_handler()` sets `g_timed_out` and
  calls `hv_vcpus_exit(&g_timeout_vcpu, 1)`.

Because the signal is delivered **to the vCPU thread while it is inside
`hv_vcpu_run`**, Apple HVF aborts the run with `HV_EXIT_REASON_UNKNOWN` (0x3)
instead of the clean `HV_EXIT_REASON_CANCELED` (0) that `hv_vcpus_exit()`
produces for a vCPU caught between runs. The run loop must therefore treat
UNKNOWN as a possible cancellation, which is ambiguous: a genuine hypervisor
fault could in principle also surface as UNKNOWN, so blindly resuming risks a
silent spin (raised by a cubic review on #76).

#76 (`fix-cross-process-signal-el0`) carries the EL0-preemption delivery fix and
routes UNKNOWN through the cancellation handler so the already-queued guest
signal is delivered instead of crashing the child. That keeps the ambiguity: the
run loop cannot tell our own preemption from a genuine fault, so it errs toward
resuming. This issue removes the ambiguity at its source rather than reasoning
about it after the fact.

Note: a real `hv_vcpu_run` API failure (non-`HV_SUCCESS` return) is *already*
caught by `HV_CHECK_CTX` (`proc.c:1113`) and crashes immediately. The only thing
that reaches the UNKNOWN branch is `HV_SUCCESS` + `exit_reason == UNKNOWN`,
i.e. our own `hv_vcpus_exit` landing mid-execution.

## Goal

Deliver every self-directed `hv_vcpus_exit` from a thread **other than** the
vCPU thread, so `hv_vcpu_run` always returns a clean `CANCELED`. Once no
legitimate path can produce UNKNOWN, the run loop can treat **any** UNKNOWN as a
hard hypervisor failure and crash with diagnostics — no heuristic needed.

### Why this is not blocked by an HVF constraint

The relevant HVF rules are:

- a **VM** (`hv_vm_create`) is per **process** (one per `elfuse --fork-child`);
- a **vCPU** is bound to the **thread** that created it (`hv_vcpu_run` must run on
  that thread);
- **`hv_vcpus_exit()` is explicitly designed to be called from another thread**
  to force a VMEXIT.

The helper thread lives in the *same* host process as the vCPU thread, so there
is no cross-process issue — this is the idiomatic HVF watchdog pattern. The
current code already calls `hv_vcpus_exit()` on a stored handle
(`g_timeout_vcpu`), proving the handle is usable off-thread; this change only
moves the call to a dedicated thread.

## Work items

- [ ] **1. Signal-mask discipline.** Block SIGUSR2 (and SIGALRM, see item 2) on
      every vCPU thread and the main thread via `pthread_sigmask`, leaving them
      unblocked only on the dedicated sigwait thread. Establish the mask before
      any vCPU thread is created, at every thread-creation site (bootstrap,
      `CLONE_THREAD` workers, fork-child bring-up), and re-establish it across
      `fork`/`exec` so children do not inherit a stale disposition. A single
      missed site silently reintroduces UNKNOWN.

- [ ] **2. Move SIGALRM onto the same path.** The "any UNKNOWN is abnormal"
      invariant only holds if *all* signal-driven `hv_vcpus_exit` calls leave the
      vCPU thread. Re-home the per-iteration timeout: either have the helper
      thread own the timeout (e.g. `sigtimedwait`/timer + `hv_vcpus_exit`) or
      replace `alarm()` with a mechanism that does not deliver SIGALRM to the
      vCPU thread. Preserve the existing `g_timed_out` → `CRASH_TIMEOUT`
      (exit 124) semantics and the guest `ITIMER_REAL` emulation that currently
      shares `alarm()`.

- [ ] **3. Live-vCPU registry.** Replace the single `g_timeout_vcpu` global with a
      per-process registry of all live vCPU handles (multi-threaded guests run
      worker vCPUs, each on its own thread — `g_timeout_vcpu` is currently
      last-writer-wins). The helper thread kicks the correct set on a transport
      event; `hv_vcpus_exit()` accepts a vCPU array, so a single call can exit
      all of them and let the signal/queue machinery sort out delivery. Register
      on vCPU create, unregister on thread exit, guard with a lock.

- [ ] **4. Empirically verify CANCELED, not UNKNOWN.** Confirm under stress that
      helper-thread `hv_vcpus_exit()` against an actively-running vCPU yields
      `CANCELED` with zero UNKNOWN across many iterations (single- and
      multi-threaded guests, and the cross-process fork case). The CANCELED-vs-
      UNKNOWN split is supported by the code's own comments, but Apple HVF has
      quirks — validate before making UNKNOWN fatal.

## Acceptance criteria

- With items 1–3 landed, the run loop treats `HV_EXIT_REASON_UNKNOWN` as fatal
  (`crash_report(CRASH_UNEXPECTED_EXIT)`), matching the original pre-#76 `else`
  branch.
- `test-fork` passes 100% over a large batch (e.g. 200+ runs); no
  `elfuse --fork-child` orphans left behind.
- `test-signal`, `test-signal-thread`, `test-mt-fork`, and the timeout=0
  validation suite stay green.
- A stress harness records 0 `HV_EXIT_REASON_UNKNOWN` exits during heavy
  cross-process signalling.

## Risks / open questions

- Signal-mask plumbing touches every thread-creation and fork/exec path; a
  missed site is silent (reintroduces UNKNOWN) — mitigated by item 4's stress
  check.
- Extra thread per guest process: small overhead, but interacts with the
  fork model (the helper must be re-created, not inherited, in each child).
- Does the helper-thread `hv_vcpus_exit` reliably interrupt a vCPU blocked in a
  host syscall issued from inside the HVC handler (e.g. `nanosleep`), or only
  one executing guest code? Confirm the cross-process wake still works for a
  child parked in a blocking host syscall.

## Out of scope

- The EL0-preemption signal-delivery fix (already on #76). This issue does not
  change `signal_deliver`.
- Any change to the `/tmp/elfuse-sig-<pid>` file transport itself.

jserv:
elfuse preempted a running vCPU by sending a host signal whose handler called hv_vcpus_exit(): SIGUSR2 for the cross-process guest-signal doorbell and SIGALRM for the per-iteration safety timeout. Because the signal landed on the vCPU thread inside hv_vcpu_run, Apple HVF aborted the run with HV_EXIT_REASON_UNKNOWN instead of the clean CANCELED that hv_vcpus_exit() produces for a vCPU caught between runs, leaving the run loop unable to tell a self-directed preemption from a genuine fault.

Block SIGUSR2 and SIGALRM on the main thread before any vCPU thread is created (proc_preempt_init), so every CLONE_THREAD worker and posix_spawn fork-child inherits the block. A dedicated thread sigwaits both and kicks every live vCPU via thread_interrupt_all() from a thread that never runs one, so hv_vcpu_run always returns CANCELED and the run loop treats any HV_EXIT_REASON_UNKNOWN as a hard hypervisor fault. The g_timed_out and g_external_guest_signal handoff flags become _Atomic with release/acquire ordering now that they cross the preempt-thread to vCPU-thread boundary.

Make a guest thread parked in a blocking host syscall reachable by a signal, which hv_vcpus_exit() alone cannot interrupt:
- attention_raise() pokes the wakeup pipe on every deliverable signal.
- ppoll/pselect/epoll_pwait break out on signal_pending_interruption.
- a blocking pipe/stdio read waits in wait_readable_or_interrupted, which returns EINTR only for a guest-visible delivery, skips O_WRONLY and O_NONBLOCK fds, and uses a bounded 200ms recheck; the non-verbose fast-path read diverts a would-block pipe/stdio read to this slow path.
- fork_child_main creates the wakeup pipe, which it skips via syscall_init.

Make the signal transport race-free: the sender holds flock around its append and the drain holds it across a full read plus ftruncate and no longer unlinks, so a concurrent append is never routed to an orphaned inode. Each process unlinks its own empty file at exit.

signal_pending_interruption() reports whether a guest-visible delivery is pending and excludes ignored and default-ignore signals, so read, poll, select, and epoll no longer wake spuriously on an ignored signal.

Close #77

---
## Summary by cubic
Route vCPU preemption (SIGUSR2/SIGALRM) through a dedicated sigwait thread so hv_vcpu_run reliably returns CANCELED on HVF, and expand interruptible blocking I/O so reads, writes, and socket ops wake only for guest-visible signals. Also harden cross-process signal transport and clean up wakeup-pipe handling.

- **Bug Fixes**
  - Preemption: block SIGUSR2/SIGALRM before any vCPU threads; a sigwait thread sets _Atomic handoff flags and kicks vCPUs; HV_EXIT_REASON_UNKNOWN now indicates a hypervisor fault.
  - Blocking I/O: add per-fd can_block and a shared io_wait_fd_or_interrupted; used by read/write/readv/writev and socket recv/accept/connect/send; the fast-path read/write diverts on would-block or probe error; sends probe with MSG_DONTWAIT and retry on EAGAIN; connect flips nonblocking, waits for POLLOUT, checks SO_ERROR; zero-length iovecs still enforce EBADF; zero-length recvmsg returns immediately.
  - Signals and waits: attention_raise pokes the wakeup pipe; ppoll/pselect/epoll_pwait break only on guest-visible deliveries; ignored/default-ignore signals no longer interrupt; wakeup-pipe init is idempotent with read/drain helpers.
  - Signal transport: sender append and receiver drain are serialized with flock; drain reads fully and ftruncate’s; each process unlinks its empty file at exit.
  - Tests: `test-poll` covers ignored SIGCHLD and that SA_RESTART read handlers run and unblock recv; `test-socket` adds zero-length/invalid recvmsg cases; `test-readv-writev` adds an all-empty-iovec access-check.

<sup>Written for commit c3de5dcc4e33e13ef41c08cdfc2fe9781a7c293d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/145?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,6 +195,11 @@ $(BUILD_DIR)/test-pthread: tests/test-pthread.c | $(BUILD_DIR)
 	@echo "  CROSS   $< (with -lpthread)"
 	$(Q)$(CROSS_COMPILE)gcc -D_GNU_SOURCE -static -O2 -o $@ $< -lpthread
 
+# test-poll uses pthread_kill to verify blocked read signal delivery.
+$(BUILD_DIR)/test-poll: tests/test-poll.c | $(BUILD_DIR)
+	@echo "  CROSS   $< (with -lpthread)"
+	$(Q)$(CROSS_COMPILE)gcc -D_GNU_SOURCE -static -O2 -o $@ $< -lpthread
+
 # test-osync-requeue drives a raw FUTEX_REQUEUE against a plain-FUTEX_WAIT
 # waiter (musl unlock_requeue pattern) to guard the os_sync wake-at-source
 # degradation; needs -lpthread.
--- a/src/main.c
+++ b/src/main.c
@@ -362,6 +362,13 @@ int main(int argc, char **argv)
     }
     proc_set_fakeroot_enabled(fakeroot);
 
+    /* Block the vCPU-preemption signals and start the sigwait thread before any
+     * vCPU thread exists, so both the normal path and the fork-child path below
+     * inherit the block on every thread they spawn.
+     */
+    if (proc_preempt_init() < 0)
+        return 1;
+
     /* Fork-child mode: receive VM state over IPC and run */
     if (fork_child_fd >= 0)
         return fork_child_main(fork_child_fd, vfork_notify_fd, verbose,
--- a/src/syscall/abi.h
+++ b/src/syscall/abi.h
@@ -740,7 +740,12 @@ typedef struct {
     int linux_flags;     /* Linux open flags (for CLOEXEC tracking) */
     void *dir;           /* DIR* for FD_DIR entries (NULL otherwise) */
     char proc_path[FD_VIRTUAL_PATH_MAX]; /* Virtual /proc dir root for *at */
-    int seals; /* F_SEAL_* bits (non-zero only for memfd_create fds) */
+    int seals;      /* F_SEAL_* bits (non-zero only for memfd_create fds) */
+    bool can_block; /* host read/write on this fd may block (pipe, socket, fifo,
+                     * char/tty); false for regular files and directories. Set
+                     * once at allocation via fstat so the interruptible wait
+                     * path can skip fds that never block.
+                     */
     sock_opt_cache_t sock; /* Socket option cache (zeroed for non-sockets) */
     void (*cleanup)(int guest_fd); /* Type-specific teardown (NULL if none) */
 } fd_entry_t;
--- a/src/syscall/fdtable.c
+++ b/src/syscall/fdtable.c
@@ -18,6 +18,7 @@
 #include <errno.h>
 #include <dirent.h>
 #include <pthread.h>
+#include <sys/stat.h>
 
 #include "utils.h"
 
@@ -72,6 +73,38 @@ static inline void fd_bitmap_set_used(int fd)
     fd_free_bitmap[fd / 64] &= ~BIT64(fd % 64);
 }
 
+/* A host read/write blocks only on non-regular, non-directory fds (pipe,
+ * socket, fifo, char/tty). Callers cache this so the interruptible wait path
+ * can skip fds that never block.
+ */
+static bool host_fd_may_block(int host_fd)
+{
+    struct stat st;
+    return host_fd >= 0 && fstat(host_fd, &st) == 0 && !S_ISREG(st.st_mode) &&
+           !S_ISDIR(st.st_mode);
+}
+
+/* Whether a read/write on a newly allocated fd of this type can block. Pipes
+ * and sockets always can; regular-file slots may actually be a fifo or char
+ * device (opened_fd_type does not split those out) and stdio may be a tty, so
+ * both need an fstat; everything else (dir, path, urandom, fuse, synthetic)
+ * never reaches the blocking wait path. Avoiding the fstat for the common
+ * pipe/socket/synthetic allocations keeps it off the fd-creation lock hold.
+ */
+static bool type_may_block(int type, int host_fd)
+{
+    switch (type) {
+    case FD_PIPE:
+    case FD_SOCKET:
+        return true;
+    case FD_STDIO:
+    case FD_REGULAR:
+        return host_fd_may_block(host_fd);
+    default:
+        return false;
+    }
+}
+
 static inline void fd_init_entry(int fd,
                                  int type,
                                  int host_fd,
@@ -85,6 +118,12 @@ static inline void fd_init_entry(int fd,
     fd_table[fd].dir = NULL;
     fd_table[fd].proc_path[0] = '\0';
     fd_table[fd].seals = 0;
+    /* Cache whether a host read/write can block so the fast-path and slow-path
+     * readers can decide whether to divert into the interruptible wait without
+     * re-stating on every call. Only regular-file and stdio slots pay an fstat
+     * here; pipes/sockets/synthetic fds resolve from the type alone.
+     */
+    fd_table[fd].can_block = type_may_block(type, host_fd);
     sock_opt_clear(&fd_table[fd]);
     fd_table[fd].cleanup = cleanup;
     /* Start conservative. Callers that set linux_flags after allocation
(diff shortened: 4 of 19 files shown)


Max042004:
below paths not covered

Sockets
- `read` on a socket fd — excluded by the `FD_PIPE || FD_STDIO` check
- `recv` / `recvfrom` / `recvmsg` — direct blocking call, `src/syscall/net.c:783`
- `accept` / `accept4` — `src/syscall/net.c:348`
- `connect` — `src/syscall/net.c:546`
- blocking `write` / `send*` on a full buffer (no write-side wait helper exists)

FIFO and pty/character devices (both verified): `opened_fd_type()`
only tests `S_ISDIR` (`src/syscall/fs.c:50`), so these fall into `FD_REGULAR`
and inherit the regular-file exclusion — but Linux makes reads on them
interruptible. `open(O_RDONLY)` on a FIFO also blocks in the host `open()`
itself.

Consequences:
- A guest-level `kill` cannot reach a process stuck in blocking `recv`/`accept`;
  only killing the host process works.
- The classic `alarm(n)` + blocking `read(sock)` timeout pattern never fires.
- Ctrl-C is ignored by programs using default-blocking sockets (CPython signal
  handling relies on EINTR per PEP 475; same for `nc`, blocking DB clients).
- Runtimes using non-blocking fds + epoll/poll (node, Go, nginx) are unaffected,
  and `SO_RCVTIMEO` passes through (`src/syscall/net-abi.c:291`), which limits
  the blast radius to default-blocking usage.

PR #146, opened 2026-07-04
Clear epoll registrations on close

### LINKED ISSUE #78: `sys_close` does not clear epoll registrations — stale `regs[fd].active` leaves the PR #73 ABA guard load-bearing
## Summary

Closing a guest fd does not remove that fd from any epoll instance's interest
table. Each epoll instance keeps its own `epoll_reg_t regs[FD_TABLE_SIZE]`
indexed by guest fd number (`src/syscall/poll.c:640-642`), but `sys_close()`
(`src/syscall/fs.c:423`) — and every other close path that funnels through
`fd_cleanup_entry()` (dup2-over-existing, the execve CLOEXEC sweep) — only tears
down the `fd_table[fd]` slot. It never touches the `regs[fd]` entry of the epoll
instances that registered that fd, so `regs[fd].active` stays `true` after the
fd is gone.

On Linux, closing the last descriptor that refers to an open file description
auto-removes it from every epoll interest list. elfuse leaves the registration
behind.

This was flagged by jserv in the PR #73 review as an *adjacent, pre-existing*
issue, explicitly to be filed separately:

> Adjacent (pre-existing, but more observable after this change): `sys_close`
> does not clear `inst->regs[fd].active` for epoll instances holding the fd, so
> a closed-then-reopened guest fd still looks active to epoll. Worth filing
> separately.

## Root cause

There is no reverse index from a target fd to the set of epoll instances that
registered it. Epoll instances are reachable only through their own `epfd` in
`fd_table[epfd].dir`. A close has the fd number in hand but no cheap way to find
"which epoll instances watch this fd," so it does nothing — the registration
tables drift out of sync with reality and only get corrected lazily, the next
time `epoll_ctl()` happens to touch that exact `(epfd, fd)` pair.

## Current mitigation (PR #73) and why it is not a fix

PR #73 added an `fd_entry_t.generation` stamp and a cross-call ABA guard in
`sys_epoll_ctl()` (`src/syscall/poll.c:743`). At the head of every
`epoll_ctl()`, a registration whose stamped generation no longer matches
`fd_table[fd].generation` is treated as stale and dropped, so DEL/MOD report
`ENOENT` and ADD starts fresh after a close+reopen of the same fd number. This
makes the **observable** close+reopen ABA correct and is covered by
`tests/test-epoll-aba.c`.

But the guard is a **point-of-use patch at the symptom layer**, not a fix for
the root cause:

- The registration table still **lies**: `regs[fd].active == true` for a fd that
  is closed (and possibly reopened as a different file). Correctness now depends
  on *every* reader re-deriving truth from `generation`.
- `sys_epoll_ctl()` does consult `generation`. **`sys_epoll_pwait()` does not.**
  The readiness loop maps `kevents[i].udata` (the guest fd number) straight back
  into `inst->regs[gfd].active` / `.oneshot_armed` with no generation check
  (`src/syscall/poll.c:1010-1012` and `1046-1051`). Today this is masked because
  a closed host fd's knote is already gone from kqueue, so no event carries the
  stale `udata` — but that is an indirect kqueue side effect, not an invariant
  the code enforces. Any future change that re-adds a host fd, or a path that
  reuses a `udata` value, resurfaces the stale entry as a wrong/mismatched-data
  readiness report.
- The guard is therefore **load-bearing**: remove or bypass the generation check
  in any one epoll read path and the original wrong-knote / wrong-data bug
  returns. That is fragile for a correctness-critical path.

Stale entries also simply linger for the epoll instance's whole lifetime
(`regs[]` is `FD_TABLE_SIZE == 1024` entries per instance), never reflecting that
the fd is gone.

## Out of scope (related, separate divergence)

elfuse keys epoll state on the guest fd *number*, while Linux keys on the open
file *description*. So `dup()`-then-close-the-original (registration should
survive via the surviving descriptor) is a distinct modeling gap that neither
the generation guard nor eager-cleanup-on-close addresses. Track separately;
do not conflate with this issue.

## Reproduction (state-level; observable symptom is currently masked by the guard)

1. `epfd = epoll_create1()`; `fd = eventfd()`.
2. `epoll_ctl(epfd, EPOLL_CTL_ADD, fd, …)` → `regs[fd].active = true`.
3. `close(fd)`.
4. Inspect the instance: `regs[fd].active` is still `true` (Linux: the
   registration is gone).

With the PR #73 guard in place the *observable* DEL/MOD/ADD behavior after a
reopen is correct (see `test-epoll-aba.c`); this issue is about the underlying
state being wrong and the guard being the only thing standing between that wrong
state and a user-visible bug.

## References

- PR #73 "Fix epoll_ctl dropping registrations in multi-threaded guests" — adds
  the generation guard and `test-epoll-aba.c`.
- `src/syscall/poll.c:743` — cross-call ABA guard (consults `generation`).
- `src/syscall/poll.c:1010-1012`, `1046-1051` — `sys_epoll_pwait` reads
  `regs[].active` / `.oneshot_armed` **without** a generation check.
- `src/syscall/poll.c:640-642` — per-instance `regs[FD_TABLE_SIZE]` table.
- `src/syscall/fs.c:423` (`sys_close`), `src/syscall/fdtable.c:457`
  (`fd_cleanup_entry`) — close teardown that does not touch epoll state.

jserv:
Closing a guest fd did not remove it from any epoll instance's interest table. Each instance keeps its own regs[FD_TABLE_SIZE] indexed by guest fd number, and every close path (sys_close, close_range, dup2 over an open slot, the execve CLOEXEC sweep) tore down only the fd_table slot, leaving regs[fd].active set for a fd that was gone. Linux auto-removes a fd from every epoll interest list when its last descriptor closes; elfuse leaned entirely on the sys_epoll_ctl generation guard to notice the stale entry lazily, and sys_epoll_pwait never re-checks generation.

epoll_note_fd_closed() now clears regs[fd] across every live instance, hooked into fd_mark_closed_unlocked() -- the single close chokepoint -- and the fd_alloc_at overwrite path that dup2/dup3 takes around it. A fd-lock-guarded epoll_live_count keeps the scan a no-op when no epoll fd exists. The kqueue knote needs no EV_DELETE: the following host fd close drops it, and clearing the software state is what makes the pwait active-check honest instead of load-bearing on the generation guard, which stays as defense in depth.

The same audit closed two adjacent multi-threaded hazards on the heap-allocated epoll instance, reachable only through fd_table[epfd].dir:
- Use-after-free: sys_epoll_ctl / sys_epoll_pwait walked inst->regs for the whole call (including across the blocking kevent() in pwait) while host_fd_ref_open pinned only the host kqueue fd. A sibling closing epfd freed inst mid-call. Fixed with a refcount guarded by fd_lock (epoll_instance_acquire / _release); the allocation is freed only once the table reference and every in-flight reference are gone. ASAN pinned the fault at sys_epoll_pwait before the fix and is clean after.
- Data race on regs[]: the new close hook writes regs[] under fd_lock while ctl/pwait accessed it locklessly. Added a per-instance mutex (mirroring the Linux eventpoll->mtx) held only for the short reg bookkeeping, never across the blocking kevent(); the close hook takes it under fd_lock (order documented in internal.h). ThreadSanitizer flagged the epoll_note_fd_closed vs sys_epoll_pwait race without the lock and is clean with it.
- Slot publication window: sys_epoll_create1 made the FD_EPOLL slot visible before attaching dir, so a close_range racing the create could tear down a half-built slot. New fdtable helper fd_alloc_dir() installs type, host_fd, dir, and flags in one fd_lock critical section, so the slot is never observable as FD_EPOLL with a NULL dir.

Close #78

---
## Summary by cubic
Fixes stale epoll watches by clearing registrations on close/dup2 and makes dup of an epoll fd share the same instance, matching Linux. Also hardens epoll lifetime to eliminate a use-after-free and data races.

- **Bug Fixes**
  - Clear epoll interest for a closed or dup2/dup3-overwritten guest fd across all instances; wired into the close chokepoint and the fd overwrite path; fast no-op when no epoll fds via epoll_live_count.
  - Remove UAF/races by refcounting epoll instances and guarding regs[] with a per-instance mutex (lock order: fd_lock → inst->lock; never held across kevent()).
  - Dup/dup3/F_DUPFD of an epoll fd now shares the eventpoll instance; slots are published atomically via `fd_alloc_dir*`, lifetime pinned via refcount, and freed with `epoll_instance_free` (source validated by host_fd + generation to reject ABA).
  - Tests added and in matrix: `test-epoll-close`, `test-epoll-dup`, `test-epoll-refcount`.

<sup>Written for commit c711243b5464bae64ca39b1d51648ebff37e25da. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/146?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/poll.c
+++ b/src/syscall/poll.c
@@ -707,7 +707,8 @@ typedef struct {
                           * close+reopen ABA: if the guest fd's current
                           * generation no longer matches, the registered open
                           * file is gone and this stale entry must not drive
-                          * kevent against the reused host fd. */
+                          * kevent against the reused host fd.
+                          */
     bool active;         /* Registered in this instance */
     bool oneshot_armed;  /* EPOLLONESHOT and event already fired,
                           * waiting for EPOLL_CTL_MOD re-arm.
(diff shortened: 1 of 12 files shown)


src/syscall/poll.c
```diff
@@ -758,19 +879,42 @@ int64_t sys_epoll_create1(int flags)
...
+    /* Count this instance for epoll_note_fd_closed()'s fast-path skip. A
+     * pathological close_range() racing this create can close gfd between the
+     * publish above and here; the close hook then frees the instance and this
+     * increment leaves the counter one too high. Each such race can add one, so
+     * the counter may stay positive with no live instance -- that only forces
+     * the scan to run when it need not (perf only, and the guarded decrement
+     * still never underflows), never a correctness or memory-safety issue.
+     */
+    pthread_mutex_lock(&fd_lock);
+    epoll_live_count++;
```
Max042004, line 987: A sibling thread that closes the brand-new fd in the gap between them frees the instance before the counter is ever incremented for it.
Once this fires, the fast-path skip in `epoll_note_fd_closed()`is defeated forever, and every subsequent `close()` of any fd in the process pays the full scan.

Suggested fix: fold the `epoll_live_count++` into the same `fd_lock` critical section that `fd_alloc_dir()` already opens to publish the slot, so "publish" and "count" become
one atomic step and the gap disappears.

src/syscall/poll.c
```diff
@@ -721,9 +721,130 @@ typedef struct {
...
+ * cleared slot's own instance (already FD_CLOSED, hence skipped) are stable.
+ * The kqueue knote itself needs no EV_DELETE: the host fd close that follows
+ * drops it automatically, and clearing the software state is what makes the
+ * pwait active-check honest.
+ */
+void epoll_note_fd_closed(int closed_fd)
+{
+    if (epoll_live_count == 0)
+        return;
+    for (int epfd = 0; epfd < FD_TABLE_SIZE; epfd++) {
```
Max042004, line 885: The fast-path skip only helps processes with *zero* epoll fds. Any long-lived
event-loop process — i.e. the common case — keeps at least one epoll fd open,
so every `close()` walks the full 1024-entry table looking for the `FD_EPOLL` slots.

Suggested fix: `epoll_live_count` is already maintained under `fd_lock`;
back it with a small compact array of live `epfd` numbers and have `epoll_note_fd_closed()` iterate that instead of the full table — turns the common case from O(1024) into O(live_instances).

Max042004 (changes requested):

Max042004:
`duplicate_guest_fd()` (`src/syscall/fs.c:716`) special-cases `FD_FUSE_*` and
`FD_EVENTFD` but not `FD_EPOLL`, so a duped epoll fd falls into the generic
path. `clone_dir_stream()` only clones for `FD_DIR` and returns `NULL`
for anything else, so `install_fd_alias_metadata_atomic()`'s `if (dir) ...`
never pass — the alias slot keeps `dir == NULL` forever.

PR #148, opened 2026-07-04
Raise host fd limit to back the guest fd table

### LINKED ISSUE #68: test-stress FD exhaustion subtest fails with default macOS file-descriptor limit
`make check` fails on `test-stress`:

```
test-stress                                   [ FAIL ] (exit 1)
    test-stress: resource limit stress tests
      32 concurrent threads          OK
      mmap/munmap churn (256 cycles) OK
      128 concurrent mmap regions    OK
      FD exhaustion (EMFILE)         FAIL: too few FDs opened before exhaustion (errno=24)
      mprotect RW->RX->RW cycling (64x) OK
      large mmap (16MiB)             OK
```

The subtest at `tests/test-stress.c:222` expects to open ≥ 500 pipe FDs before EMFILE. On a default macOS shell session that's not reliable, `pipe(2)` returns EMFILE earlier. `ulimit -n 65536` before `make check` and the test passes.

It's not an elfuse internal cap: `FD_TABLE_SIZE = 1024` (`src/syscall/abi.h:628`), comfortably above the 500 threshold.
The bottleneck is the host process `RLIMIT_NOFILE` flowing into `pipe(2)`.

Three places this could be fixed, in order of how localized:

1. `test-stress` bumps `RLIMIT_NOFILE` itself before the FD-exhaustion subtest.
2. elfuse bumps `RLIMIT_NOFILE` on the host before launching the guest.
3. `docs/testing.md` *Build Requirements* documents the required `ulimit -n`.

**Environment:** Apple M3 Max MacBook Pro, macOS 26.5.1. `kern.maxfilesperproc` = 245760, zsh interactive `ulimit -n` =
1048576, but the effective limit visible to the guest is clearly lower (`make check` is invoked from a shell where the parent
limit isn't propagated as expected).

LeoriumDev:
Fixes #68 

The test-stress FD-exhaustion subtest fails whenever make check runs from a shell with the stock macOS soft fd limit (256). The guest is told RLIMIT_NOFILE is 1024 (FD_TABLE_SIZE) and every guest fd is backed by a host fd, so once the host process runs out at ~250 the guest sees EMFILE far below its advertised limit.

Fix it in fdtable_init(): raise the host soft limit to min(OPEN_MAX, rlim_max) before any guest code runs, following the macOS setrlimit(2) man page. A limit that is already high enough is left alone, and a failed setrlimit is ignored. Host fds can now exceed FD_SETSIZE, which is fine — sys_pselect6 already falls back to poll() when a host fd doesn't fit in an fd_set.

Since dev machines usually run with raised shell limits, a regression here would go unnoticed, so this also adds a check step (test-fd-limit-stock) that reruns test-stress with the soft limit forced to 256 in a child shell.

Tested on an M3 Max: the repro above fails on unpatched main and passes with the patch; full make check stays green.

---
## Summary by cubic
Raises the host soft file descriptor limit at startup so the guest can use its full fd table without hitting EMFILE on macOS’s stock 256 limit. Adds a regression check that reruns the stress test under the stock limit.

- **Bug Fixes**
  - In `fdtable_init()`, raise `RLIMIT_NOFILE` to `min(OPEN_MAX, rlim_max)`; leave higher limits as-is and ignore `setrlimit` failures.
  - Add `test-fd-limit-stock` to run `test-stress` with `ulimit -S -n 256` to catch regressions.

<sup>Written for commit 6d66e60313afb5c5e31173843ccb5ec665a702a5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/148?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/mk/tests.mk
+++ b/mk/tests.mk
@@ -15,7 +15,7 @@
         test-sysroot-create-paths test-fork-ipc-protocol-host test-identity-override-host \
         test-proctitle-host test-proctitle-low-stack \
         test-sysroot-procfs-exec test-timeout-disable test-fuse-alpine \
-        test-sysroot-nofollow test-sysroot-chdir perf
+        test-sysroot-nofollow test-sysroot-chdir test-fd-limit-stock perf
 
 ## Build and run the assembly hello world test
 test-hello: $(ELFUSE_BIN) $(TEST_HELLO_DEP)
@@ -43,6 +43,8 @@ check: $(ELFUSE_BIN) $(TEST_DEPS) check-syscall-coverage \
 		$(BUILD_DIR)/test-fork-ipc-protocol-host \
 		$(BUILD_DIR)/test-identity-override-host
 	@bash tests/driver.sh -e $(ELFUSE_BIN) -d $(TEST_DIR) -v
+	@printf "\n$(BLUE)━━━ stock fd-limit (256) regression ━━━$(RESET)\n"
+	@$(MAKE) --no-print-directory test-fd-limit-stock
 	@printf "\n$(BLUE)━━━ TLBI RVAE1IS encoder unit test ━━━$(RESET)\n"
 	@$(BUILD_DIR)/test-tlbi-encoder-host
 	@printf "\n$(BLUE)━━━ fork IPC protocol identity unit test ━━━$(RESET)\n"
@@ -73,6 +75,15 @@ check: $(ELFUSE_BIN) $(TEST_DEPS) check-syscall-coverage \
 	@$(MAKE) --no-print-directory test-rosetta-cli
 	@printf "\n$(BLUE)━━━ hot-syscall guardrail ━━━$(RESET)\n"
 	@$(MAKE) --no-print-directory test-bench-guardrail
+
+## Rerun the fd-heavy stress test with the soft RLIMIT_NOFILE forced to the
+## stock macOS default (256): fdtable_init must raise the host limit itself.
+FD_LIMIT_STOCK_DEPS := $(ELFUSE_BIN)
+ifndef GUEST_TEST_BINARIES
+  FD_LIMIT_STOCK_DEPS += $(BUILD_DIR)/test-stress
+endif
+test-fd-limit-stock: $(FD_LIMIT_STOCK_DEPS)
+	@bash -c 'ulimit -S -n 256 && exec "$(ELFUSE_BIN)" "$(TEST_DIR)/test-stress"'
 
 ## Hot-syscall performance guardrail: ensure getpid, libc clock_gettime,
 ## and 1-byte /dev/urandom reads stay under their TODO ns/op ceilings.
--- a/src/syscall/fdtable.c
+++ b/src/syscall/fdtable.c
@@ -10,10 +10,12 @@
  * access through fd_lock.
  */
 
+#include <limits.h>
 #include <stdatomic.h>
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/resource.h>
 #include <unistd.h>
 #include <errno.h>
 #include <dirent.h>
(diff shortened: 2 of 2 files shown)


Max042004:
existing issue related to this PR:

1. `src/syscall/sys.c:684` - `getrlimit(RLIMIT_NOFILE)` reports more than the guest can use

`sys_prlimit64()` currently falls through to host `getrlimit()` on a cache miss, so after this PR the guest can observe `RLIMIT_NOFILE=10240`. The actual guest fd table is still capped at `FD_TABLE_SIZE=1024`, and `fd_alloc*()` rejects fds past that limit. This makes `getrlimit`/`ulimit -n` over-report while `/proc/self/limits` already reports the clamped guest value.

Please clamp the `RLIMIT_NOFILE` read path to the guest fd table limit, matching the existing `/proc/self/limits` special case.

2. `src/syscall/sys.c:703` - guest `setrlimit(RLIMIT_NOFILE)` can lower the host limit and undo this fix

The guest should be allowed to change its own `RLIMIT_NOFILE`, but that should be a guest-visible limit enforced by the fd table, not a direct change to the emulator process budget. Today the `RLIMIT_NOFILE` set path calls host `setrlimit()` first, then updates `fd_set_rlimit_nofile()`. A guest daemon that sets nofile to 256 can therefore lower the host process limit back to the exact stock-macOS condition this PR is trying to avoid.

Please virtualize guest `RLIMIT_NOFILE`: update the guest fd-table limit and cached guest value, but do not let guest nofile reductions reduce the host process limit that elfuse needs for internal fds and multi-host-fd emulations.

src/syscall/fdtable.c
```diff
@@ -145,11 +147,35 @@ static int fd_bitmap_find_free(int minfd)
...
+    rl.rlim_cur = want;
+    (void) setrlimit(RLIMIT_NOFILE, &rl);
+}
+
 /* Initialize the FD table and bitmap, pre-open stdin/stdout/stderr. Extracted
  * from syscall_init(); call before any guest code runs.
  */
 void fdtable_init(void)
 {
+    fd_raise_host_nofile();
```
Max042004, line 177: `fdtable_init()` is called from `syscall_init()`, and fork children call `syscall_init()` again while rebuilding state in `fork_ipc_recv_fd_table()`. That means a guest that lowers `RLIMIT_NOFILE` before `fork()` can get a child whose host soft limit is silently raised back to `OPEN_MAX`, which violates Linux rlimit inheritance semantics.

Please either run the host raise only during initial bootstrap, or serialize/restore the guest nofile state across the fork path before guest code resumes.

Max042004 (changes requested):

mk/tests.mk
```diff
@@ -74,6 +76,15 @@ check: $(ELFUSE_BIN) $(TEST_DEPS) check-syscall-coverage \
 	@printf "\n$(BLUE)━━━ hot-syscall guardrail ━━━$(RESET)\n"
 	@$(MAKE) --no-print-directory test-bench-guardrail
 
+## Rerun the fd-heavy stress test with the soft RLIMIT_NOFILE forced to the
+## stock macOS default (256): fdtable_init must raise the host limit itself.
+FD_LIMIT_STOCK_DEPS := $(ELFUSE_BIN)
+ifndef GUEST_TEST_BINARIES
+  FD_LIMIT_STOCK_DEPS += $(BUILD_DIR)/test-stress
+endif
+test-fd-limit-stock: $(FD_LIMIT_STOCK_DEPS)
```
Max042004, line 85: Please wrap this target with the same 60-second timeout used by the normal test driver, or move the lowered-rlimit case into a driver script that already has timeout handling.

Max042004 (changes requested):

Max042004:
PR #204 already addressed the issue. So close.

PR #150, opened 2026-07-04
Fix kill() broadcast and process-group signal

### LINKED ISSUE #131: Incorrect signal broadcast behavior and missing self-signaling on `kill(-1)`
### **Incorrect signal broadcast behavior and missing self-signaling on `kill(-1)`**

In the `kill(pid, sig)` system call implementation, handling of `pid = -1` (broadcast signal) is incomplete and deviates from Linux/POSIX specifications by failing to signal the calling process itself.

---

### **Location**
* `src/syscall/syscall.c` (inside the `sc_kill` handler)

---

### **Code**
```c
if (pid == -1) {
    pid_t child_pids[256];
    int count = proc_get_child_pids(child_pids, ARRAY_SIZE(child_pids));
    int delivered = 0;
    for (int i = 0; i < count; i++) {
        if (proc_send_guest_signal(child_pids[i], sig) == 0)
            delivered++;
    }
    return delivered > 0 ? 0 : -LINUX_ESRCH;
}
```

---

### **Issue**
Under Linux/POSIX specifications, calling `kill(-1, sig)` must send the signal to every process that the calling process has permission to signal (excluding `init` / PID 1). 
* The implementation correctly broadcasts the signal to emulated child processes using `proc_get_child_pids`.
* However, the **calling process itself is omitted** from receiving the signal. Since a process always has permission to signal itself, the calling process itself must also be signaled during a `kill(-1)` broadcast. 
* The previous implementation correctly queued the signal to the caller via `signal_queue(sig)`, but this was completely replaced, causing self-signaling to break.

---

### **Correct Behavior**
When `pid == -1`, `sc_kill` should iterate and deliver the signal to all emulated child processes, and additionally call `signal_queue(sig)` to deliver the signal to the calling process itself.

jserv:
sc_kill mishandled the negative-pid forms. Correct them to match Linux and the qemu-aarch64 ground truth:
- kill(-1, sig): broadcast to the emulated fork children over the cross-process transport, excluding the caller. Per kill(2), Linux does not signal the calling process on kill(-1); the prior code did the opposite (self only, no children). Return 0 on any delivery, the first transport errno if children existed but every send failed, else ESRCH.
- kill(0, sig) and kill(-pgid, sig): deliver to every process in the target group -- children, siblings, and the parent -- not just self. Because each guest process is a separate host process, group membership is shared through a per-fork-family registry file keyed by the abstract-socket namespace id, so independent sessions stay isolated. Membership is published at fork, setpgid (self and parent-side), and setsid; publish compacts the file in place so it stays bounded by live members rather than growing per event. A process syncs its own pgid from registry before acting on its group, so setsid's already-leader check and the kill group path observe the current group even after a parent-side setpgid(child).

Hardening:
- setpgid rejects a negative group id (EINVAL) and negative pid (ESRCH), reading the args as 32-bit like sc_kill.
- The signal transport tags each line with the sender's namespace so a recycled host pid on an unrelated session drops the delivery instead of applying it, and proc_send_guest_signal revalidates that the target is still this binary before signaling.

Close #131

---
## Summary by cubic
Fixes kill() negative-PID semantics to match Linux. kill(-1) now broadcasts to every member of the emulated fork family except self, and kill(0)/kill(-pgid) signal the full process group across host processes via a shared registry. Hardens signal transport to block cross-session and cross-guest PID reuse.

- **Bug Fixes**
  - kill(-1, sig) broadcasts to all fork-family peers and never to self.
  - kill(0, sig) and kill(-pgid, sig) reach the entire target group (children, siblings, parent) across processes.
  - Linux-compatible results: 0 on any delivery; ESRCH when no targets; for broadcast/group if all sends fail, return the first transport errno; sig==0 probes match Linux; setpgid rejects negative pgid (EINVAL) and negative pid (ESRCH).

- **New Features**
  - Fork-family process-group registry keyed by the abstract-socket namespace: published at fork, setpgid (self and parent-side), and setsid; processes sync pgid before group ops; setpgid’s group-existence check reads the registry; file compacted in place.
  - Hardened signal transport: tag deliveries with namespace and target guest pid; verify the target runs this binary; reader drops mismatched sessions/guests and rejects garbage; owner cleans the registry only when last member; warn when full; log failed truncates.
  - Tests added and wired into CI: `test-kill-broadcast`, `test-kill-pgroup`.

<sup>Written for commit 20d1be259b2881377dcad2a67919545a90dd4c3b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/150?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/runtime/forkipc.c
+++ b/src/runtime/forkipc.c
@@ -1528,6 +1528,14 @@ int64_t sys_clone(hv_vcpu_t vcpu,
         }
     }
 
+    /* Refresh our own process group from the registry before capturing it: a
+     * parent-side setpgid(this, ...) updates the registry but not the local
+     * cached pgid, so both the child's inherited group (hdr.pgid below) and the
+     * publish that follows would otherwise carry -- and re-stamp the registry
+     * with -- the stale group, misrouting later kill(0)/kill(-pgid).
+     */
+    proc_registry_sync_self_pgid(g);
+
     /* Snapshot of the semantic region array, populated after the memory dump
      * but before sibling vCPUs resume. Declared up front so all goto paths to
      * fail_snapshot can free it unconditionally. Header
@@ -1572,6 +1580,7 @@ int64_t sys_clone(hv_vcpu_t vcpu,
         .ctid_gva = ctid_gva,
         .shm_is_clone = (snapshot_shm_fd >= 0) ? 1 : 0,
     };
+    proc_registry_publish_self();
     if (fork_ipc_write_all(ipc_sock, &hdr, sizeof(hdr)) < 0) {
         log_error("clone: failed to send header");
         goto fail_snapshot;
@@ -1683,8 +1692,11 @@ int64_t sys_clone(hv_vcpu_t vcpu,
      * its own MAP_PRIVATE view of the same file.
      */
 
-    /* Register after successful IPC so wait4/waitid can observe the child. */
-    proc_register_child(child_host_pid, child_guest_pid);
+    /* Register after successful IPC so wait4/waitid can observe the child. Seed
+     * the child's group from the same value sent in the fork header so the
+     * parent's table matches the group the child actually inherited.
+     */
+    proc_register_child(child_host_pid, child_guest_pid, hdr.pgid);
 
     /* CLONE_VFORK suspends the parent until the child exits or execs. The
      * emulator cannot observe guest exec completion across the helper process,
--- a/src/syscall/proc-identity.c
+++ b/src/syscall/proc-identity.c
@@ -259,6 +259,14 @@ void proc_set_session(int64_t sid, int64_t pgid)
     guest_fg_pgrp = pgid;
 }
 
+void proc_set_pgid_from_registry(guest_t *g, int64_t pgid)
+{
+    pthread_mutex_lock(&session_lock);
+    guest_pgid = pgid;
+    proc_publish_pgsid_locked(g);
+    pthread_mutex_unlock(&session_lock);
+}
+
 void proc_set_fg_pgrp(int64_t pgrp)
 {
     guest_fg_pgrp = pgrp;
--- a/src/syscall/proc-pidfd.c
+++ b/src/syscall/proc-pidfd.c
@@ -245,7 +245,7 @@ int64_t sys_pidfd_send_signal(guest_t *g,
                 return -LINUX_ESRCH;
             return 0;
         }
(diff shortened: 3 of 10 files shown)


doanbaotrung:
I have tested the fix using standard native tools (`bash` and `/bin/kill`) without using any compiled test binaries:

### 1. Process Group Signaling (`kill 0`)
We spawned a background child subshell trapping `SIGUSR1`, ignored `SIGUSR1` in the parent, and sent `kill -USR1 0`:
```bash
/bin/bash -c \
  'OUT=/tmp/sig_out.txt; rm -f $OUT; ( trap "echo CHILD_RCV_USR1 >> $OUT; exit 0" USR1; sleep 10 & wait $! ) & CHILD_PID=$!; sleep 0.5; trap "" USR1; kill -USR1 0; sleep 0.5; cat $OUT'
```
Output
```
CHILD_RCV_USR1
ELF entry   : 0x33840
Load range  : 0x0 – 0x193d80
Segments    : 2
```

* **Result**: **`CHILD_RCV_USR1`** (Exit code: 0)
* **Status**: **PASSED** (The sibling/child process in the process group successfully received the signal).

### 2. Broadcast Signaling excluding Caller (`kill -1`)
We spawned a background child subshell trapping `SIGUSR2` and sent `kill -USR2 -1` via `/bin/kill` to deliver to all processes except the caller (parent):
```bash
/bin/bash -c \
  'OUT=/tmp/sig_out.txt; rm -f $OUT; ( trap "echo CHILD_RCV_USR2 >> $OUT; exit 0" USR2; sleep 10 & wait $! ) & CHILD_PID=$!; sleep 0.5; /bin/kill -USR2 -1; sleep 0.5; cat $OUT'
```
Output:
```
CHILD_RCV_USR2
ELF entry   : 0x33840
Load range  : 0x0 – 0x193d80
Segments    : 2
```

* **Result**: **`CHILD_RCV_USR2`** (Exit code: 0)
* **Status**: **PASSED** (The child process successfully received `SIGUSR2`. The parent process was correctly excluded from the broadcast signal, surviving the command without a trap and terminating successfully).

PR #153, opened 2026-07-05
Route thread-directed signals to the target thread

### LINKED ISSUE #152: tgkill discards the target thread: thread-directed signals are delivered process-wide
`sc_tgkill` validates that the target tid exists, then drops it: the signal is
queued into the single process-global pending set, and whichever thread passes
a signal-delivery point first (and does not block the signal) consumes it.
Thread-directed signals (`tgkill`, `tkill`, `rt_tgsigqueueinfo`, and therefore
`pthread_kill`) behave exactly like process-directed `kill`.

This predates PR #145 (its diff does not touch `sc_tgkill`), but that PR makes
the gap load-bearing: its purpose is to let preemption signals interrupt
threads blocked in host syscalls — and the signal now reliably interrupts the
*wrong* thread.

## Root cause

1. **Send side drops the target.** `sc_tgkill` (src/syscall/syscall.c:932-940)
   uses the `thread_find(tid)` result only to decide between `0` and
   `-ESRCH`, then calls `signal_queue(sig)`, which takes no thread argument.
   `sc_rt_tgsigqueueinfo` (src/syscall/syscall.c:961-1004) has the same shape.

2. **There is only one pending set.** `signal_queue` writes into the single
   static `sig_state` (src/syscall/signal.c:44); `signal_state_t.pending`
   (src/syscall/signal.h:182) is one process-wide bitmask. `thread_entry_t`
   (src/runtime/thread.h:53-55) only carries per-thread *blocked* masks — no
   per-thread pending set exists.

3. **Consumption is first-come-first-served.** Each vCPU thread's syscall
   epilogue (src/syscall/proc.c:1334) calls `signal_pending()`, which computes
   `sig_state.pending & ~current_thread->blocked` (src/syscall/signal.c:458-474).
   `attention_raise()` (src/syscall/signal.c:313-321) is a broadcast
   (`thread_interrupt_all()` + wakeup pipe), so every non-blocking thread races
   for the signal.

On Linux each task has *two* pending sets: `task->pending` (thread-directed;
tgkill/tkill target this) and `signal->shared_pending` (process-directed;
kill targets this). elfuse models only the shared set. The comment above
`sc_rt_sigqueueinfo` (src/syscall/syscall.c:1009-1026) documents the
process-wide queueing — which is correct for `rt_sigqueueinfo` (genuinely
process-directed on Linux), but `tgkill`/`rt_tgsigqueueinfo` share the same
downstream path and inherit the wrong routing.

**Aggravation — coalescing.** Standard signals (1-31) are one bit in the
global mask (`signal_standard_enqueue_locked`, src/syscall/signal.c:183-187).
N `tgkill`s of the same signal to N different threads collapse into a single
delivery, where Linux would keep N per-thread pending instances.

## Reproduction

A worker thread blocks in `read()` on an empty pipe; the main thread sends
`pthread_kill(worker, SIGUSR1)` and then spins in a `usleep` loop. The
`nanosleep` epilogue of the *main* thread consumes the signal: the handler
runs on main, the worker never wakes. Blocking SIGUSR1 in the main thread
first "fixes" it — routing by elimination, proving delivery targets the
process, not the tid.

## Why tests/test-poll.c does not catch this

The "SA_RESTART read handler runs" test is thread-agnostic on both axes:

- Its assertions (`got_usr1` global flag; handler writes `'h'` into the pipe
  the main thread reads, unblocking the `read` regardless of which thread ran
  the handler) pass even under misdelivery.
- Its sender thread blocks SIGUSR1 before sending, so in a two-thread process
  delivery lands on main by elimination — i.e. the test also passes under
  purely process-directed semantics.

## Impact

- **Go async preemption**: `preemptM` sends `tgkill(SIGURG)` to a specific M.
  A wrong M consumes and ignores it; the target is never preempted and
  `suspendG` retries — delayed preemption / long GC pauses, near-livelock in
  tight loops.
- **`pthread_cancel`** (glibc and musl use tgkill + SIGCANCEL): a wrong thread
  finds no cancel flag and returns; the cancel is silently lost (no retry).
- **glibc setxid / musl `__synccall`** (multithreaded `setuid()` etc.):
  broadcasts one signal per thread and waits for per-thread acks; one thread
  can consume several instances, deadlocking the caller.

jserv:
{tg,t}kill, rt_tgsigqueueinfo, and thus pthread_kill validated the target tid then queued the signal into the single process-wide pending set, so whichever thread reached a delivery point first consumed it and standard signals coalesced across threads. Go async preemption, pthread_cancel, and glibc setxid / musl __synccall all interrupted the wrong thread; interruptible blocking I/O made the gap load-bearing.

Model Linux's 2 pending sets. Extract signal_pending_t (pending bitmask, coalesced standard siginfo, per-signal RT queues); signal_state now holds one shared instance (signal->shared_pending, hit by kill) and each thread_entry_t holds its own private tpending (task->pending, hit by tgkill/tkill). signal_deliver drains the current thread's private set before the shared set, matching dequeue_signal; signal_pending, sigsuspend, rt_sigprocmask unblock, and rt_sigpending union the two.

Resolve the target under thread_lock and enqueue into its tpending under sig_lock then thread_lock so a concurrent exit cannot misroute; kill and rt_sigqueueinfo stay process-directed. Add tkill (syscall 130), which was ENOSYS. Both directed senders validate tgid/tid and treat sig 0 as an existence probe.

signalfd reads the caller's private set then the shared set with a per-entry source tag so peek and take consume from the same set; a queued directed signal wakes signalfd waiters (Linux wakes the shared sighand queue) and a blocking read re-waits on a false wake. The pending hint becomes shared.pending OR the union of every active thread's private set, recomputed on thread exit so a departed thread's unconsumed directed signal stops pinning the identity-syscall fast path off.

Close #152

---
## Summary by cubic
Route thread-directed signals to the exact target thread and match Linux delivery semantics. Adds `SYS_tkill`, corrects `tgkill`/`rt_tgsigqueueinfo` routing, and hardens signalfd to handle per-thread vs shared pending sets.

- **New Features**
  - Two pending sets: shared (process-directed) and per-thread (`tpending`). Delivery drains the thread’s private set before the shared set; `signal_pending`, `rt_sigpending`, `rt_sigsuspend`, and `rt_sigprocmask` union the two.
  - Add `SYS_tkill` (130). Route `tgkill`/`rt_tgsigqueueinfo` to the target thread with `tgid` validation; treat `sig==0` as an existence probe. Keep `kill`/`rt_sigqueueinfo` process-directed and accept a thread TID as the PID; also honor `sig==0`.
  - `signalfd`: read private then shared with a per-entry source tag so take consumes from the same set; re-wait on false wakes; wake signalfd on directed signals so target readers unblock while non-targets see `EAGAIN`.

- **Bug Fixes**
  - Fix thread-directed signals being handled by the wrong thread and stop standard-signal coalescing across threads.
  - Resolve target under `thread_lock` and enqueue into its private set under `sig_lock`→`thread_lock` to avoid misrouting during concurrent thread exit.
  - Recompute the global pending hint as (shared.pending OR union of all threads’ private masks) and refresh on thread slot deactivation/exit so stale bits do not pin fast paths off.
  - Tests: add `test-tgkill-directed` and a signalfd case ensuring a directed signal wakes the target’s signalfd.

<sup>Written for commit 217bb85cc20c900b9c6d531a5080e4a53ca602b5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/153?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -200,6 +200,11 @@ $(BUILD_DIR)/test-poll: tests/test-poll.c | $(BUILD_DIR)
 	@echo "  CROSS   $< (with -lpthread)"
 	$(Q)$(CROSS_COMPILE)gcc -D_GNU_SOURCE -static -O2 -o $@ $< -lpthread
 
+# test-tgkill-directed verifies thread-directed (pthread_kill/tgkill) routing.
+$(BUILD_DIR)/test-tgkill-directed: tests/test-tgkill-directed.c | $(BUILD_DIR)
+	@echo "  CROSS   $< (with -lpthread)"
+	$(Q)$(CROSS_COMPILE)gcc -D_GNU_SOURCE -static -O2 -o $@ $< -lpthread
+
 # test-osync-requeue drives a raw FUTEX_REQUEUE against a plain-FUTEX_WAIT
 # waiter (musl unlock_requeue pattern) to guard the os_sync wake-at-source
 # degradation; needs -lpthread.
--- a/src/runtime/forkipc.c
+++ b/src/runtime/forkipc.c
@@ -289,15 +289,13 @@ int fork_child_main(int ipc_fd,
     }
 
     /* POSIX: "Signals pending to the parent shall not be pending to the child."
-     * Clear pending bitmask and RT queue before applying state.
-     * signal_set_state() is deferred until after thread_register_main() so that
-     * current_thread is non-NULL and per-thread state (blocked mask, altstack)
-     * is properly restored.
+     * Clear the entire shared pending set before applying state. The child's
+     * single thread starts with an empty private set (thread_register_main
+     * zeroes the slot). signal_set_state() is deferred until after
+     * thread_register_main() so that current_thread is non-NULL and per-thread
+     * state (blocked mask, altstack) is properly restored.
      */
-    sig.pending = 0;
-    memset(sig.rt_queue, 0, sizeof(sig.rt_queue));
-    memset(sig.rt_head, 0, sizeof(sig.rt_head));
-    memset(sig.rt_info, 0, sizeof(sig.rt_info));
+    memset(&sig.shared, 0, sizeof(sig.shared));
 
     /* execve in the child needs the shim bytes after guest_reset clears memory.
      * Close IPC socket
--- a/src/runtime/thread.c
+++ b/src/runtime/thread.c
@@ -60,10 +60,14 @@ static _Atomic int active_thread_count = 0;
 #define THREAD_FOR_EACH(t) \
     for (thread_entry_t *t = thread_table; t < thread_table + MAX_THREADS; t++)
 
-/* Iterate active slots. Caller must hold thread_lock for stable reads. */
+/* Iterate active slots. Caller must hold thread_lock; the lock already orders
+ * these reads against the release-stores of active, so a relaxed atomic load
+ * suffices. It is used (rather than a plain read) only to keep every access to
+ * active uniformly atomic -- no mixed atomic-write / plain-read on one object.
+ */
 #define THREAD_FOR_EACH_ACTIVE(t) \
     THREAD_FOR_EACH (t)           \
-        if (t->active)
+        if (__atomic_load_n(&t->active, __ATOMIC_RELAXED))
 
 /* Iterate active slots without holding thread_lock. Uses an acquire load on the
  * active flag so the lock-free observers in thread_tid_alive() and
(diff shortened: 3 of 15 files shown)


Max042004:
tested on M4. `make check` all pass, `test-matrix` no regression.

PR #154, opened 2026-07-05
Report guest path for /proc/self/exe under Rosetta

### LINKED ISSUE #107: `readlink("/proc/self/exe")` and `readlink("/proc/self/fd/N")` return the Rosetta translator path instead of the guest binary path on x86_64 guests
### Environment

- **Host**: macOS on Apple Silicon (M-series)
- **Guest arch**: x86_64 (run via Rosetta translation)
- **Trigger**: Running any package management operation that invokes Debian/Ubuntu maintainer scripts (e.g., `apt upgrade`, `dpkg -i`)

### Problem Description

When an x86_64 guest binary is executed under Rosetta translation, `proc_intercept_readlink` has two separate code paths that both return the Rosetta interpreter path (`/Library/Apple/usr/libexec/oah/RosettaLinux/rosetta`, i.e. `ROSETTA_PATH`) instead of the actual path of the running guest binary. This breaks any guest application that uses `/proc/self/exe` or `/proc/self/fd/N` for self-identification.

#### Code path 1 — `readlink("/proc/self/exe")`

In `src/runtime/procemu.c`, `proc_intercept_readlink` had an early-exit branch at the top of the `/proc/self/exe` handler:

```c
if (!strcmp(path, "/proc/self/exe")) {
    /* Under rosetta, readlink("/proc/self/exe") points at the rosetta
     * translator (the binfmt_misc interpreter). Matches the behavior Linux
     * exposes when binfmt_misc dispatch is active.
     */
    if (proc_rosetta_active()) {
        size_t len = strlen(ROSETTA_PATH);
        if (len > bufsiz)
            len = bufsiz;
        memcpy(buf, ROSETTA_PATH, len);
        return (int) len;
    }
    /* ... normal path follows */
}
```

This was written to mimic Linux binfmt_misc behavior, where `readlink("/proc/self/exe")` in a qemu-static context would return the interpreter (qemu) rather than the guest binary. However, on Linux with native binfmt_misc + Rosetta, `/proc/self/exe` is supposed to resolve to the *real* binary, not the interpreter. The behavior that returns the interpreter was specific to older versions of the Rosetta Linux ABI and is not what most guest applications expect.

#### Code path 2 — `readlink("/proc/self/fd/N")` when N points to `/proc/self/exe`

In `proc_intercept_open`, opening `/proc/self/exe` under Rosetta opens `ROSETTA_PATH` on the host:

```c
if (!strcmp(path, "/proc/self/exe")) {
    if (g && g->is_rosetta)
        return open(ROSETTA_PATH, O_RDONLY);   /* <-- returns ROSETTA_PATH fd */
    /* ... */
}
```

This is intentional: the Rosetta VZ ioctl gate (`rosetta_ioctl_target_fd`) needs to recognize the fd it is handed as pointing to `ROSETTA_PATH`, so it accepts the ioctl. However, a side effect is that when the guest subsequently calls `readlink("/proc/self/fd/<N>")` on that descriptor, `fcntl(host_fd, F_GETPATH, fdpath)` correctly resolves the host fd to `ROSETTA_PATH`. The `/proc/self/fd/N` handler then returns that path verbatim:

```c
/* /proc/self/fd/N -> path of host fd (via fcntl F_GETPATH on macOS) */
if (!strncmp(path, "/proc/self/fd/", 14)) {
    /* ... parse N, get host_fd ... */
    char fdpath[MAXPATHLEN];
    if (fcntl(host_fd, F_GETPATH, fdpath) < 0) { ... }
    /* fdpath is now ROSETTA_PATH -- returned as-is */
    memcpy(buf, fdpath, len);
    return (int) len;
}
```

### Why this matters: GNU `coreutils` multi-call binary self-identification

Debian/Ubuntu ship a multi-call `coreutils` binary. Each applet (`rm`, `mv`, `sed`, `install`, etc.) is a hard-link to the same ELF. At startup, `coreutils` determines which applet to run by the following sequence:

1. Open `/proc/self/exe` → get fd `N`
2. Call `readlink("/proc/self/fd/N")` → expect something like `/usr/bin/rm`
3. Match the basename against the applet table

With both bugs above, step 2 returns `/Library/Apple/usr/libexec/oah/RosettaLinux/rosetta` instead. `coreutils` fails to match `rosetta` against any known applet and prints:

```
coreutils: unknown program 'rosetta'
```

and exits with status 1.

### Observed failure chain during `apt upgrade`

`apt upgrade` invokes `dpkg`, which runs maintainer scripts for every package being upgraded. The maintainer scripts call utilities such as:

- `dpkg-maintscript-helper` (a shell wrapper) which calls `rm`, `sed`, `install`, `dpkg-trigger`
- `base-files.preinst` which calls `dpkg-query`, `rm`, `ln`

All of these are hard-linked to the multi-call `coreutils` binary. Under the Rosetta-path bug, every one of them silently fails with `coreutils: unknown program 'rosetta'`, meaning no package can be configured or installed. `dpkg` reports them all as failing with exit status 1, and `apt upgrade` rolls back:

```
23:45:37 ERROR src/core/elf.c:42: /path/to/rootfs/usr/sbin/dpkg-preconfigure: not an ELF file
```

(The "not an ELF file" message appears because `dpkg-preconfigure` is itself a Perl script; when `coreutils` reports the wrong name, the outer shell tries to execute the `coreutils` binary thinking it is a new interpreter.)

jserv:
On x86_64 guests, readlink("/proc/self/exe") and
readlink("/proc/self/fd/N") (with N from opening /proc/self/exe) returned the Rosetta translator path (ROSETTA_PATH) instead of the running guest binary. GNU coreutils ships a multi-call binary whose applets self-identify by reading /proc/self/exe, opening it, then readlink-ing /proc/self/fd/N and matching the basename. Getting "rosetta" back made every applet abort with "unknown program 'rosetta'", which broke Debian/Ubuntu maintainer scripts under apt upgrade and dpkg -i.

Factor the /proc/self/exe resolution (ELF-path snapshot plus sysroot-prefix strip) into proc_readlink_self_exe. Drop the Rosetta early-return from the direct /proc/self/exe branch, and in the /proc/self/fd/N branch substitute the guest binary when the host fd resolves to ROSETTA_PATH.

proc_intercept_open("/proc/self/exe") still hands back an fd to ROSETTA_PATH so the VZ ioctl gate (rosetta_ioctl_target_fd) keeps recognising Rosetta's own self-identification; only the guest-visible readlink string changed.

Close #107

---
## Summary by cubic
Fixes wrong /proc/self/exe reporting under Rosetta. `readlink("/proc/self/exe")` and `readlink("/proc/self/fd/N")` now return the guest binary path (sysroot-stripped) so multi-call binaries no longer abort.

- **Bug Fixes**
  - Return the guest binary for `readlink("/proc/self/exe")` and `readlink("/proc/self/fd/N")` on x86_64 under Rosetta, with sysroot prefix stripped.
  - Keep `open("/proc/self/exe")` returning a host fd to `ROSETTA_PATH` so the VZ ioctl gate still recognizes Rosetta.
  - Add a regression test asserting `/proc/self/exe` resolves to the guest (basename `busybox`), not Rosetta.

- **Refactors**
  - Extract `proc_readlink_self_exe()` to resolve the ELF path and strip the sysroot; reuse in both readlink paths.
  - Remove the Rosetta early-return for `/proc/self/exe` and substitute the guest path when an fd resolves to `ROSETTA_PATH`.

<sup>Written for commit ebd6d26baa794078398a7532fbcc4b3bd0d3d39a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/154?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/runtime/procemu.c
+++ b/src/runtime/procemu.c
@@ -3667,6 +3667,50 @@ int proc_intercept_stat(const char *path, struct stat *st)
     return PROC_NOT_INTERCEPTED;
 }
 
+/* Resolve /proc/self/exe to the guest binary path in buf. Strips the sysroot
+ * prefix so a guest running under --sysroot=/opt/sr sees /bin/ls rather than
+ * /opt/sr/bin/ls, matching the chroot-like abstraction the rest of the path
+ * layer presents.
+ *
+ * Returns bytes written, or -1 with errno set. Under rosetta this still returns
+ * the real guest binary: the rosetta translator is an implementation detail,
+ * and guest apps (coreutils multi-call dispatch) expect the true path just as
+ * native binfmt_misc + rosetta exposes it.
+ */
+static int proc_readlink_self_exe(char *buf, size_t bufsiz)
+{
+    char exe_buf[LINUX_PATH_MAX];
+    if (!proc_elf_path_snapshot(exe_buf, sizeof(exe_buf))) {
+        errno = ENOENT;
+        return -1;
+    }
+    const char *exe = exe_buf;
+    char exe_real[LINUX_PATH_MAX];
+    char sysroot_snap[LINUX_PATH_MAX];
+    if (proc_sysroot_snapshot(sysroot_snap, sizeof(sysroot_snap))) {
+        /* proc_set_sysroot stores a realpath()-canonicalized form, so
+         * canonicalize exe before the prefix check or the strip fails when /var
+         * -> /private/var (and similar macOS symlinks) make the two strings
+         * diverge.
+         */
+        const char *exe_cmp = exe;
+        if (realpath(exe, exe_real))
+            exe_cmp = exe_real;
+        size_t sr_len = strlen(sysroot_snap);
+        if (sr_len > 0 && !strncmp(exe_cmp, sysroot_snap, sr_len) &&
+            (exe_cmp[sr_len] == '/' || exe_cmp[sr_len] == '\0')) {
+            exe = exe_cmp + sr_len;
+            if (*exe == '\0')
+                exe = "/";
+        }
+    }
+    size_t len = strlen(exe);
+    if (len > bufsiz)
+        len = bufsiz;
+    memcpy(buf, exe, len);
+    return (int) len;
+}
+
 int proc_intercept_readlink(const char *path, char *buf, size_t bufsiz)
 {
     {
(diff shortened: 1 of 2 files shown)


Max042004:
tested on M4. `make check` all pass, `test-matrix` no regression

PR #155, opened 2026-07-06
Fix getdents64 use-after-free on concurrent close

Max042004:
sys_getdents64 read fd_table[fd].dir with no lock and no lifetime pin,
then looped readdir()/telldir()/seekdir() on it. A sibling close() or
dup2()-overwrite of the same fd runs fd_cleanup_entry() -> closedir()
outside fd_lock and frees the stream while the read loop is still
using it -- the same use-after-free class PR #146 fixes for
FD_EPOLL's epoll_instance_t, left unaddressed for FD_DIR.

Wrap the DIR* in a refcounted dir_stream_t (fd-table's own reference
plus one per in-flight sys_getdents64, mirroring epoll_instance_t's
acquire/release), so a concurrent close/dup2/fork-restore only drops
its own reference instead of unconditionally freeing the stream.

Reproduced pre-fix under ASAN (double-free in readdir() racing
closedir() from a concurrent dup3) and confirmed clean post-fix with
tests/test-getdents-refcount.c, added to the unit suite.

---
## Summary by cubic
Fixes a getdents64 use-after-free and concurrent-walk races by refcounting and serializing directory streams. Closing, dup2/dup3, or fork-restore no longer frees a stream in use; failed restores now close the fd and return EBADF.

- **Bug Fixes**
  - Wrapped directory `DIR*` in refcounted `dir_stream_t`; `fd_table[].dir` stores the wrapper and frees on last ref.
  - `sys_getdents64` now pins the stream and holds a per-stream mutex across the readdir/telldir/seekdir walk.
  - Close/dup/fork paths updated: `fd_cleanup_entry` calls `dir_stream_release`; dup/aliasing clones the wrapper; on restore failure (dup/fdopendir/create), the child closes the slot instead of publishing `FD_DIR` with `dir=NULL`.
  - Added stress test `tests/test-getdents-refcount.c` (in `tests/manifest.txt`) that reproduced the race pre-fix and now passes.

<sup>Written for commit a900f6b634acd4c88d84785fa13793f4dd3c7350. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/155?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/runtime/fork-state.c
+++ b/src/runtime/fork-state.c
@@ -437,19 +437,40 @@ int fork_ipc_recv_fd_table(int ipc_fd, guest_t *g)
 
             if (fd_entries[i].type != FD_DIR)
                 continue;
+            /* Rebuild the directory stream. Every failure below must close
+             * the slot rather than leave it published as FD_DIR with
+             * dir=NULL: such a slot never faults, but sys_getdents64 would
+             * report ENOTDIR on it for the child's whole lifetime, masking
+             * the restore failure as a valid-but-broken fd. A closed slot
+             * yields EBADF, an honest signal that the fd did not survive
+             * the fork.
+             */
             int dir_fd = dup(host_fds[i]);
             if (dir_fd < 0) {
                 log_error("fork-child: dup failed for DIR gfd %d: %s", gfd,
                           strerror(errno));
+                close(host_fds[i]);
+                fd_mark_closed(gfd);
                 continue;
             }
             DIR *dir = fdopendir(dir_fd);
             if (!dir) {
                 close(dir_fd);
                 log_error("fork-child: fdopendir failed for gfd %d", gfd);
+                close(host_fds[i]);
+                fd_mark_closed(gfd);
                 continue;
             }
-            fd_table[gfd].dir = dir;
+            void *ds = dir_stream_create(dir);
+            if (!ds) {
+                closedir(dir);
+                log_error("fork-child: dir_stream_create failed for gfd %d",
+                          gfd);
+                close(host_fds[i]);
+                fd_mark_closed(gfd);
+                continue;
+            }
+            fd_table[gfd].dir = ds;
         }
     }
 
--- a/src/syscall/fs.c
+++ b/src/syscall/fs.c
@@ -215,6 +215,102 @@ static const char *proc_virtual_dir_path(const char *path,
     return virt;
 }
 
+/* Reference-counted wrapper around a directory stream. See the declaration in
+ * syscall/internal.h for why this exists: a raw DIR* stored in
+ * fd_table[fd].dir would let a sibling close()/dup2()/fork-restore free it via
+ * closedir() while sys_getdents64() is still mid-loop reading it. The struct
+ * itself is private to this file; every other module only ever sees the
+ * opaque void* that fd_table[].dir already stores.
+ */
+typedef struct {
+    DIR *dir;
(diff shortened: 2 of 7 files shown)


src/syscall/fs.c
```diff
@@ -1249,12 +1347,20 @@ int64_t sys_getdents64(guest_t *g, int fd, uint64_t buf_gva, uint64_t count)
     if (fd_table[fd].type == FD_PATH)
         return -LINUX_EBADF;
 
-    DIR *dir = (DIR *) fd_table[fd].dir;
-    if (!dir)
+    /* Pin the directory stream so a concurrent close()/dup2() cannot free it
+     * while this call is still walking it -- see dir_stream_t in fs.c.
+     */
+    dir_stream_t *ds = dir_stream_acquire(fd);
```
jserv, line 1367: dir_stream_acquire pins the stream's lifetime but does not serialize use of the shared DIR*. Two guest threads can pin the same wrapper and concurrently run readdir/telldir/seekdir on ds->dir in the walk below. macOS libc locks the DIR internally so this stays memory-safe after the fix, but the logical interleaving of two concurrent getdents64 on the same fd is undefined -- this matches the pre-fix raw-pointer behavior, so it is not a regression. If defined behavior is wanted, add a per-stream lock held across the telldir..seekdir walk.

src/runtime/fork-state.c
```diff
@@ -449,7 +449,14 @@ int fork_ipc_recv_fd_table(int ipc_fd, guest_t *g)
                 log_error("fork-child: fdopendir failed for gfd %d", gfd);
                 continue;
             }
-            fd_table[gfd].dir = dir;
+            void *ds = dir_stream_create(dir);
```
jserv, line 464: On dir_stream_create failure the child closedir()s and continues, leaving fd_table[gfd] published as FD_DIR with dir=NULL. It does not fault -- sys_getdents64 returns ENOTDIR via dir_stream_acquire -- but the inherited directory fd is then silently unusable for the child's lifetime. Consider marking the slot closed (or failing the fd-table restore) so the failure is not masked as a valid-but-broken FD_DIR.

src/syscall/fs.c
Max042004: Addressed in a900f6b: dir_stream_t now carries a per-stream mutex, taken in sys_getdents64 right after dir_stream_acquire() and held across the whole telldir/readdir/seekdir walk, so two concurrent getdents64 on the same fd are serialized (matching Linux's f_pos_lock semantics) instead of interleaving on the shared DIR*. The lock is only ever taken by reference holders, so the last-reference destroy path can never tear down a held mutex.

src/runtime/fork-state.c
Max042004: Addressed in a900f6b: all three restore-failure paths (dup, fdopendir, dir_stream_create) now close(host_fds[i]) and fd_mark_closed(gfd) instead of leaving the slot published as FD_DIR with dir=NULL, so the child sees EBADF on the fd rather than a silent ENOTDIR-forever slot. Chose slot-close over failing the whole fd-table restore since an OOM on one directory stream should not abort the fork.

jserv:

PR #159, opened 2026-07-06
Reap terminated worker pthreads on slot reuse

### LINKED ISSUE #157: Worker pthreads that exit before teardown are never joined, leaking thread handles under churn
Found by ThreadSanitizer (thread-leak report) while auditing PR #85's teardown:

```
WARNING: ThreadSanitizer: thread leak
  Thread T4 (tid=..., finished) created by main thread at:
    #0 pthread_create
    #1 sys_clone_thread forkipc.c:698
```

`sys_clone_thread()` creates worker pthreads joinable (the teardown join protocol needs that). A worker that exits on its own runs `thread_create_and_run` → `thread_deactivate` → returns; nothing joins it at that point. The only join sites are `thread_join_workers()` — whose snapshot takes **active** entries only, so an already-deactivated worker never enters it — and the clone startup-failure rollback. Slot reuse in `thread_register()` then memsets the entry, dropping `host_thread` without join or detach, so the terminated pthread's bookkeeping is retained by the pthread runtime forever.

Benign at process exit; real for long-running guests with thread churn (JVM/Go-style runtimes) — one leaked handle per reaped thread for the process lifetime. The vm-clone reap path (`thread_wait_vm_clone`) looks like the same pattern.

Suggested fix: `pthread_join` the stale handle at slot-reuse time in `thread_register` — `active == 0` means the pthread has already left its thread function, so the join blocks at most for the final instructions of wind-down, and slot reuse is not a hot path. The alternative (self-`pthread_detach` in `thread_deactivate`) conflicts with the teardown join: joining or detaching an already-detached pthread is undefined.

Verify: TSan build with `report_thread_leaks=1` (default) over a thread-churning test (e.g. `test-pthread`), expect zero thread-leak reports.

Line numbers from the PR #85 branch (40270cd).

Max042004:
Worker pthreads are created joinable so `thread_join_workers` can join them at teardown, but a worker that exits on its own is never joined — the teardown snapshot only sees active entries, and slot reuse in `thread_alloc` drops `host_thread` without join or detach. Each reaped thread under churn leaks one pthread handle (JVM/Go-style workloads); TSan reports 342 leaks on a 342-thread churn test.

Add a `host_thread_needs_join` flag, cleared by exactly one claimer under `thread_lock` before joining:
- `thread_alloc` joins the stale handle before reusing a slot (drops the lock during the join, since the dying worker's own wind-down can retake `thread_lock`).
- `thread_join_workers` claims at snapshot time so it can't double-join a concurrently-reused slot, and no longer joins/detaches handles it doesn't own (it previously touched detached vm-clone and main-thread handles too).
- The clone startup-failure rollback claims via `thread_claim_worker_join`, since a concurrent clone may have already reaped the slot.

Added `test-thread-churn` (150 sequential + 6×32 batched threads) to force slot reuse past the 64-entry table. TSan: 342 leaks → 0. Driver suite 107/107, churn test stable over 20 runs.

Closes #157

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,6 +196,11 @@ $(BUILD_DIR)/test-pthread: tests/test-pthread.c | $(BUILD_DIR)
 	@echo "  CROSS   $< (with -lpthread)"
 	$(Q)$(CROSS_COMPILE)gcc -D_GNU_SOURCE -static -O2 -o $@ $< -lpthread
 
+# test-thread-churn creates >64 threads to force thread-table slot reuse.
+$(BUILD_DIR)/test-thread-churn: tests/test-thread-churn.c | $(BUILD_DIR)
+	@echo "  CROSS   $< (with -lpthread)"
+	$(Q)$(CROSS_COMPILE)gcc -D_GNU_SOURCE -static -O2 -o $@ $< -lpthread
+
 # test-poll uses pthread_kill to verify blocked read signal delivery.
 $(BUILD_DIR)/test-poll: tests/test-poll.c | $(BUILD_DIR)
 	@echo "  CROSS   $< (with -lpthread)"
--- a/src/runtime/forkipc.c
+++ b/src/runtime/forkipc.c
@@ -642,6 +642,13 @@ static int64_t sys_clone_thread(hv_vcpu_t parent_vcpu,
         log_error("clone_thread: thread table full");
         return -LINUX_EAGAIN;
     }
+    /* Captured now, while the slot is guaranteed still ours (thread_alloc just
+     * marked it active, so no concurrent thread_alloc reuse-scan will touch
+     * it until our own worker deactivates it). Passed to
+     * thread_set_host_thread below so it can detect whether this exact slot
+     * got recycled to a different logical thread before that call runs.
+     */
+    uint64_t t_generation = t->generation;
 
     /* Inherit parent's signal mask (POSIX: clone inherits blocked mask) */
     if (current_thread)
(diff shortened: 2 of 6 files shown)


jserv (changes requested):
Rebase latest `main`, resolve conflicts, and refine per review messages.

Max042004:
Rebased onto latest `main` (through PR #160) and addressed both cubic findings:

1. **forkipc.c:716** (write to a possibly-recycled slot) — a worker that fails HVF bring-up can deactivate its own slot before the parent's `thread_set_host_thread` write runs, letting a concurrent clone recycle the slot first. Added a `generation` counter to `thread_entry_t`, bumped by `thread_alloc` on every reuse. The clone caller captures `t->generation` right after `thread_alloc` returns, and `thread_set_host_thread` now takes that value and rejects the write on mismatch. `sys_clone_thread`'s failure branch checks whether the write was recorded: if not, no table entry references the handle, so the parent joins it directly instead of going through the claim path.

2. **thread.c:410** (poll loop reading a recycled slot's `active` bit) — `thread_join_workers` now also snapshots each worker's generation, and the poll loop breaks out as soon as it detects a generation bump (proof the original occupant already deactivated) instead of continuing to track the replacement thread's `active` bit, then joins unconditionally rather than risking a wrong detach.

Verified with TSan over `test-thread-churn`: still 0 thread leaks, no new races introduced (the only remaining warning is the pre-existing, out-of-scope `t->vcpu` store race). Driver suite 114/114.

jserv:

PR #160, opened 2026-07-06
Order PTE publication against lock-free GVA walkers

### LINKED ISSUE #156: Data race: split_l2_block publishes L2 entry without release ordering against lock-free gva walkers
Found by ThreadSanitizer while auditing PR #85's teardown. elfuse builds and runs under TSan with Hypervisor.framework:

```sh
make elfuse BUILD_DIR=build-tsan CFLAGS="-O1 -g -fsanitize=thread -Wall -Wno-unused-parameter"
TSAN_OPTIONS="abort_on_error=0" ./build-tsan/elfuse build/test-exit-group-worker
```

```
WARNING: ThreadSanitizer: data race
  Write of size 8 by main thread (mutexes: write M0 = mmap lock):
    #0 split_l2_block guest.c:2927
    #1 guest_split_block
    #2 guest_invalidate_ptes
    #3 sys_mmap
  Previous read of size 8 by thread T7:
    #0 gva_translate_perm guest.c:1235
    #1 gva_resolve_perm
    #2 guest_ptr_bound
```

`split_l2_block()` fills 512 L3 descriptors with plain stores, then flips the L2 entry to point at the new table — all under the mmap lock. `gva_translate_perm()` walks L1→L2→L3 with plain loads, lock-free, on every syscall's guest-pointer translation, from every vCPU thread. The reader never takes the mmap lock, so the lock provides no happens-before edge to it.

Beyond the formal race, the fill-then-flip publication is only correct if the L3 stores are visible before the new L2 value. Both are plain stores, so neither the compiler nor the weakly-ordered AArch64 CPU is required to preserve that order: a concurrent walker can chase the freshly published table pointer into not-yet-visible L3 descriptors and hand a syscall handler a wrong host pointer — silent memory corruption. (The guest MMU's own walk of these tables is a separate concern already handled by TLBI/DSB in the shim; this race is host thread vs host thread.)

Suggested fix: publish table pointers with `__atomic_store_n(..., __ATOMIC_RELEASE)` and load L1/L2/L3 entries in the host-side walkers with `__ATOMIC_ACQUIRE` (`ldar` is negligible next to a syscall's total cost). The 512 L3 fills can stay plain — the release on the flip orders them. Verify by re-running the TSan build with no `race:split_l2_block` suppression.

Line numbers from the PR #85 branch (40270cd); the functions are unchanged on main (f8a06d4).

Max042004:
gva_translate_perm() walks L0->L3 with plain loads, lock-free, on every
syscall's guest-pointer translation, while the page-table mutators
(split_l2_block, guest_update_perms, guest_invalidate_ptes, ...) rewrite
descriptors with plain stores under the mmap lock. The lock serializes
mutators against each other only, so nothing orders a mutation against a
concurrent walker. In particular, split_l2_block fills 512 L3
descriptors and then flips the L2 entry to point at the new table with
two plain stores; neither the compiler nor the weakly-ordered AArch64
CPU has to keep the fills visible before the flip, so a walker can chase
the freshly published table pointer into not-yet-visible descriptors and
hand a syscall handler a wrong host pointer -- silent memory corruption.

This publishes every runtime descriptor write with a release store
(pte_store_release) and makes the walker load each level exactly once
with an acquire load (pte_load_acquire), decoding that single snapshot
instead of re-reading the entry per field (the re-reads were a second
latent hazard: a concurrent rewrite between the PT_VALID test and the
address extraction could mix two generations of the same entry).

Two spots intentionally keep plain stores: split_l2_block's 512 L3 fills
(the table page is unreachable until the release-flip publishes it, and
the pt pool's bump allocator never recycles pages while vCPU threads
run) and boot-time construction in guest_build_page_tables /
guest_init_kbuf (no vCPU thread exists yet; pthread_create provides the
happens-before edge).

Verification, ThreadSanitizer build (`make elfuse BUILD_DIR=build-tsan
CFLAGS="-O1 -g -fsanitize=thread -Wall -Wno-unused-parameter"`):

- Unpatched main (f8a06d4) reports the race on the **first** run of
  test-mprotect-mt (guest_update_perms guest.c:3163 store vs
  gva_translate_perm guest.c:1243 load). test-mprotect-mt's scenario 2
  (mprotect toggler vs readers in read(/dev/urandom, buf)) turns out to
  be a far more reliable reproducer than the test-exit-group-worker run
  cited in the issue (0 hits in 40 runs there).
- With this change: 0 data races across 20 runs of test-mprotect-mt and
  15 runs of test-exit-group-worker. The one remaining TSan warning is
  the pthread handle thread-leak tracked separately as #157.
- make check passes; the hot-syscall bench guardrail stays within its
  ceilings (read-urandom 53.3 ns/op vs 400 ceiling) -- the extra ldar
  per level is noise next to a syscall's total cost.

Closes #156

---
## Summary by cubic
Fixes a race between lock-free GVA walks and PTE mutations by making runtime PTE writes release-ordered and walker reads acquire-ordered. Prevents walkers from chasing half-published tables and returning wrong host pointers.

- **Bug Fixes**
  - Added `pte_load_acquire` and `pte_store_release` for ordered PTE access.
  - `gva_translate_perm` now loads each level once with acquire and decodes from that snapshot.
  - Switched runtime PTE publications to release stores across map, extend, invalidate, perms, and install paths; kept plain stores only for unreachable L3 fills in `split_l2_block` and single-threaded boot-time builds.
  - Verification: TSan reports 0 data races across repeated runs; tests pass; no measurable syscall overhead.

<sup>Written for commit 513eb321aa8f30d27d15c980a1e792ef199fe5f8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/160?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/core/guest.c
+++ b/src/core/guest.c
@@ -213,6 +213,45 @@ static uint64_t *pt_at(const guest_t *g, uint64_t gpa)
     return (uint64_t *) ((uint8_t *) g->host_base + gpa);
 }
 
+/* Host-side PTE accessors.
+ *
+ * gva_translate_perm() walks L0->L3 lock-free on every guest-pointer
+ * translation, from every vCPU thread, while the runtime mutators
+ * (guest_update_perms, guest_invalidate_ptes, split_l2_block, ...) rewrite
+ * descriptors under the caller's mmap lock. The lock serializes mutator
+ * against mutator only -- it provides no happens-before edge to the lock-free
+ * walker, so the descriptor accesses themselves must carry the ordering:
+ *
+ *   - pte_store_release publishes a descriptor so that everything written
+ *     before it -- most critically the 512 L3 fills of split_l2_block and the
+ *     entries of a freshly allocated L2 table -- is visible to a walker that
+ *     chases the new table pointer. Plain stores would let both the compiler
+ *     and the weakly-ordered CPU sink the fills past the flip, handing a
+ *     concurrent walker a pointer into not-yet-initialized descriptors.
+ *   - pte_load_acquire pairs with the release and doubles as the walker's
+ *     single-snapshot read: each level is loaded exactly once and decoded
+ *     from that one value, never re-read per field, so a concurrent rewrite
+ *     cannot make the walker mix two generations of the same entry.
+ *
+ * Table pages come from a bump allocator (pt_alloc_page) that is only reset
+ * on execve while all vCPU threads are quiesced, so a not-yet-published table
+ * can never be reached through a stale pointer; filling it with plain stores
+ * before the release-flip is sound. Boot-time construction also keeps plain
+ * stores -- guest_build_page_tables, guest_init_kbuf, and the rosetta kbuf
+ * user-alias install (guest_install_kbuf_user_alias /
+ * populate_kbuf_l2_blocks) all run single-threaded before any vCPU thread
+ * exists, and pthread_create provides the happens-before edge.
+ */
+static inline uint64_t pte_load_acquire(const uint64_t *entry)
+{
+    return __atomic_load_n(entry, __ATOMIC_ACQUIRE);
+}
+
+static inline void pte_store_release(uint64_t *entry, uint64_t desc)
+{
+    __atomic_store_n(entry, desc, __ATOMIC_RELEASE);
+}
+
 /* Public API */
 
 /* FEAT_TLBIRANGE probe -- runs exactly once via pthread_once. ARMv8.4
@@ -927,6 +966,11 @@ uint64_t guest_overflow_alloc(guest_t *g)
  * RW + UXN + PXN: the kbuf must stay data-only under both the kernel TTBR1
  * mirror and the user-VA TTBR0 alias to preserve the aliasing-proof W^X
  * invariant.
+ *
+ * Plain stores: both callers (guest_init_kbuf at boot,
+ * guest_install_kbuf_user_alias during rosetta bootstrap) run single-threaded
+ * before any vCPU thread exists -- see the boot-time exemption note at
+ * pte_store_release.
  */
 static void populate_kbuf_l2_blocks(uint64_t *l2, uint64_t kbuf_gpa)
 {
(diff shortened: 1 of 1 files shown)


src/core/guest.c
```diff
@@ -932,8 +970,9 @@ static void populate_kbuf_l2_blocks(uint64_t *l2, uint64_t kbuf_gpa)
 {
     for (int i = 384; i < 512; i++) {
         uint64_t ipa = kbuf_gpa + (uint64_t) (i - 384) * BLOCK_2MIB;
-        l2[i] = ipa | PT_AF | PT_SH_ISH | PT_ATTR1 | PT_AP_RW_EL0 | PT_UXN |
-                PT_PXN | PT_BLOCK;
+        pte_store_release(&l2[i], ipa | PT_AF | PT_SH_ISH | PT_ATTR1 |
```
jserv: populate_kbuf_l2_blocks now publishes with pte_store_release, but guest_install_kbuf_user_alias still writes the TTBR0 alias table pointers l0[511] and l1[511] with plain stores (lines 1172 and 1183). Both run only during single-threaded bootstrap via rosetta_finalize, so neither is a correctness hazard, yet mixing the two conventions inside the same kbuf-alias path is confusing, and this boot-only ttbr0-tree writer is not named in the header comment's exemption list. Either drop the release here to match the sibling writers (it buys nothing pre-vCPU), or switch 1172/1183 to pte_store_release and add guest_install_kbuf_user_alias to the boot-time exemption comment.

jserv (commented):
Reviewed the memory-ordering change in src/core/guest.c. The release/acquire pairing is sound: gva_translate_perm loads each level exactly once through pte_load_acquire and decodes that single snapshot, and every mutator store reachable by the lock-free walker (guest_map_va_range, guest_extend_page_tables, the split_l2_block flip, guest_invalidate_ptes, guest_update_perms, guest_install_va_pages) publishes with pte_store_release. The two plain-store exemptions are genuinely safe: the split_l2_block L3 fills go into a page that stays unreachable until the release-flip publishes the table pointer, and the boot-time constructors run before any vCPU thread exists. No missed live publication found. One consistency note inline.

src/core/guest.c
Max042004: Took the first option in 513eb32: populate_kbuf_l2_blocks is back to plain stores

jserv:

PR #161, opened 2026-07-06
Wake internal condvar parks on exit_group teardown

### LINKED ISSUE #158: exit_group teardown cannot reach workers parked on internal condvars (fork_cond, ptrace_cond)
Analysis finding from PR #85's correctness audit — a teardown-reachability gap, not a data race (no TSan report).

The exit_group teardown (exit-group flag + wakeup pipe + `hv_vcpus_exit` + the bounded futex/io flag re-checks) wakes every worker parked in a guest syscall wait within a bounded time. Two host-side condvar parks sit outside that set:

- `fork_cond` (`thread.c`): threads waiting out the fork barrier while a clone is in flight;
- `ptrace_cond` (`thread.c`): a tracer blocked waiting for its tracee to reach a ptrace-stop.

Neither is signalled by the teardown path, and their wait loops do not re-check `proc_exit_group_requested()`. A worker parked there sleeps past `thread_join_workers`' cap, gets `pthread_detach()`ed, and is still live when `guest_destroy()` unmaps the slab. The invariant "no worker touches guest memory after the unmap" then rests only on process exit following `guest_destroy()` almost immediately — a narrow-window mitigation, not a guarantee. It is the same trade-off the code already accepts for cross-thread `hv_vcpu_destroy` of abandoned workers in `thread_destroy_all_vcpus`.

Triggering requires exit_group to race an in-flight fork barrier or a ptrace stop, so it is rare; filed so the residual is tracked explicitly rather than left implicit in PR #85's argument.

Suggested fix: after setting the exit-group flag, broadcast `fork_cond` and the per-thread ptrace/resume condvars, and make those wait loops re-check `proc_exit_group_requested()` on wake — bringing both states under the same bounded-wake argument as the syscall waits.

Max042004:
Fixes #158.

exit_group teardown wakes every worker parked in a guest syscall wait, but missed three host-side condvar parks: siblings on the fork barrier (`fork_cond`), a tracer waiting for a ptrace-stop (`ptrace_cond`), and a tracee held in ptrace-stop (`resume_cond`). None were signalled by teardown, so a parked worker could outlive `thread_join_workers`' cap and still be live when `guest_destroy` unmapped guest memory — or, for a tracer waiting on a tracee that never stops, deadlock the whole process (only VM-clone exits broadcast `ptrace_cond`).

**Fix:**
- New `thread_wake_exit_waiters()`, called from both teardown entry points (`sc_exit_group`, `guest_destroy`) after the exit-group flag is set: broadcasts `fork_cond` and every live `ptrace_cond`/`resume_cond`.
- The three wait loops, plus `sys_wait4`'s blocking loop, re-check `proc_exit_group_requested()` on wake instead of looping forever.
- `thread_join_workers` now skips the main thread's slot — it never deactivates until `guest_destroy` runs, so waiting on it from a worker-initiated exit_group just burned the join cap on both sides and caused a `pthread_detach`/UAF race. This was an existing ~10% SIGSEGV on main during exit_group-vs-fork races, found while stress-testing this fix.

**Coverage:** new `tests/test-exit-group-teardown` (wait/stop/fork scenarios, `expected_rc=42`). The `wait` scenario deadlocks on unpatched main; `fork` crashes ~10% of runs. `make check` 109/109; 150 stress runs of all three scenarios, 0 failures.

### DIFF
--- a/src/runtime/thread.c
+++ b/src/runtime/thread.c
@@ -22,8 +22,9 @@
 
 #include "runtime/thread.h"
 #include "debug/log.h"
-#include "core/guest.h" /* guest_t (shim_data_base/ipa_base), BLOCK_2MIB */
-#include "hvutil.h"     /* vcpu_get_gpr, vcpu_get_sysreg */
+#include "core/guest.h"   /* guest_t (shim_data_base/ipa_base), BLOCK_2MIB */
+#include "hvutil.h"       /* vcpu_get_gpr, vcpu_get_sysreg */
+#include "syscall/proc.h" /* proc_exit_group_requested */
 
 /* From syscall/signal.h, included here directly to avoid pulling in the full
  * signal header (macOS defines sa_handler as a macro that conflicts with the
@@ -474,6 +475,16 @@ void thread_join_workers(void)
     THREAD_FOR_EACH (t) {
         if (t == current_thread || t->join_abandoned)
             continue;
+        /* Never wait for the main thread (slot 0): its entry only deactivates
+         * inside guest_destroy, which runs on the main thread AFTER its own
+         * thread_join_workers. A worker waiting here (exit_group called from
+         * a worker) would burn the full cap on it, and the resulting mutual
+         * wait (worker polling main while main polls the worker) ends in
+         * pthread_detach on both sides, leaving the loser to touch guest
+         * memory after guest_destroy unmaps it.
+         */
+        if (t == &thread_table[0])
+            continue;
         /* Inactive slots are included when they still hold an unjoined handle:
          * a worker that exited on its own shortly before teardown and whose
          * slot was never reused. Its pthread has terminated (or is in final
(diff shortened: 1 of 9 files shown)


src/syscall/syscall.c
```diff
@@ -805,6 +805,11 @@ static int64_t sc_exit_group(guest_t *g,
     proc_request_exit_group((int) x0);
     wakeup_pipe_signal();
     thread_for_each(thread_force_exit_cb, NULL);
+    /* Workers parked on internal condvars (fork barrier, ptrace stop/wait)
+     * see neither the pipe nor the vCPU kick; broadcast so they re-check the
+     * exit-group flag and terminate inside the join window below.
+     */
+    thread_wake_exit_waiters();
```
jserv, line 917: A third exit_group trigger was not given this wake: forkipc.c:1182 (last VM-clone child exiting) calls proc_request_exit_group() + thread_interrupt_all() but not thread_wake_exit_waiters(). thread_interrupt_all only kicks threads inside hv_vcpu_run, and that path broadcasts only the exiting child's ptrace_cond. A peer parked on fork_cond, or on a different slot's ptrace_cond/resume_cond, is never woken and never re-checks the flag - the same hang class this PR fixes, still open on that path. Add thread_wake_exit_waiters() right after proc_request_exit_group(exit_code) at forkipc.c:1182.

src/syscall/proc.c
```diff
@@ -714,6 +714,16 @@ int64_t sys_wait4(guest_t *g,
                 return 0;
             }
 
+            /* exit_group teardown: stop re-arming the wait. The 100ms quantum
```
jserv, line 1110: This escape was added only to the pid == -1 poll loop. The specific-PID branch below (around line 765) issues a single blocking wait4(host_pid, ..., mac_options) with no WNOHANG and no flag re-check, so a guest blocked in wait4(pid>0) during teardown is unreachable by the wakeup pipe, futex interrupt, hv_vcpus_exit, and the condvar broadcast - it hangs until that specific child dies, and if the main thread is blocked there it never reaches guest_destroy. Convert it to the same WNOHANG + pthread_cond_timedwait(&pid_cond) + proc_exit_group_requested() poll as the pid == -1 branch.

src/runtime/thread.c
```diff
@@ -875,6 +923,16 @@ int64_t thread_ptrace_wait(int64_t tracer_tid,
     pthread_mutex_lock(&thread_lock);
 
     for (;;) {
+        /* exit_group teardown: the stop/exit notifications that signal
+         * ptrace_cond stop coming once workers are being torn down. Return 0
```
jserv: Garbled sentence: "the stop/exit notifications that signal ptrace_cond stop coming once workers are being torn down" is missing a verb. Reword, e.g. "the stop/exit notifications that would signal ptrace_cond stop arriving once workers are being torn down."

jserv (changes requested):

PR #162, opened 2026-07-07
Implement times(2) syscall

Max042004:
Guests calling times(2) got -ENOSYS: the syscall was absent from
dispatch.tbl and abi.h had no SYS_times entry. Shell timing builtins and
build tools that poll times() for elapsed/CPU tick accounting fall over
on the error return.

sys_times fills the four struct tms fields from host getrusage:
RUSAGE_SELF supplies tms_utime/tms_stime and RUSAGE_CHILDREN supplies
tms_cutime/tms_cstime. Guest fork() maps 1:1 to a host fork and
sys_wait4 reaps through host waitpid, so the host's children accounting
covers exactly the guest's waited-for children. The return value is host
CLOCK_MONOTONIC converted to USER_HZ (100) ticks: Linux returns
jiffies-since-boot, and POSIX only requires an arbitrary-but-fixed
reference point that successive calls can difference. Tick conversion
uses the same 100Hz that core/stack.c advertises via AT_CLKTCK, so
libc's sysconf(_SC_CLK_TCK) scaling stays consistent. A NULL buffer is
accepted as on Linux, returning only the tick count.

## Testing

- New `test-times` (registered in manifest.txt): tick rate via
  sysconf(_SC_CLK_TCK), basic call, NULL buffer through the raw syscall,
  return value advancing across a 60ms sleep, self utime+stime growth
  against a measured 100ms CPU burn, cutime/cstime growth after reaping
  a child that burns 100ms CPU, and EFAULT on a bad pointer — 7/7.
- `make check` green (driver 81 passed / 0 failed / 3 skipped;
  skips are pre-existing fixture-absence).
- `check-syscall-coverage` passes.

---
## Summary by cubic
Add the times(2) syscall so guests get correct CPU accounting and an elapsed tick count instead of -ENOSYS. Restores shell timing and tools that poll times(), and excludes emulator helper subprocess CPU from child totals.

- **New Features**
  - Added `SYS_times` to the ABI/dispatch; wired `sc_times`.
  - Implemented `sys_times`: self `utime/stime` from `getrusage(RUSAGE_SELF)`; child `cutime/cstime` from proc accumulators updated at each guest-child reap (`sys_wait4`, `sys_waitid`, table reaper, `CLONE_VFORK`). Credit only on terminal exit/signal; ignore `WUNTRACED`, `WCONTINUED`, and `WNOWAIT` peeks; count only proc-table children, not helpers.
  - Returns `CLOCK_MONOTONIC` in 100 Hz ticks (matches `AT_CLKTCK`); accepts NULL; `-EFAULT` on bad pointer.
  - Added `test-times`: verifies `_SC_CLK_TCK==100`, tick advance, self and child CPU accumulation (including `cstime`), no credit on `waitid(WNOWAIT)`, NULL buffer, and bad-pointer `-EFAULT`.

<sup>Written for commit 7cd2c0e77357b6c2aabd28162aea84e3904f0f9d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/162?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/proc.c
+++ b/src/syscall/proc.c
@@ -81,6 +81,35 @@ static pthread_mutex_t pid_lock = PTHREAD_MUTEX_INITIALIZER; /* Lock order: 6 */
 static pthread_cond_t pid_cond =
     PTHREAD_COND_INITIALIZER; /* Signaled on child exit */
 
+/* CPU time of reaped guest children, accumulated at every host reap site and
+ * reported by times(2) as tms_cutime/tms_cstime. The emulator also waits on
+ * helper subprocesses (rosettad translate, sysroot tooling) whose CPU shows up
+ * in the host's RUSAGE_CHILDREN, so times() cannot read that aggregate; only
+ * reaps of proc_table children may land here. Relaxed atomics suffice: the
+ * counters are monotonic sums and times() tolerates reading utime/stime one
+ * reap apart.
+ */
+static _Atomic uint64_t children_utime_us;
+static _Atomic uint64_t children_stime_us;
+
+void proc_children_cpu_add(const struct rusage *ru)
+{
+    atomic_fetch_add_explicit(&children_utime_us,
+                              (uint64_t) ru->ru_utime.tv_sec * 1000000 +
+                                  (uint64_t) ru->ru_utime.tv_usec,
+                              memory_order_relaxed);
+    atomic_fetch_add_explicit(&children_stime_us,
+                              (uint64_t) ru->ru_stime.tv_sec * 1000000 +
+                                  (uint64_t) ru->ru_stime.tv_usec,
+                              memory_order_relaxed);
+}
+
+void proc_children_cpu_us(uint64_t *utime_us, uint64_t *stime_us)
+{
+    *utime_us = atomic_load_explicit(&children_utime_us, memory_order_relaxed);
+    *stime_us = atomic_load_explicit(&children_stime_us, memory_order_relaxed);
+}
+
 /* Global flag for exit_group: signals all threads to terminate. Atomic to avoid
  * undefined behavior under C11 memory model when multiple threads read/write
  * concurrently.
@@ -283,9 +312,11 @@ static int proc_reap_finished(void)
             continue;
         }
         int status;
-        pid_t ret = waitpid(proc_table[i].host_pid, &status, WNOHANG);
+        struct rusage ru;
+        pid_t ret = wait4(proc_table[i].host_pid, &status, WNOHANG, &ru);
         if (ret > 0) {
             /* Child exited; free the slot */
+            proc_children_cpu_add(&ru);
             proc_table[i].active = false;
             reaped++;
         }
@@ -1073,9 +1104,17 @@ int64_t sys_wait4(guest_t *g,
 
                 int status;
                 struct rusage ru;
-                pid_t ret = wait4(host_pid, &status, mac_options | WNOHANG,
-                                  rusage_gva ? &ru : NULL);
+                pid_t ret =
+                    wait4(host_pid, &status, mac_options | WNOHANG, &ru);
                 if (ret > 0) {
+                    /* Credit CPU only on a terminal report. mac_options may
+                     * carry WUNTRACED/WCONTINUED, and a stop/continue report
+                     * is a snapshot of a still-running child: crediting it
+                     * here would double- or triple-count the same child
+                     * across its stop, continue, and final exit reports.
+                     */
+                    if (WIFEXITED(status) || WIFSIGNALED(status))
+                        proc_children_cpu_add(&ru);
                     if (status_gva) {
                         int32_t linux_status = status;
                         if (guest_write_small(g, status_gva, &linux_status,
@@ -1146,8 +1185,7 @@ int64_t sys_wait4(guest_t *g,
             struct rusage ru;
             pid_t ret;
             if (mac_options & WNOHANG) {
-                ret = wait4(host_pid, &status, mac_options,
-                            rusage_gva ? &ru : NULL);
+                ret = wait4(host_pid, &status, mac_options, &ru);
             } else {
                 /* A bare blocking wait4 has no re-check point: a worker
                  * parked here past exit_group is invisible to
@@ -1159,8 +1197,7 @@ int64_t sys_wait4(guest_t *g,
                  * pid_cond on every host child exit.
                  */
                 for (;;) {
-                    ret = wait4(host_pid, &status, mac_options | WNOHANG,
-                                rusage_gva ? &ru : NULL);
+                    ret = wait4(host_pid, &status, mac_options | WNOHANG, &ru);
                     if (ret != 0)
                         break;
                     if (proc_exit_group_requested())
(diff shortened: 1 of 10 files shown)


src/syscall/time.c
```diff
@@ -459,6 +460,62 @@ int64_t sys_gettimeofday(guest_t *g, uint64_t tv_gva, uint64_t tz_gva)
...
+{
+    /* Linux permits a NULL buffer: only the return value (elapsed ticks) is
+     * wanted. Guest fork() maps 1:1 to a host fork and sys_wait4 reaps via
+     * host waitpid, so the host's RUSAGE_CHILDREN accounting covers exactly
+     * the guest's waited-for children.
+     */
+    if (buf_gva) {
+        struct rusage self, children;
+        if (getrusage(RUSAGE_SELF, &self) < 0 ||
+            getrusage(RUSAGE_CHILDREN, &children) < 0)
```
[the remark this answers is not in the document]
Max042004: Fixed in ac520fa. Valid finding: the emulator waitpid()s helper subprocesses (rosettad translate, sysroot tooling), so raw RUSAGE_CHILDREN would report their CPU as guest child time. times() now reads a proc-layer accumulator fed only at guest-child reap sites (sys_wait4, sys_waitid, the table-pressure reaper, and the CLONE_VFORK wait), each of which collects the child's wait4 rusage. Helper reaps never touch the accumulator.

src/syscall/proc.c
```diff
@@ -1049,9 +1080,10 @@ int64_t sys_wait4(guest_t *g,
 
                 int status;
                 struct rusage ru;
-                pid_t ret = wait4(host_pid, &status, mac_options | WNOHANG,
-                                  rusage_gva ? &ru : NULL);
+                pid_t ret =
+                    wait4(host_pid, &status, mac_options | WNOHANG, &ru);
                 if (ret > 0) {
+                    proc_children_cpu_add(&ru);
```
jserv: proc_children_cpu_add runs on every ret > 0, but sys_wait4 passes mac_options built from the guest options, which can carry WUNTRACED/WCONTINUED. A stopped or continued child returns ret > 0 with WIFSTOPPED/WIFCONTINUED and a snapshot of a still-running child, so its CPU lands in cutime/cstime before it exits and again when it is finally reaped -- one child, up to three successful wait4 calls, counted up to three times. Linux only credits child CPU on the terminal reap. Gate the add: if (WIFEXITED(status) || WIFSIGNALED(status)) proc_children_cpu_add(&ru). The WNOHANG-only site and the vfork/reaper sites are unaffected since they only observe terminations.

src/syscall/proc.c
```diff
@@ -1120,9 +1152,9 @@ int64_t sys_wait4(guest_t *g,
 
             int status;
             struct rusage ru;
-            pid_t ret =
-                wait4(host_pid, &status, mac_options, rusage_gva ? &ru : NULL);
+            pid_t ret = wait4(host_pid, &status, mac_options, &ru);
             if (ret > 0) {
+                proc_children_cpu_add(&ru);
```
jserv: Same gate needed here as the branch above: only add on WIFEXITED(status) || WIFSIGNALED(status), otherwise a stop/continue report inflates cutime/cstime.

src/syscall/proc.c
```diff
@@ -1246,7 +1278,10 @@ int64_t sys_waitid(guest_t *g,
             } else {
                 pid_t host_pid = proc_table[i].host_pid;
                 pthread_mutex_unlock(&pid_lock);
-                ret = waitpid(host_pid, &status, WNOHANG);
+                struct rusage ru;
+                ret = wait4(host_pid, &status, WNOHANG, &ru);
+                if (ret > 0)
+                    proc_children_cpu_add(&ru);
```
jserv, line 1350: waitid credits child CPU here before the WNOWAIT check at line 1335. Two problems: apply the same terminal-status gate (WIFEXITED/WIFSIGNALED); and when the guest passes WNOWAIT the child must stay waitable with its CPU uncounted until the later consuming reap, but the inner wait4 uses WNOHANG only, so the host zombie is consumed and the credit lands now. Move the add to run only when !(options & LINUX_WNOWAIT) and the status is terminal.

src/syscall/time.c
```diff
@@ -459,6 +460,65 @@ int64_t sys_gettimeofday(guest_t *g, uint64_t tv_gva, uint64_t tz_gva)
...
+    if (buf_gva) {
+        struct rusage self;
+        if (getrusage(RUSAGE_SELF, &self) < 0)
+            return linux_errno();
+
+        uint64_t cutime_us, cstime_us;
+        proc_children_cpu_us(&cutime_us, &cstime_us);
+
+        linux_tms_t tms = {
+            .tms_utime = timeval_to_clock_ticks(&self.ru_utime),
```
jserv, line 518: getrusage(RUSAGE_SELF) aggregates the whole emulator process -- every host thread plus page-table, syscall-servicing, and Rosetta overhead -- so tms_utime/tms_stime over-report relative to a real guest task. Likely unavoidable in this design, but the comment should note the caveat rather than presenting these as the guest's exact CPU times.

src/syscall/time.c
```diff
@@ -459,6 +460,65 @@ int64_t sys_gettimeofday(guest_t *g, uint64_t tv_gva, uint64_t tz_gva)
...
+            .tms_stime = timeval_to_clock_ticks(&self.ru_stime),
+            .tms_cutime = (int64_t) (cutime_us / (1000000 / LINUX_USER_HZ)),
```
jserv: cutime/cstime divide inline by (1000000 / LINUX_USER_HZ) while tms_utime/tms_stime go through timeval_to_clock_ticks. Same value today; a small usec_to_clock_ticks helper keeps both paths in step if LINUX_USER_HZ ever changes. The tv_nsec / (NSEC_PER_SEC / LINUX_USER_HZ) return expression has the same fragility.

tests/test-times.c
```diff
@@ -0,0 +1,116 @@
...
+        struct tms t2;
+        if (times(&t2) == (clock_t) -1)
+            FAIL("times failed");
+        else
+            EXPECT_TRUE(
+                t2.tms_utime + t2.tms_stime >= t1.tms_utime + t1.tms_stime + 5,
+                "utime+stime did not grow");
+    }
+
+    TEST("waited-for child CPU accumulates");
```
jserv, line 108: No regression covers the stop/continue or WNOWAIT double-count. Spawn a child, SIGSTOP it, reap with WUNTRACED, and assert cutime does not advance on the stop report; then let it exit and assert it is counted exactly once. Add a WNOWAIT peek that must not advance cutime until the consuming wait. Also assert cstime on its own -- the current checks only test the utime+cstime sum, so a regression zeroing cstime would pass.

jserv (changes requested):

PR #167, opened 2026-07-07
Silence not-an-ELF log on shebang script exec

### LINKED ISSUE #166: Noisy not an ELF file error logs during shebang script execution probing
### Description:
Every time a guest script starting with a shebang interpreter (such as `#!/bin/sh` or `#!/bin/bash`) is executed, `elfuse` prints an error-level log:
```
ERROR src/core/elf.c:44: /path/to/script: not an ELF file
```
Although shebang script execution completes successfully (falling back to the interpreter parsed by `elf_read_shebang`), the `ERROR` log is noisy and pollutes standard error logs.

### Steps to Reproduce:
1. Run any guest shebang script in the guest filesystem environment (e.g. `/usr/sbin/dpkg-preconfigure`).
2. Observe `ERROR src/core/elf.c:44: <path>: not an ELF file` in stderr.

### Root Cause Analysis:
In `src/syscall/exec.c` (and similar entry points like `main.c` bootstrap):
```c
    while (elf_load(path_host, &elf_info) < 0) {
        ...
        int rc = elf_read_shebang(path_host, interp_start, sizeof(interp_start), ...);
```
`elf_load` is called unconditionally first as a probe to check if the path is a native ELF. Inside `src/core/elf.c`, if the ELF magic check fails:
```c
    /* Reject non-ELF inputs before interpreting the rest of the header. */
    if (ehdr.e_ident[0] != ELFMAG0 || ...) {
        log_error("%s: not an ELF file", path);
        fclose(f);
        return -1;
    }
```
Because shebang scripts are expected to fail `elf_load` to enter the fallback block, this log triggers on every script execution.

### Suggested Fix:
Downgrade the log level from `log_error` to `log_debug` (or check file magic silently before logging an error), so that normal shebang resolution does not pollute standard error logs.

jserv:
elf_load() was called as a probe before shebang resolution, so it logged "not an ELF file" on every script exec. A shebang script is an expected non-ELF input at that point, not an error.

Reorder sys_execve to resolve binfmt_script first via elf_read_shebang and call elf_load exactly once after the loop, mirroring what the startup path in main.c already does. The error log now fires only on a genuine non-ELF, non-script exec.

Align the recursion depth limit across both entry points. The prior elf_load-first loop allowed a max-depth chain ending in a real ELF to load; main.c capped one level shorter and rejected such a chain. Check the limit only after a further shebang is confirmed, and share the bound via ELF_SHEBANG_MAX_DEPTH so the two paths cannot drift. Both now match the Linux exec_binprm recursion limit: five script resolutions followed by a real ELF load, a sixth returns ELOOP.

Add a shebang parser case for "env -S python3 -O" single-argument suffixes.

Close #166

---
## Summary by cubic
Silences the “not an ELF file” log when executing shebang scripts by resolving the shebang before probing ELF, and aligns recursion limits with Linux. Also adds support for `/usr/bin/env -S` single-argument suffixes.

- **Bug Fixes**
  - Resolve shebang via elf_read_shebang first and call elf_load once after the loop, so we only log on true non-ELF, non-script execs.
  - Align recursion depth across startup and execve using ELF_SHEBANG_MAX_DEPTH; check the limit only after confirming another shebang. Now allows 5 script resolutions followed by a real ELF; a 6th returns ELOOP.

- **New Features**
  - Shebang parser now supports `/usr/bin/env -S python3 -O` (passes the suffix as a single argument).

<sup>Written for commit b708ca1d280096e08d70c192b038188b76d326a8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/167?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/core/elf.h
+++ b/src/core/elf.h
@@ -137,6 +137,12 @@ void elf_resolve_interp(const char *sysroot,
                         char *out,
                         size_t out_sz);
 
+/* Maximum shebang resolutions before a chain is rejected with -ELOOP. Both the
+ * execve path and startup honor this so a max-depth chain ending in a real ELF
+ * still loads, matching the Linux kernel exec_binprm recursion limit.
+ */
+#define ELF_SHEBANG_MAX_DEPTH 5
+
 /* Read, probe, and parse a shebang script header from host_path. Writes
  * interpreter path to interp_out and the single optional argument (if present)
  * to arg_out. arg_out will be set to an empty string if there is no optional
--- a/src/main.c
+++ b/src/main.c
@@ -466,9 +466,8 @@ int main(int argc, char **argv)
     proc_set_sysroot(sysroot);
 
     int shebang_depth = 0;
-    const int max_shebang_depth = 5;
 
-    while (shebang_depth < max_shebang_depth) {
+    while (true) {
         if (resolve_guest_elf_host_path(elf_path, elf_host_path,
                                         sizeof(elf_host_path),
                                         &elf_host_temp) < 0) {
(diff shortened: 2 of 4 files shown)


doanbaotrung:
Tested. The log for "not an ELF file" was gone

PR #183, opened 2026-07-09
Add OFD lock support to fcntl and fix l_pid on conflict

### LINKED ISSUE #129: Incorrect l_pid behavior in OFD file-lock support
OFD (Open File Description) locks are a Linux extension (introduced in Linux 3.15) and have recently been added to macOS. However, their emulated implementation in `sys_fcntl` has two main problems: compilation errors on older host SDKs, and incorrect lock owner PID reporting to the guest.

---

### **Issue 1: Compilation Failure on Older macOS/Xcode SDKs**
* **Location:** `src/syscall/fs.c` (inside the `sys_fcntl` lock cases)
* **Code:**
```c
case 36:   /* F_OFD_GETLK */
case 37:   /* F_OFD_SETLK */
case 38: { /* F_OFD_SETLKW */
    ...
    switch (cmd) {
    ...
    case 36:
        mac_cmd = F_OFD_GETLK;
        break;
    case 37:
        mac_cmd = F_OFD_SETLK;
        break;
    default:
        mac_cmd = F_OFD_SETLKW;
        break;
    }
```
* **Issue:** 
OFD lock constants (`F_OFD_GETLK`, `F_OFD_SETLK`, `F_OFD_SETLKW`) are only defined on newer macOS versions (e.g., Sonoma/Sequoia SDKs). On older macOS hosts or older Xcode/CommandLineTools installations, these constants do not exist in `<sys/fcntl.h>`. Compiling the codebase on these machines results in compilation errors: `use of undeclared identifier 'F_OFD_GETLK'`.
* **Correct Behavior:** 
The code should use preprocessor guards (e.g., `#ifdef F_OFD_GETLK`) to conditionally compile OFD lock support. If the host platform's SDK does not define these constants, it should gracefully fall back to returning `-LINUX_EINVAL` or `-LINUX_ENOTSUP`.

---

### **Issue 2: Incorrect `l_pid` Returned on Conflicting OFD Locks**
* **Location:** `src/syscall/fs.c` (inside the GETLK result write-back block)
* **Code:**
```c
int32_t rp = mac_fl.l_pid;
memset(lflock, 0, sizeof(lflock));
...
memcpy(lflock + 24, &rp, 4);
```
* **Issue:** 
According to Linux specifications (`man fcntl`), when `F_OFD_GETLK` is called and a conflicting lock is found, the `l_pid` field of the returned `flock` structure must always be set to `-1`. This is because OFD locks are associated with the Open File Description (the kernel file table entry) rather than any individual process ID. The current code directly copies the host's `mac_fl.l_pid` into the guest structure, leaking host process IDs or unexpected values to the guest process.
* **Correct Behavior:** 
When processing a query for `F_OFD_GETLK` (cmd `36`), the returned pid should be forced to `-1` if a conflict exists to match Linux specifications. Guest database engines (such as SQLite) rely on this standard behavior to verify locking status.

Max042004:
Adds F_OFD_GETLK/F_OFD_SETLK/F_OFD_SETLKW (Linux cmd 36/37/38) to `sys_fcntl`, which previously fell through to EINVAL. The three cases are guarded by `#ifdef F_OFD_GETLK` etc. so hosts with older SDKs lacking these constants still build and fall back to EINVAL. Since OFD locks belong to the open file description rather than a process, `F_OFD_GETLK` now forces the returned `l_pid` to -1 on a conflict instead of leaking the host PID, matching Linux.

The Linux<->macOS struct flock translation is shared with the existing F_GETLK/F_SETLK/F_SETLKW path via a new `fcntl_flock_op()` helper rather than duplicated.

Fixes #129

## Test plan
- Added `tests/test-ofd-lock.c` (registered in `tests/manifest.txt`), covering cross-fd OFD conflict detection, `l_pid == -1` on conflict, unlock clearing the conflict, and `F_OFD_SETLKW`.
- `make check` green, including the new test and the existing `test-flock`.
- `clang-format --dry-run --Werror` clean.

---
## Summary by cubic
Adds OFD lock support to `sys_fcntl` and fixes PID reporting for lock conflicts to match Linux. Restores SQLite WAL and other OFD users; fixes #129.

- **New Features**
  - Support `F_OFD_GETLK`/`F_OFD_SETLK`/`F_OFD_SETLKW` (guarded by `#ifdef`).
  - Share Linux↔macOS `struct flock` translation via `fcntl_flock_op()`.

- **Bug Fixes**
  - For OFD `GETLK`, return `l_pid = -1` on conflict and reject nonzero request `l_pid` with `EINVAL`.
  - For classic `F_GETLK`, report the conflicting owner as the guest PID via `proc_host_to_guest_pid()`.
  - Tests added: `tests/test-ofd-lock.c`, updated `tests/test-flock.c`, and `tests/test-matrix.sh` updated to run it.

<sup>Written for commit bb90715ebfca242041763f5c85686147a61c254c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/183?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/fs.c
+++ b/src/syscall/fs.c
@@ -1017,6 +1017,130 @@ int64_t sys_dup3(int oldfd, int newfd, int linux_flags)
     return newfd;
 }
 
+/* Translate a Linux struct flock (aarch64) at `arg` to macOS layout, run
+ * fcntl(host_fd, mac_cmd, ...), and for a GETLK command write the result
+ * back translated to Linux layout. Shared by the traditional (F_GETLK/
+ * F_SETLK/F_SETLKW) and OFD (F_OFD_GETLK/F_OFD_SETLK/F_OFD_SETLKW) lock
+ * commands, which differ only in the macOS cmd values and in how l_pid is
+ * reported back for a GETLK conflict.
+ *
+ * Linux aarch64 layout: {short l_type, short l_whence,
+ *   long l_start, long l_len, int l_pid, pad[4]}
+ * macOS layout: {off_t l_start, off_t l_len, pid_t l_pid,
+ *   short l_type, short l_whence}
+ * Use guest_read/guest_write (not guest_ptr) to safely handle structs that
+ * span 2MiB page table block boundaries.
+ */
+static int64_t fcntl_flock_op(guest_t *g,
+                              host_fd_ref_t *host_ref,
+                              uint64_t arg,
+                              int mac_cmd,
+                              bool is_getlk,
+                              bool is_ofd)
+{
+    uint8_t lflock[32]; /* Linux struct flock is 32 bytes on aarch64 */
+    if (guest_read_small(g, arg, lflock, sizeof(lflock)) < 0)
+        return -LINUX_EFAULT;
+
+    int16_t l_type, l_whence;
+    int64_t l_start, l_len;
+    int32_t l_pid;
+    memcpy(&l_type, lflock + 0, 2);
+    memcpy(&l_whence, lflock + 2, 2);
+    memcpy(&l_start, lflock + 8, 8); /* offset 8 due to padding */
+    memcpy(&l_len, lflock + 16, 8);
+    memcpy(&l_pid, lflock + 24, 4);
+
+    /* Linux rejects F_OFD_GETLK/SETLK/SETLKW requests with a nonzero l_pid:
+     * OFD locks are owned by the open file description, not a process, so
+     * the field is reserved on input (fs/locks.c fcntl_getlk/fcntl_setlk
+     * both `return -EINVAL` on a nonzero request l_pid for these commands).
+     */
+    if (is_ofd && l_pid != 0)
+        return -LINUX_EINVAL;
+
+    /* l_type constants differ between Linux and macOS/BSD:
+     *   Linux: F_RDLCK=0, F_WRLCK=1, F_UNLCK=2
+     *   macOS: F_RDLCK=1, F_UNLCK=2, F_WRLCK=3
+     * Passing the Linux value straight through makes a Linux F_RDLCK (0) an
+     * invalid type on macOS, which fcntl() rejects with EINVAL. This is the
+     * lock POSIX databases (e.g. SQLite) take first, so it must map.
+     */
+    short mac_type;
+    switch (l_type) {
+    case 0: /* LINUX_F_RDLCK */
+        mac_type = F_RDLCK;
+        break;
+    case 1: /* LINUX_F_WRLCK */
+        mac_type = F_WRLCK;
+        break;
+    case 2: /* LINUX_F_UNLCK */
+        mac_type = F_UNLCK;
+        break;
+    default:
+        return -LINUX_EINVAL;
+    }
+
+    struct flock mac_fl = {
+        .l_start = l_start,
+        .l_len = l_len,
+        .l_pid = 0,
+        .l_type = mac_type,
+        .l_whence = l_whence, /* SEEK_SET=0, SEEK_CUR=1, SEEK_END=2 same */
+    };
+
+    if (fcntl(host_ref->fd, mac_cmd, &mac_fl) < 0)
+        return linux_errno();
+
+    if (!is_getlk)
+        return 0;
+
+    /* Map macOS l_type back to Linux constants (see above). */
+    int16_t rt;
+    switch (mac_fl.l_type) {
+    case F_RDLCK:
+        rt = 0; /* LINUX_F_RDLCK */
+        break;
+    case F_WRLCK:
+        rt = 1; /* LINUX_F_WRLCK */
+        break;
+    default:
+        rt = 2; /* LINUX_F_UNLCK */
+        break;
+    }
+    int16_t rw = mac_fl.l_whence;
+    int64_t rs = mac_fl.l_start, rl = mac_fl.l_len;
+    int32_t rp;
+    if (is_ofd) {
+        /* OFD locks are owned by the open file description, not a single
+         * process, so Linux always reports l_pid=-1 on a conflicting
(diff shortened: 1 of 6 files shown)


doanbaotrung:
**1. Executed python test script**

`test_ofd_lock.py` file contents:
```
import fcntl
import os
import struct

# Linux OFD Locking Constants
F_OFD_GETLK = 36
F_OFD_SETLK = 37
F_WRLCK = 1

path = '/tmp/test-ofd-python.db'

# Open two distinct open file descriptions for the same file
fd1 = os.open(path, os.O_RDWR | os.O_CREAT | os.O_TRUNC, 0o644)
fd2 = os.open(path, os.O_RDWR, 0o644)

# Pack a Linux struct flock: hhxxxxqqii (32 bytes)
# l_type = F_WRLCK (1), l_whence = SEEK_SET (0), l_start = 0, l_len = 16, l_pid = 0
lock_data = struct.pack('hhxxxxqqii', F_WRLCK, 0, 0, 16, 0, 0)

print("Taking OFD write lock on fd1...")
fcntl.fcntl(fd1, F_OFD_SETLK, lock_data)

print("Querying conflicting lock on fd2...")
query_data = struct.pack('hhxxxxqqii', F_WRLCK, 0, 0, 16, 0, 0)
res = fcntl.fcntl(fd2, F_OFD_GETLK, query_data)

# Unpack the response
l_type, l_whence, l_start, l_len, l_pid, _ = struct.unpack('hhxxxxqqii', res)

print(f"Conflicting lock type: {l_type} (Expected: 1 for F_WRLCK)")
print(f"Conflicting lock PID: {l_pid} (Expected: -1 for OFD conflict)")

if l_type == F_WRLCK and l_pid == -1:
    print("RESULT: PASS")
else:
    print("RESULT: FAIL")

os.close(fd1)
os.close(fd2)
os.unlink(path)
```
Command:
```
elfuse/build/elfuse --sysroot ~/prefixes/ubuntu-arm64/rootfs /usr/bin/python3 test_ofd_lock.py
```
Output:
```
Taking OFD write lock on fd1...
Querying conflicting lock on fd2...
Conflicting lock type: 1 (Expected: 1 for F_WRLCK)
Conflicting lock PID: -1 (Expected: -1 for OFD conflict)
RESULT: PASS
```
**2. Executed native host script**
```
clang -Ielfuse/tests third_party/elfuse/tests/test-ofd-lock.c -o /tmp/host_test_ofd
/tmp/host_test_ofd
rm /tmp/host_test_ofd
```
Output Log:
```
  F_OFD_SETLK F_WRLCK            OK
  F_OFD_GETLK reports conflict from other OFD OK
  F_OFD_GETLK l_pid is -1 on conflict OK
  F_OFD_SETLK F_UNLCK            OK
  F_OFD_GETLK reports no conflict after unlock OK
  F_OFD_SETLKW F_WRLCK           OK
OK

test-ofd-lock: 7 passed, 0 failed - PASS
```

Max042004:
Confirmed — verified against fs/locks.c (fcntl_getlk/fcntl_setlk both reject a nonzero request l_pid with EINVAL for F_OFD_GETLK/SETLK/SETLKW). Fixed in 1180fa2: fcntl_flock_op now reads the guest's l_pid and returns -EINVAL when is_ofd and l_pid != 0, with 3 new regression cases in test-ofd-lock.c.

src/syscall/fs.c
```diff
@@ -1017,6 +1017,119 @@ int64_t sys_dup3(int oldfd, int newfd, int linux_flags)
...
+     * Passing the Linux value straight through makes a Linux F_RDLCK (0) an
+     * invalid type on macOS, which fcntl() rejects with EINVAL. This is the
+     * lock POSIX databases (e.g. SQLite) take first, so it must map.
+     */
+    short mac_type;
+    switch (l_type) {
+    case 0: /* LINUX_F_RDLCK */
+        mac_type = F_RDLCK;
+        break;
+    case 1: /* LINUX_F_WRLCK */
```
doanbaotrung, line 1074: Traditional `F_GETLK` Leaks Host PID to the Guest**
* **Location:** `fcntl_flock_op` inside `src/syscall/fs.c`
* **Problem:**
  For OFD locks (`F_OFD_GETLK`), the implementation correctly overrides the conflict PID with `-1` (per Linux specification) to avoid leaking host PIDs.
  
  However, for traditional locks (`F_GETLK`), it reports:
  ```c
  rp = mac_fl.l_pid;
  ```
  This passes the macOS host process ID directly to the guest. Because `elfuse` translates and virtualizes guest PIDs, the guest program (e.g., SQLite or other locking monitors) receives a raw host PID (e.g., `40548`) instead of the child's guest PID (e.g., `2`). 
  
  If the guest tries to manage or check that PID (for example, via `kill(conflict_pid, 0)`), the PID translation gate in `sc_kill` looks for the host PID in the guest table, fails to resolve it, and returns `ESRCH`, even though the process is alive.

* **Proposed Fix:**
  Add a `proc_host_to_guest_pid` helper to map the host PID back to its corresponding guest PID via the child process table and cross-process registry.

Max042004:
Good catch — fixed in da96047. Added `proc_host_to_guest_pid()` (mirror of the existing `proc_guest_to_host_pid()`) to translate the conflicting lock's host PID back to a guest PID via the local child table + cross-process registry, so traditional `F_GETLK` now reports a PID the guest can actually resolve instead of a raw host PID. Also rebased onto main's test-matrix.sh restructuring and added a fork+pipe regression test in test-flock.c that fails without the fix.

doanbaotrung:
I think it good now.

PR #184, opened 2026-07-09
Fix openat2 ENOTDIR/ENOENT errno mismatches

### LINKED ISSUE #133: Bug: Mismatches in Ordinary Relative Path Resolution (`path_translate_at`)
### Problem
When a guest process triggers a path-based system call with an empty path `""` (e.g., `openat(dirfd, "", flags)` without `AT_EMPTY_PATH`), it is expected to fail with `ENOENT`.

However, in `path_translate_at` ([src/syscall/path.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/path.c#L188-L204)), the following condition is checked to detect ordinary relative paths:
```c
if (tx->guest_path == path && path[0] != '/') {
```
Because the empty string `""` does not start with `/` (i.e. `path[0] == '\0' != '/'`), this condition evaluates to **true**. 

The code then resolves the path relative to the directory descriptor `dirfd` using `dirfd_guest_base_path(dirfd, base, sizeof(base))`, which writes the path of the parent directory itself into `tx->guest_buf`. As a result:
- The empty path `""` is translated to the directory path itself (e.g., `/usr/bin/` or `/`).
- The subsequent host call succeeds and returns a descriptor/details for the directory, rather than returning `ENOENT`.

### Linux/POSIX Standard Behavior
Unless `AT_EMPTY_PATH` is set, passing an empty path to file-based syscalls (such as `openat`, `fstatat`, `fchmodat`, `fchownat`) must fail with `-ENOENT`.

---

### Problem
If a guest process calls a system call like `openat(dirfd, "relative_path", ...)` where `dirfd` is a valid file descriptor pointing to a **regular file** (not a directory), the call should fail with `ENOTDIR`.

In the new relative path resolution code, `dirfd_guest_base_path` is called to find the base directory. However, `dirfd_guest_base_path` rejects any non-directory file descriptors with `EBADF`:
```c
if (snap.type != FD_DIR) {
    errno = EBADF;
    return -1;
}
```
As a result:
- `path_translate_at` returns `-1` with `errno = EBADF`.
- The system call surfaces `-EBADF` (Bad File Descriptor) to the guest process instead of the correct `-ENOTDIR` (Not a Directory).

### Linux/POSIX Standard Behavior
If `dirfd` is a valid, open file descriptor but does not refer to a directory, trying to resolve a relative path through it must fail with `ENOTDIR`. `EBADF` should only be returned if the file descriptor is not open or is invalid.

Max042004:
## Summary
- `dirfd_guest_base_path()` returned EBADF instead of ENOTDIR for a valid but non-directory dirfd with no host-nameable path (pipes, sockets, etc.), surfacing through openat2()'s RESOLVE_NO_XDEV precheck.
- `sc_openat2()` let an empty pathname succeed under RESOLVE_IN_ROOT, since `path_openat2_normalize_in_root()` substitutes "." for "" and resolves to dirfd itself. openat2() has no AT_EMPTY_PATH equivalent, so empty paths now fail ENOENT up front.

Fixes #133.

## Test
Added two regression tests to `tests/test-syscall-fidelity.c`
`make check` — 120/120 passed

---
## Summary by cubic
Fix errno handling in openat2 path resolution: valid non-directory dirfds now return ENOTDIR (not EBADF), and empty pathnames return ENOENT even under RESOLVE_IN_ROOT. Added two regression tests to confirm Linux-compatible behavior.

<sup>Written for commit 9b99624c0743269ebd1fba821f12283e75284158. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/184?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/path.c
+++ b/src/syscall/path.c
@@ -875,8 +875,13 @@ static int dirfd_guest_base_path(guest_fd_t dirfd, char *out, size_t outsz)
     if (path_openat2_dirfd_host_path(dirfd, host_path, sizeof(host_path)) == 0)
         return host_path_to_guest_path(host_path, out, outsz);
 
+    /* fd_snapshot already proved dirfd is open, so a valid-but-wrong-type fd
+     * (pipe, socket, epoll, ...) belongs here, not in the "bad fd" case: Linux
+     * resolves a relative path against such a dirfd with ENOTDIR, reserving
+     * EBADF for a closed or out-of-range descriptor.
+     */
     if (snap.type != FD_DIR) {
-        errno = EBADF;
+        errno = ENOTDIR;
         return -1;
     }
 
--- a/src/syscall/syscall.c
+++ b/src/syscall/syscall.c
@@ -1944,6 +1944,14 @@ static int64_t sc_openat2(guest_t *g,
         if (guest_read_str(g, x1, path, sizeof(path)) < 0)
             return -LINUX_EFAULT;
 
+        /* openat2() has no AT_EMPTY_PATH equivalent in how.flags, so an empty
+         * pathname is always ENOENT. Without this, RESOLVE_IN_ROOT's "."
+         * substitution for an empty path (see path_openat2_normalize_in_root)
+         * would resolve to dirfd itself and let the open succeed.
+         */
+        if (path[0] == '\0')
+            return -LINUX_ENOENT;
+
         if ((resolve & (RESOLVE_NO_SYMLINKS | RESOLVE_NO_MAGICLINKS)) &&
             path_openat2_is_fd_magiclink_anchor((int) x0, path))
             return -LINUX_ELOOP;
--- a/tests/test-syscall-fidelity.c
+++ b/tests/test-syscall-fidelity.c
@@ -1289,6 +1289,58 @@ static void test_mmap_low_hint_exact(void)
     munmap(p, len);
 }
 
+/* A dirfd that is open but not a directory (and has no host-nameable path,
+ * e.g. a pipe) must resolve a relative path with ENOTDIR, not EBADF: the fd
+ * itself is valid, it just cannot anchor a lookup. Issue #133.
+ */
+static void test_openat2_resolve_no_xdev_non_directory_dirfd_rejects_enotdir(
+    void)
+{
+    TEST("openat2 RESOLVE_NO_XDEV non-directory dirfd rejects ENOTDIR");
+    int pipefd[2];
+    if (pipe(pipefd) < 0) {
+        FAIL("pipe");
+        return;
+    }
+    close(pipefd[1]);
+
+    struct open_how how = {
+        .flags = O_RDONLY, .mode = 0, .resolve = RESOLVE_NO_XDEV};
+    long fd = syscall(SYS_openat2, pipefd[0], "relative", &how, sizeof(how));
+    close(pipefd[0]);
+    if (fd >= 0) {
+        close((int) fd);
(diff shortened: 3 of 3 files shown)


doanbaotrung:
### Emulated Test Script (`test_openat2_fixes.py`)
This script uses the guest's Python 3 interpreter running under `elfuse` to directly make Linux `openat2` system calls (syscall `437`) via `ctypes` and verifies that the correct Linux error numbers (`ENOENT` and `ENOTDIR`) are returned.

Save this code as `test_openat2_fixes.py`:

```python
import ctypes
import os
import struct

# Load guest libc and define __errno_location to read thread-local errno
libc = ctypes.CDLL(None)
libc.__errno_location.restype = ctypes.POINTER(ctypes.c_int)
def get_errno():
    return libc.__errno_location().contents.value

# Linux openat2 syscall & resolve constants
SYS_openat2 = 437
RESOLVE_NO_XDEV = 0x01
RESOLVE_IN_ROOT = 0x10

# Test 1: Empty path with RESOLVE_IN_ROOT should fail with ENOENT (Linux errno 2)
print("Running Test 1: openat2 empty path with RESOLVE_IN_ROOT...")
dirfd = os.open('/tmp', os.O_RDONLY | os.O_DIRECTORY)
how = struct.pack('QQQ', os.O_RDONLY, 0, RESOLVE_IN_ROOT)
res1 = libc.syscall(SYS_openat2, dirfd, b'', how, 24)
err1 = get_errno()
os.close(dirfd)

# Test 2: Relative path with non-directory dirfd (e.g., a pipe) should fail with ENOTDIR (Linux errno 20)
print("Running Test 2: openat2 relative path with pipe fd...")
r_fd, w_fd = os.pipe()
how = struct.pack('QQQ', os.O_RDONLY, 0, RESOLVE_NO_XDEV)
res2 = libc.syscall(SYS_openat2, r_fd, b'relative', how, 24)
err2 = get_errno()
os.close(r_fd)
os.close(w_fd)

# Output results
print(f"Test 1 (Empty path)       -> Result: {res1}, Errno: {err1} (Expected: -1, 2 [ENOENT])")
print(f"Test 2 (Non-directory fd) -> Result: {res2}, Errno: {err2} (Expected: -1, 20 [ENOTDIR])")

if res1 == -1 and err1 == 2 and res2 == -1 and err2 == 20:
    print("RESULT: PASS")
else:
    print("RESULT: FAIL")
```

---

### Test Execution Command
Run the test script directly using the standalone `elfuse` binary and your guest `sysroot`:
```bash
elfuse/build/elfuse --sysroot ~/prefixes/ubuntu-arm64/rootfs /usr/bin/python3 test_openat2_fixes.py
```

---

### Output Log
```
Running Test 1: openat2 empty path with RESOLVE_IN_ROOT...
Running Test 2: openat2 relative path with pipe fd...
Test 1 (Empty path)       -> Result: -1, Errno: 2 (Expected: -1, 2 [ENOENT])
Test 2 (Non-directory fd) -> Result: -1, Errno: 20 (Expected: -1, 20 [ENOTDIR])
RESULT: PASS
```

PR #185, opened 2026-07-09
Close sysroot escape via relative symlinks

### LINKED ISSUE #130: Directory Traversal Vulnerability in `sysroot_path_is_contained` and `linkat` failures for Symlinks
### **Directory Traversal Vulnerability in `sysroot_path_is_contained` and `linkat` failures for Symlinks**

This issue outlines two symlink-related bugs: a security vulnerability that allows guest processes to escape the sysroot boundary via absolute guest symbolic links, and a system call failure on macOS when hard-linking symlinks.

---

### **Issue 1: Directory Traversal Vulnerability via Guest Absolute Symlinks**
* **Location:** `src/syscall/proc-state.c` (inside `sysroot_path_is_contained`)
* **Code:**
```c
struct stat st;
if (lstat(resolved_path, &st) == 0 && S_ISLNK(st.st_mode)) {
    char target[LINUX_PATH_MAX];
    ssize_t len = readlink(resolved_path, target, sizeof(target) - 1);
    if (len >= 0) {
        target[len] = '\0';
        if (target[0] == '/') {
            return true;
        }
    }
}
```
* **Issue:**
When `follow_final` is `true`, `sysroot_path_is_contained` is supposed to verify that the ultimate target of a path is contained within the guest sysroot. 
However, the code block above returns `true` (contained) immediately if the path is a symbolic link whose target starts with `/` (an absolute symlink in the guest rootfs). By returning `true` early, it completely bypasses the subsequent `realpath` validation. 
If an absolute symlink target contains relative parent directory elements (e.g. `/../../etc/passwd` or `/sysroot/../../../etc/passwd`), it will escape the guest sysroot and point to the host rootfs. Because of the early exit, the containment guard is bypassed, allowing emulated processes to access, create, or modify arbitrary files on the host (container escape).

* **Correct Behavior:**
If `follow_final` is `true`, symbolic links must not bypass containment checks. The path validation engine must recursively resolve the emulated guest symbolic links to their final target on the host first, and then run `realpath()` validation on that resolved path to ensure it does not escape the guest sysroot.

---

### **Issue 2: `linkat` fails with `EPERM` when linking symbolic links on macOS**
* **Location:** `src/syscall/fs.c` (inside `sys_linkat`)
* **Issue:**
macOS does not support creating a hard link to a symbolic link; calling `linkat` with a symbolic link target on macOS returns `EPERM` (or `EINVAL`). On Linux, this is a supported operation. 
* **Correct Behavior:**
To achieve Linux compatibility on macOS, if `linkat` fails with `EPERM` and the source file is a symbolic link, `sys_linkat` should fall back to creating a new symbolic link pointing to the same target (`symlinkat`).

Max042004:
## Summary
- A relative `openat(dirfd, name)` against a dirfd already inside `--sysroot` skipped containment entirely; a symlink reachable through it (absolute target, or relative with enough `..`) escaped the sysroot with no check, unlike the identical escape through an absolute path. Fixed by re-validating the reconstructed absolute path through the existing containment-checked resolver.
- Darwin's `linkat(2)` can reject a hard link to a symlink source on non-APFS filesystems (reproduced as ENOTSUP on Case-sensitive HFS+); `sys_linkat` now falls back to `symlinkat` with the same target.

Close #130

## Test
`make check` (green, includes new `test-sysroot-symlink-escape`)
`make test-linkat-symlink-fallback` (standalone, builds a scratch HFS+ image via hdiutil; green)

---
## Summary by cubic
Fixes a sysroot escape where relative dirfd `*at()` calls could traverse symlinks outside `--sysroot`. Also adds a Darwin `linkat` fallback so linking a symlink source behaves like Linux.

- **Bug Fixes**
  - For relative `*at()` under `--sysroot`, rebuild the absolute guest path from the dirfd base, run realpath + containment checks, and use the verified host path; reject absolute-target and deep-`..` symlink escapes with ELOOP, including `AT_FDCWD` after `chdir` and `faccessat`.
  - On Darwin, if `linkat` on a symlink source fails with EPERM/ENOTSUP (without `AT_SYMLINK_FOLLOW`), fall back to `symlinkat` with the same target to match Linux.
  - Tests: add `test-sysroot-symlink-escape` to `make check`, and a standalone `test-linkat-symlink-fallback` (HFS+ image; skips when unavailable).

<sup>Written for commit 65fb57485be7feb06dea2c6bbd69b2c79f52cad1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/185?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/mk/tests.mk
+++ b/mk/tests.mk
@@ -15,7 +15,8 @@
         test-sysroot-create-paths test-fork-ipc-protocol-host test-identity-override-host \
         test-proctitle-host test-proctitle-low-stack \
         test-sysroot-procfs-exec test-timeout-disable test-fuse-alpine \
-        test-sysroot-nofollow test-sysroot-chdir perf
+        test-sysroot-nofollow test-sysroot-chdir test-sysroot-symlink-escape \
+        test-linkat-symlink-fallback perf
 
 ## Build and run the assembly hello world test
 test-hello: $(ELFUSE_BIN) $(TEST_HELLO_DEP)
@@ -113,6 +114,8 @@ check: $(ELFUSE_BIN) $(TEST_DEPS) check-syscall-coverage \
 	@$(MAKE) --no-print-directory test-sysroot-host-fallback
 	@printf "\n$(BLUE)━━━ sysroot byte-exact lookup validation ━━━$(RESET)\n"
 	@$(MAKE) --no-print-directory test-sysroot-case-exact
+	@printf "\n$(BLUE)━━━ sysroot relative-dirfd symlink escape validation ━━━$(RESET)\n"
+	@$(MAKE) --no-print-directory test-sysroot-symlink-escape
 	@printf "\n$(BLUE)━━━ Alpine sysroot FUSE validation ━━━$(RESET)\n"
 	@$(MAKE) --no-print-directory test-fuse-alpine
 	@printf "\n$(BLUE)━━━ timeout=0 validation ━━━$(RESET)\n"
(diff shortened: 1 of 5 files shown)


doanbaotrung:
### 1. Directory Traversal Vulnerability Fix
* **Fix Target:** Guest absolute/relative symbolic links escaping the sysroot boundary are rejected with `ELOOP` (40) instead of being resolved to host targets.
* **Test script:** Run a Python 3 guest script that sets up an absolute symbolic link (`/abs-link -> ~`) and attempts to resolve it.

**Execution Command & Log Output:**
```bash
$elfuse/build/elfuse --sysroot ~/prefixes/ubuntu-arm64/rootfs /usr/bin/python3 -c "
import os
try:
    os.stat('/abs-link')
except OSError as e:
    print('Error:', e)
"
Error: [Errno 40] Too many levels of symbolic links: '/abs-link'
```
* **Result:** **PASS**. The guest process fails to escape the sysroot via the absolute symlink, cleanly returning `ELOOP` (40) to the guest.

---

### 2. `linkat` Fallback for Symlinks
* **Fix Target:** When the host filesystem (e.g. Case-Sensitive HFS+ on macOS) does not support hard-linking a symlink directly (without `AT_SYMLINK_FOLLOW`), `linkat()` fails with `EPERM` or `ENOTSUP`. `sys_linkat()` falls back to calling `symlinkat` to duplicate the link target, ensuring Linux-compatible success.
* **Test script:** Use Python ctypes to invoke `linkat` directly via the guest C library (`libc.so.6`) with `flags=0` (no follow) on a symbolic link, under a mocked host `linkat` failure of `ENOTSUP`.

**Execution Command & Log Output:**
```bash
$elfuse/build/elfuse --sysroot ~/prefixes/ubuntu-arm64/rootfs /usr/bin/python3 -u -c "
import ctypes
import os

path = '/tmp/d1'
with open(path + '/target.txt', 'w') as f:
    f.write('hello\n')
sym_path = path + '/sym'
os.symlink('target.txt', sym_path)

# Load libc and call linkat
libc = ctypes.CDLL('libc.so.6', use_errno=True)
libc.linkat.argtypes = [ctypes.c_int, ctypes.c_char_p, ctypes.c_int, ctypes.c_char_p, ctypes.c_int]
libc.linkat.restype = ctypes.c_int

res = libc.linkat(-100, b'/tmp/d1/sym', -100, b'/tmp/d1/hardlink', 0)
print('linkat returned:', res)

is_lnk = os.path.islink('/tmp/d1/hardlink')
tgt = os.readlink('/tmp/d1/hardlink')
print(f'Fallback result is a symlink: {is_lnk} (Expected: True)')
print(f'Fallback symlink target: {tgt} (Expected: target.txt)')
"
sys_symlinkat: target=target.txt, host_path=~/prefixes/ubuntu-arm64/rootfs/tmp/d1/sym
symlinkat returned 0, errno=0
sys_linkat entrance: oldpath=/tmp/d1/sym, newpath=/tmp/d1/hardlink, flags=0
host linkat returned -1, errno=45 (MOCKED ENOTSUP)
linkat returned: 0
Fallback result is a symlink: True (Expected: True)
Fallback symlink target: target.txt (Expected: target.txt)
```
* **Result:** **PASS**. The guest C library call to `linkat` returns `0` (success) to the guest program, and the target is cleanly recreated as a symbolic link pointing to `target.txt`.

jserv (changes requested):
Squash and refine commit messages.

Issue #31, opened 2026-05-13
OCI image support for Linux userspace execution on macOS


jserv:
## Summary
Add [OCI](https://opencontainers.org/) image support to elfuse so that OCI-packaged Linux userspace can be fetched, unpacked, and executed on macOS without a Linux kernel or VM in the loop.

This issue proposes a phased roadmap and surfaces the compatibility cliffs that need to be answered before code lands. Initial scope is intentionally limited to `linux/arm64` images consumed as Linux root filesystems - not to [OCI Runtime Spec][runtime-spec] compatibility.

## Background
OCI images have become the standard packaging format for Linux userspace, well beyond their original "container" framing. On macOS, common ways to run those images today are VM-backed:
* [Docker Desktop](https://www.docker.com/products/docker-desktop/)
* [OrbStack](https://orbstack.dev/)
* Apple's [`container`](https://github.com/apple/container) stack
* [Lima](https://github.com/lima-vm/lima)
 
All execute Linux workloads against a Linux kernel inside a virtualized environment. [`nerdctl`](https://github.com/containerd/nerdctl) is a CLI typically driven through a Linux VM environment such as Lima; it is not itself
a macOS runtime.

They differ in UX, but share one execution model:
```
OCI image -> Linux VM -> Linux kernel -> runtime -> application
```

Even OrbStack, which optimizes this aggressively, still executes Linux applications inside a Linux guest kernel.

## What elfuse Does Differently
elfuse translates Linux syscalls into macOS primitives via a runtime layer atop [Hypervisor.framework](https://developer.apple.com/documentation/hypervisor):
```
Linux ELF -> elfuse syscall/runtime -> macOS kernel
```

The goal is a Linux ABI runtime, not a complete Linux emulation: fast Linux userspace execution, low-overhead process startup, host filesystem access without a virtualized bridge, and direct visibility into the Linux ABI for education and research.

Today, running anything dynamically-linked requires hand-built sysroots fed via `--sysroot`. OCI integration replaces that with the existing Linux software distribution ecosystem.

## Relationship to Apple Container (prior art only)
Apple's [`container`](https://github.com/apple/container) and [`Containerization`](https://github.com/apple/containerization) projects are useful prior art for image distribution: registry access, image
management, and root filesystem construction. They are not an execution-model match. Apple's current design runs each Linux container inside a lightweight VM on [Virtualization.framework](https://developer.apple.com/documentation/virtualization).

[...]

PR #191, opened 2026-07-09
Add OCI image support

henrybear327:
Prior work: https://github.com/sysprog21/elfuse/pull/34
Tracking issue: https://github.com/sysprog21/elfuse/issues/31

We introduce `elfuse-oci`, a standalone Go companion binary that owns the OCI image pipeline: `pull`, `unpack`, `inspect`, `run`, `list` (alias `images`), `rmi`, and `prune`, backed by a real OCI image-layout store. elfuse itself stays a pure Linux syscall-to-Darwin runtime with **no** OCI awareness; the two binaries meet only at the existing `elfuse --sysroot <rootfs> <entrypoint> <args>` launch path.

OCI images are used here strictly as a **distribution format for Linux root filesystems** — a reproducible replacement for hand-built `--sysroot` trees. **This is not a container runtime.** There is no isolation layer: the guest shares the host network identity, PID space, and clock, and unresolved guest paths fall back to host truth. A workload that needs namespaces, cgroups, port mapping, `exec` into a running container, or a daemon needs a real container runtime, not an ELF personality.

## Why a separate Go binary

The OCI ecosystem is a Go ecosystem. Rather than grow the C runtime with an image pipeline, the acquisition/lifecycle half is ~4.4k lines of Go (plus ~7.3k lines of tests) built on `github.com/google/go-containerregistry`, and validated for on-disk conformance against `crane`, `skopeo`, and `umoci`. Keeping it out of the C tree keeps both sides small and lets each lean on its native tooling.

## Design

`docs/oci-design.md` is the source of truth: the C/Go boundary, the image-layout store and its `refs.json` pin table, hardened layer application, the `run` paths, the concurrency/locking model, and an explicit accounting of which OCI features are and are not implemented. `docs/usage.md` covers the commands; `docs/testing.md` the offline/CI split; `docs/internals.md` the host-literal path fallback.

Highlights:
- **Store**: spec-shape OCI image-layout other tools can read; refs pinned by digest; an exclusive flock serializes metadata writes so concurrent pulls cannot lose pins.
- **Unpack**: `os.Root`-bounded extraction (no symlink/hardlink escape), correct whiteout/opaque handling, staged temp-dir + atomic rename so readers never see a partial tree.
- **run**: resolves Entrypoint/Cmd/Env/User/WorkingDir with `env(1)`/Docker precedence, guarantees a PATH, resolves symbolic `--user` against the image's own `/etc/passwd`+`/etc/group` (no-follow), injects host `/etc/{resolv.conf,hosts,hostname}`, then execs elfuse in place so signals and the pid pass through.
- **macOS rootfs**: per-digest case-sensitive APFS sparsebundle with per-run clonefile COW; liveness/lifecycle decided by per-digest advisory flocks (`attach.lock`/`run.lock`), not pids. `--plain-rootfs` remains available.
- **Lifecycle GC**: `rmi`/`prune` use reachability GC (shared blobs survive while any ref reaches them) and never reclaim a cache a live run still holds — the run-lock rides through the exec into elfuse and releases exactly on guest exit.

## Try it

```
make elfuse elfuse-oci

build/elfuse-oci pull alpine:3
build/elfuse-oci inspect alpine:3
build/elfuse-oci list

build/elfuse-oci run alpine:3 /bin/sh -c 'echo hello from elfuse'
build/elfuse-oci run --entrypoint /usr/local/bin/python3 python:3.12 \
  -c 'import json,math; print(json.dumps({"pi":round(math.pi,5),"ok":True}))'

build/elfuse-oci rmi --force alpine:3
build/elfuse-oci prune --cache --all
```

Images are stored under `$ELFUSE_OCI_STORE`, or `~/.local/share/elfuse/oci` by default.

## CI

Split by runner capability
- a Linux job exercises the pure-Go store paths (pull, unpack, inspect, lifecycle) plus image-layout conformance and cross-tool interop (`crane`/`skopeo`/`umoci`) with no Hypervisor.framework
- a hosted macOS job builds and tests the darwin sparsebundle code and a run-less lifecycle smoke

---
## Summary by cubic
Adds OCI image support with a new Go CLI, `elfuse-oci`, an extracted `elfuse_launch`, and a macOS case‑sensitive APFS sparsebundle rootfs with per‑run COW. Delivers a full pull → inspect/unpack → run → rmi/prune lifecycle with reachability GC, crash‑durable store writes, and stronger path, symlink, exec, AF_UNIX, and inotify handling.

- **New Features**
  - OCI CLI/store: `pull`, `unpack`, `inspect`/`list --json`, `run`, `rmi`, `prune`; spec‑shape OCI image‑layout with refs pinned by digest; synced pin/index/blob writes; reachability GC; `--platform` (default linux/arm64); timed keychain; built on `github.com/google/go-containerregistry`.
  - Run: resolves Entrypoint/Cmd/Env/User/WorkingDir; adds `--user`/`--workdir`/`--env`/`--clear-env` with env(1) semantics; guarantees PATH; creates missing WorkingDir; resolves symbolic users from in‑image passwd/group (no‑follow); injects host `/etc/{resolv.conf,hosts,hostname}`; forwards INT/TERM/QUIT/HUP; uses `--` before guest argv; execs `elfuse --sysroot`.
  - macOS rootfs: per‑digest case‑sensitive APFS sparsebundles with per‑run `clonefile` COW; advisory flock lifecycle (`attach.lock`/`run.lock`) that survives exec; reuse vs. force‑detach of stale mounts; `--keep` sidecars; `--plain-rootfs` retained; safe prune/rmi that skip busy caches.
  - Filenames/paths: case‑escape codec + case‑exact walk; clamps `..` at guest root; guest‑namespace symlink targets; PT_INTERP resolution; `/proc/self/{exe,fd}` reverse‑map; pathname AF_UNIX translate/shorten and reverse‑map; inotify path translate/name decode; enforce `O_NOFOLLOW`/`O_NONBLOCK`; unify `/tmp`/`/var/tmp`/`~/.ccache` redirects.
  - C runtime: extracted `elfuse_launch`; new launch flags mapped to staged initial IDs, absolute Workdir validation, and conflict guards.

- **Docs & CI**
  - Docs: added `docs/oci-design.md`; updates to `docs/usage.md`, `docs/testing.md`, `docs/internals.md`.
  - Tests/CI: image‑layout conformance and cross‑tool interop (`crane`/`skopeo`/`umoci`); sparsebundle round‑trip; exec checks; end‑to‑end lifecycle; per‑image workloads (python/node/go/jvm/c). Linux runs pure‑Go store paths; hosted macOS builds/tests darwin code; self‑hosted Apple Silicon runs full HVF flows. New `hvf-elfuse-setup` composite action.

<sup>Written for commit 4ae554edca9edcf3eb1250179e208368c4fd8f3d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/191?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jserv (changes requested):
Rebase latest `main` branch, resolve conflicts, and refine per review messages.

jserv:
> Commands to try out
> ```
> build/elfuse oci list
> build/elfuse oci pull alpine:3
> build/elfuse oci inspect alpine:3
> build/elfuse oci list
> build/elfuse oci rmi --force e7a1a92a5bfe 
> build/elfuse oci list
> build/elfuse-container run --entrypoint /usr/local/bin/python3 \\n    python:3.12 -c 'import json,math; print(json.dumps({"pi":round(math.pi,5),"ok":True}))'
> build/elfuse oci run alpine:3 /bin/sh -c 'echo hello from elfuse'
> ```

Leveraging existing Go-based tools and packages is a great step toward full OCI image support. I agree with this change.

I suggest repositioning the `build/elfuse` binary as an efficient Linux syscall-to-macOS/Darwin runtime, while implementing `build/elfuse-container` in Go using OCI-related packages.

In other words, `build/elfuse oci` should not be considered a valid command. OCI-specific functionality should reside in `elfuse-container`, not in the `elfuse` executable.

henrybear327:
Putting this PR as draft for now. Experimenting with Claude code and the setup went wrong, the PoC code polluted the codebase.

jserv (changes requested):
Refine per review messages from @cubic-dev-ai and squash commits for further review.

henrybear327:
> Refine per review messages from @cubic-dev-ai and squash commits for further review.

I am in the process of resetting the changes. 

The original 8 commits would be amended for changes as those were the logically reviewable units.

jserv:
> The original 8 commits would be amended for changes as those were the logically reviewable units.

Avoid using [Semantic Commit Messages](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716), Instead, stick to [How to Write a Git Commit Message](https://cbea.ms/git-commit/),

henrybear327:
- Improved test coverage (go test cases using table tests format for better readability)
- Improved CI coverage to test the lifecycle of the OCI code
- Improved the documentation to document the design, implementation, and limitation 
- Improved the commit message

henrybear327:
I am doing a sweep of the existing AI review comments.

cmd/elfuse-container/csrun.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return nil, "", err
+	}
+	rootfs := m.rootfsDir()
+	if _, err := os.Stat(rootfs); err != nil {
+		if !os.IsNotExist(err) {
+			return nil, "", errors.Join(err, closeMount(m))
+		}
+		fmt.Fprintf(os.Stderr, "Unpacking %s -> %s\n", ref, rootfs)
+		if err := unpackImage(s, ref, rootfs); err != nil {
```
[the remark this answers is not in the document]
henrybear327: unpackImage removes a dest it created when extraction fails (deferred cleanup in unpack.go), so later runs cannot execute a partial tree; a kill-window leftover is recovered by the orphan sweep.

cmd/elfuse-oci/run.go
```diff
@@ -0,0 +1,125 @@
...
+			if ws, ok := state.Sys().(syscall.WaitStatus); ok {
+				if ws.Signaled() {
+					return 128 + int(ws.Signal()), nil
+				}
+				return ws.ExitStatus(), nil
+			}
+			return state.ExitCode(), nil
+		case sig := <-sigCh:
+			if cmd.Process != nil {
+				_ = cmd.Process.Signal(sig)
```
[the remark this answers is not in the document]
henrybear327: Working as designed: the child shares the process group so terminal signals reach it directly, and forwarding covers signals targeted at elfuse-container alone (see the spawnElfuseWait comment). Suppressing the duplicate would drop targeted-signal propagation.

cmd/elfuse-oci/gc.go
```diff
@@ -0,0 +1,304 @@
...
+	// On a last-pin removal, update index.json BEFORE committing the pin
+	// removal to refs.json. In the opposite order, a failure between the two
+	// writes strands the manifest: the ref is gone from refs.json (so rmi can
+	// no longer resolve it) while the descriptor keeps every blob live, and
+	// prune never removes descriptors -- the image becomes unreclaimable. In
+	// this order the same crash window leaves a stale pin over a removed
+	// descriptor, which a retried rmi resolves and finishes
+	// (RemoveDescriptors is a filter, so re-removing is a no-op).
+	if lastPin {
+		if err := s.removeManifestDescriptor(digest); err != nil {
```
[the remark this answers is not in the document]
henrybear327: Same disposition as the concurrent-run threads: documented single-user limitation (docs/oci-design.md "Concurrency", 15fbc0b); metadata stays lock-consistent, unpack-vs-rmi serialization is out of scope.

cmd/elfuse-oci/gc.go
```diff
@@ -0,0 +1,304 @@
...
+			break
+		}
+	}
+
+	if lastPin {
+		if !force && cacheExists(s.root, digest) {
+			return rmReport{}, fmt.Errorf("rmi: %q has an unpacked cache; pass --force or run 'elfuse-container prune --cache' first", ref)
+		}
+		if force {
+			if err := removeRefCaches(s, digest); err != nil {
```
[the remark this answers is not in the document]
henrybear327: Fixed in dcfcd8e: `rmi --force` now runs the bundle sweep first and refuses when a live run's clone (or `--no-clone` marker) still uses the volume.

cmd/elfuse-oci/gc.go
```diff
@@ -0,0 +1,304 @@
...
+// would reclaim and deletes nothing.
+//
+// gc is the shared engine behind `rmi` (called after a manifest descriptor is
+// removed from index.json, so the dropped image's config/layers surface as
+// unreachable) and `prune` (a standalone sweep that reclaims retag/partial-pull
+// orphans). It also reclaims stale temporary blob files left by interrupted
+// writes under blobs/sha256/; those filenames are not valid sha256 digests and
+// make layout.Path.GarbageCollect abort before it can sweep anything.
+func (s *store) gc(dryRun bool) (rmReport, error) {
+	live, err := s.liveBlobDigests()
```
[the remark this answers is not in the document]
henrybear327: Fixed in dcfcd8e: prune's GC now runs under the store flock, closing the scan/delete race with a concurrent pull.

cmd/elfuse-oci/run.go
```diff
@@ -0,0 +1,125 @@
...
```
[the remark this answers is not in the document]
henrybear327: Working as designed (same as the earlier signal thread): documented tradeoff -- forwarding keeps wrapper-targeted signals working; group delivery covers Ctrl-C.

cmd/elfuse-oci/etc.go
```diff
@@ -0,0 +1,98 @@
...
+	if err := writeRuntimeFile(root, "etc/hosts", []byte(hosts)); err != nil {
+		return err
+	}
+
+	// resolv.conf: copy the host's verbatim (host-truth) so the guest's DNS
+	// lookups hit the same nameservers the host uses. Fall back to a minimal
+	// default if the host file is absent or empty.
+	resolv, err := readHostResolvConfig()
+	if err != nil || len(resolv) == 0 {
+		resolv = []byte("nameserver 8.8.8.8\n")
```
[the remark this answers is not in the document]
henrybear327: Behavior kept, documented in 15fbc0b: usage.md now names the fallback nameserver (used only when the host resolver config is unreadable or empty) and its split-DNS implication.

cmd/elfuse-oci/unpack_test.go
```diff
@@ -0,0 +1,348 @@
...
+		t.Fatal(err)
+	}
+	if string(b) != "hello" {
+		t.Errorf("content: got %q, want hello", b)
+	}
+}
+
+// Ensure applyEntry reads exactly the header's Size bytes from the reader
+// (no over-read, no short read).
+func TestUnpackReadsExactSize(t *testing.T) {
```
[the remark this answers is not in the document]
henrybear327: Fixed in c75dd32: applyEntry bounds the copy to hdr.Size and errors on short bodies; the test reader carries trailing bytes and a short-read case.

cmd/elfuse-oci/run.go
```diff
@@ -0,0 +1,125 @@
...
```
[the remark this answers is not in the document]
henrybear327: Working as designed (duplicate of the earlier signal threads): documented tradeoff in the spawnElfuseWait comment.

src/main.c
```diff
@@ -611,122 +829,78 @@ int main(int argc, char **argv)
...
+     * elf_path / guest_argv above), so it must run before elfuse_launch hands
+     * control to the guest but after the shebang loop has fixed elf_path. The
+     * call only touches the original argv and the kernel procname; it does not
+     * depend on guest_bootstrap_prepare having run, so hoisting it out of the
+     * bring-up (where it previously sat between prepare and create_vcpu) is
+     * behavior-preserving.
      */
-    thread_wake_exit_waiters();
-    thread_join_workers();
+    runtime_set_process_title(argc, argv, elf_path);
```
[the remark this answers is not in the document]
henrybear327: Working as designed: the rewrite clobbers the argv block the launch args reference, so it must run after snapshotting and before elfuse_launch (documented at the call site); the old code also renamed before vCPU creation, and the failed-start window is cosmetic and brief.

henrybear327:
PR description also updated. Ready for another pass @jserv @Max042004 

Thanks.

.github/workflows/main.yml
```diff
@@ -580,6 +594,93 @@ jobs:
...
+          "$bin" pull "$IMG"
+          # No -q: grep must drain the pipe, or inspect dies of SIGPIPE
+          # (exit 141) under pipefail when grep exits at the first match.
+          "$bin" inspect "$IMG" | grep 'linux/arm64'
+          "$bin" list | grep -F "$IMG"
+
+          out="$("$bin" run --entrypoint /usr/local/bin/python3 "$IMG" \
+            -c 'import json,math; print(json.dumps({"pi":round(math.pi,5),"ok":True}))')"
+          printf 'guest said: %s\n' "$out"
+          [ "$out" = '{"pi": 3.14159, "ok": true}' ]
```
jserv: Validate with non-trivial applications.

jserv (changes requested):
Place copyright notice for each Go source file.

cmd/elfuse-oci/sparsebundle.go
```diff
@@ -0,0 +1,206 @@
...
+			return nil, fmt.Errorf("hdiutil create %s: %w: %s", image, err, out)
+		}
+	} else if err != nil {
+		return nil, err
+	}
+
+	// If a prior run left the volume attached (crash, kill), detach it first so
+	// we own a clean attach. If mountPath is not a mountpoint, ensure it is
+	// empty so hdiutil will mount onto it.
+	if isMountPointFn(mountPath) {
```
jserv, line 111: `provisionCaseSensitive` force-detaches any pre-existing mount without consulting `hasLiveClone`. A second `run` of the same digest sees the mount, calls `detachForce`, and rips the rootfs out from under a guest still executing out of a live COW clone -- the running guest hits I/O errors / crashes. Fix: check `hasLiveClone(mountPath)` before detaching (attach-and-share or fail with "image busy"), or hold a per-digest flock across the whole run so run/run and run/prune cannot interleave.

cmd/elfuse-container/csrun.go
```diff
@@ -0,0 +1,177 @@
...
+// an error on setup failure, and on a post-run cleanup failure after a guest
+// exit of zero; when the guest exits nonzero, the cleanup error is only
+// printed and the guest status still wins (os.Exit), so a guest failure code
+// is never masked by teardown.
+//
+// The clone lives in the same APFS volume as the base tree (clonefile is
+// intra-volume only), so it is instant and free until the guest writes (COW).
+// It isolates each run's mutations from the warm base, so re-runs stay clean.
+func runCaseSensitive(cf commonFlags, s *store, ref, digest string, cfg *v1.ConfigFile, rf runFlags, tail []string) error {
+	m, baseRootfs, err := ensureCaseSensitiveRootfsForRun(cf, s, ref, digest, rf.sparseSize)
```
jserv: The `run-<pid>-<ns>` liveness marker is created only after the mount is attached (here) and unpack runs. In that window the volume is mounted but has no `run-*` clone dir, so a concurrent `prune --cache --all` / `rmi --force` on the same digest finds no live clone, force-detaches the mount, and can remove the bundle mid-unpack -- the run's later `clonefile` then fails or writes into an unmounted path. Fix: create the liveness marker immediately after a successful attach, before any slow work (unpack/clone).

cmd/elfuse-container/clone_sweep.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+	}
+	return orphans
+}
+
+// reapOrphanClones removes run-<pid>-<ns> directories inside mountPath whose
+// creating pid is no longer alive. Clones of live processes are left in place.
+// Removal is best-effort: a busy entry (e.g. still unmounting) is skipped and
+// not reported. Returns the clone directories that were removed.
+func reapOrphanClones(mountPath string) []string {
```
jserv: Liveness is inferred purely from `kill(pid, 0)` on the pid embedded in the dir name. On a long-lived store, pid reuse/wraparound makes a still-wanted clone (e.g. a `--keep` run whose wrapper already exited) look dead, so `reapOrphanClones` `RemoveAll`s a clone the user asked to preserve. Fix: gate reaping on a per-clone advisory lock the live run holds `flock`'d for its lifetime -- a held lock means live regardless of pid; a free lock means reapable regardless of pid identity.

cmd/elfuse-container/unpack.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+	if err := os.MkdirAll(dest, 0o755); err != nil {
+		return err
+	}
+	// The run paths treat the rootfs path's existence as "fully unpacked", so
+	// a partial tree from a failed unpack must not survive: a later run would
+	// silently execute against it. Remove dest on failure -- but only when
+	// this call created it, so an explicit pre-existing --rootfs directory is
+	// never deleted.
+	defer func() {
```
jserv: Unpack writes in place and `defer os.RemoveAll(dest)` on failure. A concurrent `run` that already `os.Stat`'d the half-written tree as present (commands.go probe) proceeds to read `etc/passwd` etc. from a tree this failing unpack is simultaneously removing -- intermittent `resolveUser` failures / truncated reads. Fix: unpack into a temp sibling dir and atomically `rename` into the final cache path on success; never expose a partial tree under the name other processes probe.

cmd/elfuse-container/unpack.go
```diff
@@ -0,0 +1,303 @@
...
+	// Symlink mode is not portable to set across platforms; ignore mode.
+	_ = mode
+	return nil
+}
+
+// ensureParent creates name's missing parent directories with the 0o755
+// default. It never chmods: a parent that already exists may carry an exact
+// mode from its own tar entry, which must not be reset to the default here.
+func ensureParent(root *os.Root, name string) error {
+	return root.MkdirAll(filepath.Dir(name), 0o755)
```
jserv: `ensureParent`'s `MkdirAll` follows intermediate parent components that a lower layer planted as symlinks. If layer 1 ships `etc` as a symlink to `usr/etc` and layer 2 ships regular file `etc/passwd`, extraction writes `usr/etc/passwd` and leaves `/etc/passwd` unresolved -- containment holds (os.Root), but the file lands in the wrong directory and later `resolveUser` reads the wrong/missing passwd. Fix: before writing an entry, replace any symlinked/non-dir intermediate component with a real directory, matching containerd/Docker.

cmd/elfuse-container/unpack.go
```diff
@@ -0,0 +1,303 @@
...
+// to the corresponding syscall bits.
+func applyMode(root *os.Root, name string, perm, special os.FileMode) error {
+	return root.Chmod(name, perm|special)
+}
+
+// clearDirectory removes the existing children of dir (opaque whiteout),
+// keeping entries the current layer itself created: opaque markers hide
+// lower-layer content, and a marker ordered after its directory's same-layer
+// additions must not wipe them.
+func clearDirectory(root *os.Root, dir string, applied map[string]bool) error {
```
jserv: Opaque-whiteout handling does `Open`+`ReadDir` through the marker's directory without checking whether it is a symlink. If a lower layer shipped `data -> shared` and the upper layer ships `data/.wh..wh..opq`, the clear wipes `shared/`'s contents instead of `data/`'s -- silently deleting files the image author did not intend to whiteout. Fix: `Lstat` the dir first; if it is a symlink or non-dir, do not clear through it (replace it / skip), matching the non-dir handling elsewhere in this file.

cmd/elfuse-oci/etc.go
```diff
@@ -0,0 +1,104 @@
...
+// elfuse does not do network namespacing -- the guest uses the host network
+// directly -- so the host's resolver config is the correct one to hand it.
+//
+// Overwrite is intentional: these files are runtime-controlled, not image
+// content. The caller passes the final sysroot elfuse will receive as
+// --sysroot (the per-run COW clone on the case-sensitive path, or the plain
+// rootfs directory on the --plain-rootfs path), so writes are isolated to
+// this run except when the caller opted into mutating the base tree
+// (--no-clone / --plain-rootfs).
+func injectRuntimeFiles(sysroot string) error {
```
jserv, line 35: On the `--no-clone` / `--plain-rootfs` path `sysroot` is the shared warm base tree, and `writeRuntimeFile` does a non-atomic remove-then-`O_EXCL`-create with no serialization. Two concurrent `--no-clone` runs of the same digest can interleave one's `Remove` with the other's `OpenFile(O_EXCL)` -> spurious `EEXIST`/`ENOENT`, or a torn `resolv.conf`. Fix: write to a temp file in the same dir and `rename` into place so concurrent writers never observe a missing/half-written file.

cmd/elfuse-container/store.go
```diff
@@ -0,0 +1,263 @@
...
+		return "", err
+	}
+	unlock, err := s.lock()
+	if err != nil {
+		return "", err
+	}
+	defer unlock()
+	// If the image is already in the layout (re-pull of same digest), skip the
+	// append so the index doesn't accumulate duplicate descriptors.
+	if _, err := s.path.Image(h); err != nil {
```
jserv: `addImage` treats any `s.path.Image(h)` error as "not present" and proceeds to `AppendImage`. A corrupt existing descriptor or a missing blob is thus silently followed by an append, masking store corruption or duplicating descriptors. Fix: distinguish not-found from other errors -- append only on not-found, propagate the rest.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,263 @@
...
+	if err := json.Unmarshal(b, &p); err != nil {
+		return nil, fmt.Errorf("store: corrupt refs.json: %w", err)
+	}
+	if p == nil {
+		return nil, fmt.Errorf("store: corrupt refs.json: expected object")
+	}
+	return p, nil
+}
+
+func (s *store) savePins(p refPins) error {
```
jserv, line 153: `savePins` writes a temp file and `os.Rename`s it but never fsyncs the temp file or the containing directory, so the carefully-reasoned `rmi` last-pin crash ordering is not actually durable: a crash can leave `refs.json` reverted to an old pin whose descriptor `index.json` has already durably dropped, leaving `run`/`image` resolving a pin to a hash with no descriptor. Fix: `tmp.Sync()` before close and fsync the store directory after `Rename`.

src/runtime/procemu.c
```diff
@@ -2716,6 +2729,24 @@ int proc_intercept_open(const guest_t *g,
...
+        /* Linux /dev/full: read returns a NUL stream like /dev/zero, while any
+         * non-zero write must fail with ENOSPC. Back the FD with host /dev/zero
+         * so read and lseek work without extra plumbing; the proc_path tag set
+         * by resolve_virtual_path() routes writes through proc_intercept_write
+         * below for the ENOSPC short-circuit.
+         */
+        host_dev = "/dev/zero";
+        if (host_accmode == O_WRONLY)
+            host_accmode = O_RDWR;
+    } else if (!strcmp(path, "/dev/console")) {
```
jserv: `/dev/console` is unconditionally redirected to host `/dev/tty`, so `open("/dev/console", O_PATH)` (a pure path handle that touches no device) returns `ENXIO` on a session with no controlling terminal, while the synthesized stat unconditionally reports a valid `S_IFCHR` device. Result: `stat` succeeds but `open` fails, and `O_PATH` opens that Linux serves fine are rejected. Fix: for `O_PATH`, back the fd with a harmless handle (as the `/dev/ptmx` path does) rather than opening `/dev/tty`.

jserv (commented):
The dominant theme in the container-run path: `run` touches per-digest mount and clone state with no lock, while `prune`/`rmi`/other runs mutate the same state. A per-digest flock held for the lifetime of a run would close several of the inline findings at once.

Additional finding (in `src/syscall/io.c`, not part of this diff so it cannot be anchored inline): the new `/dev/full` node's ENOSPC-on-write is bypassed by the copy-path syscalls. `copy_fd_range` writes the output through the raw host fd and never routes through `proc_intercept_write`, so `sendfile`/`copy_file_range`/`splice` targeting a `/dev/full` guest fd write into the backing `/dev/zero` and report success where Linux returns -ENOSPC. Fix: thread the output guest fd into `copy_fd_range` and run write chunks tagged with a `proc_path` through `proc_intercept_write`, mirroring the read-side intercept.

Two items were considered and rejected: an apparent `AT_EMPTY_PATH` removal in `fs.c` (a diff-direction artifact -- the branch is behind main, which added that support; `main...HEAD` touches `fs.c` zero times), and a claimed uname-cache race (`cached_uname` is `static const`, resolved at compile time).

cmd/elfuse-oci/sparsebundle.go
henrybear327: Done via a per-digest flock: provision now reuses a mount a live run still holds (run.lock) and only detaches one proven stale. Never force-detaches under a live guest.
  (38bc1bf)

cmd/elfuse-container/csrun.go
henrybear327: The run now holds run.lock from before attach until exit, so sweeps skip it as busy; unpack also stages+renames atomically. (38bc1bf, 73a94ef)

cmd/elfuse-container/clone_sweep.go
henrybear327: Same idea, one per-digest run.lock instead of per-clone: held = live regardless of pid. --keep clones now carry a .elfuse-keep marker; all pid logic deleted. (73a94ef)

cmd/elfuse-container/unpack.go
henrybear327: Done: unpack into a temp sibling, os.Rename into place on success; the cache path  never shows a partial tree. (bc30db3)

cmd/elfuse-container/unpack.go
henrybear327: Fixed: parent components are Lstat'd and any symlink/non-dir replaced with a real dir, matching containerd/Docker. (bc30db3)

cmd/elfuse-container/unpack.go
henrybear327: Fixed: clearDirectory Lstats first and replaces a symlink/non-dir with an empty  real dir instead of clearing through it. (bc30db3)

cmd/elfuse-oci/etc.go
henrybear327: Fixed: write to a temp file + Root.Rename into place, so concurrent writers never  see a missing/half-written file. (bc30db3)

cmd/elfuse-container/store.go
henrybear327: Fixed: positive membership scan of the index; append only when genuinely absent, propagate read errors. (bc30db3)

cmd/elfuse-oci/store.go
henrybear327: Fixed: Sync() the temp file before close and fsync the store dir after rename.  (bc30db3)

src/runtime/procemu.c
henrybear327: Fixed: O_PATH now backs the fd with /dev/null like /dev/ptmx; stat/open stay consistent (5:1). Added a no-ctty test-opath case. (580ee72)

.github/workflows/main.yml
henrybear327: Added a python:3.12-slim sqlite+threads+fs workload and an alpine:3 gzip pipeline + COW-isolation check to the lifecycle job. (d37b116)

src/core/launch.h
```diff
@@ -0,0 +1,103 @@
...
+    bool gdb_stop_on_entry;
+    /* Per-iteration vCPU run timeout. 0 disables (no alarm()). */
+    int timeout_sec;
+    /* Fork-child IPC handles. -1 means "not a fork child". main()'s
+     * --fork-child dispatch handles the >= 0 case before reaching
+     * elfuse_launch. */
+    int fork_child_fd;
+    int vfork_notify_fd;
+    bool verbose;
+} launch_args_t;
```
jserv, line 88: Use blank lines to separate comments.

src/core/launch.h
```diff
@@ -0,0 +1,103 @@
...
+     * guest-visible entrypoint path. */
+    int guest_argc;
+    const char **guest_argv;
+    /* NULL-terminated guest environ. NULL means "use host environ". envp is
+     * char** (not const) to match the environ/guest_bootstrap_prepare
+     * convention: guest programs may mutate their environment. */
+    char **envp;
+    /* Override host uid/gid when true. Set by launchers that resolve an
+     * image User; main()'s legacy path leaves it false. */
+    bool has_creds;
```
jserv, line 70: Group similar members in a single line.

jserv (changes requested):
Improve code re-usability by introducing helpers.

src/core/launch.h
henrybear327: Refactored.

src/core/launch.h
henrybear327: Refactored.

henrybear327:
> Improve code re-usability by introducing helpers.

Attempted. Please do NOT review this PR further. I am currently breaking this PR down and sending out feature / improvement PRs that this one is based upon first.

henrybear327:
After several major iterations, I am not sure if `elfuse-container` is a good name, as I was also a bit misled by going down trying to support container execution. 

If I am not mistaken, the main idea is to help the user leverage the container images that is readily available to execute their linux binaries on. We do NOT care about a lot of the container features such as containment, etc. 

If the above statements hold, `elfuse-container` might send a signal to the user that we might be providing a container runtime, etc.

src/core/launch.c
```diff
@@ -0,0 +1,261 @@
...
+     */
+    int exit_code =
+        vcpu_run_loop(vcpu, vexit, &g, args->verbose, args->timeout_sec);
+
+    /* Tear down debugger state before joining workers: a worker parked in
+     * gdb_stub_handle_stop() stays active (not deactivated) until this
+     * broadcasts resume_cond, so joining first would just time out and
+     * detach it while it is still paused.
+     */
+    gdb_stub_shutdown();
```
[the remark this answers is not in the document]
henrybear327: Confirmed. `gdb_stub_shutdown` closed the sockets but never joined the listener thread, which could still be inside `guest_read` when `guest_destroy` unmaps the guest slab immediately afterward. Since the listener was never joined on `main` either, this is a pre-existing bug independent of this branch, so it is fixed in a separate PR: shutdown now calls `shutdown()` on both sockets (waking a listener blocked in accept/poll/recv) and joins the listener before returning. The listener never waits on a condition variable (only the vCPU threads do), so the join is bounded.

src/core/launch.h
```diff
@@ -0,0 +1,114 @@
...
+    /* NULL-terminated guest environ. NULL means "use host environ". envp is
+     * char** (not const) to match the environ/guest_bootstrap_prepare
+     * convention: guest programs may mutate their environment.
+     */
+    char **envp;
+
+    /* Override host uid/gid when true. Set by launchers that resolve an
+     * image User; main()'s legacy path leaves it false.
+     */
+    bool has_creds;
```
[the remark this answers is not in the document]
henrybear327: Confirmed mismatch. With `has_creds=false` the guest identity defaults to the fixed compile-time `GUEST_UID`/`GUEST_GID` (1000, or 0 under fakeroot), not the host uid/gid. Fixed the `launch.h` comment to document the fixed default and how a launcher opts into the host identity (pass `has_creds=true` with `getuid()`/`getgid()`); host-identity inheritance stays out of scope for this PR.

src/core/launch.h
```diff
@@ -0,0 +1,114 @@
...
+
+    /* Guest-absolute initial working directory. NULL inherits the host
+     * cwd (the caller may chdir first to control it).
+     */
+    const char *cwd_guest;
+
+    /* GDB Remote Serial Protocol port (0 disables the stub) and whether
+     * to halt before the first guest instruction.
+     */
+    int gdb_port;
```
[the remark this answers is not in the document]
henrybear327: Valid as defense in depth. `main()` is currently the only caller and already rejects `--gdb` for x86_64 guests via a static ELF probe, but the constraint belongs in `elfuse_launch` so every caller (including a future OCI run helper) inherits it. Added an `is_rosetta` gate right after bootstrap that rejects `gdb_port > 0`.

Max042004:
`unpack` fails on any image that ships setuid/setgid files when the host
process is unprivileged, which includes stock Debian-based images:

```
$ elfuse-container unpack --rootfs /tmp/gcc-rootfs gcc:14
elfuse-container: unpack: layer 0: entry "usr/bin/chage": chmodat usr/bin/chage: operation not permitted
```

`gcc:14`, `ubuntu:24.04`, and anything else carrying the shadow suite
(`chage`, `su`, `passwd`, ...) cannot be unpacked at all; the alpine/slim
images used by the CI smoke tests just happen to contain no setuid entries,
so the path is never exercised there. The `run` cache fill goes through the
same `applyMode`, so those images fail to run, too.

Root cause: `applyMode` applies the tar header mode verbatim via
`root.Chmod(name, perm|special)`. The unpacked tree is owned by the invoking
user rather than the tar's root/shadow owners, and an unprivileged chmod
that sets setuid/setgid on such files is rejected with EPERM on macOS, so
the whole unpack aborts on the first such entry.

Since elfuse remaps uid/gid and does not implement setuid-exec semantics,
the special bits carry no meaning at runtime today. I'd suggest degrading
gracefully instead of failing the unpack — retry with the plain permission
bits, and only for this specific case so genuine chmod failures still
surface

jserv:
> After several major iterations, I am not sure if `elfuse-container` is a good name, as I was also a bit misled by going down trying to support container execution.

Alternatively, name it `elfuse-oci`.

henrybear327:
> `unpack` fails on any image that ships setuid/setgid files when the host process is unprivileged, which includes stock Debian-based images:
> 
> ```
> $ elfuse-container unpack --rootfs /tmp/gcc-rootfs gcc:14
> elfuse-container: unpack: layer 0: entry "usr/bin/chage": chmodat usr/bin/chage: operation not permitted
> ```
> 
> `gcc:14`, `ubuntu:24.04`, and anything else carrying the shadow suite (`chage`, `su`, `passwd`, ...) cannot be unpacked at all; the alpine/slim images used by the CI smoke tests just happen to contain no setuid entries, so the path is never exercised there. The `run` cache fill goes through the same `applyMode`, so those images fail to run, too.
> 
> Root cause: `applyMode` applies the tar header mode verbatim via `root.Chmod(name, perm|special)`. The unpacked tree is owned by the invoking user rather than the tar's root/shadow owners, and an unprivileged chmod that sets setuid/setgid on such files is rejected with EPERM on macOS, so the whole unpack aborts on the first such entry.
> 
> Since elfuse remaps uid/gid and does not implement setuid-exec semantics, the special bits carry no meaning at runtime today. I'd suggest degrading gracefully instead of failing the unpack — retry with the plain permission bits, and only for this specific case so genuine chmod failures still surface

Thanks for testing and reporting the bug. I will address the issue by emitting an error message such that we are not silently dropping special bits without the user knowing about it.

henrybear327:
> > After several major iterations, I am not sure if `elfuse-container` is a good name, as I was also a bit misled by going down trying to support container execution.
> 
> Alternatively, name it `elfuse-oci`.

Renamed accordingly and updated the PR description.

henrybear327:
> > `unpack` fails on any image that ships setuid/setgid files when the host process is unprivileged, which includes stock Debian-based images:
> > ```
> > $ elfuse-container unpack --rootfs /tmp/gcc-rootfs gcc:14
> > elfuse-container: unpack: layer 0: entry "usr/bin/chage": chmodat usr/bin/chage: operation not permitted
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `gcc:14`, `ubuntu:24.04`, and anything else carrying the shadow suite (`chage`, `su`, `passwd`, ...) cannot be unpacked at all; the alpine/slim images used by the CI smoke tests just happen to contain no setuid entries, so the path is never exercised there. The `run` cache fill goes through the same `applyMode`, so those images fail to run, too.
> > Root cause: `applyMode` applies the tar header mode verbatim via `root.Chmod(name, perm|special)`. The unpacked tree is owned by the invoking user rather than the tar's root/shadow owners, and an unprivileged chmod that sets setuid/setgid on such files is rejected with EPERM on macOS, so the whole unpack aborts on the first such entry.
> > Since elfuse remaps uid/gid and does not implement setuid-exec semantics, the special bits carry no meaning at runtime today. I'd suggest degrading gracefully instead of failing the unpack — retry with the plain permission bits, and only for this specific case so genuine chmod failures still surface
> 
> Thanks for testing and reporting the bug. I will address the issue by emitting an error message such that we are not silently dropping special bits without the user knowing about it.

Addressed. 

Also added 5 workload images listed in https://github.com/sysprog21/elfuse/issues/224 to the CI as a test. 

Will make the CI tasks simpler for now, as the Go test currently failed due to it hitting elfuse's fixed thread-table cap, and the python test triggers an unhandled guest CPU exception in elfuse.

src/syscall/proc-identity.c
```diff
@@ -55,6 +66,16 @@ void proc_identity_init(void)
         gid = 0;
     }
 
+    /* An explicit --user request wins over the defaults and over fakeroot.
+     * It is staged before init rather than applied afterwards because
+     * build_linux_stack snapshots these values into auxv AT_UID/AT_GID; a
+     * post-init override would leave getauxval() disagreeing with getuid().
+     */
+    if (atomic_load(&initial_ids_staged)) {
```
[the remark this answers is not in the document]
henrybear327: `elfuse_launch` runs once per host process today, so it is not a reachable bug in the current use case.

But the never-cleared flag would leak a prior `--user` into a later credential-less bring-up once a multi-launch caller (the OCI run helper) exists. Adopted the atomic_exchange consume so the staged ids apply only to the bring-up they were staged for, folded into the commit that introduced `--user` staging.

henrybear327:
Depending on https://github.com/sysprog21/elfuse/pull/242 so the socket opening and closing can be functioning correctly.

cmd/elfuse-oci/gc.go
```diff
@@ -0,0 +1,415 @@
...
+
+// liveBlobDigests roots reachability at the refs.json pin set, not at every
+// index.json descriptor: an unpinned descriptor (a re-pulled tag's orphaned
+// prior manifest, or a pull that appended a manifest but crashed before
+// pinning) must not keep its blobs live. Each pinned digest's manifest,
+// config, and layers are marked live. A real gc calls pruneUnpinnedDescriptors
+// first so index.json never keeps referencing a manifest whose blobs this
+// sweep reclaims; a dry run skips that but, rooting here at pins too, still
+// reports exactly the blobs a real run would reclaim.
+func (s *store) liveBlobDigests() (map[string]bool, error) {
```
jserv: Blob GC can delete blobs a live run still needs. `liveBlobDigests` roots reachability only at current `refs.json` pins. `resolveImageForUse` drops the store lock once it holds the per-digest rootfs run-lock, so a concurrent `pull` can repin the ref to a new digest; a following `prune`/`rmi` blob sweep then removes the old digest's manifest/config/layers while a resolved-but-not-yet-unpacked run is still reading them. The cache sweep already skips busy caches via `rootfsCacheBusy` (gc.go:367), but this blob reachability pass does not. Also root liveness at digests whose rootfs run-lock is busy, the same way the cache sweep does, and add a test that combines a repull with `prune` while a run-lock is held.

cmd/elfuse-oci/unpack.go
```diff
@@ -0,0 +1,485 @@
...
+// Containment uses os.OpenRoot (Go 1.24+): every write is resolved relative
+// to the rootfs and may not escape via ".." or a symlink. os.Root forbids
+// absolute symlinks, so absolute symlink targets are rewritten to their
+// equivalent relative form, which is behavior-preserving because under
+// elfuse's --sysroot both forms resolve to the same guest path.
+//
+// Ownership is not applied here: elfuse runs as the host user and overrides
+// identity at runtime via --user, so the rootfs carries only mode bits.
+func unpackImage(img v1.Image, dest string) error {
+	if _, statErr := os.Lstat(dest); statErr == nil {
```
jserv, line 76: A symlink can redirect extraction (and execution) out of the store. `os.Lstat(dest)` treats any pre-existing entry as a mergeable rootfs and `unpackInto` then calls `os.OpenRoot(dest)`, which follows a symlinked `dest`; the run path's `os.Stat(rf.rootfs)` (commands.go:250) follows it too. A symlink planted at the digest-keyed cache path `<store>/rootfs/sha256/<hex>` redirects extraction writes, or `run --plain-rootfs` execution, to an arbitrary directory. For store-managed paths, reject a `dest` that is not a real directory (i.e. a symlink) before `OpenRoot` and the cache-exists check; keep symlink-following only for an explicit `--rootfs` if that is intended.

src/syscall/net-absock.c
```diff
@@ -196,6 +216,208 @@ static int absock_build_sun(const char *fs_path,
...
+    /* Undo the over-length shortening symlink, then map the host path back
+     * to the guest namespace so the guest reads back the spelling it bound
+     * or connected with, not the sysroot-prefixed (and possibly
+     * token-bearing) host path.
+     */
+    char host_path[LINUX_PATH_MAX];
+    str_copy_trunc(host_path, mac_path, sizeof(host_path));
+    pthread_mutex_lock(&absock_lock);
+    bool in_absock_dir = absock_dir_created &&
+                         !strncmp(host_path, absock_dir, strlen(absock_dir));
```
jserv: Prefix match lacks a path separator. `strncmp(host_path, absock_dir, strlen(absock_dir))` has no trailing `/`, so a sibling directory whose PID is a numeric prefix (`/tmp/elfuse-absock-123` vs `/tmp/elfuse-absock-12345`) false-matches and triggers an errant `readlink` on another process's path. Require the byte after the prefix to be a separator: `host_path[strlen(absock_dir)] == '/'`.

jserv (commented):
Review of the OCI pipeline and the C runtime boundary. Three findings are inline below. One more is worth flagging but sits in pre-existing code this PR does not modify, so it cannot be anchored to the diff:

`src/syscall/net-absock.c` abstract-socket bind slot reservation (pre-existing, adjacent to this PR's changes): `absock_register_locked` populates a table slot but leaves `active=false`, and `absock_lock` is dropped before the actual `bind()`; the slot is only marked live later in `absock_bind_commit`. A concurrent guest `bind()` scans for `!active`, reuses the same slot, and overwrites its `guest_fd`/name/`fs_path`, and the duplicate-name check (`absock_lookup_locked`, active-only) misses the in-flight name. Two guest fds then alias one slot, so unregister leaks one socket file and its reverse-map fails. Reserving the slot under the lock in `register_locked` (and clearing it in `absock_bind_rollback`) closes the window. Since this PR reworks the same file, it may be a good time to fix it.

.github/workflows/main.yml
```diff
@@ -7,7 +7,15 @@
 #   scan-macos    : LLVM scan-build via `make analyze`
 #   infer-macos   : Facebook Infer capture + analyze over the full build
 #   runtime-macos : HVF runtime tests on self-hosted Apple Silicon,
-#                   including release, ASAN, UBSAN, and TSAN variants
+#                   including release, ASAN, UBSAN, and TSAN variants,
+#                   plus the end-to-end OCI run and image-lifecycle checks
+#   oci-conformance : OCI image-layout conformance + cross-tool interop
+#                   (crane/skopeo/umoci) on Linux
+#   oci-macos     : elfuse-oci darwin build, unit tests, sparsebundle
```
jserv: `oci-macos` is ambiguous, as OCI instances built on macOS could also be abbreviated as oci-macos. Clarify.

sunxiaoguang:
Really looking forward to seeing this improvement. We will be able to run artifacts easier this way.

jserv (changes requested):

.github/workflows/main.yml
henrybear327: Updated to `oci-image-macos`

henrybear327:
PR ready for a final pass.

jserv (changes requested):
Rebase latest `main` branch as #272 introduces Frama-C proofs.

henrybear327:
As discussed offline, I am splitting this PR into several reviewable ones. Closing in favor of the broken down PRs

PR #193, opened 2026-07-11
Route sys_statfs through the /proc intercept

Max042004:
/proc entries are synthesized on open with no backing file under
the sysroot (see proc_intercept_open), so a raw statfs() on any
/proc path -- including the literal "/proc" -- fails with ENOENT even
though stat() and open() on the same path succeed. Confirmed by
cross-compiling a guest probe and running it under elfuse:
stat("/proc") succeeded while statfs("/proc") returned ENOENT.

Fix sys_statfs to route /proc paths through proc_intercept_open,
fstatfs the resulting host fd, and close it, mirroring how sys_newfstatat
already resolves /proc via proc_intercept_stat. The prefix check is
boundary-safe (rejects "/procfoo") to avoid reintroducing the naive-match
class of bug the issue warns about.

While reproducing this, found that proc_intercept_open's bare "/proc"
case lacked the trailing-slash variant present on all its children (e.g. "/proc/self/"),
so statfs("/proc/") missed even after the fix above; stat() already handled it
via proc_intercept_stat's separate check. Added the missing variant for consistency.

Add regression coverage to test-proc.c: statfs on "/proc", "/proc/", and a synthetic
file, plus an f_type cross-check against fstatfs on an already-open fd of the same path.

---
## Summary by cubic
Fix ENOENT from statfs on /proc paths by routing them through the /proc intercept so statfs matches stat/open, and avoid scratch-dir allocation for `/proc/self/{fd,fdinfo}`. Also handle the `/proc/` trailing-slash case and extend tests.

- **Bug Fixes**
  - sys_statfs: boundary-check `/proc` paths; try `proc_intercept_statfs` for `/proc/self/{fd,fdinfo}` (statfs `/tmp` to avoid per-open scratch), else use `proc_intercept_open` → `fstatfs`; fall back to host `statfs` if not intercepted.
  - `proc_intercept_open`: accept `"/proc/"` to match `"/proc"`.
  - Tests: `statfs("/proc")`, `statfs("/proc/")`, `statfs("/proc/self/cmdline")`; confirm `f_type` matches `fstatfs`; repeat `statfs` on `/proc/self/fd` and `/proc/self/fdinfo` past the tracking cap.

<sup>Written for commit 07955929ee0005a7394ae24eb209f0298654ab0c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/193?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/test-proc.c
+++ b/tests/test-proc.c
@@ -7,10 +7,11 @@
  *
  * Tests: /proc/self/cmdline, /proc/meminfo, /proc/stat, /proc/version,
  *        /proc/filesystems, /proc/mounts, readlink(/proc/self/exe),
+ *        statfs(/proc, /proc/self/cmdline, /proc/self/fd, /proc/self/fdinfo),
  *        /dev/null, /dev/zero, /dev/urandom
  *
  * Syscalls exercised: openat(56), read(63), write(64), readlinkat(78),
- *                     close(57)
+ *                     close(57), statfs(43), fstatfs(44)
  */
 
 #include <stdlib.h>
@@ -21,6 +22,7 @@
 #include <dirent.h>
 #include <errno.h>
 #include <sys/stat.h>
+#include <sys/statfs.h>
 
 #include "test-harness.h"
 #include "test-util.h"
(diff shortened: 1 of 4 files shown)


tests/test-proc.c
```diff
@@ -199,6 +201,83 @@ int main(void)
...
+
+    /* /proc/self/fd and /proc/self/fdinfo are directories whose *open* path
+     * allocates a fresh scratch directory with one placeholder file per live
+     * guest fd (see proc_open_fd_scratch) -- work a plain statfs() has no use
+     * for. These exercise the statfs-specific bypass (proc_intercept_statfs)
+     * added alongside issue #141's fix, including enough repeats to exceed
+     * PROC_SCRATCH_DIRS_MAX (128): pre-fix, that many statfs() calls would
+     * have leaked untracked scratch dirs in /tmp.
+     */
+    TEST("statfs /proc/self/fd");
```
jserv, line 263: The leak-regression loops don't actually test the leak. These `statfs /proc/self/fd` and `/proc/self/fdinfo` 200-iteration loops assert only that statfs() returns >= 0. If proc_intercept_statfs() were reverted (keeping the ENOENT fix), sys_statfs would fall back to proc_intercept_open -> proc_open_fd_scratch -> fstatfs -> close, which still returns 0 on every call, so all iterations still pass. The loop exercises the leak path but cannot fail on it, and the comment's claim ("pre-fix, that many statfs() calls would have leaked untracked scratch dirs") is unverified.

Caveat on any fix: the test runs inside the guest namespace, so it can't directly count host /tmp scratch dirs, and an f_type check can't distinguish the shortcut from the fallback (both live under /tmp). The realistic options are to trim the comment's leak claim to match what is actually tested (success only), or add a test-only hook exposing the scratch-dir count.

jserv (commented):
Automated review of the /proc statfs routing. The production code validated clean across all the angles checked: the open/fstatfs/close fallback leaks no scratch state (the only per-open-allocating nodes, /proc/self/fd and /proc/self/fdinfo, are shortcut before it and the fd is closed on every branch); statfs_path_is_proc runs on intercept_path while the host fallback uses host_path, so a sysroot whose real path contains /proc cannot misroute and /procfoo is rejected; proc_intercept_statfs recursion after proc_alias_self is bounded to depth 1; fstatfs is valid on both dir- and file-style synthetic fds; close_keep_errno preserves errno; and the /proc/ trailing-slash addition is genuinely reachable, not dead code. One test-coverage gap noted inline.

PR #194, opened 2026-07-11
Support AT_EMPTY_PATH in fchownat/fchmodat2

### LINKED ISSUE #128: AT_EMPTY_PATH with AT_FDCWD fails in sys_fchmodat and sys_fchownat
In Linux `*at` family system calls (such as `fchmodat`, `fchownat`, `openat`, `unlinkat`), path resolution requires a starting directory:
* **As a Sentinel:** `LINUX_AT_FDCWD` (typically `-100` on Linux) is a special sentinel value passed as the `dirfd` argument. 
* **Meaning:** It indicates that the path argument (if relative) should be resolved relative to the **Current Working Directory (CWD)** of the process, rather than relative to an open directory file descriptor.
* **Example:** If a program calls `fchmodat(LINUX_AT_FDCWD, "config.txt", mode, 0)`, the OS will look for `config.txt` in the process's current working directory.

### **AT_EMPTY_PATH with AT_FDCWD fails in sys_fchmodat and sys_fchownat**
* Location:  `src/syscall/fs.c:L2122-2126` and `fs.c:L2199-2203`
* Code
```
if (path[0] == '\0') {
    if (!(flags & LINUX_AT_EMPTY_PATH))
        return -LINUX_ENOENT;
    return sys_fchmod(dirfd, mode);
}
```
* Issue:
If `dirfd` is `LINUX_AT_FDCWD` (`-100`), the call delegates directly to `sys_fchmod(LINUX_AT_FDCWD, mode)`. `sys_fchmod` attempts to look up `LINUX_AT_FDCWD` in the guest descriptor table (fd_table), which fails and returns `-LINUX_EBADF`.

* Correct Behavior:
On Linux, `fchmodat(AT_FDCWD, "", mode, AT_EMPTY_PATH)` is a valid call that performs a chmod on the current working directory (equivalent to chmod(".", mode)). Passing `AT_FDCWD` should be handled appropriately instead of calling sys_fchmod directly.

Max042004:
fchownat() and fchmodat2() only accepted AT_SYMLINK_NOFOLLOW, so
*at(fd, "", ..., AT_EMPTY_PATH) returned EINVAL instead of operating on
fd itself. That is the only way to chown or chmod an O_PATH descriptor
(plain fchown()/fchmod() reject FD_PATH with EBADF, matching Linux), so
unlike the plain EBADF guard this flag combination needs real handling,
not just acceptance.

An empty path with AT_EMPTY_PATH now operates on dirfd directly, or the
current directory when dirfd is AT_FDCWD, mirroring the identical
AT_EMPTY_PATH branch already in fs-stat.c's stat_at_path.

Close #128

---
## Summary by cubic
Adds full `AT_EMPTY_PATH` support to fchownat and fchmodat2 so an empty path targets the dirfd (or cwd for `AT_FDCWD`), enabling chown/chmod via `O_PATH` and matching Linux/stat_at_path. Also fixes missing FUSE/proc interception and a cwd race in fchownat.

- **Bug Fixes**
  - fchownat: empty path + `AT_EMPTY_PATH` now chowns dirfd or "." (`AT_FDCWD`); returns `ENOSYS` for FUSE dirfds and `EPERM` for synthesized `/proc`; avoids a race by opening "." once for chown+stat; keep `fchown(O_PATH)` returning `EBADF`.
  - fchmodat2: empty path + `AT_EMPTY_PATH` now chmods dirfd or "." (`AT_FDCWD`); returns `ENOSYS` for FUSE dirfds and `EPERM` for synthesized `/proc`; keep `fchmod(O_PATH)` returning `EBADF`.
  - Tests: add `test-fchownat-empty-path` and `test-fchmodat-empty-path` (uses raw `fchmodat2` syscall; skips if `ENOSYS`); wire into `tests/test-matrix.sh` and update baselines.

<sup>Written for commit 74513b91f4669c4fda0fcf61bf38b36fb4fa2874. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/194?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/fs.c
+++ b/src/syscall/fs.c
@@ -2410,16 +2410,59 @@ int64_t sys_fchmodat(guest_t *g,
                      int flags)
 {
     char path[LINUX_PATH_MAX];
-    if (!validate_at_flags(flags, LINUX_AT_SYMLINK_NOFOLLOW))
+    if (!validate_at_flags(flags,
+                           LINUX_AT_SYMLINK_NOFOLLOW | LINUX_AT_EMPTY_PATH))
         return -LINUX_EINVAL;
+    if (guest_read_str(g, path_gva, path, sizeof(path)) < 0)
+        return -LINUX_EFAULT;
+
+    /* AT_EMPTY_PATH with an empty path chmods dirfd itself; see the identical
+     * branch in sys_fchownat above for the O_PATH/AT_FDCWD rationale. Neither
+     * sub-case below goes through path_translate_at, so the FUSE and /proc
+     * interception normally applied by reject_unsupported_fuse_path_op and
+     * stat_at_path's proc_intercept_stat has to be checked by hand here.
+     */
+    if ((flags & LINUX_AT_EMPTY_PATH) && path[0] == '\0') {
+        if (dirfd == LINUX_AT_FDCWD) {
+            char fuse_buf[LINUX_PATH_MAX];
+            int fuse_rc = fuse_resolve_at_path(LINUX_AT_FDCWD, ".", fuse_buf,
+                                               sizeof(fuse_buf));
+            if (fuse_rc < 0)
+                return linux_errno();
+            if (fuse_rc > 0)
+                return -LINUX_ENOSYS;
+            if (fchmodat(AT_FDCWD, ".", mode, translate_at_flags(flags)) < 0)
+                return linux_errno();
+            return 0;
+        }
+
+        fd_entry_t snap;
+        if (!fd_snapshot(dirfd, &snap))
+            return -LINUX_EBADF;
+        if (snap.type == FD_FUSE_DEV || snap.type == FD_FUSE_FILE ||
+            snap.type == FD_FUSE_DIR)
+            return -LINUX_ENOSYS;
+        if (snap.type == FD_PATH && snap.proc_path[0] != '\0')
+            return -LINUX_EPERM;
+
+        host_fd_ref_t ref;
+        if (host_dirfd_ref_open(dirfd, &ref) < 0)
+            return -LINUX_EBADF;
+        if (fchmod(ref.fd, mode) < 0) {
+            host_fd_ref_close(&ref);
+            return linux_errno();
+        }
+        host_fd_ref_close(&ref);
+        return 0;
+    }
+
     path_translation_t tx;
-    int64_t rc = read_translated_path(
-        g, dirfd, path_gva,
-        (flags & LINUX_AT_SYMLINK_NOFOLLOW) ? PATH_TR_NOFOLLOW : PATH_TR_NONE,
-        path, &tx);
-    if (rc < 0)
-        return rc;
-    rc = reject_unsupported_fuse_path_op(&tx);
+    if (path_translate_at(dirfd, path,
+                          (flags & LINUX_AT_SYMLINK_NOFOLLOW) ? PATH_TR_NOFOLLOW
+                                                              : PATH_TR_NONE,
+                          &tx) < 0)
+        return linux_errno();
+    int64_t rc = reject_unsupported_fuse_path_op(&tx);
     if (rc != INT64_MIN)
         return rc;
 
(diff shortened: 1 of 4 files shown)


jserv (changes requested):
Append `Close #128` at the end of git commit messages.

src/syscall/fs.c
```diff
@@ -2410,16 +2410,40 @@ int64_t sys_fchmodat(guest_t *g,
...
+    if (!validate_at_flags(flags,
+                           LINUX_AT_SYMLINK_NOFOLLOW | LINUX_AT_EMPTY_PATH))
         return -LINUX_EINVAL;
+    if (guest_read_str(g, path_gva, path, sizeof(path)) < 0)
+        return -LINUX_EFAULT;
+
+    /* AT_EMPTY_PATH with an empty path chmods dirfd itself; see the identical
+     * branch in sys_fchownat above for the O_PATH/AT_FDCWD rationale.
+     */
+    if ((flags & LINUX_AT_EMPTY_PATH) && path[0] == '\0') {
```
jserv, line 2425: Empty-path branch bypasses FUSE interception (mirror at :2519 in fchownat). It returns before path_translate_at + reject_unsupported_fuse_path_op, so it diverges from the named path in two ways:

- A `dirfd` inside a FUSE mount has `host_fd == -1` (fuse.c:1893), so `host_dirfd_ref_open` returns EBADF here, whereas the named path returns ENOSYS.
- If cwd is inside a FUSE mount, the AT_FDCWD sub-case calls host `fchmodat(".", ...)` directly, mutating the host backing path instead of returning ENOSYS.

Suggested fix: snapshot `dirfd` first and return `-LINUX_ENOSYS` for `FD_FUSE_*` to match reject_unsupported_fuse_path_op.

src/syscall/fs.c
```diff
@@ -2410,16 +2410,40 @@ int64_t sys_fchmodat(guest_t *g,
...
+        if (dirfd == LINUX_AT_FDCWD) {
+            if (fchmodat(AT_FDCWD, ".", mode, translate_at_flags(flags)) < 0)
+                return linux_errno();
+            return 0;
+        }
+
+        host_fd_ref_t ref;
+        if (host_dirfd_ref_open(dirfd, &ref) < 0)
```
jserv, line 2449: Proc-intercept divergence from the mirrored stat_at_path (mirror at :2533 in fchownat). stat_at_path checks `snap.type == FD_PATH && snap.proc_path[0]` and routes to proc_intercept_stat; this branch skips that and fchmod/fchowns the backing host temp fd directly.

A guest that opens a synthesized /proc path O_PATH then calls `fchownat(fd, "", ..., AT_EMPTY_PATH)` silently mutates elfuse's internal temp file and reports success (Linux would return EPERM). chown_result's EPERM-to-fakeroot-success mapping hides it further.

Suggested fix: snapshot the fd before host_dirfd_ref_open; for FD_PATH with proc_path set, return EPERM instead of touching the backing file.

src/syscall/fs.c
```diff
@@ -2479,16 +2503,54 @@ int64_t sys_fchownat(guest_t *g,
...
+     * beneath it. This is the only way to chown an O_PATH fd (plain fchown()
+     * rejects FD_PATH, matching Linux's EBADF there), so unlike the other
+     * *at() flag validation here it has to actually be handled, not just
+     * accepted. dirfd == AT_FDCWD resolves to the current directory, mirroring
+     * stat_at_path's identical AT_EMPTY_PATH branch in fs-stat.c.
+     */
+    if ((flags & LINUX_AT_EMPTY_PATH) && path[0] == '\0') {
+        int mac_flags = translate_at_flags(flags);
+        if (dirfd == LINUX_AT_FDCWD) {
+            int host_rc = fchownat(AT_FDCWD, ".", owner, group, mac_flags);
```
jserv: TOCTOU on the AT_FDCWD overlay update. `fchownat(AT_FDCWD, ".")` here and the follow-up `fstatat(AT_FDCWD, ".")` at :2526 resolve "." twice. A concurrent chdir() in another guest thread between them records the requested uid/gid overlay against a different directory's dev/ino. The dirfd sub-case is immune because it fchown+fstat the same fd.

Suggested fix: open "." once and fchown(fd)+fstat(fd) on that single descriptor, like the dirfd branch.

tests/test-fchmodat-empty-path.c
```diff
@@ -0,0 +1,166 @@
...
+#ifndef AT_EMPTY_PATH
+#define AT_EMPTY_PATH 0x1000
+#endif
+#ifndef SYS_fchmodat2
+#define SYS_fchmodat2 452
+#endif
+
+static int fchmodat2(int dirfd, const char *path, mode_t mode, int flags)
+{
+    return (int) syscall(SYS_fchmodat2, dirfd, path, (long) mode, (long) flags);
```
jserv, line 40: No ENOSYS skip guard for fchmodat2. This issues raw `fchmodat2(2)` (SYS 452, Linux 6.6+) with no fallback. On a qemu fixture running a <6.6 kernel it fails hard rather than skipping. Worth confirming the fixture kernel is >= 6.6 (the count bump to 217 assumes it).

jserv (commented):
Automated review of the AT_EMPTY_PATH handling. The common thread: the new empty-path branch returns before the normal path's reject_unsupported_fuse_path_op() / interception layer, so it treats some fd classes differently from every other *at op. errno preservation, translate_at_flags dropping AT_EMPTY_PATH, the O_PATH handling, sysroot parity, leaks, and the test-matrix count math all check out.

src/syscall/fs.c
Max042004: Fixed. `dirfd` is now snapshotted (`fd_snapshot`) before `host_dirfd_ref_open`: `FD_FUSE_DEV`/`FD_FUSE_FILE`/`FD_FUSE_DIR` return `-ENOSYS`, matching `reject_unsupported_fuse_path_op`'s behavior for the named-path case. For the `AT_FDCWD` sub-case, `fuse_resolve_at_path(LINUX_AT_FDCWD, ".", ...)` now checks whether cwd is inside a FUSE mount before touching the host path (same helper `dirfd_guest_base_path` in path.c already uses for this). Applied identically to the mirrored `fchownat` branch.

src/syscall/fs.c
Max042004: Fixed. Same `fd_snapshot` up front: `FD_PATH` with `proc_path` set now returns `-EPERM` instead of touching the backing file, mirroring `stat_at_path`'s check.

Verified the divergence and the fix with a standalone probe: pre-fix, `fchmodat2(procfd_of_/proc/self/fd, "", 0700, AT_EMPTY_PATH)` returned 0 and silently mutated elfuse's synthesized backing directory; post-fix it returns `-EPERM`.

I did try adding this as a permanent regression test, but pulled it back out — under the qemu lane (real Alpine kernel), root `fchownat` on a real procfs dentry via the same call actually succeeds (`simple_setattr`-style, no fs-specific restriction), so an EPERM assertion there would be asserting elfuse-only behavior against a real kernel and breaks qemu-aarch64 parity testing. Since this specific check protects an elfuse-internal invariant (the temp-file backing) rather than Linux-compatible behavior, I decided the code fix + the manual probe was better than a flaky/misleading cross-platform assertion. Happy to add it gated some other way if you'd prefer.

src/syscall/fs.c
Max042004: Fixed. The AT_FDCWD branch now opens "." once (`open(".", O_RDONLY)`) and does `fchown(fd)` + `fstat(fd)` on that single descriptor, same as the dirfd branch, instead of resolving "." twice via `fchownat` then `fstatat`.

tests/test-fchmodat-empty-path.c
Max042004: Fixed. Added an `fchmodat2_supported()` probe at the start of `main()` (via the same raw syscall, against the already-created fixture file with a no-op mode); if it returns `ENOSYS` the test now prints a skip message and returns 0 instead of running the rest and hard-failing.

For the record, the qemu lane's fixture is Alpine 3.21's `linux-virt` kernel, which is currently 6.12.x (>= 6.6, so fchmodat2 is present and the 217 count holds) — but it's fetched live and unpinned in-repo, so this guards against a future/older kernel there without needing to touch the count.

PR #200, opened 2026-07-13
chown-overlay fix ID-space mismatch in set and apply guards

### LINKED ISSUE #136: Guest User Cannot Own Files when Host UID Equals Guest UID (`GUEST_UID = 1000`)
### Description
In [chown-overlay.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/chown-overlay.c), the function [chown_overlay_apply](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/chown-overlay.c#L158-L180) maps files owned by the host user (`getuid()`) to guest root (`0`) by default:
```c
if (st->st_uid == getuid())
    st->st_uid = 0;
```
However, in [chown_overlay_set](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/chown-overlay.c#L96-L141), early exit and entry removal checks use the raw host IDs (`cur_uid`/`cur_gid`) instead of the virtualized guest IDs:
```c
if (!e) {
    uint32_t next_uid = new_owner == (uint32_t) -1 ? cur_uid : new_owner;
    ...
    if (next_uid == cur_uid && next_gid == cur_gid)
        goto out; // Exits without creating an entry
}
...
if (e->uid == cur_uid && e->gid == cur_gid)
    remove_locked(dev, ino); // Removes existing entry
```

If the host user running `elfuse` has a UID of `1000` (which is standard on most Linux systems and matches `GUEST_UID`), `getuid()` returns `1000`. When the guest tries to `chown` a file to UID `1000`:
1. `new_owner` is `1000`.
2. `cur_uid` (host owner) is `1000`.
3. Since `new_owner == cur_uid`, the function exits early without storing an overlay entry.
4. Subsequent `stat` calls fall back to the default rule: `if (st->st_uid == getuid()) st->st_uid = 0;`.
5. The guest sees the file as owned by `0` (root).

### Impact
Any guest process running as `GUEST_UID` (1000) will be unable to own any files. Files they explicitly `chown` to themselves will appear as owned by root (`0`), breaking standard applications and file managers.

doanbaotrung:
chown-overlay: fix ID-space mismatch in set/apply guards
chown_overlay_set receives cur_uid/cur_gid from chown_result where
they were taken directly from host_st->st_uid/st_gid (the raw macOS
stat values).  new_owner/new_group are guest-virtual UIDs/GIDs.
These two live in different ID namespaces, so comparing them:

  next_uid = new_owner == -1 ? cur_uid : new_owner;
  if (next_uid == cur_uid && next_gid == cur_gid) goto out;

produces wrong answers whenever host UID != GUEST_UID (e.g. host
user has UID 501, guest runs as GUEST_UID 1000), or whenever the
file already has an overlay entry that virtualises its owner to a
different value.

Two complementary fixes:

fs.c - chown_result:
  Apply the existing overlay to a temporary stat copy before
  extracting cur_uid/cur_gid.  This makes the values passed to
  chown_overlay_set reflect what the guest currently sees for the
  file rather than the physical host owner.

chown-overlay.c - chown_overlay_set / chown_overlay_apply:
  Initialise new overlay entries with the sentinel (uint32_t)-1
  ("no override for this field") instead of cur_uid/cur_gid.
  Resolve sentinels to cur before the stale-entry removal check
  so both sides of the comparison use guest-virtual IDs.
  Skip sentinel fields in chown_overlay_apply so a partial chown
  does not corrupt the untouched field with a stale value.

fix #136

---
## Summary by cubic
Fixes ID-space mismatches in the chown overlay by evaluating guards in the guest-visible UID/GID space to avoid false no-ops and stale entries. Corrects ownership updates on macOS and with virtualized owners; fixes #136.

- **Bug Fixes**
  - fs.c: Apply the existing overlay to a temp stat to get guest-visible uid/gid before calling chown_overlay_set; return ENOMEM if overlay allocation fails to avoid reporting a state the guest can’t observe.
  - chown-overlay.c: Initialize new entries with sentinel (uint32_t)-1, resolve sentinels before no-op/stale-removal checks, and skip sentinel fields in apply so partial chown doesn’t overwrite untouched values.

<sup>Written for commit a124c8d59b81630855909df1223017c9ea98602f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/200?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/chown-overlay.c
+++ b/src/syscall/chown-overlay.c
@@ -117,6 +117,11 @@ int chown_overlay_set(uint64_t dev,
         }
         e->dev = dev;
         e->ino = ino;
+        /* Initialise with the sentinel (uint32_t)-1, meaning "no override
+         * recorded for this field yet".  The apply path skips sentinel
+         * fields, so a partial chown that only touches one of uid/gid does
+         * not corrupt the other field with a stale physical host value.
+         */
         e->uid = (uint32_t) -1;
         e->gid = (uint32_t) -1;
         unsigned int idx = bucket_index(dev, ino);
@@ -131,6 +136,10 @@ int chown_overlay_set(uint64_t dev,
     if (new_group != (uint32_t) -1)
         e->gid = new_group;
 
+    /* A sentinel field means "keep whatever cur reports".  Resolve before
+     * comparing so the no-op check is evaluated in the guest-virtual ID
+     * space on both sides.
+     */
     uint32_t effective_uid = e->uid == (uint32_t) -1 ? cur_uid : e->uid;
     uint32_t effective_gid = e->gid == (uint32_t) -1 ? cur_gid : e->gid;
     if (effective_uid == cur_uid && effective_gid == cur_gid)
@@ -166,6 +175,9 @@ void chown_overlay_apply(struct stat *st)
     overlay_entry_t *e =
         find_locked((uint64_t) st->st_dev, (uint64_t) st->st_ino);
     if (e) {
+        /* Skip sentinel fields: (uint32_t)-1 means "no override recorded"
+         * for that field, so leave the host value untouched.
+         */
         if (e->uid != (uint32_t) -1)
             st->st_uid = e->uid;
         if (e->gid != (uint32_t) -1)
--- a/src/syscall/fs.c
+++ b/src/syscall/fs.c
@@ -2458,15 +2458,31 @@ static int64_t chown_result(int host_rc,
 {
     if (host_rc < 0 && errno != EPERM)
         return linux_errno();
-    if (host_st && chown_overlay_set((uint64_t) host_st->st_dev,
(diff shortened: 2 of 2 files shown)


Max042004 (approved):

jserv (changes requested):
Rebase latest `main` and resolve conflicts.

doanbaotrung:
Rebased and fix conflicted

PR #201, opened 2026-07-13
Signal implement SYS_rt_sigtimedwait syscall 137

### LINKED ISSUE #196: Implement `SYS_rt_sigtimedwait` to support synchronous signal handling `sigwaitinfo`/`sigtimedwait`
**Issue: Implement `SYS_rt_sigtimedwait` to support synchronous signal handling (`sigwaitinfo`/`sigtimedwait`)**

#### **1. Summary**
In Linux guest applications running under `elfuse`, utilizing synchronous signal handling functions such as `sigwait()`, `sigwaitinfo()`, or `sigtimedwait()` results in failures or runtime crashes. This occurs because the underlying system call `SYS_rt_sigtimedwait` (syscall 179) is unimplemented and returns `-ENOSYS`.

#### **2. Why it is Needed**
In modern multi-threaded Linux applications, handling signals asynchronously via signal handlers (using `sigaction`) is often discouraged due to safety concerns (reentrancy limits in signal-handler contexts). Instead, the standard Linux design pattern is to block signals in all threads via `pthread_sigmask`, spawn a dedicated signal-management thread, and wait for signals synchronously using:
*   `sigwait(const sigset_t *set, int *sig)`
*   `sigwaitinfo(const sigset_t *set, siginfo_t *info)`
*   `sigtimedwait(const sigset_t *set, siginfo_t *info, const struct timespec *timeout)`

In standard Linux C libraries (both `glibc` and `musl`), all three of these APIs wrapper the `rt_sigtimedwait` system call. If `SYS_rt_sigtimedwait` is missing and returns `ENOSYS`:
- Applications attempting to handle daemon reload signals (e.g. `SIGHUP`), termination signals (e.g. `SIGTERM`, `SIGINT`), or worker synchronization signals synchronously fail to initialize their signal helper loops.
- POSIX-compliant multi-threaded software cannot coordinate signals correctly.

#### **3. Implementation Overview**
The system call can be emulated in `elfuse` by registering `SYS_rt_sigtimedwait` in `dispatch.tbl` and implementing the lookup logic in the signal subsystem:

*   **Signature**:
    ```c
    int64_t signal_rt_sigtimedwait(guest_t *g, uint64_t set_gva, uint64_t info_gva, uint64_t timeout_gva, uint64_t sigsetsize)
    ```
*   **Behavior**:
    1.  Read the target signal set (`set_gva`) and validate parameters (such as `sigsetsize == 8`).
    2.  Check if any signals in the requested set are currently pending.
    3.  If a signal is pending, consume it, populate the `siginfo_t` struct at `info_gva` (if provided), and return the signal number.
    4.  If no signal is pending:
        - If a timeout (`timeout_gva`) is provided, return `-EAGAIN` (or sleep for the specified duration and check again).
        - If no timeout is specified, block the guest thread (e.g., in a sleep loop) until a signal is delivered.

doanbaotrung:
Add rt_sigtimedwait so that sigwait(3), sigwaitinfo(3), and sigtimedwait(3) work inside the guest.  All three are thin wrappers over this syscall in musl and glibc; returning -ENOSYS breaks any application that uses a dedicated signal-management thread.

abi.h:
  Define SYS_rt_sigtimedwait = 137, which sits between
  rt_sigpending (136) and rt_sigqueueinfo (138) in the aarch64
  Linux syscall table.

signal.c:
  sigtimedwait_try_dequeue() - acquires sig_lock and dequeues one
    signal matching the caller's mask.  Thread-directed (private)
    set is drained before the shared set, mirroring Linux
    dequeue_signal() priority.  Reuses signal_rt_dequeue_locked()
    and signal_standard_peek_locked() from signal_deliver().
  signal_rt_sigtimedwait() - reads and validates the guest mask
    and optional timeout, then polls in 1 ms sleep chunks (same
    pattern as interruptible_sleep_ns in time.c) until:
      - a matching signal is consumed -> return signum
      - timeout expires             -> return -EAGAIN
      - exit_group requested        -> return -EINTR
      - unblocked non-waited signal -> return -EINTR
    SIGKILL/SIGSTOP are silently removed from the wait mask,
    matching Linux do_sigtimedwait behavior.
    NULL timeout_gva blocks indefinitely; {0,0} polls once.
    info_gva = 0 skips the siginfo_t write (sigwait(3) path).

signal.h:     declare signal_rt_sigtimedwait.
dispatch.tbl: register sc_rt_sigtimedwait with guest-ptr flag.
syscall.c:    SC_FORWARD x0..x3 -> set, info, timeout, sigsetsize.

Fix #196

---
## Summary by cubic
Implements `SYS_rt_sigtimedwait` (137) so `sigwait(3)`, `sigwaitinfo(3)`, and `sigtimedwait(3)` work inside the guest, matching Linux behavior and unblocking apps that use a signal-manager thread.

- **New Features**
  - Wires up `SYS_rt_sigtimedwait`: defines 137, registers `sc_rt_sigtimedwait`, forwards args in `syscall.c`, and updates coverage mapping (`sigwait`, `sigwaitinfo`, `sigtimedwait`).
  - Consumes one pending signal (thread-directed before process-directed); optionally writes `siginfo_t`; removes `SIGKILL`/`SIGSTOP` from the wait mask; `NULL` timeout blocks, `{0,0}` polls once, finite waits in 1 ms chunks; returns signum on success, `-EAGAIN` on timeout, `-EINTR` on exit-group or an unblocked deliverable unrelated signal.
  - Adds `tests/test-sigtimedwait.c` and a Makefile target (links `-lpthread`) covering `sigwait`, `sigwaitinfo` with `siginfo_t`, zero-time poll, short timeout, and cross-thread delivery.

<sup>Written for commit 34616d57dadecd0847cd7e75c4e1a0836da4cd59. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/201?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/signal.c
+++ b/src/syscall/signal.c
@@ -1332,7 +1332,189 @@ int64_t signal_rt_sigpending(guest_t *g, uint64_t set_gva, uint64_t sigsetsize)
     return 0;
 }
 
-/* sigaltstack. */
+/* rt_sigtimedwait. */
+
+/* Try to consume one signal from the set under sig_lock.
+ * Returns the signal number if one was found and dequeued, 0 otherwise.
+ * Populates *info_out with the queued siginfo metadata.
+ */
+static int sigtimedwait_try_dequeue(uint64_t mask, signal_rt_info_t *info_out)
+{
+    pthread_mutex_lock(&sig_lock);
+
+    /* Private (thread-directed) set first, then shared (process-directed),
+     * matching Linux dequeue_signal() priority.
+     */
+    signal_pending_t *tp = current_thread ? &current_thread->tpending : NULL;
+    uint64_t thread_m = tp ? (tp->pending & mask) : 0;
+    uint64_t shared_m = sig_state.shared.pending & mask;
+
+    if ((thread_m | shared_m) == 0) {
+        pthread_mutex_unlock(&sig_lock);
+        return 0;
+    }
+
+    int signum;
+    signal_pending_t *src;
+    if (thread_m) {
+        signum = bit_ctz64(thread_m) + 1;
+        src = tp;
+    } else {
+        signum = bit_ctz64(shared_m) + 1;
+        src = &sig_state.shared;
+    }
+
+    /* Dequeue: same logic as signal_deliver. */
+    if (signum >= LINUX_SIGRTMIN) {
+        signal_rt_dequeue_locked(src, signum, info_out);
+    } else {
+        *info_out = signal_standard_peek_locked(src, signum);
+        src->std_info_valid[signum - 1] = false;
+        src->pending &= ~sig_bit(signum);
+    }
+    refresh_pending_hint_locked();
+
+    pthread_mutex_unlock(&sig_lock);
+    return signum;
+}
+
+int64_t signal_rt_sigtimedwait(guest_t *g,
+                               uint64_t set_gva,
+                               uint64_t info_gva,
+                               uint64_t timeout_gva,
+                               uint64_t sigsetsize)
+{
+    if (sigsetsize != 8)
+        return -LINUX_EINVAL;
+    if (!set_gva)
+        return -LINUX_EFAULT;
+
+    uint64_t mask;
+    if (guest_read_small(g, set_gva, &mask, sizeof(mask)) < 0)
+        return -LINUX_EFAULT;
+
+    /* SIGKILL and SIGSTOP cannot be caught or waited for. Remove them from
+     * the wait mask silently, matching Linux do_sigtimedwait behavior.
+     */
+    uint64_t unmaskable = sig_bit(LINUX_SIGKILL) | sig_bit(LINUX_SIGSTOP);
+    mask &= ~unmaskable;
+
+    /* Determine deadline: NULL timeout_gva means block indefinitely.
+     * Zero timespec means poll once.
+     */
+    bool has_timeout = (timeout_gva != 0);
+    int64_t remaining_ns = 0;
+
+    if (has_timeout) {
+        linux_timespec_t lts;
+        if (guest_read_small(g, timeout_gva, &lts, sizeof(lts)) < 0)
+            return -LINUX_EFAULT;
+        /* Negative tv_sec is invalid (matches kernel hrtimer validation). */
+        if (lts.tv_sec < 0 || lts.tv_nsec < 0 || lts.tv_nsec >= 1000000000LL)
+            return -LINUX_EINVAL;
+        /* Saturate to INT64_MAX to avoid overflow. */
+        if (lts.tv_sec > (INT64_MAX / 1000000000LL) ||
+            (lts.tv_sec == (INT64_MAX / 1000000000LL) &&
+             lts.tv_nsec > (INT64_MAX % 1000000000LL))) {
+            remaining_ns = INT64_MAX;
+        } else {
+            remaining_ns = lts.tv_sec * 1000000000LL + lts.tv_nsec;
+        }
+    }
+
+    /* Poll/wait loop. */
+#define SIGWAIT_CHUNK_NS 1000000LL /* 1ms chunks */
+
+    while (1) {
+        signal_rt_info_t info;
+        int signum = sigtimedwait_try_dequeue(mask, &info);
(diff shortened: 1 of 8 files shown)


src/syscall/signal.c
```diff
@@ -1332,7 +1332,162 @@ int64_t signal_rt_sigpending(guest_t *g, uint64_t set_gva, uint64_t sigsetsize)
...
+        if (proc_exit_group_requested())
+            return -LINUX_EINTR;
+
+        /* If a non-waited signal is pending, return -EINTR so the caller's
+         * library (musl/glibc) can restart or propagate.
+         */
+        uint64_t *blocked = thread_blocked_ptr();
+        uint64_t deliverable;
+        pthread_mutex_lock(&sig_lock);
+        deliverable = self_pending_locked() & ~*blocked & ~mask;
```
Max042004: `signal_rt_sigtimedwait`'s EINTR check doesn't filter by signal disposition, so signals that should silently discard (SIG_IGN, or default-ignore/continue/stop like SIGCHLD/SIGURG/SIGWINCH/SIGCONT) spuriously wake the wait when they're not in `mask`

signal_pending_interruption() (signal.c:611) already implements the correct disposition filter. Suggest reusing that filtering logic here

Max042004 (changes requested):

src/syscall/signal.c
doanbaotrung: Thanks for the review.

Fixed. The old check used a raw bitmask self_pending_locked() & ~blocked & ~mask which would spuriously wake the wait for signals like SIGCHLD, SIGURG, SIGWINCH, and SIGCONT — all of which are silently discarded by signal_deliver with no guest-visible effect.

Replaced it with a disposition-aware filter loop that mirrors signal_pending_interruption() exactly: a candidate signal only causes -EINTR if it has a real user-visible effect — an explicit non-SIG_IGN handler, or a SIG_DFL disposition of TERM/CORE. Signals with SIG_IGN or default IGN/CONT/STOP dispositions now correctly fall through and do not interrupt the wait.

jserv (changes requested):
Refined based on the review from @cubic-dev-ai . If you disagree with any of its feedback, discuss it with the reviewer to reach a consensus and consolidate the shared knowledge base.

src/syscall/signal.c
```diff
@@ -1332,7 +1332,162 @@ int64_t signal_rt_sigpending(guest_t *g, uint64_t set_gva, uint64_t sigsetsize)
...
+        if (guest_read_small(g, timeout_gva, &lts, sizeof(lts)) < 0)
+            return -LINUX_EFAULT;
+        /* Negative tv_sec is invalid (matches kernel hrtimer validation). */
+        if (lts.tv_sec < 0 || lts.tv_nsec < 0 || lts.tv_nsec >= 1000000000LL)
+            return -LINUX_EINVAL;
+        /* Saturate to a large positive to avoid overflow. */
+        if (lts.tv_sec > (INT64_MAX / 1000000000LL))
+            remaining_ns = INT64_MAX;
+        else
+            remaining_ns = lts.tv_sec * 1000000000LL + lts.tv_nsec;
```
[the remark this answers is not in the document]
doanbaotrung: ### Fix
I changed the saturation check condition to precisely verify the overflow threshold on both the `tv_sec` and `tv_nsec` components:
```c
        /* Saturate to INT64_MAX to avoid overflow. */
        if (lts.tv_sec > (INT64_MAX / 1000000000LL) ||
            (lts.tv_sec == (INT64_MAX / 1000000000LL) &&
             lts.tv_nsec > (INT64_MAX % 1000000000LL))) {
            remaining_ns = INT64_MAX;
        } else {
            remaining_ns = lts.tv_sec * 1000000000LL + lts.tv_nsec;
        }
```

jserv:

PR #203, opened 2026-07-13
Optimize anonymous mmap and page-fault paths

### LINKED ISSUE #165: sys_mmap: eager zero-fill on MAP_ANONYMOUS makes sparse mappings ~5x slower than Linux
## Summary

`sys_mmap` zeroes the entire region synchronously for every `MAP_ANONYMOUS` mapping (`src/syscall/mem.c:2439`):

```c
/* Zero the mapped region */
memset((uint8_t *) g->host_base + result_off, 0, length);
```

This is functionally correct, but pays the full zero-fill cost up front. Linux backs new anonymous pages with the shared zero page and zeroes lazily on first touch, so workloads that map a region and touch only a fraction of it (e.g. CPython pymalloc 256 KiB arenas) pay for pages they never use.

## Measurements (main @ e6b3455, Apple M-series)

Static aarch64 binary looping `mmap(256 KiB, MAP_ANON|MAP_PRIVATE) → touch → munmap`, 5000 iterations:

| Workload | elfuse | OrbStack (Linux) |
|---|---:|---:|
| 1-byte touch | 7,041 ns/cycle | 1,392 ns/cycle |
| touch every 4 KiB page (64 pages) | 6,934 ns/cycle | 8,904 ns/cycle |

On elfuse the 1-byte-touch cost equals the full-touch cost — the whole region is paid for at mmap time. On Linux the cost scales with pages actually touched.

Downstream effect: CPython `for _ in range(5M): pass` runs ~11.5 ns/iter on elfuse vs ~7.3 on OrbStack with the same python3.12 binary; the gap mostly disappears with `PYTHONMALLOC=malloc` (bypassing pymalloc's mmap-backed arenas), confirming it is mmap-bound rather than interpreter-bound.

## Notes toward a fix

- The lazy machinery already exists: `guest_materialize_lazy()` (`src/core/guest.c:3265`) plus the translation-fault path in `src/syscall/proc.c` already perform fault-time PTE creation, per-2MiB-block zeroing, and TLBI — but only for regions carrying the `noreserve` flag. Extending this to all anonymous mappings is the natural fix. Open questions: stale-PTE invalidation cost on address reuse (currently `guest_invalidate_ptes` per mmap on the NORESERVE path), and concurrent faults on the same range from multiple vCPUs.
- Host `madvise(MADV_DONTNEED)` is not a shortcut: on macOS it behaves like `MADV_FREE`, and replacing the memset with it regressed the benchmark to 14–30 µs/cycle in a local experiment.
- The memset itself is only ~25% of the gap (measured by unsoundly disabling it: ~5.6 µs/cycle remains). The rest is the per-syscall region-tracking / `guest_extend_page_tables` / `guest_update_perms` / TLBI path — lazy materialization also skips most of that work for pages never touched.

## Reproduction

```c
#include <stdio.h>
#include <stdlib.h>
#include <sys/mman.h>
#include <time.h>

    int N = (argc > 1) ? atoi(argv[1]) : 5000;
    size_t SZ = 256 * 1024;
    struct timespec t0, t1;
    clock_gettime(CLOCK_MONOTONIC, &t0);
    for (int i = 0; i < N; i++) {
        char *p = mmap(NULL, SZ, PROT_READ | PROT_WRITE,
                       MAP_ANON | MAP_PRIVATE, -1, 0);
        p[0] = 1; /* or: for (size_t j = 0; j < SZ; j += 4096) p[j] = 1; */
        munmap(p, SZ);
    }
    clock_gettime(CLOCK_MONOTONIC, &t1);
    double ns = (t1.tv_sec - t0.tv_sec) * 1e9 + (t1.tv_nsec - t0.tv_nsec);
    printf("%.1f ns/cycle\n", ns / N);
    return 0;
}
```

```sh
aarch64-linux-musl-gcc -O2 -static -o bench_mmap bench_mmap.c
build/elfuse bench_mmap 5000
```

Max042004:
Anonymous memory is reworked on a lazy foundation: sys_mmap only records the region, while page tables, zeroing and host-memory commit are deferred to first touch. Two optimization layers sit on top.

mmap: the common shape (addr == 0, MAP_PRIVATE|MAP_ANONYMOUS, RW) is served entirely at EL1 with no vCPU exit and no lock. The shim stays stateless: it bump-allocates from a host-carved per-vCPU VA arena and reports each grant through a per-vCPU SPSC ring that the host drains whenever it takes mmap_lock, so no cross-vCPU synchronization exists. Arenas recycle freed address space through the gap allocator, refill automatically when a request does not fit (the new arena sized to cover the triggering request, with an ARENA_MAX guard so oversized requests fall back without churning the arena), and decay back to workload scale based on served-length history.

Page fault: a per-2MiB-block dirty map lets materialization skip zeroing blocks whose slab bytes are known zero, moving the host page commit off the fault and out of mmap_lock onto the guest's own first writes. The TLBI protocol emits a single range op per materialized block instead of per-page invalidations, and a repeat fault on an already-valid block returns without re-zeroing.

Host-side access to untouched lazy memory (I/O buffers, futex words, signal frames, shmat/shmdt copies) materializes on demand with lock-order-safe pre-faulting, and PROT_NONE stays a pure faulting reservation. tests/test-mmap-lazy.c and tests/test-mmap-fastpath.c pin the contract: zero-on-reuse, huge sparse reservations, host access to untouched memory, PROT_NONE round trips, fork, racing first touch, and arena refill/revocation.

Close #165

---
## Summary by cubic
Implements lazy anonymous mmap with an EL1 fast path for common anon RW calls and EL1 deferred munmap. Cuts mmap/munmap latency, reduces contention, and fixes mremap edge cases. Closes #165.

- **New Features**
  - Lazy anon memory with a per‑2 MiB dirty map: skip zeroing clean blocks, batch TLBI per block, ignore repeat faults; bootstrap marks ELF/shim/stack blocks dirty to preserve zero‑on‑reuse on MAP_FIXED.
  - EL1 mmap fast path for `mmap(NULL, len, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, ...)` up to 32 GiB: per‑vCPU VA arenas + SPSC rings with adaptive refill and automatic fallback.
  - EL1 deferred `munmap`: retire with broadcast TLBI and defer host cleanup; drain on VM‑exit and on `mmap_lock` acquire; per‑2 MiB PTE‑present index accelerates teardown and arena recycle.
  - Host access pre‑faulting for futex words, signal frames, `exec` argv/env copying, and SysV SHM; `PROT_NONE` remains a pure reservation.
  - Robust `mremap`: preserve neighbor PTEs on cross‑block extend and fully roll back destination on ENOMEM; new tests cover splits, grow/move, and adjacent‑page safety.
  - Timing and tooling: enable CNTVCT reads on worker vCPUs; add `bench-mmap` and `bench-mmap-lazy`, plus fast‑path/dirty‑map stats scripts and counters (incl. TLBI wire mode).

- **Migration**
  - EL1 mmap/munmap fast paths are on by default and fall back when shape unsupported, arenas exhausted, or rings full. Set `ELFUSE_MMAP_FASTPATH=0` to disable; verbose tracing, the syscall histogram, GDB, and Rosetta force the host path.
  - Fork IPC protocol bumped to ELFM. Update both parent and child when mixing versions.

<sup>Written for commit a58f5bbc5f229168ca8759b01cc07e1fc39d954e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/203?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/core/shim.S
+++ b/src/core/shim.S
@@ -251,6 +251,28 @@
 .Lctr_skip_\@:
 .endm
 
+/* Per-vCPU mmap counters live after the 16 x 24-byte publication ring in the
+ * mmap control block. Each vCPU is the sole writer of its own slots. Keep the
+ * shared stats gate so normal fast paths do not touch diagnostic cache lines.
+ */
+.equ SHIM_MMAP_COUNTERS_OFF, 0x1C0
+.equ MC_SHAPE_MISS,          0
+.equ MC_CAPACITY_MISS,       1
+.equ MC_RING_FULL,           2
+.equ MC_GENERATION_STALE,    3
+.equ MC_ATTENTION,           4
+.equ MC_HIT,                 5
+
+.macro MMAP_COUNTER_INC slot
+    mrs x29, tpidr_el1
+    ldrb w30, [x29, #SHIM_STATS_EN_OFF]
+    cbz w30, .Lmmap_ctr_skip_\@
+    ldr x29, [x12, #(SHIM_MMAP_COUNTERS_OFF + 8 * \slot)]
+    add x29, x29, #1
+    str x29, [x12, #(SHIM_MMAP_COUNTERS_OFF + 8 * \slot)]
+.Lmmap_ctr_skip_\@:
+.endm
+
 /* BAD_VEC: vector-table entry that reports an unexpected exception.
  * Each table slot is 128 bytes; the leading .align 7 places this entry at the
  * next 128-byte boundary.
@@ -440,6 +462,118 @@ svc_handler:
     b.eq getsid_fast
     cmp x10, #278                      /* SYS_getrandom? */
     b.eq getrandom_fast
+    cmp x10, #222                      /* SYS_mmap? */
+    b.eq mmap_anon_fast
+    b handle_svc_0
+
+/* mmap_anon_fast: consume a host-prepared, per-vCPU VA arena for the exact
+ * anonymous RW shape used by allocators:
+ *
+ *   mmap(NULL, len, PROT_READ|PROT_WRITE,
+ *        MAP_PRIVATE|MAP_ANONYMOUS[|MAP_NORESERVE], fd, off)
+ *
+ * The control is selected without another system register.  SP_EL1 stack tops
+ * are shim_data_end - slot*4KiB; controls are shim_data+0x20000 + slot*2KiB,
+ * so (shim_data_end - ALIGN_UP(sp,4KiB))/2 is the control-array displacement.
+ * Each vCPU is the sole producer of its ring and cursor.  The host acquire-
+ * drains all rings whenever it takes mmap_lock.
+ */
+mmap_anon_fast:
+    mrs x12, tpidr_el1                 /* shared shim-data base */
+    ldar w13, [x12]                    /* tracing/signal attention gate */
+
+    add x14, sp, #0xfff
+    and x14, x14, #~0xfff              /* this vCPU's stack top */
+    add x15, x12, #0x200, lsl #12      /* shim_data + 2MiB */
+    sub x15, x15, x14                  /* slot * 4KiB */
+    lsr x15, x15, #1                   /* slot * 2KiB */
+    add x12, x12, x15
+    add x12, x12, #0x20, lsl #12       /* control base + 0x20000 */
+
+    cbnz w13, mmap_attention_bail
+
+    ldr x14, [sp, #0]                  /* addr */
+    cbnz x14, mmap_shape_bail
+    ldr x16, [sp, #16]                 /* prot */
+    cmp x16, #3                        /* PROT_READ|PROT_WRITE */
+    b.ne mmap_shape_bail
+    ldr x17, [sp, #24]                 /* flags */
+    bic x18, x17, #0x4000              /* tolerate MAP_NORESERVE */
+    cmp x18, #0x22                     /* MAP_PRIVATE|MAP_ANONYMOUS */
+    b.ne mmap_shape_bail
+
+    ldr x15, [sp, #8]                  /* len */
+    cbz x15, mmap_shape_bail
+    adds x15, x15, #0xfff
+    b.cs mmap_shape_bail                /* PAGE_ALIGN_UP overflow */
+    and x15, x15, #~0xfff
+
+    ldar w13, [x12]                    /* descriptor generation */
+    ldr w14, [x12, #4]                 /* consumer generation */
+    cmp w13, w14
+    b.ne mmap_generation_bail
+    ldr w13, [x12, #8]                 /* SHIM_MMAP_CTRL_ENABLED */
+    tbz w13, #0, mmap_capacity_bail
+
+    ldr x19, [x12, #40]                /* private bump cursor */
+    mov x20, #0x200000
+    cmp x15, x20
+    b.lo 1f
+    sub x21, x20, #1
+    adds x19, x19, x21
+    b.cs mmap_capacity_bail
+    bic x19, x19, x21                  /* >=2MiB: 2MiB-align */
+1:  ldr x20, [x12, #32]                /* arena limit */
+    adds x21, x19, x15                 /* new cursor */
+    b.cs mmap_capacity_bail
+    cmp x21, x20
+    b.hi mmap_capacity_bail
+
+    add x23, x12, #12                  /* &head */
+    ldar w22, [x23]
+    ldr w23, [x12, #16]                /* producer-owned tail */
+    sub w24, w23, w22
+    cmp w24, #16                       /* SHIM_MMAP_RING_SIZE */
+    b.hs mmap_ring_full_bail
+
+    and w24, w23, #15
+    add x24, x24, x24, lsl #1          /* index * 3 */
+    add x25, x12, #64                   /* ring base */
+    add x25, x25, x24, lsl #3          /* index * 24 */
+    str x19, [x25, #0]
+    str x15, [x25, #8]
+    str x16, [x25, #16]
+    str x21, [x12, #40]                /* publish cursor before entry */
+    add w23, w23, #1
+    add x25, x12, #16
+    stlr w23, [x25]                    /* release-publish ring entry */
+    mov x0, x19
+    MMAP_COUNTER_INC MC_HIT
+    b svc_restore_eret
+
+mmap_shape_bail:
+    MMAP_COUNTER_INC MC_SHAPE_MISS
+    b handle_svc_0
+
+mmap_capacity_bail:
+    MMAP_COUNTER_INC MC_CAPACITY_MISS
+    b handle_svc_0
+
+mmap_ring_full_bail:
+    MMAP_COUNTER_INC MC_RING_FULL
+    b handle_svc_0
+
+mmap_attention_bail:
+    MMAP_COUNTER_INC MC_ATTENTION
+    b attn_bail
+
+mmap_generation_bail:
+    /* A revocation deliberately leaves the consumer generation stale.  Ack
+     * only after the acquire load above, then make this syscall take HVC; the
+     * host will either leave the control disabled or publish a fresh arena.
+     */
+    MMAP_COUNTER_INC MC_GENERATION_STALE
+    str w13, [x12, #4]
     b handle_svc_0
 
 identity_class_fast:
@@ -1203,6 +1337,7 @@ handle_el0_fault:
 
 .Lel0_fault_tlbi_full:
     /* Broadcast TLB + conditional I-cache flush. X11=0 skips IC IALLU. */
+    dsb ishst
     tlbi vmalle1is
     dsb ish
     cbz x11, .Lel0_fault_full_no_ic
@@ -1223,6 +1358,7 @@ handle_el0_fault:
     mov x13, x11
     ubfx x11, x9, #12, #44
     mov x12, x10
+    dsb ishst
 4:  tlbi vae1is, x11
     add x11, x11, #1
     subs x12, x12, #1
@@ -1237,7 +1373,9 @@ handle_el0_fault:
 
 .Lel0_fault_tlbi_rvae:
     /* Single-shot TLBI RVAE1IS (FEAT_TLBIRANGE). X9 carries the pre-encoded
-     * operand (baddr | NUM<<39 | TG=01<<46); X11 the I-cache hint. */
+     * operand (baddr | NUM<<39 | SCALE<<44 | TG=01<<46); X11 the I-cache
+     * hint. */
+    dsb ishst
     tlbi rvae1is, x9
     dsb ish
     cbz x11, .Lel0_fault_rvae_no_ic
@@ -1281,6 +1419,7 @@ tlbi_restore_eret:
      * leak VA bits into the TTL [47:44] or ASID [63:48] operand fields.
      */
     ubfx x0, x0, #12, #44
+    dsb ishst
     tlbi vae1is, x0
     dsb ish
     ic iallu
@@ -1356,6 +1495,7 @@ handle_svc_0:
      *        1 = broadcast TLBI VMALLE1IS
      *        2 = execve replaced register state (drop frame + flush)
      *        3 = selective TLBI VAE1IS over X10 pages starting at X9
+     *        4 = single-shot TLBI RVAE1IS with encoded operand in X9
      *   5. Resume vCPU (execution continues below)
      */
     hvc #5
@@ -1384,6 +1524,7 @@ tlbi_full:
      * include exec), so the shim must IC IALLU; zero means a data-only
      * PT change and the I-cache invalidation is skipped.
      */
+    dsb ishst
     tlbi vmalle1is
     dsb ish
     cbz x11, .Ltlbi_full_skip_ic
@@ -1419,6 +1560,7 @@ tlbi_selective:
      * leak VA bits into the TTL [47:44] or ASID [63:48] operand fields. */
     ubfx x11, x9, #12, #44    /* x11 = VA[55:12] (current page operand) */
     mov x12, x10              /* x12 = remaining page counter */
+    dsb ishst
 3:  tlbi vae1is, x11
     add x11, x11, #1          /* next page (operand is in 4 KiB units) */
     subs x12, x12, #1
@@ -1434,11 +1576,11 @@ tlbi_selective:
 tlbi_range_large:
     /* Single-shot TLBI RVAE1IS (FEAT_TLBIRANGE, ARMv8.4+). The host has
      * encoded the full operand in X9: baddr (VA >> 12), TTL=0, NUM in bits
-     * [43:39], SCALE=0, ASID=0. One instruction covers up to 64 pages,
-     * avoiding the broadcast TLBI VMALLE1IS that the prior selective cap
-     * forced for 17..64-page ranges. X11 carries the I-cache hint as in
-     * tlbi_full / tlbi_selective.
+     * [43:39], SCALE in bits [45:44], ASID=0. One instruction covers up to
+     * TLBI_RVAE_MAX_PAGES. X11 carries the I-cache hint as in tlbi_full /
+     * tlbi_selective.
      */
+    dsb ishst
     tlbi rvae1is, x9
     dsb ish
     cbz x11, .Ltlbi_rvae_skip_ic
--- a/src/core/shim.S
+++ b/src/core/shim.S
@@ -26,7 +26,7 @@
  *                            4 = single-shot TLBI RVAE1IS (FEAT_TLBIRANGE);
  *                                X9 carries the pre-encoded RVAE1IS operand
  *                                (baddr | NUM<<39 | SCALE<<44 | TTL<<37 |
- *                                ASID<<48; SCALE=0, TTL=0, ASID=0 today)
+ *                                ASID<<48; TTL=0 and ASID=0)
  *                          X11 carries the I-cache hint for X8 in {1, 3, 4}:
  *                          1 = IC IALLU after the TLBI sequence (new
  *                          executable content visible to EL0), 0 = skip the
@@ -251,11 +251,11 @@
 .Lctr_skip_\@:
 .endm
 
-/* Per-vCPU mmap counters live after the 16 x 24-byte publication ring in the
+/* Per-vCPU mmap counters live after the 32 x 24-byte publication ring in the
  * mmap control block. Each vCPU is the sole writer of its own slots. Keep the
  * shared stats gate so normal fast paths do not touch diagnostic cache lines.
  */
-.equ SHIM_MMAP_COUNTERS_OFF, 0x1C0
+.equ SHIM_MMAP_COUNTERS_OFF, 0x340
 .equ MC_SHAPE_MISS,          0
 .equ MC_CAPACITY_MISS,       1
 .equ MC_RING_FULL,           2
@@ -462,6 +462,8 @@ svc_handler:
     b.eq getsid_fast
     cmp x10, #278                      /* SYS_getrandom? */
     b.eq getrandom_fast
+    cmp x10, #215                      /* SYS_munmap? */
+    b.eq munmap_anon_fast
     cmp x10, #222                      /* SYS_mmap? */
     b.eq mmap_anon_fast
     b handle_svc_0
@@ -533,10 +535,10 @@ mmap_anon_fast:
     ldar w22, [x23]
     ldr w23, [x12, #16]                /* producer-owned tail */
     sub w24, w23, w22
-    cmp w24, #16                       /* SHIM_MMAP_RING_SIZE */
+    cmp w24, #32                       /* SHIM_MMAP_RING_SIZE */
     b.hs mmap_ring_full_bail
 
-    and w24, w23, #15
+    and w24, w23, #31
     add x24, x24, x24, lsl #1          /* index * 3 */
     add x25, x12, #64                   /* ring base */
     add x25, x25, x24, lsl #3          /* index * 24 */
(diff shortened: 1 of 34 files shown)


jserv:
Provide comprehensive benchmarks.

src/core/shim.S
```diff
@@ -440,6 +462,118 @@ svc_handler:
...
+    b.hs mmap_ring_full_bail
+
+    and w24, w23, #15
+    add x24, x24, x24, lsl #1          /* index * 3 */
+    add x25, x12, #64                   /* ring base */
+    add x25, x25, x24, lsl #3          /* index * 24 */
+    str x19, [x25, #0]
+    str x15, [x25, #8]
+    str x16, [x25, #16]
+    str x21, [x12, #40]                /* publish cursor before entry */
```
jserv: The EL1 fast path publishes the bump cursor (`str x21, [x12, #40]`) before the release `stlr` that publishes the ring tail. Because the host consumer runs concurrently on another thread and `mmap_fastpath_skip_reserved` (src/syscall/mem.c:417) only excludes the unconsumed tail `[cursor, arena_limit)`, there is a window where a just-granted span leaks out as free VA:

vCPU A (in `mmap_anon_fast`) writes its ring slot for `[old_cursor, new_cursor)` and stores the new cursor, but has not yet executed the `stlr` tail. vCPU B takes a slow-path `sys_mmap`; `mmap_lock_acquire` drains A's ring but sees the old tail, so no region is recorded. `find_free_gap_inner` then observes A's new cursor and protects only `[new_cursor, arena_limit)`, so `[old_cursor, new_cursor)` is neither in `regions[]` nor reserved, and gets handed to B. Two guest mappings alias the same VA -- data corruption and stale-byte exposure.

The window is real regardless of store order, because seeing a fresh `cursor` never implies the covering ring entry is visible. The robust fix is host-side: have `mmap_fastpath_skip_reserved` (and the recycle scan in `mmap_fastpath_refill_thread_locked`) exclude the entire enabled arena `[arena_base, arena_limit)` for every sibling slot, not just `[cursor, limit)`. Reordering the shim so the cursor store follows the `stlr` narrows the window but does not close it under weak memory unless the cursor also carries release semantics paired with the consumer's acquire.

src/syscall/mem.c
```diff
@@ -35,6 +37,392 @@
...
+                addr > arena_limit || len > arena_limit - addr ||
+                e->prot != (LINUX_PROT_READ | LINUX_PROT_WRITE)) {
+                log_fatal(
+                    "mmap fast path: invalid entry in vCPU slot %d "
+                    "(addr=0x%llx len=0x%llx arena=0x%llx..0x%llx)",
+                    slot, (unsigned long long) addr, (unsigned long long) len,
+                    (unsigned long long) arena_base,
+                    (unsigned long long) arena_limit);
+            }
+            if (guest_region_add_ex(g, addr, addr + len,
```
jserv, line 449: The drain fails closed only on ring corruption, arena-bounds violation, and table-full. `guest_region_add_ex_owned_gpa` (src/core/guest.c:2106) inserts a sorted region with no overlap check, so if a bad arena base or the cursor race in the fast path ever produces an overlapping grant, this path silently inserts a duplicate region over live VA instead of detecting the invariant break. Adding an overlap assertion here (the insertion neighbor is already known via `region_lower_bound_start`) turns a silent corruption into a fail-closed abort -- cheap defense-in-depth for the fast path's "EL1 only bump-allocates disjoint VA" contract.

Max042004:
# elfuse vs OrbStack — mmap / Page-Fault Benchmark

benchmark suite provide in:
https://gist.github.com/Max042004/ada7bcb7b14dd392ae35633386e7d6b7
https://gist.github.com/Max042004/17ba8e373b45a0a89ab7865969406446

- bench machine: M4
- Baseline: commit `718f908` (pre-optimization), measured with 5 s per-invocation timeout
- All values: median of 10 iterations, per-operation ns (32-op batched CNTVCT timing, ~1.3 ns resolution)
- Fault pages are 16 KiB mappings so consecutive arena-contiguous allocations never share a macOS host page — every touch is a genuine fault on both platforms; one untimed warmup batch precedes each measurement
- `timeout (>5s)` = the whole 10-iteration invocation exceeded 5 s (baseline eagerly zeroes the entire mapping at mmap() time)

## mmap (single-threaded)

| Size | 718f908 (ns) | optimized (ns) | speedup | OrbStack (ns) | optimized / OrbStack |
|---|---|---|---|---|---|
| 4 KiB | 1,480.5 | 87.2 | 17.0x | 162.8 | 0.54x |
| 128 KiB | 2,175.8 | 72.9 | 29.8x | 162.8 | 0.45x |
| 4 MiB | 36,694.0 | 126.3 | 290.5x | 166.7 | 0.76x |
| 128 MiB | timeout (>5s) | 213.5 | — | 169.3 | 1.26x |
| 1 GiB | timeout (>5s) | 722.7 | — | 169.3 | 4.27x |
| 4 GiB | timeout (>5s) | 1,208.3 | — | 171.9 | 7.03x |
| 32 GiB | timeout (>5s) | 1,541.7 | — | 208.3 | 7.40x |

## Page fault (single-threaded, touch 1 byte of a fresh page)

| | 718f908 (ns) | optimized (ns) | OrbStack (ns) | optimized / OrbStack |
|---|---|---|---|---|
| touch cost | 10.4* | 1,931.0 | 401.0 | 4.82x |

\* Not comparable: the baseline commits (memsets) every page eagerly inside mmap() itself, so by the time the benchmark touches the page there is no fault left to take — the 10.4 ns is a plain cached store. The baseline's fault cost is what makes its mmap column above slow/timeout.

## Multithread mmap

| Size | Threads | 718f908 (ns) | optimized (ns) | speedup | OrbStack (ns) | optimized / OrbStack |
|---|---|---|---|---|---|---|
| 4 KiB | 2 | 1,734.4 | 651.0 | 2.66x | 793.0 | 0.82x |
| 4 KiB | 4 | 2,682.3 | 1,141.9 | 2.35x | 1,260.4 | 0.91x |
| 128 KiB | 2 | 5,886.7 | 821.6 | 7.16x | 932.3 | 0.88x |
| 128 KiB | 4 | 10,252.6 | 881.5 | 11.6x | 1,209.6 | 0.73x |
| 4 MiB | 2 | 77,003.9 | 849.0 | 90.7x | 373.7 | 2.27x |
| 4 MiB | 4 | 161,036.5 | 1,819.0 | 88.5x | 1,428.4 | 1.27x |
| 128 MiB | 2 | timeout (>5s) | 988.3 | — | 1,059.9 | 0.93x |
| 128 MiB | 4 | timeout (>5s) | 2,066.4 | — | 562.5 | 3.67x |
| 1 GiB | 2 | timeout (>5s) | 2,287.8 | — | 588.5 | 3.89x |
| 1 GiB | 4 | timeout (>5s) | 6,290.4 | — | 910.2 | 6.91x |
| 4 GiB | 2 | timeout (>5s) | 4,583.3 | — | 1,505.2 | 3.04x |
| 4 GiB | 4 | timeout (>5s) | 8,265.6 | — | 2,369.8 | 3.49x |
| 32 GiB | 2 | timeout (>5s) | 9,208.3 | — | 8,458.3 | 1.09x |
| 32 GiB | 4 | timeout (>5s) | 24,416.7 | — | 15,458.3 | 1.58x |

## Multithread page fault

| Threads | 718f908 (ns)* | optimized (ns) | OrbStack (ns) | optimized / OrbStack |
|---|---|---|---|---|
| 2 | 2,061.2 | 2,841.1 | 3,882.8 | 0.73x |
| 4 | 3,255.2 | 4,950.5 | 5,621.1 | 0.88x |

\* Same caveat as the single-threaded table: baseline pages are pre-committed at mmap() time, so its numbers are barrier-synchronized cached stores, not faults.

jserv:
I have a follow-up branch [elfuse-mmap-eval](https://github.com/sysprog21/elfuse/tree/elfuse-mmap-eval) with two changes on top of the mmap and page-fault work here, plus a reusable mmap microbenchmark.

### 1. mremap: preserve neighbor PTEs on cross-block extend (correctness)

`mremap_extend_range()` applied `guest_update_perms()` over the whole 2 MiB-aligned extension, which zeroed the lazy PTEs of unrelated mappings that shared an edge block. It now snapshots per-block presence before extending, then splits and selectively invalidates only the non-kept subranges of freshly created blocks; pre-existing blocks keep their PTEs. A single idempotent rollback path restores the page-table high-water mark on failure, the ENOMEM `sys_mremap` path restores the destination snapshot page tables, and the 5 open-coded RX/RW arena tests + 4 copy-pasted high-water bumps collapse into `mmap_pt_end_for_off()` / `mmap_bump_next()`. `guest_va_block_mapped()` now also reports L3 table descriptors as mapped. New tests: adjacent-page-fault, neighbor-L3 preservation, hinted-tail zero.

I also dropped a fresh-range invalidate-skip micro-optimization after A/B benchmarking showed it was a no-op: the invalidate it skipped is already free on an unmapped range (nothing to walk), at every size from 4 KiB to 32 GiB.

### 2. CNTKCTL_EL1 on worker vCPUs (perf bug the benchmark surfaced)

Bootstrap sets `CNTKCTL_EL1.EL0VCTEN|EL0PCTEN` on the primary vCPU, but the CLONE_THREAD and CLONE_VM worker bring-up paths omitted it. Worker EL0 read 0 from `CNTVCT_EL0`, so the vDSO `clock_gettime` trampoline's seqlock/anchor check failed and silently fell back to the SVC clock on every non-main thread. Dynamic glibc, `clock_gettime(CLOCK_MONOTONIC)` averaged over 200k calls:

|         | no fix    | with fix |
|---------|-----------|----------|
| main    | 8.1 ns    | 9.2 ns   |
| worker  | 2898.5 ns | 9.2 ns   |

~315x on worker threads. Static glibc guests SVC `clock_gettime` even on the main thread, which is why this stayed hidden.

### Benchmark (`tests/bench-mmap.c`, Apple M1 / MacBookAir10,1)

Timing reads `CNTVCT_EL0` directly at EL0 (`isb; mrs`), not a `clock_gettime` SVC that would swamp sub-microsecond ops; the 24 MHz counter (~41.7 ns/tick) is amortized over an adaptive batch. One untimed warmup per case, median (and min) over runs.

```
== A. mmap / munmap latency vs size (steady state, NULL hint) ==
size        batch      mmap ns    munmap ns
4 KiB          64        212.9       2414.1
16 KiB         64        212.9       2397.1
64 KiB         64        213.5       2403.6
256 KiB        64        212.9       2399.1
1 MiB          64        263.0       2399.1
2 MiB          64        278.6       2398.4
8 MiB          32        290.4       2432.3
64 MiB          4         72.9       2666.7
256 MiB         1         83.3       3416.7
1 GiB           1         83.3       2500.0
4 GiB           1       3666.7       6291.7
16 GiB          1       3666.7       6375.0
32 GiB          1       3708.3       6458.3

== B. fresh bump-tail mmap, per-mmap ns ==
size          count  fresh mmap ns
4 KiB          1000          259.5
2 MiB          1000          283.0
128 MiB          16          286.5
32 GiB            1         2666.7

== C. first-touch fault (16 KiB stride, 512 pages) ==
  per-fault: median 5058.0 ns  min 4855.1 ns

== D. mprotect split (2 MiB block -> L3, protect middle 4 KiB) ==
  split: median 4083.3 ns  min 3500.0 ns

== E. mremap grow 4 KiB -> 8 KiB ==
  in-place: median 3250.0 ns  min 3041.7 ns
  move:     median 4291.7 ns  min 3750.0 ns

== F. multi-threaded fresh mmap, per-op ns (mmap_lock contention) ==
size        threads      mean ns       max ns
4 KiB             2       1403.0       1623.3
4 KiB             4       2641.0       3541.3
2 MiB             2        861.0        939.3
2 MiB             4       1709.0       1937.3
```

Notes on reading this: mmap latency is flat with size on the lazy path (~210 ns to 32 GiB); the real allocation-path cost is `munmap` (~2400 ns, ~10x mmap). To A/B a page-table change, build elfuse with and without it and run the identical benchmark binary under each: overlapping median ranges across repeated runs mean no measurable effect, a consistent gap that scales with size means a real one. Section F times with `clock_gettime` because `CNTVCT_EL0` needs per-vCPU `EL0VCTEN` (see change 2).

Build/run:
```sh
make build/bench-mmap
build/elfuse build/bench-mmap
```

@Max042004 , please measure on Apple M4 and analyze carefully.

Max042004:
# Deferred EL1 `munmap`: Flow, Synchronization, and Long Tails
The goal of this design is not to eliminate all `munmap` work, but to divide it
into two stages with different correctness deadlines:

1. **Guest-visible retirement** must complete before `munmap()` returns. EL1
   clears the Stage-1 descriptor, broadcasts a TLB invalidation to all vCPUs,
   and only then publishes the retire record.
2. **Host bookkeeping and backing reclamation** do not affect the address-space
   semantics immediately visible to the guest, so they can be batched at the
   next natural VM exit.

As a result, the common anonymous `munmap()` path requires no HVC, worker
thread, or synchronous wait for macOS memory cleanup.

## The Complete Flow at a Glance

```mermaid
flowchart TD
    U[EL0: munmap addr, len] --> E{EL1 fast path available?}
    E -- No --> S[HVC slow path]
    E -- Yes --> R[Reserve this vCPU's retire-ring slot<br/>Fill data without publishing tail]
    R --> P[Validate and clear Stage-1 PTE / L2 block]
    P --> T[DSB ISHST<br/>broadcast TLBI<br/>DSB ISH + ISB]
    T --> Q[release-store tail<br/>Publish retire record]
    Q --> Z[Return directly to EL0]

    Z -. Any later natural VM exit .-> H[Host acquire-snapshots<br/>all retire tails]
    H --> M[Drain mmap publications first]
    M --> C[Commit retire metadata]
    C --> B{Complete, exclusive dirty run<br/>and materialized >= 32 MiB?}
    B -- Yes --> X[Quiesce sibling vCPUs<br/>HVF detach + fresh backing + remap]
    B -- No --> Y[memset dirty fragments]
    X --> A[Remove region metadata<br/>Release quarantine]
    Y --> A
```

The solid path in the diagram is the work that must complete during this
`munmap()`. Host cleanup below the dotted edge occurs at a later VM exit. The
cleanup has not disappeared; it is simply no longer included in the
guest-visible `munmap` latency.

## Stage One: EL1 Immediately Completes the Guest-Visible Unmap

### Fast-Path Eligibility

EL1 accepts only requests with a simple enough shape to complete safely
without allocating host metadata:

- `addr` and `length` are both 4 KiB aligned, `length != 0`, and
  `addr + length` does not overflow.
- The entire range belongs to one still-valid anonymous mmap fast-arena
  generation.
- The range is not a stack, file overlay, `MAP_SHARED` mapping, or elfuse
  infrastructure.
- If a partial 2 MiB block must be cleared page by page, its L3 table already
  exists. EL1 does not split blocks or allocate new page-table pages.
- This vCPU's retire ring is not near-full.
- The host page-table gate remains open.

If any condition fails, execution returns to the existing `SVC -> HVC` slow
path. A fallback is a normal correctness path, not an error.

```mermaid
flowchart TD
    A[Receive munmap] --> V{Aligned, nonzero, no overflow?}
    V -- No --> H[HVC slow path]
    V -- Yes --> F{Retire-ring occupancy < 31?}
    F -- No --> H
    F -- Yes --> G{Join PT-gate handshake?}
    G -- No --> H
    G -- Yes --> N{One valid arena generation<br/>contains the whole range?}
    N -- No --> H
    N -- Yes --> L{Does every partial L2 already have L3?}
    L -- No --> H
    L -- Yes --> R[Reserve slot]
    R --> W[Walk / clear descriptors]
    W --> T[TLBI + completion barriers]
    T --> P[Publish tail]
    P --> O[Return 0]
```

### Why the Ring Slot Must Be Reserved First

Each vCPU owns a 32-entry SPSC ring:

```c
struct munmap_retire_entry {
    uint64_t addr;
    uint64_t length;
    uint32_t arena_generation;
    uint32_t flags; /* arena slot + charged materialized pages */
};
```

EL1 is the ring's only producer, and the host is its only consumer. The
producer first checks `head` and `tail` and fills the slot, but does not advance
`tail`. This prevents the unrecoverable state in which a PTE has already been
cleared before EL1 discovers that no space remains to record the metadata
retirement.

A per-vCPU SPSC design has two additional benefits: producers never write the
same tail or cause cache-line bouncing on a shared MPSC bitmap, and host
allocator state remains exclusively host-owned.

### The Two Page-Table Walk Paths

EL1 validates first and mutates second. The generic path checks L0, L1, and L2,
clears one descriptor for a complete L2 block, and clears the corresponding
4 KiB leaves in an existing L3 table. The two-pass walk ensures that mutation
cannot begin before discovering that a 2 MiB block would need to be split.

The common large, 2 MiB-aligned case uses the L2-only path:

- L0 is walked only once.
- One L1 entry is looked up per 1 GiB window.
- Consecutive L2 descriptors are validated and cleared linearly.
- After validation, EL1 uses ordinary aligned 64-bit stores. The host gate has
  already excluded host writers, and a sibling fast producer can only write
  zero to the same descriptor.
- All descriptor stores share one final `DSB ISHST`.

This means a 1 GiB retirement does not need to visit 262,144 individual 4 KiB
PTEs. If the range consists of 2 MiB blocks, EL1 processes only 512 L2
descriptors.

### Translation Ordering Required Before Return

The critical order is:

```text
clear Stage-1 descriptors
        ↓
DSB ISHST
        ↓
TLBI ...IS                 (inner-shareable broadcast)
        ↓
DSB ISH
ISB
        ↓
release-store retire.tail
        ↓
return to EL0
```

Small ranges use per-page `VAE1IS`. Larger ranges use `RVAE1IS` when the CPU
supports FEAT_TLBIRANGE, directly encoding SCALE 0 through 3. One SCALE=3
instruction can cover up to 8 GiB. Ranges larger than one instruction's
coverage are emitted as adjacent batches that share the final completion
barrier. Without range-TLBI support, the path falls back to `VMALLE1IS`.

The release-store to `tail` occurs after TLBI completion. Once the host sees
the record with an acquire-load, it therefore also knows that:

- all entry fields are fully visible;
- the descriptor stores have completed;
- stale translations on sibling vCPUs have been invalidated; and
- any subsequent EL0 access to the range must fault.

These are the Linux address-space semantics that must be guaranteed when
`munmap()` returns.

## Stage Two: Host State Is Reconciled at the Next Natural VM Exit

The host drains after every return from `hv_vcpu_run()` and **before**
dispatching the exit reason. Syscalls, page faults, signals, `MAP_FIXED`,
fork/exec, and process exit therefore cannot inspect stale region metadata
first.

This ordering is especially important for page faults. If the fault handler
consulted old metadata first, it could rematerialize the mapping that EL1 had
just cleared, invalidating the semantics of an already successful `munmap()`.

The host commit flow is:

1. Acquire-snapshot the retire tails of all vCPUs.
2. Drain every mmap publication ring.
3. Validate each retire entry's range, arena slot, and generation.
4. Wait for overlapping lazy materialization to finish.
5. Zero or replace the backing.
6. Remove `guest_region` metadata and update dirty state, gap hints, and the
   page-table cache generation.
7. Release-store the ring head, allowing the producer to reuse the slot.
8. Release the VA quarantine only after the range has been fully committed.

The host does not clear PTEs or execute TLBI again, because EL1 has already
completed both operations.

### How Large Zeroing Is Delegated to macOS

"Delegating zeroing to macOS" more precisely means discarding the old backing
and allowing macOS to provide clean pages later through zero-fill-on-demand.
It does not mean synchronously asking macOS to write zero to every byte.

A dirty run uses backing replacement only if all of the following conditions
hold:

- EL1 counted at least 32 MiB of actually materialized backing. A large virtual
  range with only one touched page does not qualify.
- The run consists of contiguous, complete, 2 MiB-aligned dirty blocks.
- No live alias remains for the backing; the retirement owns the entire run
  exclusively.

Replacement first quiesces sibling vCPUs, splits only the two HVF segment
boundaries of the range, and then performs:

```text
hv_vm_unmap(target IPA)
        ↓
shared slab: F_PUNCHHOLE + MAP_SHARED | MAP_FIXED
private slab: fresh MAP_ANON | MAP_PRIVATE | MAP_FIXED
        ↓
hv_vm_map(same IPA, fresh host backing)
        ↓
resume sibling vCPUs
```

Here, `hv_vm_unmap()` temporarily detaches the Stage-2/HVF mapping; it does not
execute guest `munmap` a second time. If the range is too short, unaligned,
aliased, impossible to split safely at its segment boundaries, or any
replacement step fails, the host conservatively falls back to `memset` for the
dirty fragments. The 32 MiB threshold is a crossover selected from
measurements on Apple M4, not a correctness constant.

## VA Quarantine: TLBI Completion Does Not Make the Address Reusable

EL1 has made the old VA inaccessible, but host metadata has not yet committed.
If the mmap fast path issued the same address now, an old retire record could
remove the new mapping as well. The address must therefore pass through an
explicit state transition:

```mermaid
stateDiagram-v2
    [*] --> Mapped
    Mapped --> Retired: EL1 clears PTE + TLBI
    Retired --> Retired: host has not drained<br/>VA quarantined
    Retired --> Reusable: metadata/backing commit
    Reusable --> Mapped: mmap allocation
```

While in the `Retired` state:

- the mmap arena allocator must skip the range;
- arena rewind may advance only past committed retire entries;
- host paths such as `MAP_FIXED`, `mprotect`, and `mremap` that may inspect
  metadata or reuse an address must drain first; and
- process teardown must perform a final drain.

If the guest runs for a long time without any natural VM exit, metadata may
remain temporarily without violating correctness. When ring occupancy reaches
31, the next `munmap` falls back to HVC and drains the batch. A separate
**256 MiB materialized pending-byte advisory** sets `cleanup_requested` when
crossed, but does not force the current vCPU to exit. Ring near-full remains
the only hard backpressure from pending retirement volume.

## Synchronization

### Host Writers and EL1 Producers: The PT Gate

Per-vCPU rings remove contention among producers, but the host may still run
`mmap`, `mprotect`, page-table extension, or backing replacement concurrently.
These host writers and EL1 PTE writers coordinate through one shared gate and
one private `producer_active` word per vCPU:

```mermaid
sequenceDiagram
    participant E as EL1 producer
    participant G as shared PT gate
    participant H as host writer

    E->>G: acquire-load gate == 0
    E->>E: release-store producer_active = 1
    E->>G: acquire-load gate again
    alt gate is still 0
        E->>E: validate / clear PTE / TLBI / publish
        E->>E: release-store producer_active = 0
    else host has closed the gate
        E->>E: producer_active = 0, fallback to HVC
    end

    H->>G: acq_rel increment (close)
    H->>E: acquire-wait producer_active == 0
    H->>H: drain / modify page tables
    H->>G: release decrement (open)
```

The second gate read closes the race between checking the gate and announcing
an active producer. The host closes the gate before waiting for every active
producer to leave, thereby acquiring page-table writer exclusion. Producers
do not lock one another because their writes to overlapping descriptors are
idempotent zero stores.

### vCPU A Performs mmap, vCPU B Performs munmap: Cross-Ring Causality

Suppose A fast-mmaps a region and transfers the pointer to B through a
release/acquire handoff, after which B immediately fast-munmaps it. A's mmap
publication and B's retirement reside in different SPSC rings, so the host
cannot drain them in an arbitrary order.

```mermaid
sequenceDiagram
    participant A as vCPU A
    participant B as vCPU B
    participant H as host

    A->>A: publish mmap entry
    A->>B: release/acquire handoff pointer
    B->>B: clear PTE + TLBI
    B->>B: release-publish retire tail
    H->>B: acquire-snapshot retire tail
    H->>A: drain all mmap publications
    H->>H: region metadata now exists
    H->>B: commit snapshotted retirement
```

The host must perform these operations in order:

1. Acquire-snapshot **all** retire tails.
2. Drain **all** mmap publications.
3. Process only the retire entries included in the snapshots.

The acquire of B's tail establishes transitive causal ordering. The host then
creates A's semantic region before removing it. `arena_generation` prevents a
delayed record from damaging a newly provisioned arena.

## Latency and Long Tails

### What `munmap` Timing Actually Includes

```mermaid
flowchart LR
    subgraph Timed[guest-visible munmap latency]
        V[validate] --> D[descriptor scan/clear]
        D --> T[TLBI issue + completion]
        T --> P[ring publish]
    end
    P -. later VM exit .-> C[metadata commit]
    C --> R[backing discard / memset]
```

Fast-path latency still includes two size-dependent costs: descriptor
scan/clear and TLBI completion. It does not include the later region removal,
`memset`, sibling quiescence, or HVF unmap/remap.

Moving this work out of `munmap` does not make it free. Deferred cleanup adds
service time to the next natural VM exit. If the retire ring becomes near-full,
the `munmap` that triggers the fallback also observes the long tail of a batch
drain.

### Why TLBI Has a Long Tail

`DSB ISH` must wait for the inner-shareable invalidation to complete; TLBI is
not merely a fire-and-forget instruction. Its latency is affected by:

- whether sibling vCPUs hold translations for the range;
- the SCALE/NUM encoding and instruction batch count of the range TLBI;
- current activity on the coherency fabric and other cores; and
- macOS preempting the vCPU host thread between the two `CNTVCT` reads.

The first two items are translation-correctness costs that the architecture
requires EL1 to wait for. The final item is wall-clock measurement noise, but
it is still a real long tail observed by the application. Latency therefore
need not increase monotonically with size. For example, 256 MiB may lie at a
SCALE=2 boundary while a larger envelope switches to SCALE=3.

### The Correct Microbenchmark Method

The old H benchmark used adaptive batches: a sample ended when accumulated
counter ticks reached a threshold or accumulated dirty bytes reached a cap.
This allowed the first slow operation to end the sample immediately. For
example, one 1.1 ms long tail at 256 MiB became the entire sample, while the few
1 GiB operations happened not to be preempted and appeared to take only about
2.4 us. This was not a sound comparison of steady-state costs.

The corrected benchmark uses a fixed operation count for each size, records
every latency, and reports p50, the R-7 interpolated p95, and max. One
representative run on Apple M4 produced:

| Size | Fixed ops | p50 | p95 | max |
|------|----------:|----:|----:|----:|
| 64 MiB | 32 | 3.60 us | 16.53 us | 19.07 us |
| 256 MiB | 24 | 5.37 us | 8.13 us | 45.57 us |
| 1 GiB | 8 | 2.01 us | 7.97 us | 8.99 us |

The 1 GiB case uses only eight operations because dirtying 24 GiB page by page
would hit elfuse's 10-second vCPU watchdog. Its p95 should therefore be treated
as directional because of the small sample count. The important distinction
is:

- **p50**: normal EL1 descriptor/TLBI fast-path latency;
- **p95**: common TLBI-completion variation; and
- **max**: rare TLBI, host-scheduling, or ring-fallback long tails.

Dirty stores must remain outside the timed interval. Their page faults are
natural VM exits that drain the previous retirement before the next timed
operation. Section H therefore measures one guest-visible `munmap`, not a
continually accumulating retire ring.

## Correctness Invariants

Implementation and review should continuously verify the following
invariants:

1. No PTE may be modified without a reserved retire slot.
2. The retire tail may not be published, and EL0 may not resume, before the
   broadcast TLBI has completed.
3. Before the host can observe a retire record, both the entry and translation
   retirement must have completed.
4. Every metadata-reading host path must drain first.
5. Mmap publications must commit before snapshotted retirements.
6. A retired VA may not be reallocated before host commit.
7. Before fresh backing replacement, the target must have no live PTE or
   external alias, and sibling vCPUs must be quiesced.
8. Replacement failure must fall back safely and must not leave an HVF segment
   detached.

Together, these invariants distinguish two moments: `munmap()` return means
that the guest translation has been revoked; host drain completion means that
metadata, backing, and VA reuse have all committed.

## Implementation Locations and Validation

On the `elfuse-mmap-eval` branch, the primary code is located in:

- `src/core/mmap-fastpath.h`: per-vCPU mmap/retire-ring ABI, the 32-entry ring,
  and the 256 MiB pending-byte advisory.
- `src/core/shim.S`: EL1 validation, the L2-only scan, descriptor clearing,
  scaled `RVAE1IS`, and release publication.
- `src/syscall/mem.c`: the PT gate, cross-ring drain ordering, metadata commit,
  and the 32 MiB backing-replacement policy.
- `src/syscall/proc.c`: draining after every natural HVF return and before
  dispatch.
- `tests/test-mmap-fastpath.c`: cross-vCPU handoff and immediate SCALE=3 TLBI
  invalidation.
- `tests/test-mmap-lazy.c`: large backing replacement, neighbor preservation,
  zero reuse, and fork snapshots.

In addition to functional tests, section H of `tests/bench-mmap.c` should retain
its fixed operation counts and per-operation latency statistics so that
adaptive sampling cannot again misreport a single long tail as the steady-state
cost for an entire size.

Max042004:
# mmap microbenchmark: elfuse vs. OrbStack

Source: Apple M4 / MacBook Air. Both runs report a 41.67 ns `CNTVCT_EL0` tick.
Values are nanoseconds unless stated otherwise.

## A. Steady-state `mmap` / `munmap` (NULL hint)

| Size | elfuse ops | elfuse mmap | elfuse munmap | OrbStack ops | OrbStack mmap | OrbStack munmap |
| 4 KiB | 4217 | 31.2 | 146.9 | 914 | 184.4 | 267.6 |
| 16 KiB | 4143 | 31.2 | 147.0 | 939 | 176.2 | 265.1 |
| 64 KiB | 4154 | 31.2 | 148.0 | 935 | 173.9 | 272.1 |
| 256 KiB | 4125 | 31.0 | 146.7 | 944 | 173.9 | 279.0 |
| 1 MiB | 4101 | 31.1 | 147.0 | 949 | 173.7 | 324.2 |
| 2 MiB | 4193 | 30.8 | 147.3 | 939 | 173.6 | 246.4 |
| 8 MiB | 4134 | 313.0 | 30.5 | 947 | 173.4 | 252.2 |
| 64 MiB | 3948 | 1841.8 | 33.3 | 896 | 171.9 | 298.6 |
| 256 MiB | 4183 | 30.6 | 140.3 | 935 | 172.4 | 489.5 |
| 1 GiB | 4214 | 310.8 | 30.6 | 936 | 173.3 | 1127.0 |
| 4 GiB | 4135 | 956.1 | 31.9 | 895 | 173.4 | 1140.4 |
| 16 GiB | 4127 | 959.4 | 31.6 | 845 | 186.4 | 1217.0 |
| 32 GiB | 3852 | 1828.4 | 33.2 | 912 | 184.8 | 1229.8 |

## B. Fresh bump-tail `mmap` (`lazy_fresh_range` path)

| Size | Count | elfuse fresh mmap | OrbStack fresh mmap |
|---|---:|---:|---:|
| 4 KiB | 1000 | 105.6 | 205.2 |
| 64 KiB | 1000 | 103.5 | 176.5 |
| 1 MiB | 1000 | 103.8 | 176.0 |
| 2 MiB | 1000 | 103.6 | 183.3 |
| 8 MiB | 256 | 98.1 | 207.5 |
| 128 MiB | 16 | 31.2 | 210.9 |
| 1 GiB | 8 | 31.2 | 239.6 |
| 8 GiB | 4 | 593.8 | 239.6 |
| 32 GiB | 2 | 2020.8 | 250.0 |

## C. First-touch page-fault cost

Touches 512 fresh pages with a 16 KiB stride.

| Metric | elfuse | OrbStack |
|---|---:|---:|
| Median ns/fault | 2223.0 | 456.2 |
| Minimum ns/fault | 2135.3 | 436.3 |

## D. `mprotect` split

Splits a 2 MiB block into L3 entries and protects the middle 4 KiB.

| Metric | elfuse | OrbStack |
|---|---:|---:|
| Median ns | 3416.7 | 458.3 |
| Minimum ns | 3208.3 | 375.0 |

## E. `mremap` grow: 4 KiB to 8 KiB

| Mode | elfuse median | elfuse minimum | OrbStack median | OrbStack minimum |
|---|---:|---:|---:|---:|
| In-place | 2750.0 | 2541.7 | 375.0 | 333.3 |
| Move | 2916.7 | 2791.7 | 375.0 | 333.3 |

## F. Multi-threaded fresh `mmap`

| Size | Threads | elfuse mean | elfuse max | OrbStack mean | OrbStack max |
|---|---:|---:|---:|---:|---:|
| 4 KiB | 2 | 233.7 | 235.3 | 599.1 | 615.8 |
| 4 KiB | 4 | 603.0 | 740.0 | 1875.4 | 2043.0 |
| 2 MiB | 2 | 348.0 | 360.0 | 612.5 | 641.1 |
| 2 MiB | 4 | 776.3 | 800.0 | 1533.5 | 1581.7 |

## G. `munmap` after materializing one 4 KiB page

| Size | elfuse ops | elfuse munmap | OrbStack ops | OrbStack munmap |
|---|---:|---:|---:|---:|
| 4 KiB | 522 | 357.5 | 324 | 512.2 |
| 16 KiB | 333 | 587.1 | 232 | 735.0 |
| 64 KiB | 518 | 330.9 | 254 | 582.7 |
| 256 KiB | 418 | 374.8 | 275 | 560.9 |
| 1 MiB | 312 | 537.1 | 142 | 1265.5 |
| 2 MiB | 234 | 738.6 | 159 | 1065.0 |
| 8 MiB | 230 | 711.1 | 95 | 1045.4 |
| 64 MiB | 253 | 665.8 | 108 | 1623.3 |
| 256 MiB | 508 | 352.1 | 106 | 2101.2 |
| 1 GiB | 450 | 348.3 | 83 | 2167.3 |
| 4 GiB | 513 | 326.5 | 88 | 2025.9 |
| 16 GiB | 530 | 328.0 | 73 | 2414.3 |
| 32 GiB | 463 | 333.6 | 68 | 2311.5 |

## H. `munmap` after dirtying every 4 KiB page

| Size | elfuse ops | elfuse p50 | elfuse p95 | elfuse max | OrbStack ops | OrbStack p50 | OrbStack p95 | OrbStack max |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| 4 KiB | 2048 | 324.5 | 657.8 | 2991.1 | 2048 | 449.5 | 782.8 | 5324.5 |
| 16 KiB | 2048 | 282.8 | 574.5 | 7907.8 | 2048 | 491.1 | 741.1 | 5366.1 |
| 64 KiB | 2048 | 366.1 | 616.1 | 6449.5 | 2048 | 782.8 | 1199.5 | 7949.5 |
| 256 KiB | 1024 | 366.1 | 699.5 | 6824.5 | 1024 | 1657.8 | 2574.5 | 10032.8 |
| 1 MiB | 512 | 532.8 | 824.5 | 3074.5 | 512 | 5491.1 | 7574.5 | 17824.5 |
| 2 MiB | 256 | 741.1 | 1116.1 | 2616.1 | 256 | 9741.1 | 13001.5 | 16407.8 |
| 8 MiB | 128 | 1095.3 | 1949.5 | 2074.5 | 128 | 44199.5 | 51132.8 | 59657.8 |
| 64 MiB | 32 | 3887.0 | 18509.9 | 21282.8 | 32 | 357178.6 | 385453.6 | 418032.8 |
| 256 MiB | 24 | 5720.3 | 8045.3 | 50491.1 | 24 | 1470928.6 | 1670295.3 | 1854991.1 |
| 1 GiB | 8 | 1803.6 | 6024.5 | 6199.5 | 8 | 6277970.3 | 6356257.8 | 6357949.5 |

Max042004:
# elfuse vs. OrbStack: mmap benchmark

`elfuse`: `build/elfuse build/bench-mmap`  
`OrbStack`: `orb ./build/bench-mmap`

Both runs report `CNTVCT 41.67 ns/tick`. Timings are in ns unless marked `us`.
The tables intentionally omit iteration counts to keep the elfuse-vs.-OrbStack
comparison immediately visible.

## A. Steady-state mmap / munmap latency

| Size | mmap: elfuse | mmap: OrbStack | munmap: elfuse | munmap: OrbStack |
|---|---:|---:|---:|---:|
| 4 KiB | 49.8 | 177.7 | 211.6 | 283.6 |
| 16 KiB | 37.8 | 176.0 | 169.7 | 279.9 |
| 64 KiB | 30.2 | 176.6 | 143.6 | 280.0 |
| 256 KiB | 24.4 | 176.2 | 126.2 | 297.8 |
| 1 MiB | 21.6 | 176.1 | 115.9 | 370.4 |
| 2 MiB | 20.9 | 179.5 | 112.4 | 274.7 |
| 8 MiB | 18.2 | 176.2 | 103.2 | 262.9 |
| 64 MiB | 18.0 | 175.5 | 102.4 | 313.3 |
| 256 MiB | 17.8 | 177.6 | 101.8 | 401.2 |
| 1 GiB | 17.9 | 178.9 | 102.6 | 403.8 |
| 4 GiB | 197.8 | 174.8 | 16.8 | 390.9 |
| 16 GiB | 577.7 | 192.9 | 18.2 | 421.9 |
| 32 GiB | 1083.2 | 181.3 | 19.5 | 477.1 |

## B. Fresh bump-tail mmap

| Size | elfuse | OrbStack |
|---|---:|---:|
| 4 KiB | 90.0 | 180.0 |
| 64 KiB | 89.7 | 173.9 |
| 1 MiB | 89.8 | 188.5 |
| 2 MiB | 89.8 | 180.2 |
| 8 MiB | 80.2 | 211.8 |
| 128 MiB | 39.1 | 244.8 |
| 1 GiB | 36.5 | 244.8 |
| 8 GiB | 583.3 | 239.6 |
| 32 GiB | 1875.0 | 854.2 |

## C. First-touch fault cost

| Metric | elfuse | OrbStack |
|---|---:|---:|
| Sweep p50 | 79.4 us | 240.0 us |
| Sweep p95 | 734.3 us | 457.7 us |
| Sweep max | 760.6 us | 477.1 us |
| Amortized/touch p50 | 155.0 | 468.8 |
| Amortized/touch p95 | 1434.1 | 893.9 |
| Amortized/touch max | 1485.5 | 931.9 |

## D. mprotect split: 2 MiB block to L3, middle 4 KiB

| Operation | elfuse median | OrbStack median | elfuse min | OrbStack min |
|---|---:|---:|---:|---:|
| Split | 791.7 | 1708.3 | 291.7 | 958.3 |
| Fast leaf | 375.0 | 500.0 | 291.7 | 458.3 |

## E. mremap: 4 KiB to 8 KiB

| Operation | elfuse median | OrbStack median | elfuse min | OrbStack min |
|---|---:|---:|---:|---:|
| In-place grow | 1625.0 | 375.0 | 1500.0 | 333.3 |
| Move | 2041.7 | 375.0 | 1500.0 | 333.3 |

## F. Multi-threaded fresh mmap

| Size | Threads | elfuse mean | OrbStack mean | elfuse max | OrbStack max |
|---|---:|---:|---:|---:|---:|
| 4 KiB | 2 | 206.3 | 591.1 | 229.3 | 615.8 |
| 4 KiB | 4 | 736.3 | 386.7 | 898.7 | 591.7 |
| 2 MiB | 2 | 308.7 | 627.3 | 329.3 | 636.9 |
| 2 MiB | 4 | 901.3 | 1029.0 | 1065.3 | 1180.2 |

## G. munmap after materializing one 4 KiB page

| Size | elfuse | OrbStack |
|---|---:|---:|
| 4 KiB | 310.0 | 422.9 |
| 16 KiB | 360.1 | 391.8 |
| 64 KiB | 246.4 | 403.6 |
| 256 KiB | 293.6 | 433.5 |
| 1 MiB | 479.6 | 786.8 |
| 2 MiB | 615.0 | 962.6 |
| 8 MiB | 626.4 | 902.8 |
| 64 MiB | 285.6 | 1037.2 |
| 256 MiB | 289.8 | 1218.0 |
| 1 GiB | 328.1 | 1833.1 |
| 4 GiB | 297.1 | 2329.7 |
| 16 GiB | 289.5 | 1847.2 |
| 32 GiB | 287.8 | 2531.0 |

## H. munmap after dirtying every 4 KiB page

| Size | p50: elfuse | p50: OrbStack | p95: elfuse | p95: OrbStack | max: elfuse | max: OrbStack |
| 4 KiB | 199.5 | 490.8 | 449.5 | 824.2 | 9574.5 | 6240.8 |
| 16 KiB | 241.1 | 449.2 | 616.1 | 740.8 | 8699.5 | 15324.2 |
| 64 KiB | 282.8 | 740.8 | 699.5 | 907.5 | 13491.1 | 11574.2 |
| 256 KiB | 282.8 | 1657.5 | 407.8 | 2199.2 | 2366.1 | 11282.5 |
| 1 MiB | 491.1 | 5449.2 | 843.2 | 6426.2 | 11616.1 | 32115.8 |
| 2 MiB | 532.8 | 9740.8 | 616.1 | 10730.4 | 12032.8 | 22365.8 |
| 8 MiB | 574.5 | 44095.0 | 1003.6 | 47961.7 | 2241.1 | 58865.8 |
| 64 MiB | 824.5 | 368074.2 | 2449.5 | 401917.9 | 3449.5 | 406657.5 |
| 256 MiB | 1928.6 | 1427011.7 | 3730.7 | 1467692.9 | 4699.5 | 1643324.2 |
| 1 GiB | 2866.1 | 5997178.3 | 5132.8 | 6121607.5 | 5657.8 | 6146865.8 |

sunxiaoguang:
When starting a Spring service with JVM ZGC under elfuse, the startup phase can make the entire host system extremely sluggish or temporarily unresponsive.

ZGC reserves large virtual address ranges during initialization, while only a small fraction of those pages are immediately accessed. elfuse currently performs substantial work for the whole mapping up front, even when the mapping is only an address-space reservation.

This is not specific to ZGC. The same issue can affect any managed runtime that reserves large sparse anonymous mappings.

Hopefully this PR could help with this issue.

jserv (changes requested):
Rebase latest `main` branch as #272 introduces Frama-C proofs.

jserv:
Close in favor of #372

PR #206, opened 2026-07-15
proc-identity fix saved-set-ID update in setreuid and setregid

### LINKED ISSUE #140: `setreuid` / `setregid` Saved ID is Not Updated on Real ID Change when `e == -1`
### Description
In [proc-identity.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/proc-identity.c), the macro `DEFINE_SETRE` handles setting the real and effective IDs:
```c
        if (e != (uint32_t) -1) {
            eff = e;
            if (r != (uint32_t) -1 || e != old_real)
                saved = e;
        }
```
If `e == -1` but `r != -1`, the entire `if (e != (uint32_t) -1)` block is skipped.

According to Linux behavior, if the real UID/GID is set, the saved ID must be updated to the new effective ID, even if the effective ID itself was not changed (`e == -1`).

### Impact
Applications that rotate/swap real and effective IDs (a standard pattern in security-sensitive programs using `setreuid`) will experience incorrect saved-set-user-ID state, leading to permissions discrepancies or unexpected `EPERM` failures on subsequent calls.

doanbaotrung:
sys: guard saved ID update in setreuid/setregid with euid/egid != -1

Modify DEFINE_SETRE macro to ensure that the saved-set-ID is only
updated by the effective ID change check when the new effective ID
is actually being set (i.e. e != -1).

This prevents setreuid(-1, -1) and setregid(-1, -1) from spuriously
clobbering the saved-set-ID when the current effective ID differs
from the real ID, matching Linux behavior.

Fix #140 

---
## Summary by cubic
Fix saved-set-ID handling in setreuid(2) and setregid(2) to match Linux rules. The saved ID now updates from the new effective ID after applying real/effective changes, including when only the real ID is set.

- **Bug Fixes**
  - Updated `DEFINE_SETRE` to compute `saved` from the final `eff` after independently applying `r` and `e`.
  - Fixes the missed setreuid(r, -1) path; same change applies to setregid. Fixes #140.

<sup>Written for commit a6d01d9d4bfc060bd0122488aa8b9408c32c8581. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/206?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/proc-identity.c
+++ b/src/syscall/proc-identity.c
@@ -131,22 +131,28 @@ int64_t proc_sys_setgid(uint32_t gid)
     return 0;
 }
 
-#define DEFINE_SETRE(suffix, real, eff, saved, perm_fn)    \
-    int64_t proc_sys_setre##suffix(uint32_t r, uint32_t e) \
-    {                                                      \
-        uint32_t old_real = real;                          \
-        if (r != (uint32_t) -1 && r != real && r != eff)   \
-            return -LINUX_EPERM;                           \
-        if (e != (uint32_t) -1 && !perm_fn(e))             \
-            return -LINUX_EPERM;                           \
-        if (r != (uint32_t) -1)                            \
-            real = r;                                      \
-        if (e != (uint32_t) -1) {                          \
-            eff = e;                                       \
-            if (r != (uint32_t) -1 || e != old_real)       \
-                saved = e;                                 \
-        }                                                  \
-        return 0;                                          \
+#define DEFINE_SETRE(suffix, real, eff, saved, perm_fn)                    \
+    int64_t proc_sys_setre##suffix(uint32_t r, uint32_t e)                 \
+    {                                                                      \
+        uint32_t old_real = real;                                          \
+        if (r != (uint32_t) -1 && r != real && r != eff)                   \
+            return -LINUX_EPERM;                                           \
+        if (e != (uint32_t) -1 && !perm_fn(e))                             \
+            return -LINUX_EPERM;                                           \
+        if (r != (uint32_t) -1)                                            \
+            real = r;                                                      \
+        if (e != (uint32_t) -1)                                            \
+            eff = e;                                                       \
+        /* Linux setreuid(2): saved-set-ID is updated to                   \
+         * the new effective ID when the real ID is set                    \
+         * (r != -1) OR when the new effective ID differs                  \
+         * from the old real ID.  Both conditions use the                  \
+         * *new* effective value (eff, which may be                        \
+         * unchanged if e == -1).                                          \
+         */                                                                \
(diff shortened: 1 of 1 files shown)


src/syscall/proc-identity.c
```diff
@@ -141,11 +141,17 @@ int64_t proc_sys_setgid(uint32_t gid)
...
-                saved = e;                                 \
-        }                                                  \
+        /* Linux setreuid(2): saved-set-ID is updated to   \
+         * the new effective ID when the real ID is set    \
+         * (r != -1) OR when the new effective ID differs  \
+         * from the old real ID.  Both conditions use the  \
+         * *new* effective value (eff, which may be        \
+         * unchanged if e == -1).                          \
+         */                                                \
+        if (r != (uint32_t) -1 || eff != old_real)         \
```
jserv: Saved-set-ID condition drops the `e != -1` guard.

This clause fires even when the effective ID is not being changed. When `e == -1`, `eff` still holds the *current, unchanged* effective ID; if it differs from the real ID, `saved` gets spuriously rewritten. So `setreuid(-1, -1)` on a process whose `euid != ruid` (e.g. `ruid=1000, euid=0, suid=500` reached via a prior `setresuid`) clobbers the saved-set-ID even though a `(-1, -1)` call must be a no-op — contradicting this PR's own description table (`setreuid(-1,-1) -> no change`). This is privilege-relevant because `uid_is_permitted`/`gid_is_permitted` later key off `emu_suid`/`emu_sgid`.

Linux (`kernel/sys.c` `__sys_setreuid`, and the `setreuid(2)` man page) guards the second clause with `euid != -1`:

```c
if (r != (uint32_t) -1 || (e != (uint32_t) -1 && eff != old_real))
    saved = eff;
```

The core intent of the PR — updating the saved-set-ID for `setreuid(r, -1)` — is correct; this is the one remaining defect, and it applies to `setregid` too via the shared macro.

doanbaotrung: Fixed.

Modify DEFINE_SETRE macro to ensure that the saved-set-ID is only
updated by the effective ID change check when the new effective ID
is actually being set (i.e. e != -1).

This prevents setreuid(-1, -1) and setregid(-1, -1) from spuriously
clobbering the saved-set-ID when the current effective ID differs
from the real ID, matching Linux behavior.

jserv:

PR #207, opened 2026-07-15
Intercept fstatfs and synthesize procfs magic for all /proc files

### LINKED ISSUE #142: Missing `proc_path` for Dynamic/Non-Stateful `procfs` Files in `sys_fstatfs`
### Description
In `sys_fstatfs` ([fs-stat.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/fs-stat.c#L404)), the implementation verifies if the snapshot's `proc_path` points to a `/proc` prefix:

```c
int64_t sys_fstatfs(guest_t *g, int fd, uint64_t buf_gva)
{
    fd_entry_t snap;
    memset(&snap, 0, sizeof(snap));
    bool have_snap = fd_snapshot(fd, &snap);
    if (have_snap &&
        (!strcmp(snap.proc_path, "/proc") ||
         !strncmp(snap.proc_path, "/proc/", 6))) {
        linux_statfs_t proc_st;
        fill_proc_statfs(&proc_st);
        if (guest_write_small(g, buf_gva, &proc_st, sizeof(proc_st)) < 0)
            return -LINUX_EFAULT;
        return 0;
    }
```

However, the `proc_path` metadata is only set on file descriptors during `openat` if `resolve_virtual_path` returning true. `resolve_virtual_path` ([fs.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/fs.c#L133)) is currently constrained to virtual directories and stateful files:

```c
static bool resolve_virtual_path(const char *path, char *out, size_t out_size)
{
    ...
    if (strncmp(path, "/proc", 5) != 0)
        return false;

    char virt_buf[64];
    const char *virt = proc_virtual_dir_path(path, virt_buf, sizeof(virt_buf));
    if (!virt)
        virt = proc_stateful_file_path(path);
    if (!virt)
        return false;

    str_copy_trunc(out, virt, out_size);
    return true;
}
```

### Problem & Impact
For standard/dynamic read-only `proc` files like `/proc/self/maps`, `/proc/uptime`, `/proc/loadavg`, etc., `resolve_virtual_path` returns `false`, causing the descriptor's `proc_path` in `fd_table` to remain empty (`""`). 

When calling `fstatfs` on these open file descriptors, `sys_fstatfs` fails to detect that they are `procfs` files and falls back to running `fstatfs` on the host descriptor (which is a temporary backing file in `/tmp`). This returns the host volume's filesystem magic (e.g. APFS/HFS+ on macOS) instead of `LINUX_PROC_SUPER_MAGIC`.

doanbaotrung:
When fstatfs is called on open file descriptors pointing to /proc files (e.g., /proc/self/maps, /proc/uptime, /proc/loadavg), it returned the host filesystem magic (e.g. APFS) instead of LINUX_PROC_SUPER_MAGIC (0x9fa0).

Fix this by:
- Modifying resolve_virtual_path in fs.c to resolve and normalize all paths under /proc (not just virtual directories/stateful files) and record them in the fd table's proc_path field.
- Introducing a fill_proc_statfs helper in fs-stat.c that populates a static linux_statfs_t layout with PROC_SUPER_MAGIC (0x9fa0) and zero blocks/free space.
- Intercepting both sys_statfs and sys_fstatfs when the target path or fd's proc_path points to /proc, and directly writing the synthesized procfs metadata back to the guest.

Fix #142

---
## Summary by cubic
Return the correct procfs magic for `/proc` files by intercepting `statfs`/`fstatfs` and normalizing `/proc` paths, while skipping procfs symlinks so only real proc entries report procfs. Fixes #142.

- **Bug Fixes**
  - Normalize and record `/proc` paths in `resolve_virtual_path`, mapping `/proc/<pid>` (for the current pid) to `/proc/self`, and exclude procfs symlinks (`/proc/self/{exe,cwd,root}`, `/proc/self/fd/<n>`, and task equivalents) so they report the target filesystem.
  - Add `fill_proc_statfs` to synthesize a `linux_statfs_t` with `PROC_SUPER_MAGIC` (0x9fa0), 4K block size, and zeroed counts.
  - Intercept `sys_statfs` and `sys_fstatfs` for `/proc` targets and return the synthesized procfs stats.

<sup>Written for commit 9718690712aee86dd27b529885f8f4470e9f2821. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/207?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/fs-stat.c
+++ b/src/syscall/fs-stat.c
@@ -367,6 +367,20 @@ static bool statfs_path_is_proc(const char *path)
     return !strncmp(path, "/proc", 5) && (path[5] == '\0' || path[5] == '/');
 }
 
+static void fill_proc_statfs(linux_statfs_t *lin)
+{
+    memset(lin, 0, sizeof(*lin));
+    lin->f_type = 0x9fa0; /* PROC_SUPER_MAGIC */
+    lin->f_bsize = 4096;
+    lin->f_blocks = 0;
+    lin->f_bfree = 0;
+    lin->f_bavail = 0;
+    lin->f_files = 0;
+    lin->f_ffree = 0;
+    lin->f_namelen = 255;
+    lin->f_frsize = 4096;
+}
+
 int64_t sys_statfs(guest_t *g, uint64_t path_gva, uint64_t buf_gva)
 {
     char path[LINUX_PATH_MAX];
@@ -379,37 +393,17 @@ int64_t sys_statfs(guest_t *g, uint64_t path_gva, uint64_t buf_gva)
     if (tx.fuse_path)
         return -LINUX_ENOSYS;
 
-    struct statfs mac_st;
-
-    /* /proc has no host-filesystem counterpart under the sysroot: every entry
-     * is synthesized on open (see proc_intercept_open), so a raw statfs() on
-     * the guest path always misses with ENOENT even for "/proc" itself. Try
-     * the lightweight statfs-specific resolver first (it shortcuts nodes whose
-     * open path would otherwise allocate scratch state statfs never uses),
-     * then fall back to the same intercept open/fstatfs/close that
-     * sys_newfstatat uses via proc_intercept_stat.
-     */
     if (statfs_path_is_proc(tx.intercept_path)) {
-        int intercepted = proc_intercept_statfs(tx.intercept_path, &mac_st);
-        if (intercepted == PROC_NOT_INTERCEPTED) {
-            int host_fd = proc_intercept_open(g, tx.intercept_path, 0, 0);
-            if (host_fd == PROC_NOT_INTERCEPTED) {
-                if (statfs(tx.host_path, &mac_st) < 0)
-                    return linux_errno();
-            } else if (host_fd < 0) {
-                return linux_errno();
-            } else {
-                int rc = fstatfs(host_fd, &mac_st);
-                close_keep_errno(host_fd);
-                if (rc < 0)
-                    return linux_errno();
-            }
-        } else if (intercepted < 0) {
-            return linux_errno();
-        }
-    } else if (statfs(tx.host_path, &mac_st) < 0) {
-        return linux_errno();
+        linux_statfs_t lin_st;
+        fill_proc_statfs(&lin_st);
+        if (guest_write_small(g, buf_gva, &lin_st, sizeof(lin_st)) < 0)
+            return -LINUX_EFAULT;
+        return 0;
     }
+
+    struct statfs mac_st;
+    if (statfs(tx.host_path, &mac_st) < 0)
+        return linux_errno();
 
     linux_statfs_t lin_st;
     translate_statfs(&mac_st, &lin_st);
@@ -421,6 +415,16 @@ int64_t sys_statfs(guest_t *g, uint64_t path_gva, uint64_t buf_gva)
 
 int64_t sys_fstatfs(guest_t *g, int fd, uint64_t buf_gva)
 {
+    fd_entry_t snap;
+    memset(&snap, 0, sizeof(snap));
+    if (fd_snapshot(fd, &snap) && statfs_path_is_proc(snap.proc_path)) {
+        linux_statfs_t proc_st;
+        fill_proc_statfs(&proc_st);
+        if (guest_write_small(g, buf_gva, &proc_st, sizeof(proc_st)) < 0)
+            return -LINUX_EFAULT;
+        return 0;
+    }
+
     host_fd_ref_t host_ref;
     if (host_fd_ref_open(fd, &host_ref) < 0)
         return -LINUX_EBADF;
--- a/src/syscall/fs.c
+++ b/src/syscall/fs.c
@@ -127,6 +127,43 @@ static const char *proc_stateful_file_path(const char *path)
     return NULL;
 }
 
+static bool proc_path_is_symlink(const char *path)
+{
+    if (!path)
+        return false;
+
+    if (!strcmp(path, "/proc/self/exe") || !strcmp(path, "/proc/self/cwd") ||
+        !strcmp(path, "/proc/self/root")) {
+        return true;
+    }
+
+    if (!strncmp(path, "/proc/self/fd/", 14)) {
+        char *endp;
+        long n = strtol(path + 14, &endp, 10);
+        if (endp != path + 14 && *endp == '\0' && n >= 0)
+            return true;
+    }
+
+    if (!strncmp(path, "/proc/self/task/", 16)) {
+        char *endp;
+        strtol(path + 16, &endp, 10);
+        if (endp != path + 16 && *endp == '/') {
+            const char *sub = endp + 1;
+            if (!strcmp(sub, "exe") || !strcmp(sub, "cwd") ||
+                !strcmp(sub, "root")) {
+                return true;
+            }
+            if (!strncmp(sub, "fd/", 3)) {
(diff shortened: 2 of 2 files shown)


src/syscall/fs-stat.c
```diff
@@ -379,38 +393,18 @@ int64_t sys_statfs(guest_t *g, uint64_t path_gva, uint64_t buf_gva)
...
-
-    /* /proc has no host-filesystem counterpart under the sysroot: every entry
-     * is synthesized on open (see proc_intercept_open), so a raw statfs() on
-     * the guest path always misses with ENOENT even for "/proc" itself. Try
-     * the lightweight statfs-specific resolver first (it shortcuts nodes whose
-     * open path would otherwise allocate scratch state statfs never uses),
-     * then fall back to the same intercept open/fstatfs/close that
-     * sys_newfstatat uses via proc_intercept_stat.
-     */
     if (statfs_path_is_proc(tx.intercept_path)) {
```
jserv, line 396: statfs on a nonexistent /proc path now returns success instead of ENOENT. statfs_path_is_proc() matches any /proc-prefixed path, so this block fills PROC_SUPER_MAGIC and returns 0 without confirming the entry exists. The previous code fell through proc_intercept_statfs -> proc_intercept_open -> statfs(host_path), so a miss propagated ENOENT.

Trigger: statfs("/proc/bogus", &buf) or statfs("/proc/self/not_a_file", &buf) -- real Linux and pre-PR return -ENOENT; this returns 0 with f_type=0x9fa0. Programs that use statfs as an existence/feature probe are misled.

Fix: gate synthesis on a positive existence check (only synthesize when proc_intercept_stat(tx.intercept_path, ...) recognizes the path, otherwise fall through to the host statfs that yields ENOENT). Only f_type needed to change; existence semantics should be preserved.

src/syscall/fs.c
```diff
@@ -149,11 +186,32 @@ static bool resolve_virtual_path(const char *path, char *out, size_t out_size)
...
-        return false;
 
-    str_copy_trunc(out, virt, out_size);
-    return true;
+    if (virt) {
+        str_copy_trunc(out, virt, out_size);
+        return true;
+    }
+
+    /* If it has a valid /proc prefix, normalize it and record it. */
```
jserv, line 195: Stamping proc_path for every non-symlink /proc/* entry -- including regular files like /proc/cpuinfo and /proc/self/maps -- leaks dirfd semantics into consumers that key off proc_path. dirfd_guest_base_path (path.c:883) accepts any fd with a non-empty proc_path as a resolution base and returns before the snap.type != FD_DIR -> ENOTDIR guard at path.c:909 -- unlike resolve_proc_dirfd_path (path.c:522), which requires FD_DIR.

Trigger: open /proc/cpuinfo as fd 5, then openat(5, "x", ...). Real Linux returns ENOTDIR; this builds /proc/cpuinfo/x and resolves it (wrong errno / wrong target). It also reroutes plain fstat on such fds through proc_intercept_stat.

Fix: only stamp proc_path for /proc entries that are directories (the cases where the fd is a legitimate dirfd), or record a separate proc-mount flag used solely by the statfs path rather than overloading proc_path, which carries dirfd-resolution meaning.

src/syscall/fs-stat.c
```diff
@@ -367,6 +367,20 @@ static bool statfs_path_is_proc(const char *path)
     return !strncmp(path, "/proc", 5) && (path[5] == '\0' || path[5] == '/');
 }
 
+static void fill_proc_statfs(linux_statfs_t *lin)
```
jserv, line 370: The linux_statfs_t layout here is complete and correct (all fields covered by memset(0); f_type/f_bsize/f_namelen/f_frsize consistent with a real procfs). No regression test asserts the magic, though -- the existing tests only compare statfs vs fstatfs, not the value 0x9fa0. Worth adding one assertion for f_type == 0x9fa0 on both statfs("/proc/...") and fstatfs(open("/proc/...")).

Max042004:
The new `proc_path_is_symlink()` exclusion only feeds `resolve_virtual_path()`
in `fs.c`, which is what stamps a *file descriptor's* `proc_path` at open
time. `sys_statfs()` (the path-based syscall, `fs-stat.c:384`) never goes
through that function — it gates purely on
`statfs_path_is_proc(tx.intercept_path)` (`fs-stat.c:363`), a plain `/proc`
prefix check with no symlink exclusion

```c
if (statfs_path_is_proc(tx.intercept_path)) {
    linux_statfs_t lin_st;
    fill_proc_statfs(&lin_st);
    if (guest_write_small(g, buf_gva, &lin_st, sizeof(lin_st)) < 0)
        return -LINUX_EFAULT;
    return 0;
}
```

This replaced the previous fallback chain that called `proc_intercept_open()`
+ real `fstatfs()` on paths `proc_intercept_statfs()` didn't own — which is
exactly what used to make `/proc/self/exe` report the real backing
filesystem, since `statfs(2)` (unlike `lstat`) always follows the symlink on
real Linux.

Max042004:
`sys_statfs()` no longer calls it (see #1), and nothing else in the tree does
either — it only calls itself recursively via the `/proc/<pid>` self-alias
branch, which is now unreachable. It can be deleted

src/syscall/fs-stat.c
doanbaotrung: **Fix:** I modified `sys_statfs` to execute a positive existence check first:
  1. Check via `proc_intercept_stat` if it is a synthetic `/proc` node.
  2. If `proc_intercept_stat` returns `PROC_NOT_INTERCEPTED`, check if the path exists on the host's sysroot mapping via `stat(tx.host_path)`.
  3. Only synthesize `PROC_SUPER_MAGIC` statfs if either check confirms the file/directory exists; otherwise, fall through to let host `statfs` yield `ENOENT`.

src/syscall/fs.c
doanbaotrung: **Fix:** I updated `dirfd_guest_base_path` to assert that the file descriptor is of type `FD_DIR`. If a regular emulated file descriptor under `/proc` (e.g. `/proc/cpuinfo`) is passed as a `dirfd`, it now correctly throws `ENOTDIR` and returns `-1` (matching real Linux). This resolves the leaking `dirfd` semantics while allowing `fstat` and emulated read/write slow paths to continue referencing `proc_path` for regular files.

src/syscall/fs-stat.c
doanbaotrung: **Fix:** Added test assertions verifying that `st.f_type == 0x9fa0` (representing `PROC_SUPER_MAGIC`) for `/proc`, `/proc/` (trailing slash), and `/proc/self/cmdline` under both `statfs` and `fstatfs`.

doanbaotrung:
### Cleanup of `proc_intercept_statfs` (procemu.c / procemu.h)
* **Fix:** Since `sys_statfs` now performs direct existence resolution and no longer calls `proc_intercept_statfs`, and nothing else in the tree calls it, I completely deleted the declaration and definition of `proc_intercept_statfs`.

jserv:

PR #210, opened 2026-07-16
Share one path component iterator across syscall/

henrybear327:
path.c and sidecar.c each carry a private copy of the same '/'-separated component walk, differing only in name and return type.

Promote path.c's path_next_component to external linkage and switch sidecar.c to it so the syscall layer keeps a single walker and future path predicates do not grow another copy.

---
## Summary by cubic
Deduplicated path component iteration in the syscall layer by exporting `path_next_component` and switching `sidecar.c` to use it. This keeps one '/'-component walker and aligns behavior across callers with no functional changes.

- **Refactors**
  - Exposed `path_next_component(const char **pathp, const char **comp, size_t *len)` in `path.c` and declared it in `path.h` with docs.
  - Removed `sidecar_next_component` and updated call sites in `sidecar_translate_lookup_at` to use the shared helper.
  - Preserved semantics: skips repeated slashes and returns false at end of path.

<sup>Written for commit 3451457be71c69896c6d7894477ef3a5a03d7112. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/210?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/path.c
+++ b/src/syscall/path.c
@@ -284,9 +284,7 @@ int path_translate_dirent_name(guest_fd_t dirfd,
     return 0;
 }
 
-static bool path_next_component(const char **pathp,
-                                const char **comp,
-                                size_t *len)
+bool path_next_component(const char **pathp, const char **comp, size_t *len)
 {
     const char *p = *pathp;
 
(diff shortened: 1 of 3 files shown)


jserv:

PR #212, opened 2026-07-17
Fix /tmp cleanup and verify absent parent ENOENT

### LINKED ISSUE #187: Path containment checks fail with `ELOOP` on non-existent parent directories
### Description
When a guest process queries or attempts to create a file in a parent directory that does not exist yet (e.g., calling `open(..., O_CREAT)` on `/etc/sysusers.d/basic.conf` where `/etc/sysusers.d` is absent), the syscall is incorrectly aborted with an `ELOOP` ("Too many levels of symbolic links") error instead of returning `ENOENT`. 

This causes standard Linux tools (like `systemd-sysusers` or package post-install scripts) to crash because they expect an `ENOENT` error (which they can handle by creating the parent directory) but encounter `ELOOP` instead.

### Steps to Reproduce
Run any guest command under `elfuse` that attempts to open, check, or create a path whose parent directory does not exist, e.g.:
```bash
# Assuming /etc/sysusers.d does not exist in the sysroot
touch /etc/sysusers.d/basic.conf
```
* **Expected behavior:** Syscall fails with `ENOENT`.
* **Actual behavior:** Syscall fails with `ELOOP`.

---

### Root Cause
In [proc-state.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/proc-state.c#L445-L508), `sysroot_path_is_contained` calls `realpath` on the resolved host path to verify it sits within the guest `sysroot`:

```c
if (follow_final) {
    if (!realpath(resolved_path, real_path))
        return false; // <--- FAILS HERE on ENOENT/ENOTDIR
}
```

Since the parent directory does not exist, `realpath` fails on the host and sets `errno = ENOENT`. However, `sysroot_path_is_contained` treats any `realpath` failure as a containment violation and returns `false`. `elfuse` then maps containment violations to `ELOOP` to abort the syscall, hiding the legitimate `ENOENT` error.

---

### Proposed Solution
If `realpath` fails with `ENOENT` or `ENOTDIR`, the containment check should recursively walk up the directory tree until it finds the closest existing parent directory, verify its containment, and permit the syscall if the parent is contained.

#### Suggested Patch:
Add a parent-recursive lookup helper in `src/syscall/proc-state.c`:

```c
static bool realpath_existing_parent(const char *path, char *resolved_path)
{
    char temp[LINUX_PATH_MAX];
    str_copy_trunc(temp, path, sizeof(temp));

    while (1) {
        if (realpath(temp, resolved_path)) {
            return true;
        }
        if (errno != ENOENT && errno != ENOTDIR) {
            return false;
        }
        char *slash = strrchr(temp, '/');
        if (!slash) {
            return false;
        }
        if (slash == temp) {
            str_copy_trunc(temp, "/", sizeof(temp));
            if (realpath(temp, resolved_path)) {
                return true;
            }
            return false;
        }
        *slash = '\0'; // Move up to parent directory
    }
}
```

And update `sysroot_path_is_contained` to fallback to this helper when `realpath` returns `ENOENT` or `ENOTDIR`:

```c
if (follow_final) {
    if (!realpath(resolved_path, real_path)) {
        if (errno == ENOENT || errno == ENOTDIR) {
            if (!realpath_existing_parent(resolved_path, real_path))
                return false;
        } else {
            return false;
        }
    }
}
```

doanbaotrung:
Fix /tmp cleanup and verify absent parent ENOENT

A preceding test left ${sysroot}/tmp as a symlink to host-out, causing
downstream creation tests to hit the symlink-rejecting directory
creation path and fail with ELOOP. We now restore the /tmp directory
at the end of the test.

Since containment checks are only run on paths verified to exist,
checks on absent intermediate directories only occur via a TOCTOU race.
The recursive parent resolution walk-up in sysroot_path_is_contained
has been reverted to a secure, fail-closed behavior to prevent any
fail-open vulnerabilities.

A deterministic test is added to verify that opening a path with a
non-existent parent returns ENOENT (and not ELOOP). The TOCTOU racer
is kept as a non-fatal stress check, asserting that no artifacts are
written outside of the sysroot mount.

Fix #187

---
## Summary by cubic
Return ENOENT/ENOTDIR for paths with missing or non-directory components instead of ELOOP, and harden tests to prevent false ELOOPs. Fixes #187.

- **Bug Fixes**
  - Fail-closed: only validate containment on paths that exist; remove parent-walk fallback to avoid fail-open behavior.
  - Tests: added deterministic ENOENT check for a missing parent; TOCTOU stress ensures no ELOOP and no artifacts outside the sysroot.
  - Cleanup: restore `/tmp` inside the sysroot after tests to remove a leftover symlink that caused spurious ELOOPs.

<sup>Written for commit 3066dd4dc9a712053af0f0d38411332b9f216c66. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/212?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/proc-state.c
+++ b/src/syscall/proc-state.c
@@ -441,7 +441,6 @@ bool proc_sysroot_casefold_enabled(void)
     pthread_mutex_unlock(&sysroot_lock);
     return enabled;
 }
-
 /* Confirm @resolved_path canonicalizes inside @sysroot. This is a
  * check-then-use sequence: callers issue the actual syscall after this returns,
  * so a symlink swapped in between will not be re-validated. openat2
--- a/tests/test-sysroot-create-paths.c
+++ b/tests/test-sysroot-create-paths.c
@@ -7,10 +7,12 @@
 
 #include <errno.h>
 #include <fcntl.h>
+#include <signal.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <sys/wait.h>
 #include <sys/xattr.h>
 #include <unistd.h>
 
@@ -293,7 +295,81 @@ int main(int argc, char **argv)
                 PASS();
             }
         }
-    unlink_symlink_escape_done:;
+    unlink_symlink_escape_done:
+        unlink(sysroot_tmp);
+        mkdir(sysroot_tmp, 0755);
+    }
+
+    TEST("open file in non-existent parent returns ENOENT, not ELOOP");
+    {
+        int fd = open("/etc/nonexistent_parent_dir_123/sub/file.txt",
+                      O_CREAT | O_WRONLY, 0644);
+        EXPECT_ERRNO(fd, ENOENT,
+                     "open on non-existent parent failed to return ENOENT");
+        if (fd >= 0)
+            close(fd);
+    }
+
+    TEST("open file in non-existent parent TOCTOU race stress pass");
+    {
+        char race_dir[512];
+        char race_file[512];
+        snprintf(race_dir, sizeof(race_dir), "%s/tmp/race_dir",
+                 mounted_sysroot_root);
+        snprintf(race_file, sizeof(race_file), "/tmp/race_dir/file.txt");
+
+        pid_t pid = fork();
+        if (pid < 0)
+            FAIL("fork failed");
+        else if (pid == 0) {
+            /* Child process: continuously create and delete the directory */
+            while (1) {
+                mkdir(race_dir, 0755);
+                rmdir(race_dir);
+            }
+            _exit(0);
+        }
+
+        int hit_eloop = 0;
+        int iterations = 0;
+        while (iterations < 10000) {
+            iterations++;
+            int fd = open("/tmp/race_dir/file.txt", O_CREAT | O_WRONLY, 0644);
+            if (fd >= 0) {
+                close(fd);
+                unlink("/tmp/race_dir/file.txt");
+            } else if (errno == ELOOP) {
+                hit_eloop = 1;
+                break;
+            }
+        }
+
+        kill(pid, SIGKILL);
+        int status;
+        waitpid(pid, &status, 0);
+
+        /* Cleanup guest paths */
+        unlink("/tmp/race_dir/file.txt");
+        rmdir("/tmp/race_dir");
+
+        /* Cleanup host paths inside sysroot */
+        char host_race_file[512], host_race_dir[512];
+        snprintf(host_race_file, sizeof(host_race_file),
+                 "%s/tmp/race_dir/file.txt", mounted_sysroot_root);
+        snprintf(host_race_dir, sizeof(host_race_dir), "%s/tmp/race_dir",
+                 mounted_sysroot_root);
+        unlink(host_race_file);
+        rmdir(host_race_dir);
+
+        /* Assert that no artifact exists outside mounted_sysroot_root on the
+         * host */
+        struct stat host_st;
+        if (stat("/tmp/race_dir", &host_st) == 0)
+            FAIL("TOCTOU race created directory outside of sysroot");
+        else if (hit_eloop)
+            FAIL("TOCTOU race returned ELOOP");
+        else
+            PASS();
     }
 
     SUMMARY("test-sysroot-create-paths");


src/syscall/proc-state.c
```diff
@@ -449,6 +449,29 @@ bool proc_sysroot_casefold_enabled(void)
  * the legacy *at() surface this is best-effort defense at the boundary; the
  * guest is in the host user's trust domain.
  */
+static bool realpath_existing_parent(const char *path, char *resolved_path)
```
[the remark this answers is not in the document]
doanbaotrung: Updated

tests/test-sysroot-create-paths.c
```diff
@@ -295,6 +295,15 @@ int main(int argc, char **argv)
         }
     unlink_symlink_escape_done:;
     }
+    TEST("open file in non-existent parent returns ENOENT, not ELOOP");
```
Max042004, line 303: The new regression test currently fails:

make test-sysroot-create-paths open file in non-existent parent returns ENOENT, not ELOOP ... errno=40
There are two independent problems with this test. First, the preceding failed unlink rejects redirected tmp symlink escape case leaves ${sysroot}/tmp as a symlink to the live host-out directory. The next /tmp/... create therefore reaches the symlink-rejecting directory creation path and returns ELOOP.

More importantly, even after cleaning up that symlink, this open(O_CREAT) path is routed with PATH_TR_CREATE_PARENTS. For /tmp, missing parents are intentionally created by sysroot_ensure_dir_exists(), so the open succeeds rather than returning ENOENT. Thus the test does not exercise realpath_existing_parent().

Max042004 (changes requested):

Max042004:
Looking at the callers, containment is invoked only after sysroot_path_exists(), after
 access(parent), or after sysroot_ensure_dir_exists(parent). In a stable filesystem each guarantees
 that the subsequent realpath() has an existing path to resolve. The new fallback appears reachable
 only through a TOCTOU change between those operations and realpath().

 The issue’s /etc/sysusers.d/basic.conf example also does not take this path today: when the
 sysroot parent is absent and the path is outside the forced /tmp, /var/tmp, and .ccache routing
 set, proc_resolve_sysroot_create_path() returns the literal host path before containment is
 checked.

 Please provide a concrete reproducer that reaches this fallback, or clarify that the change only
 addresses a TOCTOU case and add a deterministic test for it. The current regression test should
 also restore the /tmp fixture state.

doanbaotrung:
### Actions Completed
1. **Restored `/tmp` Fixture State:** Cleaned up the leftover `${sysroot}/tmp` symlink in `test-sysroot-create-paths.c` by unlinking it and recreating `/tmp` as a normal directory at the end of the symlink escape test (`unlink_symlink_escape_done`). This prevents downstream `/tmp` operations from hitting the symlink-rejecting creation code path and triggering unexpected `ELOOP` failures.
2. **Clarified and Tested TOCTOU Mitigation:** As you correctly noted, under stable filesystem conditions the resolver callers guarantee existence before calling `sysroot_path_is_contained`, making `realpath_existing_parent` only reachable via a TOCTOU race (where a directory is removed/replaced in between the check and `realpath`). 
   To verify this mitigation deterministically, we implemented a concurrent TOCTOU race test in test-sysroot-create-paths.c. The test spawns a child process that loops creating and deleting `/tmp/race_dir` while the parent attempts `open(O_CREAT)` on `/tmp/race_dir/file.txt`. The test asserts that the race eventually triggers `ENOENT` (due to the recursive parent resolution fallback) and *never* returns `ELOOP` (which would signify a containment check violation).

src/syscall/proc-state.c
```diff
@@ -488,8 +520,12 @@ static bool sysroot_path_is_contained(const char *resolved_path,
                 return false;
 
             *slash = '\0';
-            if (!realpath(parent, real_path))
-                return false;
+            if (!realpath(parent, real_path)) {
+                if (errno != ENOENT && errno != ENOTDIR)
+                    return false;
```
jserv: When `realpath(parent)` fails and `realpath_existing_parent` climbs more than one level (a missing intermediate component, not just the leaf's parent), the `snprintf` below reattaches only the final basename (`slash + 1`) onto the shallower ancestor. For `${sr}/a/b/c` with `b` missing it validates `${sr}/a/c`, not the `${sr}/a/b/c` the syscall runs on — the middle is silently dropped. This authorizes the wrong object even without an adversary, and converts a case that previously failed closed (returned false -> `ELOOP`) into fail-open: if `b` is later created as a symlink out of the sysroot, the syscall escapes. Either keep the fail-closed return when `realpath(parent)` fails, or validate the entire stripped remainder (reject `..`, confirm the full reconstructed path is prefix-contained) rather than reattaching one component.

tests/test-sysroot-create-paths.c
```diff
@@ -293,7 +295,59 @@ int main(int argc, char **argv)
...
+
+    TEST("open file in non-existent parent TOCTOU race");
+    {
+        char race_dir[512];
+        char race_file[512];
+        snprintf(race_dir, sizeof(race_dir), "%s/tmp/race_dir",
+                 mounted_sysroot_root);
+        snprintf(race_file, sizeof(race_file), "/tmp/race_dir/file.txt");
+
+        pid_t pid = fork();
```
jserv, line 321: `fork()`'s return isn't checked. On `-1` the child branch is skipped and `kill(pid, SIGKILL)` becomes `kill(-1, SIGKILL)` — SIGKILL to every process the user can signal, then `waitpid(-1, ...)` reaps an arbitrary child. Add `if (pid < 0) FAIL(...)` before the split.

tests/test-sysroot-create-paths.c
```diff
@@ -293,7 +295,59 @@ int main(int argc, char **argv)
...
+                mkdir(race_dir, 0755);
+                rmdir(race_dir);
+            }
+            _exit(0);
+        }
+
+        int hit_enoent = 0;
+        int hit_eloop = 0;
+        int iterations = 0;
+        while (hit_enoent == 0 && iterations < 50000) {
```
jserv: The test only `PASS`es if it wins the race and observes `ENOENT`, so it's flaky on a single core or loaded CI (`hit_enoent` stays 0 -> spurious `FAIL`). It also asserts only the errno mapping, never that nothing was created outside the sysroot — so it can't catch the fail-open case in the parent-only branch. Make the `ENOENT`-mapping check deterministic (open through a statically missing intermediate, assert `ENOENT` and not `ELOOP`), keep the racer as a non-fatal stress pass, and add an assertion that no artifact exists outside `mounted_sysroot_root`.

jserv (commented):
The new walk-up in `sysroot_path_is_contained` fires only under the single-window TOCTOU that the module comment (`proc-state.c:469`) already declares out of scope, so the actionable issue is the fail-open regression noted inline, not the race itself. Two lower-priority items without their own anchor: `str_copy_trunc` in `realpath_existing_parent` (`:448`) silently truncates an over-long path and could authorize a different, shorter contained path — security code should fail closed with `ENAMETOOLONG`; and the test's `rmdir(race_dir)` cleanup can leave residue (`ENOTEMPTY`) if a race iteration created `file.txt` inside.

jserv (changes requested):
Enforce rules specified in https://cbea.ms/git-commit/ .

doanbaotrung:
Updated commit message

doanbaotrung:
#### 1. Reverted to Fail-Closed containment behavior in [proc-state.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/proc-state.c#L458-L545)
* Completely removed the `realpath_existing_parent` helper function and its usage inside `sysroot_path_is_contained`.
* Reverted the `sysroot_path_is_contained` implementation to standard fail-closed behavior (returning `false` immediately if any `realpath` call fails). This eliminates the vulnerability where intermediate path components could resolve incorrectly, potentially creating a fail-open escape if they were later replaced by symlinks pointing outside the sysroot.

#### 2. Robust Test Process Fork Handling in [test-sysroot-create-paths.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/tests/test-sysroot-create-paths.c#L318-L322)
* Checked the return value of `fork()`. If `pid < 0`, we immediately fail the test with a clean error message rather than calling `kill(-1, SIGKILL)`.

#### 3. Deterministic `ENOENT` and Leak checks in [test-sysroot-create-paths.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/tests/test-sysroot-create-paths.c#L302-L373)
* **Deterministic check:** Added a test checking that opening `/etc/nonexistent_parent_dir_123/sub/file.txt` with `O_CREAT` returns `ENOENT` and not `ELOOP` deterministically. Since this path is outside of `/tmp`, no directory auto-creation is attempted, ensuring stable and reliable testing.
* **Non-fatal stress racer:** The TOCTOU racer now runs up to 10,000 iterations without failing if it does not win the timing window to observe `ENOENT`, removing flakiness in CI.
* **Robust cleanup:** Cleaned up guest and host paths inside the sysroot for `race_file` and `race_dir` to ensure the directory is empty and `rmdir` succeeds without `ENOTEMPTY` residues.
* **Leak verification:** Added an assertion checking that no directories/files were created in `/tmp/race_dir` on the host outside of the sysroot mount.

PR #215, opened 2026-07-17
sysroot bypass hardcoded /etc/passwd and /etc/group intercepts

### LINKED ISSUE #188: Hardcoded `/etc/passwd` and `/etc/group` intercepts break guest name resolution
### Description
In emulation environments using a custom guest filesystem (`sysroot`), standard name-resolution lookups (e.g. calling `getgrnam("messagebus")` or querying `grp.getgrnam` in Python) fail to resolve newly added users or groups. 

This occurs because `elfuse` intercepts all open operations on `/etc/passwd` and `/etc/group` and returns a hardcoded mock database. This shadows the guest's real user and group databases inside the `sysroot`, making any users or groups added during package provisioning completely invisible.

### Steps to Reproduce
1. Provision a full guest rootfs prefix (e.g., Ubuntu).
2. Create or add a new system group (e.g., `messagebus` with GID `997`) using `groupadd` or `systemd-sysusers`.
3. Try to resolve the group inside the guest environment using python or `getent`:
   ```bash
   python3 -c "import grp; print(grp.getgrnam('messagebus'))"
   ```
* **Expected behavior:** The group resolves successfully, returning the GID.
* **Actual behavior:** Lookup fails with `KeyError: "getgrnam(): name not found: 'messagebus'"`.

---

### Root Cause
1. In [path.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/path.c#L59-L60), `path_might_use_open_intercept` intercepts `/etc/passwd` and `/etc/group`:
   ```c
   if (!strcmp(path, "/etc/mtab") || !strcmp(path, "/etc/passwd") ||
       !strcmp(path, "/etc/group"))
       return true;
   ```
2. In [procemu.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/runtime/procemu.c#L3345-L3356), `elfuse` intercepts the open call and serves a hardcoded synthetic literal:
   ```c
   if (!strcmp(path, "/etc/group")) {
       return proc_emit_literal(
           "root:x:0:\n"
           "staff:x:20:\n"
           "user:x:1000:\n");
   }
   ```
This synthetic literal completely overrides the actual `/etc/group` file located in the guest's `sysroot` prefix.

---

### Proposed Solution
If a custom guest `sysroot` is configured, `elfuse` should bypass these synthetic intercepts and allow the guest to read the real `/etc/passwd` and `/etc/group` files inside the rootfs.

#### Suggested Patch:
Modify `path_might_use_open_intercept` in `src/syscall/path.c` to check if a sysroot is snapshot-active and bypass the intercept:

```c
    if (!strcmp(path, "/etc/mtab"))
        return true;
    if (!strcmp(path, "/etc/passwd") || !strcmp(path, "/etc/group")) {
        char sr[LINUX_PATH_MAX];
        if (proc_sysroot_snapshot(sr, sizeof(sr)))
            return false; // Bypass synthetic intercept and read from guest sysroot
        return true;
    }
```

doanbaotrung:
When a custom guest sysroot is configured, hardcoded intercepts on /etc/passwd and /etc/group shadow the actual files in the guest rootfs. This prevents guest processes from resolving users and groups that were added during package provisioning.

We now check if a sysroot is active in path_might_use_open_intercept() and bypass the synthetic intercepts in that case, allowing the guest to read the actual user/group database from its rootfs.

Fix #188

---
## Summary by cubic
Prefer the guest sysroot’s `/etc/passwd` and `/etc/group` over synthetic intercepts when those files exist; otherwise fall back to intercepts. Fixes #188.

- **Bug Fixes**
  - Updated path_might_use_open_intercept to check the active sysroot via proc_sysroot_snapshot and stat the target; bypass intercept only if the sysroot file exists.
  - Extended tests to assert `/etc/passwd` in the sysroot wins and `/etc/group` falls back when absent; test setup now creates `sysroot/etc/passwd`.

<sup>Written for commit f1fd55d06bc3660a784d0af5af9b3e2459184c04. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/215?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/path.c
+++ b/src/syscall/path.c
@@ -56,9 +56,21 @@ bool path_might_use_open_intercept(const char *path)
         return true;
     if (path_prefix_match(path, SYSFS_CPU_PREFIX, sizeof(SYSFS_CPU_PREFIX) - 1))
         return true;
-    if (!strcmp(path, "/etc/mtab") || !strcmp(path, "/etc/passwd") ||
-        !strcmp(path, "/etc/group"))
+    if (!strcmp(path, "/etc/mtab"))
         return true;
+    if (!strcmp(path, "/etc/passwd") || !strcmp(path, "/etc/group")) {
+        char sr[LINUX_PATH_MAX];
+        if (proc_sysroot_snapshot(sr, sizeof(sr))) {
+            char sysroot_file[LINUX_PATH_MAX];
+            if (snprintf(sysroot_file, sizeof(sysroot_file), "%s%s", sr, path) <
+                (int) sizeof(sysroot_file)) {
+                struct stat st;
+                if (stat(sysroot_file, &st) == 0)
+                    return false;
+            }
+        }
+        return true;
+    }
     if (!strcmp(path, "/var/run/utmp") || !strcmp(path, "/run/utmp"))
         return true;
 
--- a/mk/tests.mk
+++ b/mk/tests.mk
@@ -257,6 +257,8 @@ test-sysroot-host-fallback: $(ELFUSE_BIN) $(BUILD_DIR)/test-sysroot-host-fallbac
 	printf 'host-final\n' > "$$mirror/final.txt"; \
 	printf 'host-loses\n' > "$$mirror/both.txt"; \
 	printf 'sysroot-wins\n' > "$$sysroot$$mirror/both.txt"; \
+	mkdir -p "$$sysroot/etc"; \
+	printf 'guest-passwd\n' > "$$sysroot/etc/passwd"; \
 	$(ELFUSE_BIN) --sysroot "$$sysroot" \
 	    $(BUILD_DIR)/test-sysroot-host-fallback \
 	    "$$hostdir" "$$mirror/final.txt" "$$mirror/both.txt"
--- a/tests/test-sysroot-host-fallback.c
+++ b/tests/test-sysroot-host-fallback.c
@@ -133,6 +133,22 @@ int main(int argc, char **argv)
     else
         PASS();
 
+    TEST("sysroot passwd file wins over synthetic intercept");
+    if (read_file_nul("/etc/passwd", buf, sizeof(buf)) <= 0)
+        FAIL("read /etc/passwd failed");
+    else if (strcmp(buf, "guest-passwd\n"))
+        FAIL("did not bypass /etc/passwd synthetic intercept");
+    else
+        PASS();
+
+    TEST("absent sysroot group file falls back to synthetic intercept");
+    if (read_file_nul("/etc/group", buf, sizeof(buf)) <= 0)
+        FAIL("read /etc/group failed");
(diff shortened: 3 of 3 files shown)


src/syscall/path.c
```diff
@@ -56,9 +56,14 @@ bool path_might_use_open_intercept(const char *path)
         return true;
     if (path_prefix_match(path, SYSFS_CPU_PREFIX, sizeof(SYSFS_CPU_PREFIX) - 1))
         return true;
-    if (!strcmp(path, "/etc/mtab") || !strcmp(path, "/etc/passwd") ||
-        !strcmp(path, "/etc/group"))
+    if (!strcmp(path, "/etc/mtab"))
         return true;
+    if (!strcmp(path, "/etc/passwd") || !strcmp(path, "/etc/group")) {
```
Max042004, line 61: path_might_use_open_intercept() only checks that a sysroot is configured
(proc_sysroot_snapshot()), not that the sysroot actually contains
/etc/passwd / /etc/group. When it doesn't, the request falls through to
proc_resolve_sysroot_path_flags()'s host-literal fallback, which resolves
the literal host path — on macOS that's the real /etc/passwd /
/etc/group on the machine running elfuse.

Max042004 (changes requested):

src/syscall/path.c
doanbaotrung: ### Fix Details
1. **Added Existence Checks:** Updated `path_might_use_open_intercept` in [path.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/path.c#L60-L73) to perform a `stat` existence check on the guest sysroot-resolved files (`${sysroot}/etc/passwd` / `${sysroot}/etc/group`). The synthetic intercept is bypassed only when the file is present in the sysroot.
2. **Updated Test Harness:** Modified [tests.mk](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/mk/tests.mk#L259-L264) to only stage `passwd` in the fallback test's guest sysroot, leaving `group` absent.
3. **Added Assertions:** Updated [test-sysroot-host-fallback.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/tests/test-sysroot-host-fallback.c#L141-L151) to verify that the absent guest `group` file falls back to the synthetic group intercept instead of resolving to the host machine's `/etc/group`.

Max042004 (approved):

PR #216, opened 2026-07-17
Emulate setgroups for fakeroot guests

### LINKED ISSUE #189: Missing `SYS_setgroups` (Syscall 159) on AArch64 breaks privilege-dropping guest processes
### Description
In emulation environments, guest processes that attempt to drop privileges (for example, standard package managers like `apt-get` spawning a helper process as user `_apt` or GID `65534`) crash or fail to execute with the following error:
`E: setgroups 65534 failed - setgroups (38: Function not implemented)`
`E: Method gave invalid 400 URI Failure message: Failed to setgroups - setgroups (38: Function not implemented)`

This is because system call `159` (`setgroups` on Linux AArch64) is unimplemented in `elfuse`, returning `ENOSYS`.

### Steps to Reproduce
1. Run a guest process under `elfuse` that drops group privileges (e.g. running `apt-get install` when the `_apt` sandbox user exists).
2. The guest process invokes the standard glibc `setgroups` function.
* **Expected behavior:** The group list is successfully cleared or set, allowing the process to continue.
* **Actual behavior:** The call fails with `ENOSYS` ("Function not implemented").

---

### Root Cause
1. System call `159` is not defined in `src/syscall/abi.h`.
2. The syscall is not registered in the generated dispatch header `src/syscall/dispatch.tbl`.
3. There is no implementation wrapper in `src/syscall/sys.c` or `src/syscall/syscall.c`.

---

### Proposed Solution
Under root or fakeroot emulation (standard for local dev container runtimes), `setgroups` cannot actually be executed on the host. It should be emulated as a no-op that returns success (`0`) to guest processes.

#### Suggested Patch:

1. Define `SYS_setgroups` in [src/syscall/abi.h](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/abi.h):
   ```c
   #define SYS_setgroups 159
   ```

2. Register the syscall in [src/syscall/dispatch.tbl](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/dispatch.tbl):
   ```
   SYS_setgroups sc_setgroups 0
   ```

3. Declare the syscall handler in [src/syscall/sys.h](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/sys.h):
   ```c
   int64_t sys_setgroups(guest_t *g, int size, uint64_t list_gva);
   ```

4. Define the no-op implementation in [src/syscall/sys.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/sys.c):
   ```c
   int64_t sys_setgroups(guest_t *g, int size, uint64_t list_gva)
   {
       /* Under fakeroot/root emulation, allow setgroups to succeed as a no-op.
        * Non-root guests get EPERM.
        */
       if (proc_get_euid() == 0 || proc_fakeroot_enabled()) {
           return 0;
       }
       return -LINUX_EPERM;
   }
   ```

5. Register the wrapper in [src/syscall/syscall.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/syscall.c):
   ```c
   SC_FORWARD(sc_setgroups, sys_setgroups(g, (int) x0, x1))
   ```

doanbaotrung:
Wire aarch64 SYS_setgroups through the syscall dispatch table and emulate it in the guest identity layer. Allow root/fakeroot callers to set or clear supplementary groups as a no-op after validating the guest group list, while non-root callers receive EPERM.

Extend credential coverage for root/fakeroot success, non-root denial, bad pointers, zero-count calls, and invalid sizes.

Fix #189

---
## Summary by cubic
Emulates setgroups(2) for aarch64 guests by wiring `SYS_setgroups` and making root/fakeroot calls a validated no-op. Non-root calls return `-EPERM`. Fixes #189.

- **New Features**
  - Defined `SYS_setgroups` (159) in `abi.h`; wired in `dispatch.tbl`; forwarded `sc_setgroups` to `sys_setgroups`.
  - Root or fakeroot: validate group list and return 0 (including size == 0). Others: `-EPERM`.
  - Errors: size < 0 or > 65536 → `-EINVAL`; bad pointer → `-EFAULT`; alloc failure → `-ENOMEM`.
  - Extended `tests/test-credentials.c` to cover root/fakeroot success, non-root denial, bad pointer, zero count, and invalid size.

<sup>Written for commit b4b693c7fde5b8f041d5b8c5a54f28d25481c307. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src/syscall/sys.c
```diff
@@ -517,6 +517,27 @@ int64_t sys_getgroups(guest_t *g, int size, uint64_t list_gva)
     return ngroups;
 }
 
+#define LINUX_NGROUPS_MAX 65536
+
+int64_t sys_setgroups(guest_t *g, int size, uint64_t list_gva)
+{
+    if (size < 0 || size > LINUX_NGROUPS_MAX)
```
Max042004: sys_setgroups() checks argument validity before privilege, which is the reverse of the real kernel's order. Linux's sys_setgroups (kernel/groups.c) checks CAP_SETGID first, then validates gidsetsize.

Confirmed with the default (non-root, GUEST_UID=1000) guest: setgroups((unsigned)-1, 0) currently returns -EINVAL, but a real kernel returns -EPERM here since the capability check runs first. The new "setgroups invalid size" test in tests/test-credentials.c bakes in this reversed order (it expects -EINVAL unconditionally, regardless of caller privilege), so it won't catch a fix.

Max042004 (changes requested):

src/syscall/sys.c
doanbaotrung: Changed:

sys.c:

- `sys_setgroups()` now checks privilege first, then validates gidsetsize, matching Linux behavior.

test-credentials.c:

- invalid-size test now expects:root/fakeroot: `-EINVAL`
- non-root: `-EPERM`

Max042004 (approved):

PR #217, opened 2026-07-17
proc allow root and fakeroot guests to switch to any UID/GID

### LINKED ISSUE #190: UID/GID permission switching restrictions under Fakeroot return `EPERM`
### Description
In emulation environments running in `fakeroot` mode or as the virtual guest `root` user, guest processes that attempt to change their UID or GID (such as invoking `seteuid` or `setegid` to drop privileges to a non-root system user) fail and abort with the following error:
`E: seteuid 42 failed - seteuid (1: Operation not permitted)`

This occurs because `elfuse`'s internal permission checks restrict UID/GID changes strictly to the currently configured guest UID/GID values, ignoring the fact that a root or fakeroot guest process is allowed to switch to any arbitrary user ID.

### Steps to Reproduce
1. Start the guest environment under `fakeroot` or as guest `root` (UID 0).
2. Execute a guest program that drops privileges to another user ID (e.g. user `_apt` with UID `42`):
   ```bash
   python3 -c "import os; os.seteuid(42)"
   ```
* **Expected behavior:** The UID change succeeds (since the process is running as root/fakeroot).
* **Actual behavior:** The call fails with `OSError: [Errno 1] Operation not permitted`.

---

### Root Cause
In [proc-identity.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/proc-identity.c#L108-L117), `uid_is_permitted` and `gid_is_permitted` only allow switching to IDs that match the process's active credentials:

```c
static bool uid_is_permitted(uint32_t val)
{
    return val == emu_uid || val == emu_euid || val == emu_suid;
}

static bool gid_is_permitted(uint32_t val)
{
    return val == emu_gid || val == emu_egid || val == emu_sgid;
}
```

Under Linux, a process with effective UID `0` (or running in a fakeroot sandbox) is allowed to change its credentials to any arbitrary UID/GID. Since `elfuse` blocks these switches, guest processes cannot drop privileges.

---

### Proposed Solution
Modify `uid_is_permitted` and `gid_is_permitted` in `src/syscall/proc-identity.c` to allow any UID/GID switches when running under `fakeroot` or if the guest process has root privileges (`emu_euid == 0`).

#### Suggested Patch:

```diff
diff --git a/src/syscall/proc-identity.c b/src/syscall/proc-identity.c
index 859a07c..27c71a3 100644
--- a/src/syscall/proc-identity.c
+++ b/src/syscall/proc-identity.c
@@ -107,11 +107,15 @@
 
 static bool uid_is_permitted(uint32_t val)
 {
+    if (emu_euid == 0 || fakeroot_enabled)
+        return true;
     return val == emu_uid || val == emu_euid || val == emu_suid;
 }
 
 static bool gid_is_permitted(uint32_t val)
 {
+    if (emu_euid == 0 || fakeroot_enabled)
+        return true;
     return val == emu_gid || val == emu_egid || val == emu_sgid;
 }
```

doanbaotrung:
Guest processes running as root (effective UID 0) or in fakeroot mode should have the privilege to change their UID or GID to any value. Previously, elfuse restricted UID/GID changes strictly to the currently configured active credentials (emu_uid, emu_euid, etc.), aborting privilege-dropping calls with EPERM.

We now permit any UID/GID switches in uid_is_permitted() and gid_is_permitted() if the guest process has effective UID 0 or fakeroot is active.

Fix #190

---
## Summary by cubic
Allow root (euid 0) and fakeroot guests to switch to any UID/GID; privileged setuid/setgid now reset real, effective and saved IDs, and setreuid/setregid may change the real ID, matching Linux and avoiding EPERM when dropping privileges. Fixes #190.

<sup>Written for commit e39d92fe4dabd139ef014aa9f5f04e294d8e628b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/217?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/proc-identity.c
+++ b/src/syscall/proc-identity.c
@@ -107,11 +107,15 @@ uint32_t proc_get_sgid(void)
 
 static bool uid_is_permitted(uint32_t val)
 {
+    if (emu_euid == 0 || proc_fakeroot_enabled())
+        return true;
     return val == emu_uid || val == emu_euid || val == emu_suid;
 }
 
 static bool gid_is_permitted(uint32_t val)
 {
+    if (emu_euid == 0 || proc_fakeroot_enabled())
+        return true;
     return val == emu_gid || val == emu_egid || val == emu_sgid;
 }
 
@@ -119,7 +123,12 @@ int64_t proc_sys_setuid(uint32_t uid)
 {
     if (!uid_is_permitted(uid))
         return -LINUX_EPERM;
-    emu_euid = uid;
+    if (emu_euid == 0 || proc_fakeroot_enabled()) {
+        emu_uid = uid;
+        emu_euid = uid;
+        emu_suid = uid;
+    } else
+        emu_euid = uid;
     return 0;
 }
 
@@ -127,7 +136,12 @@ int64_t proc_sys_setgid(uint32_t gid)
 {
     if (!gid_is_permitted(gid))
         return -LINUX_EPERM;
-    emu_egid = gid;
+    if (emu_euid == 0 || proc_fakeroot_enabled()) {
+        emu_gid = gid;
+        emu_egid = gid;
+        emu_sgid = gid;
+    } else
+        emu_egid = gid;
     return 0;
 }
 
@@ -135,7 +149,8 @@ int64_t proc_sys_setgid(uint32_t gid)
     int64_t proc_sys_setre##suffix(uint32_t r, uint32_t e)                 \
     {                                                                      \
         uint32_t old_real = real;                                          \
-        if (r != (uint32_t) -1 && r != real && r != eff)                   \
+        if (r != (uint32_t) -1 && r != real && r != eff &&                 \
+            !(emu_euid == 0 || proc_fakeroot_enabled()))                   \
             return -LINUX_EPERM;                                           \
         if (e != (uint32_t) -1 && !perm_fn(e))                             \
             return -LINUX_EPERM;                                           \
--- a/tests/test-credentials.c
+++ b/tests/test-credentials.c
@@ -10,6 +10,7 @@
  */
 
 #include <sys/types.h> /* gid_t */
+#include <sys/wait.h>  /* waitpid */
 #include <unistd.h>    /* getgroups */
 
 #include "test-harness.h"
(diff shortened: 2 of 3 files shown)


src/syscall/proc-identity.c
```diff
@@ -107,35 +107,49 @@ uint32_t proc_get_sgid(void)
...
+        emu_egid = gid;
     return 0;
 }
 
 #define DEFINE_SETRE(suffix, real, eff, saved, perm_fn)                    \
     int64_t proc_sys_setre##suffix(uint32_t r, uint32_t e)                 \
     {                                                                      \
         uint32_t old_real = real;                                          \
-        if (r != (uint32_t) -1 && r != real && r != eff)                   \
+        if (r != (uint32_t) -1 && r != real && r != eff && !perm_fn(r))    \
```
Max042004: setreuid/setregid's real-ID (r) check now falls back to perm_fn(r), which for unprivileged callers also accepts the saved ID via uid_is_permitted/gid_is_permitted. Real Linux setreuid(2)/setregid(2) restrict the real-ID argument to {current real, current effective} only — saved ID is only valid there via setresuid/setresgid.

Verified against a real Linux kernel (aarch64, 7.0.5): after setresuid(1000,1000,0) (drop euid, keep suid=0), setreuid(0,-1) returns EPERM on Linux but succeeds here, letting an unprivileged process restore its real UID through a call a real kernel would reject. Suggest a narrower permission check for r (real/effective only, bypassed when privileged) instead of reusing uid_is_permitted/gid_is_permitted as-is.

tests/test-credentials.c
```diff
@@ -238,6 +271,29 @@ int main(void)
         EXPECT_TRUE(rc > 0 && (mask & 1), "CPU0 not in mask");
     }
 
+    if (expected_id == 0) {
```
Max042004, line 279: These root-only assertions never run in CI: test-credentials isn't in tests/manifest.txt (not part of make check), test-matrix.sh invokes it without --fakeroot (default guest UID 1000), and it's listed in QEMU_SKIP for the real-root qemu lane — so expected_id == 0 is unreachable everywhere.

Running it manually with --fakeroot fails 7/27 cases: the earlier setuid(other)/setgid(other) root-branch calls (lines 73/156) now permanently drop identity to 1000 per this PR's own "reset all three fields" fix, but the later setfsuid/setfsgid/getgroups assertions still assume the stale expected_id == 0 identity. Please wire --fakeroot into a test runner so this doesn't silently bit-rot, and fix the stale-identity assumption in those later checks.

Max042004 (changes requested):

doanbaotrung:
I have addressed both review items:

1. **Narrowed Real-ID Check:** Replaced the reuse of `perm_fn(r)` for real-ID validation in `DEFINE_SETRE` at [proc-identity.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/src/syscall/proc-identity.c#L151-L154) with a narrow check `!(emu_euid == 0 || proc_fakeroot_enabled())`. This restricts unprivileged callers to setting the real ID strictly to the current real or effective ID (preventing them from using the saved ID via `setreuid`/`setregid`), while preserving unrestricted switches for privileged callers.
2. **Prevented down-stream test pollution via `fork`:** Refactored credential-dropping and privilege tests in [test-credentials.c](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/tests/test-credentials.c#L74-L313) to run in fork-isolated child processes. This ensures the parent process retains its initial credentials throughout the entire test suite execution, fixing the errors when running the test under `--fakeroot`.
3. **Enabled testing in CI:** Added `test-credentials-fakeroot` to [test-matrix.sh](file:///Users/dbaotrung/personal/muplar/third_party/elfuse/tests/test-matrix.sh#L277) (and listed it under `QEMU_SKIP` to skip it on native QEMU runners which do not support the elfuse fakeroot flag). This ensures the fakeroot mode identity assertions are run and validated under the Rosetta matrix.

tests/test-credentials.c
doanbaotrung: Updated

src/syscall/proc-identity.c
doanbaotrung: Updated

Max042004 (approved):

PR #223, opened 2026-07-18
Add CI performance benchmark suite

### LINKED ISSUE #195: CI/CD performance benchmarking
## Motivation

Performance is one of the primary goals of elfuse. While functional correctness is verified through automated testing, continuous performance evaluation is equally important to detect regressions and quantify improvements over time.

This proposal introduces a lightweight benchmark suite that can be executed automatically in CI/CD and provides a consistent basis for comparing elfuse against native Linux and OrbStack.

Objectives:

* Detect performance regressions in pull requests.
* Provide reproducible measurements across releases.
* Keep CI execution time short.
* Focus on workloads representative of Linux application execution.

## Benchmark Strategy

Two complementary tiers:

1. **Tier 1: Microbenchmarks** using lmbench to isolate low-level operating system overheads.
2. **Tier 2: Application Benchmarks** using widely adopted command-line tools to measure real-world developer workflows.

## Tier 1: lmbench Microbenchmarks

Rather than adopting large suites such as SPEC CPU or Phoronix Test Suite, this proposal selects three focused lmbench benchmarks representing the most critical execution paths of elfuse.

| Benchmark       | Purpose                             | Relevance                                                                           |
| --------------- | ----------------------------------- | ----------------------------------------------------------------------------------- |
| `lat_syscall`   | Measure system call latency         | Evaluates the overhead of the Linux ABI and syscall forwarding path.                |
| `lat_proc exec` | Measure process startup latency     | Evaluates ELF loading, dynamic linking, memory mapping, and process initialization. |
| `lat_fs`        | Measure filesystem metadata latency | Evaluates pathname translation and filesystem-intensive workloads.                  |

### `lat_syscall`

Measures the latency of fundamental system calls. Recommended operations: `null`, `read`, `write`, `stat`, `open`. These represent the basic cost of entering the Linux execution environment and are exercised by nearly every Linux application.

### `lat_proc exec`

Measures end-to-end process startup latency via `exec()`. For elfuse, it covers nearly the entire execution path: ELF parsing, program loading, dynamic linker initialization, memory mapping, initial process setup, and process termination. Primary indicator of Linux application startup performance.

### `lat_fs`

Measures filesystem metadata operations (file creation, deletion, directory traversal). Metadata-intensive workloads are common in Git, Make, CMake, package managers, and shell utilities, so `lat_fs` approximates many software development workloads.

## Tier 2: Application Benchmarks

Microbenchmarks explain where time is spent; developers ultimately care about application performance. The suite includes representative command-line tools exercising common Linux development workflows.

| Tool                   | Primary workload          | Primary subsystem                                        |
| ---------------------- | ------------------------- | -------------------------------------------------------- |
| `python3 -c pass`      | Process startup           | ELF loader, dynamic linker, interpreter initialization   |
| `git status`           | Filesystem metadata       | `open()`, `stat()`, `readdir()`                          |
| `rg` (ripgrep)         | Large-scale source search | `mmap()`, `read()`, page cache                           |
| `zstd`                 | Compression               | CPU computation and filesystem I/O                       |
| `make` (small project) | Software build            | Process creation, filesystem access, compiler invocation |

These were selected because they are representative of daily Linux developer workflows, small enough to execute in CI, stable and reproducible, and portable across native Linux, OrbStack, and elfuse.

## CI/CD Integration

The complete suite should run for every pull request:

```text
Build elfuse
        │
Run lmbench benchmarks
        │
Run application benchmarks
        │
Collect benchmark results
        │
Generate JSON report
        │
Compare with baseline
        │
Upload artifacts
```

Expected execution time remains only a few minutes.

## Result Format

Benchmark results exported as structured JSON to simplify regression analysis:

```json
{
  "lmbench": {
    "lat_syscall": { "null": 0.23, "open": 1.98, "stat": 2.45 },
    "lat_proc": { "exec": 705 },
    "lat_fs": { "create": 47, "delete": 34 }
  },
  "applications": {
    "python_startup_ms": 24.1,
    "git_status_ms": 132.7,
    "ripgrep_ms": 218.4,
    "zstd_ms": 645.8,
    "make_ms": 1038.2
  }
}
```

## Comparison Targets

Compare three execution environments: Native Linux (baseline), OrbStack, and elfuse. Each benchmark reports both absolute time and the ratio relative to native Linux.

| Benchmark           | Native Linux | OrbStack | elfuse |
| ------------------- | ------------ | -------- | ------ |
| `lat_proc exec`     | 520 µs       | 730 µs   | 680 µs |
| `lat_syscall(open)` | 1.4 µs       | 1.8 µs   | 2.0 µs |
| `lat_fs(create)`    | 41 µs        | 46 µs    | 44 µs  |
| `python3 -c pass`   | 21 ms        | 26 ms    | 24 ms  |
| `git status`        | 118 ms       | 126 ms   | 121 ms |
| `rg`                | 205 ms       | 214 ms   | 208 ms |
| `zstd`              | 630 ms       | 651 ms   | 642 ms |
| `make`              | 1.02 s       | 1.08 s   | 1.04 s |

(Numbers above are illustrative placeholders, not measured results.)

## Implementation Plan

Concrete, self-contained tasks. Each is independently mergeable and grounded in the current tree (`tests/bench-*`, `tests/test-matrix.sh`, `tests/fetch-fixtures.sh`, `.github/workflows/main.yml`).

### Constraint: where benchmarks can run

elfuse requires the Hypervisor.framework entitlement, so any benchmark that executes a guest *through* elfuse only runs on the existing `[self-hosted, macOS, arm64]` runner (the `runtime-macos` job in `main.yml`), not on GitHub-hosted `ubuntu-24.04` / `macos-15`. Consequence: native-Linux and OrbStack baselines cannot be produced inside GitHub-hosted CI. They are captured manually on a reference host and checked in as static baseline JSON; CI measures only the `elfuse-aarch64` (and, where available, `qemu-aarch64`) columns and compares against the checked-in baseline.

### Task 1: Vendor/fetch benchmark guest binaries

* [ ] Add lmbench aarch64-linux-musl binaries (`lat_syscall`, `lat_proc`, `lat_fs`) to the fixtures flow in `tests/fetch-fixtures.sh` (same Alpine-package pattern already used for busybox/coreutils). Pin the package version and record the SHA.
* [ ] Add Tier-2 guest binaries as static aarch64-musl fixtures where possible: `git`, `ripgrep` (`rg`), `zstd`, `make`, and a Python interpreter. Where a static build is impractical, document the dynamic + `--sysroot` invocation instead.
* [ ] Land a fixed source tree for the `git status` / `rg` / `make` targets (a small committed corpus under `tests/bench-corpus/`) so measurements are reproducible run to run.
* [ ] Verify each fixture runs under elfuse manually: `./elfuse <fixture> --help` (or equivalent) exits 0.

### Task 2: Benchmark runner + JSON emitter

* [ ] Add `tests/bench-suite.sh` that runs the Tier-1 and Tier-2 workloads and writes one JSON document matching the schema in "Result Format". Reuse `tests/lib/test-runner.sh` (`run`/`run_check`) so every invocation inherits the existing `timeout $TEST_TIMEOUT` wrapping.
* [ ] Drive each workload N iterations (default N=10), discard warmup, and report median (not mean) to reduce noise on a shared runner. Emit iteration count and raw samples alongside the median.
* [ ] Parameterize the execution environment via an arg/env (`BENCH_ENV=elfuse-aarch64|native|orbstack`) so the same script produces every column. Reuse `detect_x86_64_host_class` / the mode plumbing from `tests/test-matrix.sh` rather than duplicating host detection.
* [ ] Add a `make bench` target (and a CI-scoped `make bench-ci`) that builds the harness and runs `tests/bench-suite.sh`, writing `build/bench-results.json`.

### Task 3: Baseline + regression comparison

* [ ] Add `scripts/bench-compare.py`: takes `build/bench-results.json` and a checked-in `tests/bench-baseline.json`, prints a per-metric ratio table, and exits non-zero when any metric regresses beyond a threshold (default: +20%, configurable via env). Reuse the CLI/reporting style of `scripts/check-syscall-coverage.py`.
* [ ] Commit an initial `tests/bench-baseline.json` captured on the self-hosted runner (elfuse column) plus manually captured native-Linux and OrbStack columns on a documented reference host. Record host class / CPU / date in the file so stale baselines are visible.
* [ ] Document the baseline-refresh procedure (when and how to regenerate) in `tests/README` or the suite script header.

### Task 4: CI/CD integration

* [ ] Add a `benchmark` leg to the `runtime-macos` matrix in `.github/workflows/main.yml` (self-hosted arm64), gated the same way as the existing legs (`github.repository == 'sysprog21/elfuse'`), running `make bench-ci` then `scripts/bench-compare.py`.
* [ ] Upload `build/bench-results.json` via `actions/upload-artifact@v7` (mirror the existing `infer-*` artifact step) with a short retention.
* [ ] Make regression a soft signal first: run `bench-compare.py` in report-only mode (`|| true` or a `continue-on-error` step) for an initial settling period, then flip to hard-fail once the runner's variance is characterized. State the flip criterion in the workflow comment.
* [ ] Keep the leg off the sanitizer legs and out of GitHub-hosted jobs; confirm total added wall-clock stays within the existing `runtime-macos` `timeout-minutes` budget.

### Acceptance criteria

* [ ] `make bench` produces valid JSON matching the documented schema on a dev machine.
* [ ] `scripts/bench-compare.py` correctly flags an injected 2x regression and passes on baseline.
* [ ] The CI leg runs green on a PR, uploads the artifact, and its added runtime is within a few minutes.
* [ ] Native / OrbStack baseline capture is documented and reproducible.

Max042004:
## Summary

Add a reproducible two-tier performance benchmark suite for detecting
regressions in pull requests.

- Tier 1 runs pinned lmbench workloads for syscall, process, and
filesystem latency.
- Tier 2 measures Python startup, Git status, ripgrep, zstd, and Make
against a deterministic committed corpus.
- Results include medians and raw samples in structured JSON.
- Baseline columns are prepared for elfuse, QEMU-based native Linux,
and OrbStack.

## CI integration

Add a benchmark leg to the self-hosted Apple Silicon runtime matrix.
The leg runs exclusively on the host to avoid interference from sibling
jobs, compares results in report-only mode, and uploads the JSON artifact.

Missing baseline metrics are treated as failures, preventing broken or
timed-out benchmarks from silently passing. Benchmark preparation and
cleanup commands are also bounded by timeouts.

The checked-in baseline has not yet been captured on the designated
self-hosted CI machine. It will be refreshed in a follow-up after the
benchmark leg is deployed and runner variance has been characterized.
Comparison therefore remains report-only for now.

## Validation

- Full cold run: approximately 10m 19s
- Cached CI-equivalent run: approximately 9m 20s
- Baseline match exits successfully
- An injected 2x regression exits non-zero
- A missing baseline metric exits non-zero
- Benchmark output contains 9 lmbench and 6 application metrics

Closes #195

---
## Summary by cubic
Adds a two-tier performance benchmark suite with a deterministic workload to catch syscall, process-startup, and app-level regressions. Runs on the self-hosted Apple Silicon runner under an exclusive host lock and posts a sanitized PR comment with benchmark deltas. (Issue #195)

- New Features
  - `tests/bench-suite.sh`: Tier 1 `lmbench` (`lat_syscall`/`lat_proc`/`lat_fs`) and Tier 2 app workloads (`git status`, `rg`, `make`, `zstd`) with warmups; guest-side timing for `qemu-aarch64` via `tests/bench-timeit.c`; writes medians+samples to `build/bench-results*.json`.
  - Deterministic `tests/bench-corpus` for stable Tier 2 inputs; excluded from formatting to preserve baselines. Fixtures fetch app tools and cross-compile pinned `lmbench`.
  - Baselines in `tests/bench-baseline.json` for `elfuse-aarch64`, `qemu-aarch64`, and `orbstack`. `scripts/bench-compare.py` flags regressions (tunable via `BENCH_REGRESSION_THRESHOLD`) and treats missing metrics as NEW by default; `scripts/bench-promote.py` refreshes baselines.
  - CI: `make bench` / `make bench-ci`; `BENCH_ENV` selects `elfuse-aarch64` (default) | `qemu-aarch64` | `orbstack` | `native`. Benchmark leg isolates measurements with `scripts/ci-host-lock.sh`. `.github/workflows/bench-report.yml` uses `scripts/bench-report.py` to publish a sanitized advisory PR comment from the uploaded report artifact. Docs in `docs/testing.md`.

- Migration
  - Run: `make bench` (or `BENCH_ENV=qemu-aarch64 make bench-ci`).
  - Compare: `scripts/bench-compare.py --report-only`.
  - Refresh baselines: run `make bench-ci` per env, then `scripts/bench-promote.py`; commit the updated `tests/bench-baseline.json`.

<sup>Written for commit c2c8cbe41a6ede5de741109e0b7744901bbc3f1a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/223?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tests/bench-corpus/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+The content is deterministic output of a one-shot generator and is part
+of the benchmark definition: editing, reformatting, or regenerating it
+changes every measurement and invalidates `tests/bench-baseline.json`.
+Do not modify these files; if the corpus ever must change, regenerate
+the baseline in the same change (see the header of
+`tests/bench-suite.sh` for the refresh procedure).
+
+Files use the `.src` suffix (C-like content) so repository lint
+(clang-format, newline checks) never rewrites the corpus.
```
jserv: Is it necessary?

jserv (commented):
The generated `tests/bench-corpus/*.c` fixtures were excluded from review.

scripts/ci-host-lock.sh
```diff
@@ -0,0 +1,110 @@
...
+lock_path, mode, pid_path, wait_min, hold_min = sys.argv[1:6]
+flags = fcntl.LOCK_EX if mode == "exclusive" else fcntl.LOCK_SH
+
+r, w = os.pipe()
+child = os.fork()
+if child == 0:
+    # Holder: acquire (blocking), report through the pipe, then keep
+    # the descriptor open until release kills us or we self-expire.
+    os.close(r)
+    os.setsid()
```
jserv, line 105: The detached lock holder calls setsid() and closes the pipe but never redirects fds 1/2, so it keeps the CI step's stdout/stderr open. GitHub Actions never sees pipe EOF on the "Acquire host lock" step and it hangs until the holder self-expires (CI_HOST_LOCK_MAX_HOLD_MIN, up to 120 min); "Release host lock" cannot run until acquire returns, so the job deadlocks.

Fix: after os.setsid(), redirect fds 0/1/2 to /dev/null (os.open + os.dup2 over 0,1,2) before writing readiness and sleeping.

.github/workflows/main.yml
```diff
@@ -359,7 +366,8 @@ jobs:
             ubsan_options: ''
             tsan_options: ''
             test_timeout: ''
-            job_timeout: 20
+            job_timeout: 35
+            lock_wait: 10
```
jserv: The host lock is acquired at checkout and released only at job end, so every leg holds its lock for its entire runtime. The exclusive benchmark leg then blocks shared siblings far longer than this lock_wait=10 budget (its real work is ~15-25 min, up to a 60-min job), so a sibling that queues behind it exhausts 10 min of wait and fails spuriously.

Fix: acquire the exclusive lock immediately around `make bench-ci` only (held just for the measurement window), or raise every sibling's lock_wait above the benchmark leg's worst-case hold time.

scripts/bench-compare.py
```diff
@@ -0,0 +1,206 @@
...
+                        help="make missing baseline metrics fail even in "
+                             "report-only mode")
+    args = parser.parse_args()
+
+    results = load_json(args.results)
+    baseline = load_json(args.baseline)
+
+    env = results.get("meta", {}).get("env", "elfuse-aarch64")
+    column = (baseline.get("environments") or {}).get(env)
+    if not column:
```
jserv: A missing, empty, or misspelled meta.env resolves to no baseline column and returns 0 before any regression/missing check, silently disabling all gating including --fail-on-missing. Also, .get("env", "elfuse-aarch64") only defaults on an absent key, not a present empty string.

Fix: treat a missing/empty column as a hard error (non-zero exit) when --fail-on-missing is set, and reject an empty meta.env explicitly.

scripts/ci-host-lock.sh
```diff
@@ -0,0 +1,110 @@
...
+except ProcessLookupError:
+    pass
+sys.exit(f"ci-host-lock: {mode} lock on {lock_path} not acquired within "
+         f"{wait_min} min (holder died or a sibling job is wedged; "
+         f"inspect with: lsof {lock_path})")
+PY
+        ;;
+    release)
+        if [ -s "$PID_FILE" ]; then
+            kill "$(cat "$PID_FILE")" 2> /dev/null || true
```
jserv: Release runs kill on the recorded PID with no check that it still names the lock holder. After the holder self-expires (CI_HOST_LOCK_MAX_HOLD_MIN) or crashes and the OS recycles its PID, this signals an unrelated process.

Fix: verify holder identity before killing (pgid/start-time), or kill the setsid process group and confirm it still owns the flock.

scripts/bench-compare.py
```diff
@@ -0,0 +1,206 @@
...
+def flatten(tree: dict) -> dict[str, float]:
+    """Flatten a results/baseline document into {dotted-path: median}.
+
+    Result files store {"median": x, "samples": [...]} leaves; baseline
+    columns store plain numbers. Both shapes are accepted so a captured
+    results file can be pasted into the baseline as-is if desired.
+    """
+    flat: dict[str, float] = {}
+
+    def walk(prefix: str, node) -> None:
```
jserv, line 64: float(node["median"]) accepts NaN/Infinity from Python's permissive JSON parser. A NaN median makes every ratio comparison false (nan > 1.0+threshold and nan < 1.0-threshold are both False), so a corrupt result passes as green.

Fix: validate each metric with math.isfinite() and reject non-finite (and non-positive baseline) values.

.github/workflows/main.yml
```diff
@@ -392,10 +402,35 @@ jobs:
...
+          # the release leg via run_bench.
+          - name: Benchmark
+            sanitizer: bench
+            extra_cflags: ''
+            asan_options: ''
+            ubsan_options: ''
+            tsan_options: ''
+            test_timeout: ''
+            job_timeout: 60
+            lock_wait: 45
```
jserv: This benchmark leg lock_wait=45 is below the TSAN sibling's job_timeout=60. Since each leg holds its lock for the whole job, a TSAN run lasting >45 min holds the shared lock long enough that the benchmark leg's exclusive acquire times out and the leg goes red; the "longest sibling (~45 min)" assumption is inconsistent with the 60-min TSAN budget.

Fix: set the benchmark leg's lock_wait to at least the maximum sibling job_timeout (60), and widen its job_timeout accordingly.

tests/bench-suite.sh
```diff
@@ -0,0 +1,1202 @@
...
+        os.remove(path)
+        os.symlink(rel, path)
+PYEOF
+}
+
+# Populate a corpus work area at host path $1: a source tree, 512 copied
+# source-tree shards for ripgrep (~87 MiB / tens of thousands of files), and a zstd
+# input grown deterministically to ~87 MiB. The larger fixed workloads keep
+# short-run scheduling and command noise from dominating their medians.
+prepare_corpus()
```
jserv, line 666: The script runs with set -u -o pipefail but not set -e, and prepare_corpus() ignores failures from cp/cat/mv (and run_tier2_tool's $pre_each at line 801 runs unchecked). A disk-full or copy failure leaves fewer rg shards / a truncated input, or a failed make_clean_out turns make into a no-op build, and the suite then measures a smaller workload and reports artificially fast numbers instead of failing.

Fix: check each prep command (cp ... || bench_die, etc.) and "$pre_each" || bench_die.

tests/bench-suite.sh
```diff
@@ -0,0 +1,1202 @@
...
+    esac
+    if [ "$rc" -ne 0 ]; then
+        echo "bench-suite: sample '$label' failed (rc=$rc, out='$out'):" >&2
+        echo "  $*" >&2
+        exit 1
+    fi
+    echo "$out"
+}
+
+run_tier2_tool()
```
jserv, line 774: For native/qemu-aarch64/orbstack, timeout wraps bench-timeit but the timed workload runs as bench-timeit's child, not in its own process group. timeout kills bench-timeit, not necessarily the forked workload, so a hung git/make/zstd can survive and distort later samples or future runs.

Fix: place the child in its own process group and kill the group on timeout (or run timeout inside bench-timeit's child group).

scripts/bench-compare.py
```diff
@@ -0,0 +1,206 @@
...
+    missing: list[str] = []
+    improvements: list[str] = []
+    rows = [("metric", "baseline", "current", "ratio", "")]
+    for metric in sorted(base):
+        if metric not in current:
+            rows.append((metric, f"{base[metric]:g}", "-", "-",
+                         "MISSING (measurement failed)"))
+            missing.append(metric)
+            continue
+        ratio = (current[metric] / base[metric]) if base[metric] else 0.0
```
jserv: ratio = current/base if base[metric] else 0.0 forces ratio to 0.0 when the baseline is exactly 0, and 0.0 then classifies as "improved" (ratio < 1.0 - threshold) rather than uncomparable.

Fix: when base[metric] == 0 emit an explicit n/a verdict and skip both regression and improvement classification.

scripts/bench-compare.py
```diff
@@ -0,0 +1,206 @@
...
+        default=os.environ.get("BENCH_REGRESSION_THRESHOLD", "0.20"),
+        help="regression threshold as a fraction (default 0.20 = +20%%)")
+    parser.add_argument("--report-only", action="store_true",
+                        default=os.environ.get("BENCH_REPORT_ONLY") == "1",
+                        help="print the report but always exit 0")
+    parser.add_argument("--fail-on-missing", action="store_true",
+                        help="make missing baseline metrics fail even in "
+                             "report-only mode")
+    args = parser.parse_args()
+
```
jserv, line 177: load_json guards only FileNotFoundError/JSONDecodeError, so a syntactically valid top-level JSON that isn't an object ([], 42, null) reaches results.get(...) and raises an uncaught AttributeError with a raw traceback instead of a clean exit.

Fix: assert the parsed document is a dict and route non-objects through the same invalid-JSON exit path (also applies to bench-promote.py).

scripts/bench-promote.py
```diff
@@ -0,0 +1,206 @@
...
+                out[name] = sub
+    return out
+
+
+def merge_tree(dest: dict, src: dict, prefix: str,
+               added: list, updated: list) -> None:
+    for name, value in src.items():
+        tag = f"{prefix}{name}"
+        if isinstance(value, dict):
+            merge_tree(dest.setdefault(name, {}), value, f"{tag}.",
```
jserv: merge_tree(dest.setdefault(name, {}), value, ...) crashes with AttributeError if the existing baseline value at name is a scalar while the results value is a dict, and conversely silently overwrites a nested subtree with a scalar when shapes are reversed.

Fix: detect a scalar-vs-dict shape mismatch between dest[name] and value and error out (or replace and report) instead of blindly setdefault-ing.

scripts/bench-promote.py
```diff
@@ -0,0 +1,206 @@
...
+        print("bench-promote: --dry-run, not writing", file=sys.stderr)
+        return 0
+
+    fd, tmp_name = tempfile.mkstemp(prefix=f".{args.baseline.name}.",
+                                    suffix=".tmp", dir=args.baseline.parent,
+                                    text=True)
+    try:
+        os.fchmod(fd, stat.S_IMODE(args.baseline.stat().st_mode))
+        with os.fdopen(fd, "w", encoding="utf-8") as tmp:
+            tmp.write(json.dumps(baseline, indent=2) + "\n")
```
jserv: json.dumps(baseline, indent=2) uses sort_keys=False, so newly promoted metrics append in results iteration order and churn the hand-authored alphabetized baseline, producing noisy order-dependent diffs.

Fix: pass sort_keys=True.

.github/workflows/main.yml
```diff
@@ -585,6 +638,37 @@ jobs:
...
+
+      - name: Compare against baseline (soft regressions)
+        # Soft signal for now: the self-hosted runner's run-to-run variance
+        # is not yet characterized, so a regression prints loudly in the log
+        # but does not fail the leg. Flip criterion: once the leg has run
+        # for ~2 weeks / ~20 runs with every ratio inside the +20% threshold,
+        # drop BENCH_REPORT_ONLY=1 so regressions hard-fail.
+        if: ${{ matrix.run_bench }}
+        run: |
+          BENCH_REPORT_ONLY=1 python3 scripts/bench-compare.py --fail-on-missing
```
jserv: `--fail-on-missing` combined with BENCH_REPORT_ONLY=1 hard-fails on any missing metric, so a single transient lmbench ERR/hang (exactly what the retry+ERR machinery exists for) reds this leg despite the "soft signal for now" intent. This is deliberate per the PR description, but worth reconsidering during the ~2-week ramp: consider dropping --fail-on-missing until the flip that removes BENCH_REPORT_ONLY.

tests/bench-suite.sh
```diff
@@ -0,0 +1,1202 @@
...
+    tier2_run "${gitc[@]}" add -A > /dev/null 2>&1 \
+        || bench_die "git add failed"
+    tier2_run "${gitc[@]}" commit -q -m corpus-snapshot \
+        > /dev/null 2>&1 || bench_die "git commit failed"
+
+    run_startup_metric
+    run_tier2_tool python_startup_ms "" python3 -c pass
+    run_tier2_tool git_status_ms "" git -c 'safe.directory=*' \
+        --no-optional-locks -C "${GUEST_WORK}/corpus" status --porcelain
+    run_tier2_tool ripgrep_ms "" rg --no-config -c syscall_entry_ \
```
jserv: `rg -c syscall_entry_` exits 1 when it finds zero matches, and the timeit wrapper treats any non-zero child status as fatal, so a future corpus edit that drops the literal would abort the whole benchmark run instead of recording a metric.

Fix: use a pattern guaranteed to match, or tolerate rg's exit-1 no-match case explicitly.

scripts/ci-host-lock.sh
```diff
@@ -0,0 +1,110 @@
...
+
+r, w = os.pipe()
+child = os.fork()
+if child == 0:
+    # Holder: acquire (blocking), report through the pipe, then keep
+    # the descriptor open until release kills us or we self-expire.
+    os.close(r)
+    os.setsid()
+    fd = os.open(lock_path, os.O_RDWR | os.O_CREAT, 0o666)
+    fcntl.flock(fd, flags)
```
jserv: flock(2) provides no writer priority, so under overlapping workflow runs a steady stream of shared (normal-leg) acquirers can starve the benchmark leg's exclusive request until its wait budget expires.

Fix: gate new shared acquisitions behind a writer-waiting flag once an exclusive request is pending.

tests/bench-suite.sh
```diff
@@ -0,0 +1,1202 @@
...
+
+bench_note()
+{
+    printf "%s %s\n" "${BLUE}bench:${RESET}" "$*"
+}
+
+# Run one timed sample: prints elapsed microseconds on stdout.
+# Any non-zero exit (including timeout's 124) is fatal -- a failing
+# sample must never be smoothed into a fast measurement.
+bench_sample_us()
```
jserv: bench_sample_us() is defined but never called (Tier-2 timing goes through timeit_sample_us), leaving dead code that can rot out of sync with the real sampler.

Fix: delete bench_sample_us or wire it in where a host-side timed sample is actually needed.

jserv:
Once the benchmark is complete, upload the results to GitHub Pages, as [rv32emu-bench](https://sysprog21.github.io/rv32emu-bench/) does.

.github/workflows/bench-report.yml
```diff
@@ -0,0 +1,172 @@
...
+jobs:
+  comment:
+    name: Comment on benchmark changes
+    if: ${{ github.event.workflow_run.event == 'pull_request' }}
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    steps:
+      - name: Find benchmark report artifact
+        id: artifact
+        uses: actions/github-script@v8
```
jserv: Bump to v9, as https://github.com/actions/github-script

tests/bench-corpus/README.md
```diff
@@ -0,0 +1,19 @@
+# bench-corpus
+
+Fixed synthetic source tree consumed by `tests/bench-suite.sh` for the
+Tier-2 application benchmarks (issue #195):
```
jserv: Don't mention issue numbers.

tests/bench-corpus/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+The content is deterministic output of a one-shot generator and is part
+of the benchmark definition: editing, reformatting, or regenerating it
+changes every measurement and invalidates `tests/bench-baseline.json`.
+Do not modify these files; if the corpus ever must change, regenerate
+the baseline in the same change (see the header of
+`tests/bench-suite.sh` for the refresh procedure).
+
+Files use conventional `.c` suffixes. The formatter explicitly excludes this
+directory because its C-like content is benchmark data, not compilation input.
```
jserv: Drop this.

.github/workflows/main.yml
henrybear327: nit: rewrite the entire comment to make it clearer

.github/workflows/main.yml
```diff
@@ -392,10 +402,36 @@ jobs:
             ubsan_options: ''
             tsan_options: halt_on_error=1
             test_timeout: '60'
-            job_timeout: 45
+            job_timeout: 60
+            shared_lock_wait: 15
             run_matrix: false
             check_target: check-sanitizer
             brew_pkgs: binutils
+          # Benchmark leg (issue #195): release build, two-tier suite
```
henrybear327: ditto

docs/testing.md
```diff
@@ -148,6 +150,211 @@ or run all matrix modes back-to-back with `make test-matrix`.
 green `make check` covers BusyBox validation. Use `make test-busybox` to
 iterate on a single applet failure without rerunning the unit suite.
 
+## Performance Benchmarks
```
henrybear327, line 145: nit: make the text concise

I think the current `testing.md` has grown to the size that's already too big and too complicated to read

henrybear327 (approved):
In general, LGTM. Just some nits, as the text / comment is bloating a bit I feel. 

Thanks.

jserv:
> Once the benchmark is complete, upload the results to GitHub Pages, as [rv32emu-bench](https://sysprog21.github.io/rv32emu-bench/) does.

I create [elfuse-bench](https://github.com/sysprog21/elfuse-bench) and @Max042004 can maintain it directly.

jserv:
Instead of using hard-to-understand filenames such as `mod_00.c`, use more descriptive names. If the files need to be ordered, filenames such as `01-function.c` would work well instead.

scripts/ci-host-lock.sh
```diff
@@ -0,0 +1,278 @@
...
+            try:
+                os.unlink(marker_path)
+            except FileNotFoundError:
+                pass
+        if marker_fd >= 0:
+            os.close(marker_fd)
+        if listener is not None:
+            listener.close()
+        try:
+            os.unlink(control_path)
```
jserv, line 203: Control-socket cleanup race that can orphan a live lock holder. The detached holder unconditionally `os.unlink(control_path)` here even on the clean, socket-driven release path (flock already dropped, `released=True`). The bench step in main.yml does `release` (holder A) immediately followed by `acquire exclusive` (holder B) reusing the same fixed `CONTROL_FILE` path. If holder A is descheduled -- exactly the busy-host scenario this lock targets -- so its unlink lands after holder B has re-`bind()`ed the recreated socket, A deletes B's socket. B's later EXIT-trap release then hits FileNotFoundError, reports nothing to release, and B keeps the exclusive flock until the 120-min self-expiry, wedging every sibling leg's shared acquire. Suggest guarding this unlink with `if not released:` so a successor holder that re-bound the same path is never clobbered.

scripts/ci-host-lock.sh
```diff
@@ -0,0 +1,278 @@
...
+    client = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    client.settimeout(5)
+    client.connect(control_path)
+    client.sendall(b"release")
+    released = client.recv(16) == b"released"
+    client.close()
+except (FileNotFoundError, ConnectionError, socket.timeout):
+    pass
+
+for path in (pid_path, control_path):
```
jserv, line 263: Same class as the holder-side race: the release client unlinks the control socket unconditionally, including when it did not confirm a release. A `socket.timeout` or `ConnectionError` on connect/recv is swallowed above, then this loop removes a still-live holder's socket path -- if the holder was briefly stuck before `accept()`, it never gets the release, never drops its flock, and is now permanently unreachable for any future release (held until self-expiry). Suggest unlinking `control_path` only when `released` is true; leave a still-live holder's socket in place so a retry can reach it.

tests/bench-suite.sh
```diff
@@ -0,0 +1,1194 @@
...
+            GUEST_WORK="/bench-work"
+            local prep
+            prep='rm -rf /bench-work && mkdir -p /bench-work'
+            prep="$prep && cp -R /mnt/host/tests/bench-corpus /bench-work/corpus"
+            prep="$prep && mkdir -p /bench-work/rg-corpus"
+            prep="$prep && i=0 && while [ \$i -lt $RG_SHARDS ]; do cp -R /bench-work/corpus /bench-work/rg-corpus/\$i || exit 1; i=\$((i + 1)); done"
+            prep="$prep && cat /bench-work/corpus/*/*.c > /bench-work/zin"
+            prep="$prep && i=0 && while [ \$i -lt $ZSTD_DOUBLINGS ]; do cat /bench-work/zin /bench-work/zin > /bench-work/zin.2 && mv /bench-work/zin.2 /bench-work/zin || exit 1; i=\$((i + 1)); done"
+            qemu_exec timeout "$TEST_TIMEOUT" sh -c "$prep" \
+                > /dev/null 2>&1 \
```
jserv, line 941: The qemu Tier-2 corpus prep (512-shard `cp -R` plus the zstd doublings) runs under `timeout "$TEST_TIMEOUT"` -- the per-sample bound -- and `bench_die`s the whole run on a slow-but-valid guest. The native (`prepare_corpus`) and orbstack prep paths do the same heavy copy work untimed, so the transports bound prep inconsistently and a legitimately slow guest FS becomes a hard failure rather than a measurement. Suggest bounding prep with its own (larger) timeout applied consistently across transports, separate from the per-sample `TEST_TIMEOUT`.

tests/bench-suite.sh
```diff
@@ -0,0 +1,1194 @@
...
+            out[metric] = {
+                "median": round(statistics.median(timed), 6),
+                "samples": timed,
+                "unit": "us",
+            }
+
+# Tier-2 rows are bench-timeit's workload-only microseconds; report ms.
+apps = data.get("applications", {})
+for bench, rows in sorted(apps.items()):
+    vals = rows[bench]
```
jserv, line 1156: `vals = rows[bench]` assumes the metric key equals the bench key. Every current Tier-2 producer emits `applications <metric> <metric> <value>` so it holds today, but it is an unguarded invariant: any future row where bench != metric raises KeyError and crashes the JSON emitter with an opaque traceback instead of a diagnosable error. Suggest iterating `rows.items()` (emit each metric under `bench`) or asserting the invariant with a clear message.

tests/bench-timeit.c
```diff
@@ -0,0 +1,129 @@
...
+        return 1;
+    }
+    uint64_t elapsed = monotonic_ns() - start;
+
+    if (!WIFEXITED(status) || WEXITSTATUS(status) != 0) {
+        fprintf(stderr, "bench-timeit: command failed (status 0x%x)\n", status);
+        return 1;
+    }
+
+    printf("%llu\n", (unsigned long long) (elapsed / 1000));
```
jserv, line 127: `elapsed / 1000` floors nanoseconds to microseconds, so any workload finishing under 1 microsecond prints 0. Downstream `summarize()` accepts that as a valid fast sample and bench-compare scores it as a large improvement rather than a broken/instant measurement. Harmless for the current ms-scale corpus, but it removes the only guard against a workload that does no work reading as improved. Suggest emitting higher resolution (keep nanoseconds) or rejecting `elapsed < 1000` as an invalid sample.

jserv (commented):
Checked and found correct: the reader/writer flock logic never over-grants (shared is refused while an exclusive lock is held or pending, and the double writer_pending() check closes the publish race); bench-compare regression math and its baseline<=0 / division guards; threshold parsing; missing-metric-as-hard-failure under --fail-on-missing; bench-promote's atomic tempfile+fsync+os.replace and partial-capture provenance; and the workflow's EXIT-trap releases plus the always() net.

jserv:
This work should include #292 .

jserv (changes requested):
Rebase latest `main` branch and resolve conflict. Meanwhile, refine per review messages.

PR #227, opened 2026-07-18
Report startstack in procfs stat

### LINKED ISSUE #226: `/proc/self/stat` reports startstack as 0 that breaking stack discovery
## Summary

`/proc/self/stat` reports field 28, `startstack`, as `0`.

Some Linux runtimes and crash/unwind paths read `/proc/self/stat` field 28, then look up that address in `/proc/self/maps` to identify the process stack VMA. Under `elfuse`, that lookup fails because address `0` is not inside any stack mapping, even though `/proc/self/maps` correctly contains a `[stack]` entry.

## Observed Behavior

`/proc/self/maps` contains a stack VMA, for example:

    7801000-8000000 rw-p 00000000 00:00 0    [stack]

But `/proc/self/stat` emits zeros for `startcode`, `endcode`, and `startstack`:

    field 25 rsslim     = 18446744073709551615
    field 26 startcode  = 0
    field 27 endcode    = 0
    field 28 startstack = 0

A runtime then fails to find stack address `0` in `/proc/self/maps` and reports:

    stack not found in /proc/self/maps

In the traced run, that failure caused the process to raise `SIGABRT`:

    rt_sigqueueinfo(tid=1, sig=6, si_code=-1)

## Expected Behavior

`/proc/self/stat` field 28 should contain the initial user stack pointer, or at minimum an address inside the tracked `[stack]` VMA.

It should not be `0` while `/proc/self/maps` reports a valid `[stack]` range.

## Actual Behavior

`/proc/self/stat` field 28 is always emitted as `0` by the synthetic stat formatter.

## Relevant Code

In `src/runtime/procemu.c`, the `/proc/self/stat` formatter includes:

    "18446744073709551615 0 0 0 0 0 0 "

Those fields correspond to:

    rsslim startcode endcode startstack kstkesp kstkeip ...

So `startstack` is currently hard-coded to `0`.

## Suggested Fix

Record the initial guest user stack pointer during bootstrap/exec and emit it as field 28 in `/proc/self/stat`.

If `elfuse` does not yet store the exact initial SP, use a conservative address inside the tracked stack VMA until exact tracking is added.

The `/proc/self/task/<tid>/stat` formatter should also avoid reporting `startstack` as `0` if consumers may read task stat instead of process stat.

doanbaotrung:
Record the initial guest user stack pointer and expose it as field 28 in /proc/self/stat and /proc/self/task/<tid>/stat.

Carry the value across fork IPC and bump the private protocol magic so mixed parent/child builds reject the incompatible header layout.

Add procfs regression coverage to ensure startstack points inside the reported [stack] VMA.

Fix #226

---
## Summary by cubic
Expose the initial user stack pointer as startstack (field 28) in /proc/self/stat and /proc/self/task/<tid>/stat. Persist it across exec and fork, and add tests to confirm it’s the recorded SP and inside the [stack] VMA. Fixes #226.

- New Features
  - Record `start_stack` during bootstrap and exec; emit it as field 28 in `/proc/self/stat` and `/proc/self/task/<tid>/stat` (with a safe fallback if unavailable).
  - Preserve `start_stack` through fork IPC; bump `FORK_IPC_PROTOCOL_MAGIC` to ELFN to reject mixed builds.
  - Add procfs tests for self and task stats to verify it lies within [stack], uses the recorded SP (not the fallback), and is preserved across fork.

- Migration
  - Parent and child must run matching builds due to the protocol bump. Rebuild both together.

<sup>Written for commit 353416193d90909f52617e3f5045772f0f561fd1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/227?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/runtime/forkipc.c
+++ b/src/runtime/forkipc.c
@@ -230,6 +230,7 @@ int fork_child_main(int ipc_fd,
     g.elf_load_min = hdr.elf_load_min;
     g.stack_base = hdr.stack_base;
     g.stack_top = hdr.stack_top;
+    g.start_stack = hdr.start_stack;
     g.mmap_next = hdr.mmap_next;
     g.mmap_end = hdr.mmap_end;
     g.pt_pool_next = hdr.pt_pool_next;
@@ -1678,6 +1679,7 @@ int64_t sys_clone(hv_vcpu_t vcpu,
         .brk_current = g->brk_current,
         .stack_base = g->stack_base,
         .stack_top = g->stack_top,
+        .start_stack = g->start_stack,
         .mmap_next = g->mmap_next,
         .mmap_end = g->mmap_end,
         .pt_pool_next = g->pt_pool_next,
--- a/src/runtime/procemu.c
+++ b/src/runtime/procemu.c
@@ -2538,7 +2538,18 @@ static int proc_open_meminfo(void)
  * the tid is unparseable or the leaf is unknown. Split out of
  * proc_intercept_open to keep that dispatcher readable.
  */
-static int proc_open_self_task_node(const char *path, int linux_flags)
+static uint64_t proc_start_stack(const guest_t *g)
+{
+    if (g->start_stack != 0)
+        return g->start_stack;
+    if (g->stack_top > g->stack_base)
+        return g->stack_top - 16;
+    return 0;
+}
+
+static int proc_open_self_task_node(const guest_t *g,
+                                    const char *path,
+                                    int linux_flags)
 {
     char *endp;
     long tid = strtol(path + 16, &endp, 10);
@@ -2552,13 +2563,17 @@ static int proc_open_self_task_node(const char *path, int linux_flags)
     }
 
     if (!strcmp(endp, "/stat")) {
+        uint64_t start_stack = proc_start_stack(g);
         return proc_emit_fmt(
-            "%ld (%.15s) R %lld %lld %lld 0 0 0 0 0 0 0 0 0 0 0 "
-            "20 0 %d 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 "
-            "0 0 0 0 0 0 0 0\n",
+            "%ld (%.15s) R %lld %lld %lld 0 0 0 "          /* 1-9 */
+            "0 0 0 0 0 0 0 0 "                             /* 10-17 */
+            "20 0 %d 0 0 0 0 "                             /* 18-24 */
+            "18446744073709551615 0 0 %llu 0 0 0 "         /* 25-31 */
+            "0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0\n", /* 32-52 */
             tid, proc_comm_name(), (long long) proc_get_ppid(),
             (long long) proc_get_pid(), /* pgid */
-            (long long) proc_get_sid(), thread_active_count());
+            (long long) proc_get_sid(), thread_active_count(),
+            (unsigned long long) start_stack);
     }
 
     if (!strcmp(endp, "/status")) {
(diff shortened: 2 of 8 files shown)


tests/test-procfs.c
```diff
@@ -137,6 +192,40 @@ int main(void)
         }
     }
 
+    TEST("procfs: stat startstack is in [stack]");
```
jserv, line 229: This test (and the task-stat one below) can pass vacuously. It only asserts `startstack` lies inside `[stack]`, but `proc_start_stack()` falls back to `stack_top - 16` when `g->start_stack == 0`, and that fallback is *also* inside `[stack]`. So if start_stack propagation ever breaks (left 0 on some path), the membership check still passes and the regression goes undetected. The test does catch a mis-placed stat field (that would read a neighboring literal 0), but not a lost value. Consider asserting the reported value differs from the `stack_top-16` fallback, or comparing against an independently derived SP.

src/runtime/forkipc.c
```diff
@@ -1678,6 +1679,7 @@ int64_t sys_clone(hv_vcpu_t vcpu,
         .brk_current = g->brk_current,
```
jserv, line 1679: This new wire-format field is the riskiest part of the PR, but nothing tests that it survives the fork IPC round-trip. `test-fork-ipc-protocol-host.c` only static-asserts the magic bumped to ELFN; no test forks a child and reads the child's `/proc/self/stat` field 28 to confirm `start_stack` packs/unpacks at the right offset. A fork→child field-28 check would close this.

src/runtime/procemu.c
```diff
@@ -2552,13 +2563,17 @@ static int proc_open_self_task_node(const char *path, int linux_flags)
     }
 
     if (!strcmp(endp, "/stat")) {
+        uint64_t start_stack = proc_start_stack(g);
```
jserv, line 2566: `/proc/self/task/<tid>/stat` reports the process-wide `g->start_stack` for every tid, whereas real Linux reports each thread's own stack start. Harmless today (single `guest_t`, no per-thread stack tracking, the test only exercises the main tid), but a latent gap for `CLONE_VM` threads reading a non-main tid's field 28.

tests/test-procfs.c
```diff
@@ -137,6 +192,40 @@ int main(void)
...
+    {
+        char statbuf[1024];
+        char mapsbuf[4096];
```
jserv: `mapsbuf[4096]` truncates silently at `bufsz-1`. `[stack]` is typically the last VMA in `/proc/self/maps`, so if the emitted maps ever exceed 4095 bytes the `[stack]` line is cut and `parse_stack_range` returns false → spurious test failure. This fails safe (won't mask a bug), just a latent flakiness source as the emulated region set grows.

jserv (commented):
The core change is correct (field 28 lands in the right column in both emitters, fallback keeps 16-byte SP alignment, fork-IPC struct insertion is 8-byte aligned, all four stack-establishing paths set start_stack, and the ELFM->ELFN magic bump rejects mixed builds).

tests/test-procfs.c
```diff
@@ -27,6 +28,60 @@ int passes = 0, fails = 0;
...
+    if (!p)
+        return false;
+    p++;
+
+    int field = 3;
+    while (*p == ' ')
+        p++;
+    while (*p != '\0' && field <= target) {
+        char *endp;
+        if (field == 3) {
```
jserv, line 49: `parse_proc_stat_field_u64` special-cases field 3 (`endp = p+1`, no `field == target` check), so it can't return field 3 and assumes state is exactly one character. Not exercised here (target is always 28), so no live impact — flagging only as a latent constraint if this helper gets reused.

doanbaotrung:
Fixed base on comments. What changed:

- The startstack tests no longer pass vacuously on the stack_top - 16 fallback.
- Added a fork test that confirms the child sees the same field 28 value after fork IPC restore.
- Increased /proc/self/maps buffer from 4 KiB to 64 KiB.
- Tightened parse_proc_stat_field_u64() so it explicitly rejects unsupported fields < 4 instead of pretending it can parse field 3.

jserv:

PR #228, opened 2026-07-19
Resolve /dev/shm host paths in one place

henrybear327:
macOS has no `/dev/shm`, so elfuse backs POSIX shared memory with a per-UID host directory, `/tmp/elfuse-shm-<uid>/<name>`, via the validated resolver `proc_dev_shm_resolve` (flat leaf only; "..", embedded '/', and empty names rejected). 

Only three syscalls performed that redirect by hand: open, stat, and unlinkat. Every other path syscall fell through path_translate_at, which prepends the sysroot instead, so the same guest path resolved two different ways:
```
    /dev/shm/foo
      open  -> /tmp/elfuse-shm-1000/foo   (backing dir, created)
      chmod -> <sysroot>/dev/shm/foo      (nothing here -> ENOENT)
```

LTP's library startup (lib/tst_test.c setup_ipc) is failing to run all tests due to the following:
- An open(O_CREAT|O_EXCL) on a /dev/shm path followed by SAFE_CHMOD of the same path
- The open landed in the backing dir and succeeded
- The chmod resolved the sysroot path, hit ENOENT, and aborted the process in common setup, so all 24 conformance tests failed before any test body ran.

The fix is to do the redirect once, in path_translate_at, through the same resolver and a new shm_redirect flag, so chmod, chown, truncate, utimensat, rename, link, symlink, mknod, readlink, mkdir, statfs, and the xattr family all inherit the backing path from one choke point.

The backing path is absolute and POSIX ignores dirfd for absolute paths, so path_tx_dirfd() selects AT_FDCWD and the hand-rolled unlinkat rewrite folds away.

---
## Summary by cubic
Centralized /dev/shm path resolution so all path syscalls and exec use the same per-UID backing directory on macOS. Enforced never-follow and tmpfs reporting so shm symlinks cannot escape to the host; fixes LTP setup_ipc open(O_CREAT) → chmod failure.

- **Bug Fixes**
  - Redirect /dev/shm/<name> in `path_translate_at` via the validated resolver and set `is_dev_shm`; use `path_translation_dirfd` for absolute redirects; remove the `unlinkat` special-case.
  - Enforce never-follow on shm paths: `path_translation_at_flags` adds `AT_SYMLINK_NOFOLLOW` on *at and xattr; lstat for stat and statfs leaf checks; exec image/interpreter opened with `O_NOFOLLOW`; truncate/chdir reach the leaf via a nofollow fd (`O_NONBLOCK` for FIFOs); clear `AT_SYMLINK_FOLLOW` on `linkat`; report `/dev/shm` and leaves as tmpfs (synthetic `statfs`, `TMPFS_MAGIC`).
  - Tighten name gate: allow flat names containing ".."; reject exact "..", '/', and empty with `EACCES`; return `ENAMETOOLONG` on overflow; use lstat in proc stat; refuse a `/dev/shm` interpreter during bootstrap.
  - Add `test-dev-shm-paths`: covers open→chmod, metadata ops, chdir/getcwd, rename/link/unlink, tmpfs statfs for root/leaves and FIFOs, symlink containment, exec of symlink refused and real binary runs, and name-shape/oversize gates; enabled in `tests/manifest.txt`.

<sup>Written for commit a89a59c1f714c824e7bafaa965f482248dcaff9f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/228?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/runtime/procemu.c
+++ b/src/runtime/procemu.c
@@ -683,11 +683,35 @@ static int proc_parse_fd_index(const char *path,
     return (int) n;
 }
 
-/* Resolve a /dev/shm/<suffix> guest path to a host path inside the per-UID shm
- * dir. Rejects empty, traversing, or compound suffixes with EACCES; reports
- * ENAMETOOLONG when the host path overflows. The same validation runs in
- * proc_intercept_open and proc_intercept_stat, so the helper is one source of
- * truth for the security gate.
+/* Map a guest /dev/shm/<name> path to its host backing path, and gate the name.
+ *
+ * macOS has no /dev/shm, so elfuse backs POSIX shared memory with a per-UID
+ * host directory (/tmp/elfuse-shm-<uid>/). This is the single source of truth
+ * for that mapping. Callers proc_intercept_open, proc_intercept_stat, and
+ * path_translate_at (which records the hit in path_translation_t.is_dev_shm)
+ * all resolve through here, so one guest shm path never resolves two ways
+ * (e.g. open landing in the backing dir while chmod falls through to the
+ * sysroot).
+ *
+ * A POSIX shm name is always a single flat component: glibc's __shm_get_name
+ * (posix/shm-directory.c) strips the leading slash and rejects an empty name or
+ * any embedded '/' with EINVAL. This enforces the same shape, also rejects the
+ * ".." component (whole-component compare, so a flat name like "a..b" is fine),
+ * and returns EACCES (ENAMETOOLONG on overflow).
+ *
+ * The never-follow invariant lives here. On Linux /dev/shm is an in-namespace
+ * tmpfs, so a symlink planted at a shm leaf resolves inside that namespace.
+ * elfuse's backing store is a plain host directory, so the same symlink would
+ * resolve onto the host filesystem, escaping the sandbox. A symlink leaf is
+ * never legitimate anyway: glibc's shm_open (sysdeps/posix/shm_open.c) opens
+ * objects with O_NOFOLLOW. So every shm op must act on the leaf without
+ * following it. Because this resolver hands back an absolute host path that
+ * bypasses the sysroot, that duty is spread across syscall families and
+ * funneled through: path_translation_at_flags() (AT_SYMLINK_NOFOLLOW on the *at
+ * metadata calls), shm_open_leaf() (O_NOFOLLOW fd for truncate/chdir), proc
+ * open (O_NOFOLLOW), xattr (XATTR_NOFOLLOW), stat (lstat), linkat (clears
+ * AT_SYMLINK_FOLLOW), and statfs (answered synthetically, never touching the
+ * leaf).
  */
 static int dev_shm_resolve_path(const char *guest_suffix,
                                 char *host_path,
@@ -696,8 +720,8 @@ static int dev_shm_resolve_path(const char *guest_suffix,
     const char *shm = shm_dir_path();
     if (!shm)
         return -1;
-    if (strstr(guest_suffix, "..") || strchr(guest_suffix, '/') ||
-        guest_suffix[0] == '\0') {
+    if (guest_suffix[0] == '\0' || strchr(guest_suffix, '/') ||
+        !strcmp(guest_suffix, "..")) {
         errno = EACCES;
         return -1;
     }
@@ -3497,12 +3521,15 @@ int proc_intercept_stat(const char *path, struct stat *st)
                            path); /* sticky bit, like real /dev/shm */
         return 0;
     }
-    /* /dev/shm/<name> files: check the host temp dir */
+    /* /dev/shm/<name> files: check the host backing dir, and lstat rather than
+     * stat so a planted symlink leaf is never followed (see
+     * dev_shm_resolve_path).
+     */
     if (!strncmp(path, "/dev/shm/", 9)) {
         char host_path[512];
         if (dev_shm_resolve_path(path + 9, host_path, sizeof(host_path)) < 0)
             return -1;
-        return stat(host_path, st);
+        return lstat(host_path, st);
     }
 
     /* /dev/pts directory and /dev/pts/N slave entries. glibc ptsname(3) stats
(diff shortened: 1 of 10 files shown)


src/syscall/path.c
```diff
@@ -200,6 +202,26 @@ int path_translate_at(guest_fd_t dirfd,
...
+     * path is absolute, self-contained, and must never be sidecar-mapped.
+     * shm_redirect signals the redirect to callers, which must force nofollow
+     * on the host call; see dev_shm_resolve_path() for that invariant.
+     */
+    if (!strncmp(tx->guest_path, "/dev/shm/", 9) && tx->guest_path[9] != '\0') {
+        if (proc_dev_shm_resolve(tx->guest_path + 9, tx->host_buf,
+                                 sizeof(tx->host_buf)) < 0)
+            return -1;
+        tx->host_path = tx->host_buf;
+        tx->shm_redirect = true;
```
jserv: Redirecting here also reaches follow-capable callers outside the nofollow-forced metadata set: `statfs()` (`fs-stat.c:434`), `chdir()` (`fs.c:1746`/`1752`), `execve()`/interp resolution (`exec.c:175`,`598`), and bootstrap interp (`bootstrap.c:214`). Each dereferences a symlink leaf onto the host, escaping the backing dir. Either carry `shm_redirect` into those calls and force nofollow (`fstatfs`/`fchdir` via an `O_NOFOLLOW` open, `O_NOFOLLOW` on exec), or reject a symlink shm leaf up front.

jserv (commented):
Centralizing the `/dev/shm` redirect in `path_translate_at` is the right call, but the shared primitive now reaches follow-capable callers that were never taught to force nofollow. Inline notes below. Test gap: `tests/test-dev-shm-paths.c` exercises the symlink-not-followed case only for the metadata ops, not for `statfs`/`chdir`/`execve` on a symlink leaf.

src/syscall/fs.c
```diff
@@ -2374,7 +2374,8 @@ int64_t sys_faccessat(guest_t *g,
     }
 
     int mac_flags = translate_faccessat_flags(flags);
-    if (faccessat(dir_ref.fd, tx.host_path, mode, mac_flags) < 0) {
+    if (faccessat(path_tx_dirfd(&tx, &dir_ref), tx.host_path, mode, mac_flags) <
```
jserv: `faccessat` doesn't set `AT_SYMLINK_NOFOLLOW` on `shm_redirect`, unlike `fchmodat`/`fchownat`/`utimensat`. Add `if (tx.shm_redirect) mac_flags |= AT_SYMLINK_NOFOLLOW;` so the leaf stays non-followed in the ENOENT-at-stat-time race window.

src/syscall/fs.c
```diff
@@ -2437,6 +2438,21 @@ int64_t sys_truncate(guest_t *g, uint64_t path_gva, int64_t length)
     if (rc != INT64_MIN)
         return rc;
 
+    /* truncate(2) has no nofollow variant, so reach a shm leaf via open
+     * O_NOFOLLOW + ftruncate (see dev_shm_resolve_path).
+     */
+    if (tx.shm_redirect) {
+        int fd = open(tx.host_path, O_WRONLY | O_NOFOLLOW);
```
jserv: `O_NOFOLLOW` guards symlinks but not FIFOs: `mknod` a shm FIFO, then `truncate` it, and this `open(O_WRONLY)` blocks the vCPU thread until a reader appears — Linux returns `EINVAL` immediately. Add `O_NONBLOCK` (fails fast with `ENXIO`), plus `O_CLOEXEC` on the short-lived fd.

src/runtime/procemu.c
```diff
@@ -678,11 +678,27 @@ static int proc_parse_fd_index(const char *path,
...
- * dir. Rejects empty, traversing, or compound suffixes with EACCES; reports
- * ENAMETOOLONG when the host path overflows. The same validation runs in
- * proc_intercept_open and proc_intercept_stat, so the helper is one source of
- * truth for the security gate.
+/* Resolve a /dev/shm/<suffix> guest path to a host path inside the per-UID
+ * shm backing dir. The suffix must be a single flat component: glibc's
+ * __shm_get_name (posix/shm-directory.c) strips the leading slash and rejects
+ * an empty name or any embedded '/' with EINVAL, so a real POSIX shm name is
+ * always a leaf. This helper enforces the same shape, additionally rejects
+ * "..", and reports EACCES (ENAMETOOLONG when the host path overflows).
```
jserv: The `..` gate below uses `strstr(guest_suffix, "..")` (line 710), which rejects any name containing `..`, so valid flat names like `a..b` now fail with `EACCES` across every redirected syscall. If the goal is only traversal prevention, compare the whole component: `strcmp(guest_suffix, "..") == 0`.

tests/test-dev-shm-paths.c
```diff
@@ -0,0 +1,276 @@
...
+
+static void name_fixtures(void)
+{
+    int pid = (int) getpid();
+    snprintf(shm_path, sizeof(shm_path), "/dev/shm/elfuse_paths_%d", pid);
+    snprintf(shm_path2, sizeof(shm_path2), "/dev/shm/elfuse_paths2_%d", pid);
+    snprintf(shm_link, sizeof(shm_link), "/dev/shm/elfuse_link_%d", pid);
+    snprintf(shm_evil, sizeof(shm_evil), "/dev/shm/elfuse_evil_%d", pid);
+    snprintf(victim_path, sizeof(victim_path), "/tmp/elfuse-shm-victim-%d",
+             pid);
```
jserv: Use macro to avoid duplicate `/dev/shm/` path.

henrybear327: Refactored.

jserv:

PR #230, opened 2026-07-19
Enable CNTVCT on cloned vCPUs

### LINKED ISSUE #229: Cloned vCPUs do not preserve EL0 virtual counter access
# Cloned vCPUs do not preserve EL0 virtual counter access

## Summary

A guest AArch64 process can read `CNTVCT_EL0` successfully on the initial
vCPU, but the same read from a thread created with `clone(CLONE_THREAD | ...)`
traps as an unknown MRS register and returns `0`.

This causes multithreaded guest programs that rely on the Linux/vDSO clock path
to see an invalid virtual counter on worker threads. In larger programs this can
lead to heavy `clock_gettime()` fallback loops, timer/futex stalls, and
secondary crash-reporting failures.

## Environment

- Host: macOS on Apple Silicon
- Runner: elfuse using Hypervisor.framework
- Guest arch: AArch64 Linux ELF

## Observed Behavior

For the initial vCPU, a tiny guest program that reads `CNTVCT_EL0` exits
successfully.

For a cloned thread, the child vCPU produces:

```text
clone(flags=0x50f00, stack=..., ptid=0x0, tls=0x0, ctid=0x0)
thread tid=2 starting on vCPU
MRS trap: unknown reg Op0=3 Op1=3 CRn=14 CRm=0 Op2=2 (hv_reg=0xdf02) -> 0
```

`Op0=3 Op1=3 CRn=14 CRm=0 Op2=2` is `CNTVCT_EL0`.

The child observes `CNTVCT_EL0 == 0`.

## Expected Behavior

All guest vCPUs should have the same EL0 access to the virtual counter as the
initial vCPU. A guest thread should be able to read `CNTVCT_EL0`, and the value
should be nonzero and monotonic.

## Root Cause

The initial vCPU setup enables EL0 access to physical/virtual counters:

```c
/* CNTKCTL_EL1.EL0VCTEN | EL0PCTEN */
HV_CHECK(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_CNTKCTL_EL1, 0x3ULL));
```

This is done in `src/core/bootstrap.c`.

However, worker vCPUs created for `clone(CLONE_THREAD | ...)` replay many parent
system registers in `src/runtime/forkipc.c`, but they do not set or copy
`HV_SYS_REG_CNTKCTL_EL1`.

The same omission appears to apply to the VM-clone child setup path.

As a result, EL0 reads of `CNTVCT_EL0` on cloned vCPUs trap into elfuse's MRS
emulation path. That path currently has a host fallback for `CNTFRQ_EL0`
`(3,3,14,0,0)`, but no fallback for `CNTVCT_EL0` `(3,3,14,0,2)`, so it logs
"unknown reg" and returns `0`.

## Minimal Reproducer

This assembly exits `0` if the cloned child can read a valid virtual counter,
and exits `42` if the child sees `CNTVCT_EL0 == 0`.

```asm
.global _start

.equ SYS_clone, 220
.equ SYS_exit_group, 94

.text
_start:
    adrp x1, child_stack_top
    add  x1, x1, :lo12:child_stack_top

    // CLONE_VM | CLONE_FS | CLONE_FILES | CLONE_SIGHAND |
    // CLONE_THREAD | CLONE_SYSVSEM
    movz x0, #0x0f00
    movk x0, #0x5, lsl #16

    mov x2, #0
    mov x3, #0
    mov x4, #0
    mov x8, #SYS_clone
    svc #0

    cbz x0, child

parent_spin:
    b parent_spin

child:
    mrs x0, cntvct_el0
    cbz x0, bad

    mrs x1, cntvct_el0
    cmp x1, x0
    b.lo bad

    mov x0, #0
    mov x8, #SYS_exit_group
    svc #0

bad:
    mov x0, #42
    mov x8, #SYS_exit_group
    svc #0

.bss
.balign 16
child_stack:
    .skip 65536
child_stack_top:
```

Build:

```sh
clang -target aarch64-linux-gnu -nostdlib -static -fuse-ld=lld \
  cntvct-clone-test.S -o cntvct-clone-test
```

Run:

```sh
build/elfuse --verbose ./cntvct-clone-test
echo $?
```

Observed exit is `42`, with the `CNTVCT_EL0` unknown-register log shown above.

## Suggested Fix

Set `HV_SYS_REG_CNTKCTL_EL1` for every newly-created vCPU, not only the initial
vCPU.

Likely affected paths:

- `thread_create_and_run()` in `src/runtime/forkipc.c`
- `vm_clone_thread_run()` in `src/runtime/forkipc.c`

For example, after creating/configuring the worker vCPU, apply the same setting
used during initial bootstrap:

```c
hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_CNTKCTL_EL1, 0x3ULL);
```

A regression test should cover reading `CNTVCT_EL0` in a cloned thread, not only
on the initial vCPU.
```

### PR DESCRIPTION
**Pull request by doanbaotrung**
Worker vCPUs created for clone threads did not get CNTKCTL_EL1 configured, even though the initial vCPU enables EL0 access to the architected counter.

Set CNTKCTL_EL1 during cloned vCPU setup so CNTVCT_EL0 reads work from guest threads as they do on the main vCPU. Add a pthread regression test that reads CNTVCT_EL0 from both main and worker vCPUs.

Fix #229

---
## Summary by cubic
Enable EL0 reads of the architected timers on all vCPUs by using a single CNTKCTL_EL1 mask in bootstrap and both cloned-vCPU paths (CLONE_THREAD and CLONE_VM). Restores CNTVCT_EL0/CNTPCT_EL0 for guest threads and adds a regression test. Fixes #229.

- **Bug Fixes**
  - Define CNTKCTL_EL1_EL0_TIMER_EN in hvutil.h; apply it in bootstrap, thread_create_and_run, and vm_clone_thread_run to enable CNTVCT_EL0 and CNTPCT_EL0 at EL0.
  - Add tests/test-cntvct-thread.c; verify CNTVCT_EL0 on the main vCPU and a CLONE_THREAD worker. Add to test matrix and update baselines (elfuse-aarch64 238, qemu-aarch64 218).

<sup>Written for commit 37c3c53ee52ab7784f97f349aaa0c13b6f835161. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/230?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/runtime/forkipc.c
+++ b/src/runtime/forkipc.c
@@ -863,6 +863,8 @@ static void *thread_create_and_run(void *arg)
     WORKER_HV(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_TCR_EL1, tca->tcr));
     WORKER_HV(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_TTBR0_EL1, tca->ttbr0));
     WORKER_HV(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_CPACR_EL1, tca->cpacr));
+    WORKER_HV(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_CNTKCTL_EL1,
+                                  CNTKCTL_EL1_EL0_TIMER_EN));
 
     /* All worker vCPUs in the process share the same shim_globals base (one VM
      * per process); a fresh TPIDR_EL1 set is still required because HVF created
@@ -1212,6 +1214,8 @@ static void *vm_clone_thread_run(void *arg)
     HV_CHECK(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_TCR_EL1, tca->tcr));
     HV_CHECK(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_TTBR0_EL1, tca->ttbr0));
     HV_CHECK(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_CPACR_EL1, tca->cpacr));
+    HV_CHECK(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_CNTKCTL_EL1,
+                                 CNTKCTL_EL1_EL0_TIMER_EN));
     HV_CHECK(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_SCTLR_EL1, tca->sctlr));
     HV_CHECK(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_SP_EL1, tca->sp_el1));
     HV_CHECK(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_SP_EL0, tca->child_stack));
--- a/tests/test-cntvct-thread.c
+++ b/tests/test-cntvct-thread.c
@@ -1,0 +1,105 @@
+/*
+ * test-cntvct-thread.c -- EL0 timer counter access on cloned vCPUs
+ *
+ * Copyright 2026 elfuse contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Linux exposes EL0 timer counter reads on AArch64. elfuse's vDSO fast path
+ * depends on CNTVCT_EL0, so this test verifies worker vCPUs created by
+ * clone(CLONE_THREAD) get the same virtual-counter configuration as the
+ * bootstrap vCPU.
+ *
+ * This does not exercise elfuse's CLONE_VM-without-CLONE_THREAD worker path.
+ * The production fix uses the same CNTKCTL_EL1_EL0_TIMER_EN constant for both
+ * clone paths so they cannot drift, but this regression test stays on the
+ * pthread/CLONE_THREAD path to remain valid on the QEMU reference lane too.
+ */
+
+#include <pthread.h>
+#include <stdint.h>
+
+#include "test-harness.h"
+
+#if !defined(__aarch64__)
+#error "test-cntvct-thread is aarch64-only"
+#endif
+
+int passes = 0, fails = 0;
+
+typedef struct {
+    uint64_t first;
+    uint64_t second;
+} timer_sample_t;
+
+static inline uint64_t read_cntvct(void)
+{
+    uint64_t value;
+    __asm__ volatile("mrs %0, cntvct_el0" : "=r"(value));
+    return value;
+}
+
+static void expect_timer_sample(const char *label, const timer_sample_t *sample)
+{
+    char msg[128];
+
+    /* Today elfuse's trapped-MRS fallback returns 0 for CNTVCT_EL0, so the
+     * nonzero check catches missing CNTKCTL_EL1.EL0VCTEN on cloned vCPUs. If a
+     * CNTVCT_EL0 host fallback is added later, this regression test must grow a
+     * stronger signal that EL0 direct counter access was configured.
+     */
+    snprintf(msg, sizeof(msg), "%s CNTVCT_EL0 was zero or non-monotonic",
+             label);
+    EXPECT_TRUE(sample->first != 0 && sample->second >= sample->first, msg);
+}
+
+static void *read_timer_thread(void *arg)
+{
+    timer_sample_t *sample = (timer_sample_t *) arg;
+
+    sample->first = read_cntvct();
+    sample->second = read_cntvct();
+
+    return NULL;
+}
+
+static void test_main_vcpu_timers(void)
+{
+    TEST("main vCPU EL0 timers");
+
+    timer_sample_t sample = {0};
+    sample.first = read_cntvct();
+    sample.second = read_cntvct();
+    expect_timer_sample("main vCPU", &sample);
+}
+
+static void test_clone_thread_vcpu_timers(void)
+{
+    TEST("clone(CLONE_THREAD) vCPU EL0 timers");
+
+    timer_sample_t sample = {0};
+    pthread_t thread;
+    int err = pthread_create(&thread, NULL, read_timer_thread, &sample);
+    if (err != 0) {
+        FAIL("pthread_create failed");
+        return;
+    }
+
+    err = pthread_join(thread, NULL);
+    if (err != 0) {
+        FAIL("pthread_join failed");
+        return;
+    }
+
+    expect_timer_sample("clone(CLONE_THREAD) vCPU", &sample);
+}
+
+int main(void)
+{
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: src/runtime/forkipc.c [hunk 1]
```diff
@@ -863,6 +863,7 @@ static void *thread_create_and_run(void *arg)
     WORKER_HV(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_TCR_EL1, tca->tcr));
     WORKER_HV(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_TTBR0_EL1, tca->ttbr0));
     WORKER_HV(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_CPACR_EL1, tca->cpacr));
+    WORKER_HV(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_CNTKCTL_EL1, 0x3ULL));
```
jserv: The literal `0x3ULL` now lives in three places (here, `vm_clone_thread_run` below, and `bootstrap.c:740`) with an explanatory comment only at the bootstrap site. Consider a shared named constant, e.g. `#define CNTKCTL_EL1_EL0_TIMER_EN 0x3ULL /* EL0VCTEN | EL0PCTEN */`, used by all three, so the initial and cloned vCPU configs cannot drift. At minimum add a one-line comment pointing back to the bootstrap rationale.

src/runtime/forkipc.c
```diff
@@ -1212,6 +1213,7 @@ static void *vm_clone_thread_run(void *arg)
     HV_CHECK(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_TCR_EL1, tca->tcr));
     HV_CHECK(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_TTBR0_EL1, tca->ttbr0));
     HV_CHECK(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_CPACR_EL1, tca->cpacr));
+    HV_CHECK(hv_vcpu_set_sys_reg(vcpu, HV_SYS_REG_CNTKCTL_EL1, 0x3ULL));
```
jserv: Same `0x3ULL` magic as the `thread_create_and_run` path and `bootstrap.c:740`. Fold all three into one named constant (or add a back-reference comment) so a future change to EL0 timer access updates every vCPU consistently.

tests/test-cntvct-thread.c
```diff
@@ -0,0 +1,81 @@
...
+                "main vCPU CNTVCT_EL0 was zero or non-monotonic");
+}
+
+static void test_worker_vcpu_cntvct(void)
+{
+    TEST("worker vCPU CNTVCT_EL0");
+
+    uint64_t values[2] = {0, 0};
+    pthread_t thread;
+    int err = pthread_create(&thread, NULL, read_cntvct_thread, values);
```
jserv: `pthread_create` maps to `clone(CLONE_THREAD)`, so this test only exercises the `thread_create_and_run` path (`forkipc.c:866`). The other patched site, `vm_clone_thread_run` (`forkipc.c:1216`, the CLONE_VM-without-THREAD path), is never hit — a regression dropping only that line would still pass here. Consider adding a raw `clone(CLONE_VM)` variant (as `test-simd-clone.c` does) to cover it, or note the gap in the header.

tests/test-cntvct-thread.c
```diff
@@ -0,0 +1,81 @@
...
+    return 1;
+#endif
+}
+
+static void *read_cntvct_thread(void *arg)
+{
+    uint64_t *values = (uint64_t *) arg;
+
+    values[0] = read_cntvct();
+    values[1] = read_cntvct();
```
jserv: The fix sets `0x3` = EL0VCTEN | EL0PCTEN, but the test reads only `CNTVCT_EL0`. A regression to `0x2` (dropping EL0PCTEN) would still pass. Read `CNTPCT_EL0` here as well and assert the same properties to cover both bits.

tests/test-cntvct-thread.c
```diff
@@ -0,0 +1,81 @@
...
+        return;
+    }
+
+    err = pthread_join(thread, NULL);
+    if (err != 0) {
+        FAIL("pthread_join failed");
+        return;
+    }
+
+    EXPECT_TRUE(values[0] != 0 && values[1] >= values[0],
```
jserv: This `!= 0` check discriminates the bug only because the MRS-trap handler currently returns 0 for CNTVCT (proc.c has a host fallback only for CNTFRQ_EL0). If a CNTVCT_EL0 host fallback is later added next to it, trap emulation would return a nonzero value and this test would pass even with the vCPU fix reverted. Worth a comment tying the two together so the coupling is not silently broken.

tests/test-cntvct-thread.c
```diff
@@ -0,0 +1,81 @@
...
+int passes = 0, fails = 0;
+
+static inline uint64_t read_cntvct(void)
+{
+#if defined(__aarch64__)
+    uint64_t value;
+    __asm__ volatile("mrs %0, cntvct_el0" : "=r"(value));
+    return value;
+#else
+    return 1;
```
jserv: The non-aarch64 `return 1;` makes both assertions pass trivially — a silent green rather than a skip. Prefer `return 0;` (visible FAIL) or `#error "aarch64-only test"` so an accidental native build fails loudly. Low priority: the Makefile always cross-compiles for aarch64, so this branch is dead in shipped builds.

doanbaotrung:
Implemented the reviewed fixes in `third_party/elfuse`:

- Added shared `CNTKCTL_EL1_EL0_TIMER_EN` in `hvutil.h`.
- Replaced all three `0x3ULL` CNTKCTL sites in `bootstrap.c` and `forkipc.c`.
- Expanded `test-cntvct-thread.c` to:
  - read both `CNTVCT_EL0` and `CNTPCT_EL0`
  - cover `pthread_create` / `CLONE_THREAD`
  - cover raw `clone(CLONE_VM | SIGCHLD)` for `vm_clone_thread_run`
  - fail loudly on non-aarch64 builds with `#error`
  - document the current “trapped MRS returns 0” coupling

jserv:

PR #232, opened 2026-07-20
Accept live thread TIDs in priority syscalls

### LINKED ISSUE #231: getpriority(PRIO_PROCESS, tid) returns ESRCH for live guest threads
### Description

A multithreaded guest program can abort when it calls
`getpriority(PRIO_PROCESS, tid)` for a live guest thread id.

On Linux, `PRIO_PROCESS` accepts a process id, and Linux thread ids are valid
targets because threads are represented as tasks. If `tid` refers to a live
thread in the same process, `getpriority(PRIO_PROCESS, tid)` should return that
task's nice value.

In elfuse, `getpriority` currently appears to accept only `who == 0` or
`who == guest_pid`. When a guest runtime calls `getpriority(PRIO_PROCESS, 2)`
for a live guest thread, elfuse returns `-ESRCH`.

### Observed Behavior

Verbose elfuse log:

```text
syscall 141 ... (0x0, 0x2, ...)
  -> -3
```

On AArch64:

```text
syscall 141 = getpriority
x0 = 0 = PRIO_PROCESS
x1 = 2 = target guest tid
return -3 = -ESRCH
```

The guest then aborts because `getpriority` unexpectedly failed with
`ESRCH`:

```text
Check failed: niceness != -1 || errno == 0  No such process
```

### Expected Behavior

If `who` names a live guest thread, elfuse should treat it as a valid
`PRIO_PROCESS` target and return the emulated nice value.

For example:

```text
getpriority(PRIO_PROCESS, 0)          -> current nice value
getpriority(PRIO_PROCESS, guest_pid)  -> current nice value
getpriority(PRIO_PROCESS, live_tid)   -> current nice value
getpriority(PRIO_PROCESS, dead_tid)   -> -ESRCH
```

`setpriority(PRIO_PROCESS, tid, prio)` should likely follow the same target
validation rules.

### Suspected Root Cause

`proc_sys_getpriority()` only accepts self or the process id:

```c
if (who != 0 && who != (int) guest_pid)
    return -LINUX_ESRCH;
```

This rejects valid guest thread ids even though elfuse tracks them in the thread
table.

### Suggested Fix

For `PRIO_PROCESS`, accept:

- `who == 0`
- `who == guest_pid`
- `thread_tid_alive(who)`

Then return `20 - emu_nice`.

Apply equivalent target validation to `proc_sys_setpriority()`.

### Test Suggestion

Add a regression test that creates a guest thread and calls:

```c
getpriority(PRIO_PROCESS, child_tid)
```

Expected result should be the current nice value, not `-ESRCH`.

Also test:

```c
setpriority(PRIO_PROCESS, child_tid, 5);
getpriority(PRIO_PROCESS, child_tid);
```

Expected result should reflect the emulated nice value.

doanbaotrung:
Linux treats thread IDs as valid PRIO_PROCESS targets because each thread is a task. elfuse only accepted who 0 and the guest process ID, so getpriority(PRIO_PROCESS, tid) returned ESRCH for live guest threads.

Share PRIO_PROCESS target validation between getpriority and setpriority, accepting live guest TIDs from the thread table.

Add regression coverage for live thread getpriority, live thread setpriority, and dead thread ESRCH behavior.

Fix #231

---
## Summary by cubic
Accept live guest TIDs as PRIO_PROCESS targets in getpriority, matching Linux and avoiding ESRCH for running threads. Add a shared `proc_pid_alive` used by getpriority and sched_* queries; keep setpriority self-only while nice is process-wide; add tests for live TID get, non-self set ESRCH, and dead TID ESRCH. Fixes #231.

<sup>Written for commit 41fa2d2f2bea9d6ceef2ddd1f412bbf6e4c820a5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/232?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/proc-identity.c
+++ b/src/syscall/proc-identity.c
@@ -13,6 +13,7 @@
 
 #include "syscall/abi.h"
 #include "core/shim-globals.h"
+#include "runtime/thread.h"
 #include "syscall/proc-identity.h"
 #include "syscall/proc.h"
 
@@ -223,10 +224,20 @@ void proc_set_nice(int32_t val)
     emu_nice = val;
 }
 
+bool proc_pid_alive(int pid)
+{
+    if (pid == 0 || pid == (int) guest_pid)
+        return true;
+    return thread_tid_alive((int64_t) pid) != 0;
+}
+
 int64_t proc_sys_setpriority(int which, int who, int prio)
 {
     if (which != 0)
         return -LINUX_EINVAL;
+    /* emu_nice is process-global. Reject non-self task IDs until nice is
+     * tracked per task rather than silently changing every thread's value.
+     */
     if (who != 0 && who != (int) guest_pid)
         return -LINUX_ESRCH;
     if (prio < -20)
@@ -241,7 +252,7 @@ int64_t proc_sys_getpriority(int which, int who)
 {
     if (which != 0)
         return -LINUX_EINVAL;
-    if (who != 0 && who != (int) guest_pid)
+    if (!proc_pid_alive(who))
         return -LINUX_ESRCH;
     return 20 - emu_nice;
 }
--- a/tests/test-credentials.c
+++ b/tests/test-credentials.c
@@ -34,6 +34,97 @@
 #define __NR_geteuid 175
 #define __NR_getgid 176
 #define __NR_setgroups 159
+#define __NR_sched_yield 124
+#define __NR_clock_gettime 113
+
+#define CLOCK_MONOTONIC 1
+
+#define CLONE_VM 0x00000100UL
+#define CLONE_FS 0x00000200UL
+#define CLONE_FILES 0x00000400UL
+#define CLONE_SIGHAND 0x00000800UL
+#define CLONE_THREAD 0x00010000UL
+#define CLONE_SYSVSEM 0x00040000UL
+#define CLONE_PARENT_SETTID 0x00100000UL
+#define CLONE_CHILD_CLEARTID 0x00200000UL
+#define CLONE_DETACHED 0x00400000UL
+
+typedef struct {
+    long tv_sec;
+    long tv_nsec;
+} test_timespec_t;
+
+static volatile int prio_child_tid = 0;
+static volatile int prio_child_ready = 0;
+static volatile int prio_child_release = 0;
+static int prio_dead_tid = 0;
+static char prio_child_stack[8192] __attribute__((aligned(16)));
+
+static long monotonic_ms(void)
+{
+    test_timespec_t ts = {0};
+    long rc = raw_syscall2(__NR_clock_gettime, CLOCK_MONOTONIC, (long) &ts);
+    if (rc != 0)
+        return -1;
+    return ts.tv_sec * 1000 + ts.tv_nsec / 1000000;
+}
+
+static void priority_child_work(void)
+{
+    prio_child_ready = 1;
+    raw_futex_wake((int *) &prio_child_ready, 1);
+
+    while (prio_child_release == 0)
+        raw_futex_wait((int *) &prio_child_release, 0);
+
+    raw_exit(0);
+}
+
+static long spawn_priority_child(void)
+{
+    prio_child_tid = 0;
+    prio_child_ready = 0;
+    prio_child_release = 0;
+
+    unsigned long flags = CLONE_VM | CLONE_FS | CLONE_FILES | CLONE_SIGHAND |
+                          CLONE_THREAD | CLONE_SYSVSEM | CLONE_PARENT_SETTID |
+                          CLONE_CHILD_CLEARTID | CLONE_DETACHED;
+    void *stack_top = prio_child_stack + sizeof(prio_child_stack);
+    long ret =
+        raw_clone(flags, stack_top, (int *) &prio_child_tid, /* parent_tid */
+                  0,                                         /* tls */
+                  (int *) &prio_child_tid);                  /* child_tid */
+    if (ret == 0) {
+        priority_child_work();
+        __builtin_unreachable();
+    }
+    return ret;
+}
+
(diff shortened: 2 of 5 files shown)


src/syscall/proc-identity.c
```diff
@@ -223,11 +224,18 @@ void proc_set_nice(int32_t val)
     emu_nice = val;
 }
 
+static bool prio_process_target_alive(int who)
```
jserv: `prio_process_target_alive()` is a near-verbatim copy of `sched_pid_alive()` in `src/syscall/sys.c:284`: accept `who==0`, accept the process pid, else defer to `thread_tid_alive()`. The only difference is `guest_pid` vs `proc_get_pid()`, which resolve to the same value. Hoist one shared helper and call it from both sites.

src/syscall/proc-identity.c
```diff
@@ -223,11 +224,18 @@ void proc_set_nice(int32_t val)
...
+        return true;
+    return thread_tid_alive((int64_t) who) != 0;
+}
+
 int64_t proc_sys_setpriority(int which, int who, int prio)
 {
     if (which != 0)
         return -LINUX_EINVAL;
-    if (who != 0 && who != (int) guest_pid)
+    if (!prio_process_target_alive(who))
```
jserv: `setpriority` writes the single process-global `emu_nice` regardless of which live TID is targeted, so setting nice on one thread changes the priority `getpriority` reports for every thread and for self. Linux gives each task its own nice. Widening the accepted `who` to any live TID makes this shared-state model observably wrong for the first time.

Either model per-TID nice, or keep the single-nice model but document it and reject non-self TIDs rather than silently accepting a TID and writing the global.

Note the new test can't catch this: it sets and reads back the same TID (`tests/test-credentials.c:359-361`), so a global and a per-thread implementation both yield 15.

tests/test-credentials.c
```diff
@@ -284,6 +332,50 @@ int main(void)
...
+                raw_futex_wait((int *) &prio_child_ready, 0);
+
+            int tid = prio_child_tid;
+            long rc = raw_syscall3(__NR_setpriority, 0, tid, 5);
+            long prio = raw_syscall2(__NR_getpriority, 0, tid);
+            EXPECT_TRUE(rc == 0 && prio == 15,
+                        "setpriority/getpriority(live tid) mismatch");
+            release_priority_child();
+
+            TEST("getpriority dead thread TID");
```
jserv: `TEST("getpriority dead thread TID")` is nested inside the `else` of the clone-failure check. If `spawn_priority_child()` fails, this sub-test's label and pass/fail count vanish entirely, skewing the totals. Hoist it to its own top-level `TEST(...)` block after the live-TID block.

tests/test-credentials.c
```diff
@@ -34,6 +34,54 @@
...
+{
+    prio_child_tid = 0;
+    prio_child_ready = 0;
+    prio_child_release = 0;
+
+    /* CLONE_VM | CLONE_FS | CLONE_FILES | CLONE_SIGHAND | CLONE_THREAD |
+     * CLONE_SYSVSEM | CLONE_SETTLS | CLONE_PARENT_SETTID |
+     * CLONE_CHILD_CLEARTID | CLONE_DETACHED.
+     */
+    unsigned long flags = 0x7d0f00;
```
jserv: Clone flags `0x7d0f00` include `CLONE_SETTLS` (0x80000) but the tls argument is `0`, so the TLS base is deliberately set to 0. The child only touches globals and raw syscalls, so it's harmless, but the flag is pointless and contradicts the comment. Drop `CLONE_SETTLS` (`0x7c0f00`) or pass a real tls pointer.

tests/test-credentials.c
```diff
@@ -284,6 +332,50 @@ int main(void)
...
+            int tid = prio_child_tid;
+            long rc = raw_syscall3(__NR_setpriority, 0, tid, 5);
+            long prio = raw_syscall2(__NR_getpriority, 0, tid);
+            EXPECT_TRUE(rc == 0 && prio == 15,
+                        "setpriority/getpriority(live tid) mismatch");
+            release_priority_child();
+
+            TEST("getpriority dead thread TID");
+            long dead_prio = 0;
+            for (int i = 0; i < 1000; i++) {
```
jserv: The dead-thread poll is the right shape (elfuse's `thread_deactivate()` legitimately lags the `CLONE_CHILD_CLEARTID` futex, so the retry is required), but the hard 1000-iteration `sched_yield` cap can expire under load and produce a spurious FAIL. Consider looping unbounded with a yield, or a time bound, instead of a fixed count.

doanbaotrung:
Fixed the review comments.

Changed:
- Hoisted the shared PID/TID liveness helper to `proc_pid_alive()` in `proc-identity.c`.
- Replaced the scheduler-local `sched_pid_alive()` in `sys.c` with `proc_pid_alive()`.
- Kept `getpriority(PRIO_PROCESS, live_tid)` supported.
- Changed `setpriority(PRIO_PROCESS, live_tid, ...)` to return `-ESRCH` for non-self TIDs, since `emu_nice` is still process-global.
- Updated the comment in `proc.h` to document that model.
- Fixed the tests:
  - removed pointless `CLONE_SETTLS`
  - used named clone flags
  - made dead-TID test top-level
  - changed fixed retry count to a 2s monotonic time bound
  - updated setpriority coverage to assert rejection and no global nice mutation

jserv:

PR #241, opened 2026-07-23
Restrict sysroot fallback to exclude system paths

### LINKED ISSUE #238: Guest system paths leak/escape to macOS host paths when absent in the sysroot
### Guest system paths leak/escape to macOS host paths when absent in the sysroot

**Problem Description:**  
The sysroot path translation functions (`proc_resolve_sysroot_path_flags` and `proc_resolve_sysroot_create_path`) are designed to fall back to the host literal path if a path (or its parent directory) does not exist in the guest sysroot. This fallback is useful for accessing host resources like `/etc/resolv.conf` or `/tmp`. 

However, for standard Linux system paths (e.g., `/usr/`, `/bin/`, `/sbin/`, `/lib/`, `/lib64/`, `/var/`, `/opt/`, `/etc/`), falling back to the host is incorrect. If a guest program checks or writes to a system path (like `/usr/bin/python3` or `/etc/ssh/...`) before it has been unpacked or created in the sysroot, it is redirected to the host macOS system directory.

**Implications:**  
1. **Permission Denied / SIP Errors**: Guest operations attempt to read/write from macOS host system directories, encountering permission issues (`EPERM`/`EACCES`) due to macOS root ownership and System Integrity Protection (SIP).
2. **Host Contamination**: Guest tools could see or load host binaries/libraries instead of failing with `ENOENT`.

**Proposed Solution:**  
Introduce a helper (e.g., `is_guest_system_path`) to check if a guest path belongs to standard system directories:
* `/usr/`, `/bin/`, `/sbin/`, `/lib/`, `/lib64/`, `/var/`, `/opt/`, `/boot/`, `/srv/`, `/root/`, `/home/`, `/etc/` (excluding `/etc/resolv.conf` and `/etc/hosts` to keep DNS/network resolution working).

If a path matches these system directories, **never fall back to the host**. Instead, force the path to resolve to the sysroot, allowing the guest system call to fail cleanly with `ENOENT` or auto-create parent directories inside the sysroot.

doanbaotrung:
When absolute paths do not exist in the guest sysroot, translation falls back to the host literal path. For standard system directories (such as /usr, /bin, /lib, /sbin, /var, /opt, and /etc), this fallback leaks guest actions to the host.

On macOS, system directories are protected by System Integrity Protection (SIP) and root ownership, causing permission failures (EPERM/EACCES) when guest tools attempt writes or backups.

Prevent fallback for system paths. Keep fallback for network configurations like /etc/resolv.conf to support DNS resolution.

Fix #238

---
## Summary by cubic
Restricts sysroot fallback so guest paths under system directories never escape to the host, including dot/dot-dot variants. Fixes #238 and avoids SIP permission errors on macOS while keeping DNS working.

- **Bug Fixes**
  - Block fallback for system dirs (/usr, /bin, /sbin, /lib, /lib64, /var, /opt, /boot, /srv, /root, /home) and /etc except /etc/resolv.conf and /etc/hosts; allow macOS /var/folders. Normalize absolute paths to catch cases like "/tmp/../usr".
  - Enforce the system-path check in both resolve and create flows; add tests for blocked /usr and dot-component paths, and allowed /etc/hosts.

<sup>Written for commit b1e4a21c873f5e690855126c779c40a2eb510731. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/241?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/proc-state.c
+++ b/src/syscall/proc-state.c
@@ -24,6 +24,7 @@
 #include "syscall/internal.h"
 #include "syscall/proc.h"
 #include "syscall/proc-state.h"
+#include "syscall/path.h"
 
 /* Shim blob reference (set by startup/bootstrap) */
 static const unsigned char *shim_blob_ptr = NULL;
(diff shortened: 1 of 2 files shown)


src/syscall/proc-state.c
```diff
@@ -553,6 +578,14 @@ static const char *proc_resolve_sysroot_path_flags(const char *path,
         errno = ENAMETOOLONG;
         return NULL;
     }
+
+    /* Prevent escaping guest system paths to macOS host paths, which leads
+     * to host contamination and permission failures (e.g. SIP/EPERM).
+     */
+    if (is_guest_system_path(path)) {
```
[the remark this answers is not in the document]
doanbaotrung: 1. **Lexical path normalization**: Added a helper `lexical_normalize_absolute_path` to resolve dot components syntactically before validating against the guest system path restrictions (for both path flags resolution and parent dir creation).
2. **Added verification tests**: Updated `tests/test-sysroot-host-fallback.c` with new test cases to guarantee that:
   * Blocked guest system paths (`/usr/bin/env`) fail fallback (`ENOENT`).
   * Dot-component bypass paths (`/tmp/../usr/bin/env`) also fail fallback (`ENOENT`).
   * Allowed network configuration paths (`/etc/hosts`) fallback correctly.

src/syscall/proc-state.c
```diff
@@ -521,6 +522,86 @@ static bool sysroot_path_exists(const char *resolved_path, bool follow_final)
...
+    }
+    if (strncmp(path, "/etc/", 5) == 0) {
+        /* Allow fallback for network resolution config files */
+        if (strcmp(path, "/etc/resolv.conf") == 0 ||
+            strcmp(path, "/etc/hosts") == 0)
+            return false;
+        return true;
+    }
+    return false;
+}
```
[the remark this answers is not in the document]
doanbaotrung: I have updated the `is_guest_system_path` helper function on the `system-path-leaks` branch as suggested:

1. **Bare Directory Matching**: Added matching for exact bare system directories (e.g., matching both `/usr/` and `/usr`, `/var/` and `/var`, `/etc/` and `/etc`, etc.).
2. **Proper macOS Exclusions**: Ensured that the macOS host temp path exclusion matches both `/var/folders/` and its bare directory counterpart `/var/folders`.

src/syscall/proc-state.c
```diff
@@ -521,6 +522,94 @@ static bool sysroot_path_exists(const char *resolved_path, bool follow_final)
...
+        if (strncmp(path, "/var/folders/", 13) == 0 ||
+            strcmp(path, "/var/folders") == 0)
+            return false;
+        return true;
+    }
+    if (strncmp(path, "/etc/", 5) == 0 || strcmp(path, "/etc") == 0) {
+        if (strcmp(path, "/etc/resolv.conf") == 0 ||
+            strcmp(path, "/etc/hosts") == 0)
+            return false;
+        return true;
```
jserv: Replace `strcmp(...) == 0` with `!strcmp(...)`.

jserv: Add comment: `/var/folders` (resolving to `/private/var/folders`) is a macOS system directory that securely stores per-user temporary files and caches.

doanbaotrung: Updated

jserv:

PR #242, opened 2026-07-23
Translate guest paths through the sysroot sidecar

### LINKED ISSUE #233: sysroot/sidecar path-translation gaps
## Summary

Two syscall paths under `--sysroot` forget to convert between the name the guest uses and the name stored on disk. 

On a case-insensitive host volume (the default APFS) the case-fold sidecar stores guest-created names as opaque
`.ef_<hex>` tokens, so any path that skips the conversion breaks in one of two directions:
- **Bug 1 (disk -> guest):** after `chdir` into a guest-created directory,   `getcwd()` and `readlink("/proc/self/cwd")` report the on-disk `.ef_<hex>` token path instead of the guest path, in both the calling process and forked children.
- **Bug 2 (guest -> disk):** a relative `access(2)` on a guest-created file returns `ENOENT`, even though `open()` and `stat()` on the same name in the same directory succeed.

Both are 100% reproducible on a case-insensitive sysroot. It was caught during OCI image PR CI runs. 

### Reproducer

```c
/* poc-cwd.c */
#include <fcntl.h>
#include <stdio.h>
#include <string.h>
#include <sys/stat.h>
#include <sys/wait.h>
#include <unistd.h>

#define WORK_DIR   "/exec-cwd-work"
#define NESTED_DIR WORK_DIR "/Nested"

static int check_cwd(const char *ctx)
{
    int bad = 0;
    char cwd[512], proc_cwd[512];

    if (!getcwd(cwd, sizeof cwd)) { perror("getcwd"); return 1; }
    printf("[%s] getcwd()          = %s\n", ctx, cwd);
    if (strcmp(cwd, NESTED_DIR) != 0) { printf("   >> WRONG: expected %s\n", NESTED_DIR); bad = 1; }
    if (strstr(cwd, ".ef_"))          { printf("   >> LEAK: sidecar token in cwd\n"); bad = 1; }

    ssize_t n = readlink("/proc/self/cwd", proc_cwd, sizeof proc_cwd - 1);
    if (n < 0) { perror("readlink /proc/self/cwd"); return 1; }
    proc_cwd[n] = '\0';
    printf("[%s] /proc/self/cwd    = %s\n", ctx, proc_cwd);
    if (strcmp(proc_cwd, NESTED_DIR) != 0) { printf("   >> WRONG: expected %s\n", NESTED_DIR); bad = 1; }
    if (strstr(proc_cwd, ".ef_"))          { printf("   >> LEAK: sidecar token in /proc/self/cwd\n"); bad = 1; }
    return bad;
}

int main(int argc, char **argv)
{
    if (argc > 1 && !strcmp(argv[1], "--check"))   /* forked-child re-check */
        return check_cwd("child");

    if (mkdir(WORK_DIR, 0755) < 0 && access(WORK_DIR, F_OK) != 0) { perror("mkdir work"); return 1; }
    if (mkdir(NESTED_DIR, 0755) < 0 && access(NESTED_DIR, F_OK) != 0) { perror("mkdir nested"); return 1; }
    if (chdir(NESTED_DIR) < 0) { perror("chdir"); return 1; }

    int bad = check_cwd("parent");

    char self[512];
    ssize_t n = readlink("/proc/self/exe", self, sizeof self - 1);
    if (n < 0) { perror("readlink /proc/self/exe"); return 1; }
    self[n] = '\0';

    pid_t pid = fork();
    if (pid < 0) { perror("fork"); return 1; }
    if (pid == 0) {
        char *cargv[] = { self, "--check", NULL };
        execve(self, cargv, NULL);
        perror("execve"); _exit(127);
    }
    int status = 0;
    waitpid(pid, &status, 0);
    if (!WIFEXITED(status) || WEXITSTATUS(status) != 0) {
        printf(">> forked child observed the token leak (child exit=%d)\n",
               WIFEXITED(status) ? WEXITSTATUS(status) : -1);
        bad = 1;
    }
    return bad;
}
```

The guest must create the directories itself: only guest-created names get sidecar tokens, and the token is what leaks. Two nested levels exercise the multi-component case. The `fork` + `execve` re-check matters because `fork` relaunches elfuse, whose lazily recomputed cwd is where the token leaks most visibly.

### Build and run

```sh
aarch64-linux-gnu-gcc -static -O0 poc-cwd.c -o poc-cwd
d=$(mktemp -d); mkdir -p "$d/bin"; cp poc-cwd "$d/bin/poc-cwd"
./build/elfuse --sysroot "$d" "$d/bin/poc-cwd"
ls -a "$d"                      # guest dir is on disk as .ef_<hex>, not "exec-cwd-work"
rm -rf "$d"
```

### Observed (main @ 6cee917)

```
[parent] getcwd()          = /.ef_587aff5e13cf3813/.ef_f2ea99c58566b230
   >> WRONG: expected /exec-cwd-work/Nested
   >> LEAK: sidecar token in cwd
[parent] /proc/self/cwd    = /.ef_587aff5e13cf3813/.ef_f2ea99c58566b230
   >> WRONG: expected /exec-cwd-work/Nested
   >> LEAK: sidecar token in /proc/self/cwd
[child] getcwd()          = /.ef_587aff5e13cf3813/.ef_f2ea99c58566b230
   >> LEAK ...
[child] /proc/self/cwd    = /.ef_587aff5e13cf3813/.ef_f2ea99c58566b230
   >> LEAK ...
>> forked child observed the token leak (child exit=1)
```

`ls -a "$d"` confirms the guest directory really is stored as
`.ef_587aff5e13cf3813` on disk.

### Expected

```
getcwd() = /exec-cwd-work/Nested
readlink("/proc/self/cwd") = /exec-cwd-work/Nested
```

### Why it matters (real workload)

The JVM computes `user.dir` from `getcwd()` at startup; `javac` then resolves source files against it. With `user.dir` set to a `.ef_<hex>` path that does not exist as a guest path, javac fails. Shell builtins mask the bug (`sh`'s `pwd` answers from `$PWD`), so only a fresh process that asks the kernel (`/bin/pwd`, the JVM, or the forked child above) exposes it.

### Reproducer

```c
/* poc-access.c */
#include <fcntl.h>
#include <stdio.h>
#include <string.h>
#include <sys/stat.h>
#include <unistd.h>

#define WORK_DIR "/access-work"

int main(void)
{
    if (mkdir(WORK_DIR, 0755) < 0 && access(WORK_DIR, F_OK) != 0) { perror("mkdir"); return 1; }
    if (chdir(WORK_DIR) < 0) { perror("chdir"); return 1; }

    int fd = open("T.java", O_CREAT | O_WRONLY, 0644);   /* guest creates it -> tokenized */
    if (fd < 0) { perror("open(create)"); return 1; }
    close(fd);

    struct stat st;
    printf("open/stat see T.java: stat=%d (size ok=%d)\n",
           stat("T.java", &st), st.st_mode != 0);

    if (access("T.java", F_OK) != 0) {
        perror("access(\"T.java\", F_OK)");
        printf(">> BUG: access() reports ENOENT for a file open()/stat() just saw\n");
        printf(">> (this is exactly javac's \"error: file not found: T.java\")\n");
        return 1;
    }
    printf("access(\"T.java\", F_OK) = 0 (ok)\n");
    return 0;
}
```

### Build and run

```sh
aarch64-linux-gnu-gcc -static -O0 poc-access.c -o poc-access
d=$(mktemp -d); mkdir -p "$d/bin"; cp poc-access "$d/bin/poc-access"
./build/elfuse --sysroot "$d" "$d/bin/poc-access"
rm -rf "$d"
```

### Observed (main @ 6cee917)

```
open/stat see T.java: stat=0 (size ok=1)
>> BUG: access() reports ENOENT for a file open()/stat() just saw
>> (this is exactly javac's "error: file not found: T.java")
```

### Expected

`access("T.java", F_OK)` should return `0`, matching `open`/`stat` on the same
name.

## Common root cause and status

Both bugs share one shape: the sidecar makes "the name the guest uses" and "the name on disk" different, and any path that forgets to convert between them, in either direction, breaks. Bug 1 forgot disk -> guest (reading cwd back); bug 2 forgot guest -> disk (checking access), or rather threw away a conversion it had already computed.

henrybear327:
Consolidate the sysroot path-translation handlers so guest-visible paths are reconstructed and reverse-mapped consistently across syscalls.

Closes https://github.com/sysprog21/elfuse/issues/233

---
## Summary by cubic
Unifies sysroot path translation through the sidecar so guest paths resolve and reverse‑map consistently across exec, filesystem, inotify, and networking. Normalizes classification once per lookup, tightens `/dev/shm` and interpreter handling, and updates AF_UNIX bind and cleanup behavior.

- **New Features**
  - Centralized resolver in `path.c` with a one‑pass normalizer, a dirfd‑relative fast path, an `AT_EMPTY_PATH` helper, and `/dev/shm` opens forced via `path_translation_oflags` (`O_NOFOLLOW|O_NONBLOCK`); short‑lived descriptors use `O_CLOEXEC`.
  - Exec: interpreter fallback only if the rewritten path differs and exists; `execveat` translates dirfd‑relative names and opens `/dev/shm` leaves with `O_NOFOLLOW`; `getcwd` and `/proc/self/cwd` reverse‑map sidecar tokens after fork/exec.
  - AF_UNIX: pathname translation with over‑length shortening via symlinks in a shared absock dir; bind now matches stat/open (no spurious `EADDRINUSE` on sidecar names); cleanup retires only shortening symlinks so a forked child's sockets keep working; addresses round‑trip across bind/connect/sendmsg/accept/recvmsg.
  - Inotify: snapshots hold the sidecar index open, fail on unreadable indexes (keeping the baseline), emit guest‑visible names, and hide sidecar artifacts; orphan tokens remain listable by their on‑disk spelling.

- **Docs and Tests**
  - Added `docs/sysroot.md` and a full suite: path‑translation matrix (casefold and case‑sensitive lanes), exec‑cwd regression, `/dev/shm` `execveat` no‑follow, absock lifecycle and cleanup, orphan token visibility, unreadable‑index snapshot behavior, plus a sidecar state checker wired into `make check`; removed the `listxattr` indirect‑coverage exemption.

<sup>Written for commit 68886eb650d18930878b58363e72a8cf3257894d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/242?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/internals.md
+++ b/docs/internals.md
@@ -96,7 +96,7 @@ Key files:
 | `src/syscall/mem.c` | `brk`, `mmap`, `mprotect`, `mremap`, `madvise`, `msync` |
 | `src/syscall/fs.c`, `fs-stat.c`, `fs-xattr.c` | filesystem syscalls |
 | `src/syscall/io.c`, `poll.c`, `fd.c`, `fdtable.c` | I/O, polling, FD lifecycle and table |
-| `src/syscall/path.c` | centralized guest-to-host path resolution |
+| `src/syscall/path.c` | centralized guest-to-host path resolution (see also [sysroot.md](sysroot.md)) |
 | `src/syscall/sidecar.c` | case-fold sidecar tokens for case-insensitive macOS volumes |
 | `src/syscall/fuse.c` | guest-internal FUSE transport and minimal VFS |
 | `src/syscall/inotify.c` | inotify via kqueue `EVFILT_VNODE` |
@@ -406,6 +406,8 @@ Socket syscalls are translated in `src/syscall/net.c` and friends:
   type argument; both bits must be extracted before calling `socket()`.
 - `SOL_SOCKET` option numbers (`SO_TYPE`, `SO_SNDBUF`, `SO_RCVBUF`, …)
   differ between platforms and are remapped per option.
+- AF_UNIX pathname and abstract socket addresses are translated through the
+  sysroot; see [sysroot.md](sysroot.md#af_unix-socket-addresses).
 
 ### Stack Alignment
 
@@ -806,10 +808,13 @@ Only a non-empty flat leaf is redirected. Bare `/dev/shm` and `/dev/shm/` stay
 on the sysroot path so the synthetic-directory intercepts keep answering for
 them, and `statfs` on a shm leaf or on `/dev/shm` reports `TMPFS_MAGIC`
 synthetically rather than the host filesystem's type. Because the backing path
-is absolute, two inline helpers in `src/syscall/path.h` adapt the `*at()` calls:
-`path_translation_dirfd()` returns `AT_FDCWD` (POSIX ignores `dirfd` for an
-absolute path), and `path_translation_at_flags()` forces the nofollow flag
-described next.
+is absolute, three inline helpers in `src/syscall/path.h` adapt the calls that
+receive it: `path_translation_dirfd()` returns `AT_FDCWD` (POSIX ignores
+`dirfd` for an absolute path), `path_translation_at_flags()` forces the
+nofollow flag described next onto an `*at()` call, and
+`path_translation_oflags()` forces the equivalent open flags onto an `open()`.
+Keeping each requirement in one helper is what stops a new call site from
+picking up the backing path without also picking up the rules that go with it.
 
 ### The Never-Follow Invariant
 
@@ -825,12 +830,20 @@ is spread across the syscall families, one mechanism each:
 | Operation family | Never-follow mechanism |
 |------------------|------------------------|
 | `*at()` metadata (chmod, chown, stat, utimensat, access) | `path_translation_at_flags()` adds `AT_SYMLINK_NOFOLLOW` |
-| open for truncate/chdir | `shm_open_leaf()` opens `O_NOFOLLOW` |
+| open for truncate/chdir, and `execveat` | `path_translation_oflags()` adds `O_NOFOLLOW` |
 | proc open | `O_NOFOLLOW` |
 | xattr get/set/list/remove | `XATTR_NOFOLLOW` |
 | stat | `lstat`, not `stat` |
 | linkat | clears `AT_SYMLINK_FOLLOW` |
 | statfs | nofollow `lstat` existence probe, then a synthetic reply |
+
+The open flavors carry a second requirement for the same reason. A shm leaf that
+is a FIFO would block the opening thread until a peer arrives, and that thread is
+a vCPU thread, so the whole guest would stall inside one syscall. Every open of a
+shm leaf therefore adds `O_NONBLOCK` alongside `O_NOFOLLOW`, which turns the
+stall into an immediate return: `truncate` reports `EINVAL`, matching Linux
+`truncate(2)` on a FIFO. `path_translation_oflags()` carries both flags together
+so a caller cannot take one without the other.
 
 The name gate lives with the resolver. A POSIX shm name is always a single flat
 component: glibc's `__shm_get_name` (`posix/shm-directory.c`) strips the leading
@@ -841,9 +854,10 @@ rejected name (`ENAMETOOLONG` if the backing path overflows).
 
 Related implementation: `src/runtime/procemu.c` (`dev_shm_resolve_path`),
 `src/syscall/path.c` and `path.h` (`path_translate_at`, `is_dev_shm`,
-`path_translation_dirfd`, `path_translation_at_flags`), and the metadata
-handlers in `src/syscall/fs.c`, `fs-stat.c`, and `fs-xattr.c`. Validation:
-`tests/test-dev-shm-paths.c`.
+`path_translation_dirfd`, `path_translation_at_flags`,
+`path_translation_oflags`), and the metadata handlers in `src/syscall/fs.c`,
+`fs-stat.c`, and `fs-xattr.c`. Validation: `tests/test-dev-shm-paths.c` and
+`tests/test-execveat-shm.c`.
 
 ## Dynamic Linking
 
--- a/src/runtime/procemu.c
+++ b/src/runtime/procemu.c
@@ -63,6 +63,7 @@
 #include "syscall/fuse.h"
 #include "syscall/internal.h"
 #include "syscall/net-identity.h"
+#include "syscall/path.h"
 #include "syscall/proc.h"
 #include "syscall/sys.h"
 
@@ -3783,24 +3784,27 @@ static int proc_readlink_self_exe(char *buf, size_t bufsiz)
     }
     const char *exe = exe_buf;
     char exe_real[LINUX_PATH_MAX];
-    char sysroot_snap[LINUX_PATH_MAX];
-    if (proc_sysroot_snapshot(sysroot_snap, sizeof(sysroot_snap))) {
-        /* proc_set_sysroot stores a realpath()-canonicalized form, so
-         * canonicalize exe before the prefix check or the strip fails when /var
-         * -> /private/var (and similar macOS symlinks) make the two strings
-         * diverge.
-         */
-        const char *exe_cmp = exe;
-        if (realpath(exe, exe_real))
-            exe_cmp = exe_real;
-        size_t sr_len = strlen(sysroot_snap);
-        if (sr_len > 0 && !strncmp(exe_cmp, sysroot_snap, sr_len) &&
-            (exe_cmp[sr_len] == '/' || exe_cmp[sr_len] == '\0')) {
-            exe = exe_cmp + sr_len;
-            if (*exe == '\0')
-                exe = "/";
-        }
-    }
+    char exe_guest[LINUX_PATH_MAX];
+    /* proc_set_sysroot stores a realpath()-canonicalized form, so canonicalize
+     * exe before the reverse map or the sysroot strip fails when /var ->
+     * /private/var (and similar macOS symlinks) make the two strings diverge.
+     * path_host_to_guest also rewrites sidecar token components back to guest
+     * spellings, which a bare prefix strip would leak. Only an actual rewrite
+     * is adopted: an identity result keeps the original spelling, so a
+     * host-literal exe is not silently canonicalized (/tmp -> /private/tmp).
+     *
+     * With no sysroot configured the reverse map is an identity copy, so the
+     * canonicalization would be computed and then discarded; skip it there
+     * rather than pay one lstat per path component on every readlink. Testing
+     * proc_get_sysroot() for NULL without a snapshot is sanctioned (see
+     * proc.h).
+     */
+    const char *exe_cmp = exe;
+    if (proc_get_sysroot() && realpath(exe, exe_real))
+        exe_cmp = exe_real;
+    if (path_host_to_guest(exe_cmp, exe_guest, sizeof(exe_guest)) == 0 &&
+        strcmp(exe_guest, exe_cmp))
+        exe = exe_guest;
     size_t len = strlen(exe);
     if (len > bufsiz)
         len = bufsiz;
@@ -3863,10 +3867,17 @@ int proc_intercept_readlink(const char *path, char *buf, size_t bufsiz)
          */
         if (proc_rosetta_active() && !strcmp(fdpath, ROSETTA_PATH))
             return proc_readlink_self_exe(buf, bufsiz);
-        size_t len = strlen(fdpath);
+        /* F_GETPATH reports the raw host path; the guest must see its own
+         * namespace (sysroot stripped, sidecar tokens reverse-mapped).
+         */
+        char guest_view[LINUX_PATH_MAX];
+        const char *report = fdpath;
+        if (path_host_to_guest(fdpath, guest_view, sizeof(guest_view)) == 0)
+            report = guest_view;
+        size_t len = strlen(report);
         if (len > bufsiz)
             len = bufsiz;
-        memcpy(buf, fdpath, len);
+        memcpy(buf, report, len);
         return (int) len;
     }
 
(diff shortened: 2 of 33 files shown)


docs/internals.md
```diff
@@ -96,7 +96,7 @@ Key files:
 | `src/syscall/mem.c` | `brk`, `mmap`, `mprotect`, `mremap`, `madvise`, `msync` |
 | `src/syscall/fs.c`, `fs-stat.c`, `fs-xattr.c` | filesystem syscalls |
 | `src/syscall/io.c`, `poll.c`, `fd.c`, `fdtable.c` | I/O, polling, FD lifecycle and table |
-| `src/syscall/path.c` | centralized guest-to-host path resolution |
+| `src/syscall/path.c` | centralized guest-to-host path resolution (model: `docs/sysroot.md`) |
```
jserv: Why did you place "model" here?

jserv: Did you mean "See also"?

src/runtime/procemu.c
```diff
@@ -3788,24 +3789,21 @@ static int proc_readlink_self_exe(char *buf, size_t bufsiz)
...
+     * path_host_to_guest also rewrites sidecar token components back to guest
+     * spellings, which a bare prefix strip would leak. Only an actual rewrite
+     * is adopted: an identity result keeps the original spelling, so a
+     * host-literal exe is not silently canonicalized (/tmp -> /private/tmp).
+     */
+    const char *exe_cmp = exe;
+    if (realpath(exe, exe_real))
+        exe_cmp = exe_real;
+    if (path_host_to_guest(exe_cmp, exe_guest, sizeof(exe_guest)) == 0 &&
+        strcmp(exe_guest, exe_cmp) != 0)
```
jserv: Replace `strcmp(...) != 0` with `strcmp(...)`.

docs/internals.md
henrybear327: It's a mistake. I had a huge merge conflict and resolving it went wrong.

henrybear327: It was originally something like "refer to the modelling at ..."

henrybear327:
Documentation rewriting in progress.

src/syscall/proc-state.c
```diff
@@ -514,10 +514,63 @@ static bool sysroot_path_exists(const char *resolved_path, bool follow_final)
...
+ * name /tmp/x) are classified identically. Normalization collapses ".", "..",
+ * and duplicate slashes and yields a leading-slash-free component list ("." for
+ * the root), matched component by component so "/var/tmpfoo" and "/.ccachefoo"
+ * cannot false-positive.
+ */
+static bool comp_eq(const char *comp, size_t len, const char *lit)
+{
+    size_t n = strlen(lit);
+    return len == n && !memcmp(comp, lit, n);
+}
```
jserv: See if this function is worthy to be promoted as one of static inline functions in utility header.

henrybear327: Promoted to `src/syscall/path.h` and also refactored more use cases to use the new wrapper function.

jserv (changes requested):

src/syscall/syscall.c
```diff
@@ -2153,10 +2153,27 @@ static int64_t sc_execveat(guest_t *g,
...
         host_fd_ref_t dir_ref;
-        if (host_fd_ref_open(dirfd, &dir_ref) < 0)
+        if (host_dirfd_ref_open(dirfd, &dir_ref) < 0)
             return -LINUX_EBADF;
-        int tmp_fd = openat(dir_ref.fd, pathname, O_RDONLY);
+        /* O_CLOEXEC: the descriptor is closed a few lines below, but a
+         * concurrent execve on another vCPU thread inside that window would
+         * otherwise leak it into the new image.
+         */
+        int tmp_fd = openat(path_translation_dirfd(&tx, &dir_ref), tx.host_path,
```
jserv, line 2175: execveat resolves the target with `F_GETPATH` here and passes only the host backing path into `sys_execve` (exec.c:433), which rebuilds the guest identity via `path_host_to_guest()`. That function cannot invert the `/dev/shm` per-UID backing dir or a FUSE-materialized copy, so `execveat(dirfd, "/dev/shm/foo", ...)` (or a FUSE-backed binary) makes `/proc/self/exe` report the private host path -- the exact leak this PR exists to prevent. Suggest carrying the guest spelling forward (join `tx.guest_path` against the dirfd/cwd to an absolute guest path) and passing that as the exec identity, so non-invertible legs keep the guest name.

src/syscall/inotify.c
```diff
@@ -638,11 +687,22 @@ int64_t sys_inotify_add_watch(guest_t *g,
...
+    if (tx.fuse_path || tx.proc_resolved != 0 ||
+        path_might_use_open_intercept(tx.intercept_path))
+        return -LINUX_ENOSYS;
+
     /* Open the path for event monitoring. O_EVTONLY is macOS-specific: opens
      * for event notification only, does not prevent unmount or require read
      * access to the file contents.
      */
-    int host_fd = open(path, O_EVTONLY);
+    int host_fd = open(tx.host_path, O_EVTONLY);
```
jserv, line 705: This opens the translated leaf with `open(tx.host_path, O_EVTONLY)` and never applies `path_translation_oflags(&tx, ...)`, so a guest symlink under `/dev/shm` is followed out of the namespace and an unrelated host file gets watched -- breaking the never-follow invariant the rest of the PR enforces. Suggest `open(tx.host_path, path_translation_oflags(&tx, O_EVTONLY))`.

src/syscall/exec.c
```diff
@@ -209,31 +209,67 @@ static int exec_resolve_guest_host_path(const char *guest_path,
...
+ * sidecar-mapped prefix whose loader suffix is absent: without it a
+ * differs-but-missing path would be accepted and the /lib/<basename> fallback
+ * skipped, so a store-style interpreter (e.g. a /nix/.../ld-musl.so.1 that
+ * ships the loader under /lib) would fail to launch.
+ */
+static bool exec_translated_usable(const char *host,
+                                   const char *guest,
+                                   bool temp)
+{
+    return temp || (strcmp(host, guest) && access(host, F_OK) == 0);
```
jserv, line 224: `exec_translated_usable()` probes the interpreter with `access(host, F_OK)`, which follows the final symlink. For a PT_INTERP redirected out of `/dev/shm` this dereferences the symlink and then accepts a leaf that the later `O_NOFOLLOW` open rejects with ELOOP, so the `/lib/<basename>` fallback is skipped and a resolvable interpreter fails to launch. The already-computed `shm_nofollow` is not threaded in. Suggest passing the shm flag through and probing shm paths with `lstat()` or `open(..., O_NOFOLLOW)`.

src/syscall/net-absock.c
```diff
@@ -283,19 +514,53 @@ void absock_bind_rollback(int idx)
...
+    /* The shortening links are untracked, so no table owns them and the
+     * process that minted the namespace retires them on its way out. Sweep
+     * only symlinks: the abstract-socket backing files sharing this directory
+     * belong to whichever process bound them, including children that outlive
+     * the owner, and each unlinks its own above. Unlinking those too would
+     * destroy a socket a live child still has bound. A live child that loses
+     * a shortening link only degrades to the raw host link path, never to a
+     * missing socket. See docs/sysroot.md for the full lifecycle.
+     */
+    if ((uint64_t) getpid() == absock_get_namespace_id()) {
```
jserv, line 538: Cleanup ownership gap: when only a forked child uses AF_UNIX pathname sockets, the child inherits the parent's pid as `namespace_id`, creates the dir plus any over-length shortening symlinks, but its cleanup runs with `getpid() != absock_get_namespace_id()`, so this symlink sweep is skipped. The parent registered no `atexit` because it never created the dir. The symlinks leak in /tmp, and the trailing `rmdir` then fails because the dir is not empty. Suggest letting the process that actually created the directory own the sweep (record the creator pid), or tracking shortening links per-process and unlinking them in every `absock_cleanup`.

src/syscall/net-absock.c
```diff
@@ -196,6 +217,216 @@ static int absock_build_sun(const char *fs_path,
...
+    /* Undo the over-length shortening symlink, then map the host path back
+     * to the guest namespace so the guest reads back the spelling it bound
+     * or connected with, not the sysroot-prefixed (and possibly
+     * token-bearing) host path.
+     */
+    char host_path[LINUX_PATH_MAX];
+    str_copy_trunc(host_path, mac_path, sizeof(host_path));
+    pthread_mutex_lock(&absock_lock);
+    bool in_absock_dir = absock_dir_created &&
+                         !strncmp(host_path, absock_dir, strlen(absock_dir));
```
jserv, line 297: The `in_absock_dir` test does a raw `strncmp(host_path, absock_dir, strlen(absock_dir))` with no component boundary, so a sibling path like `/tmp/elfuse-absock-1234extra/...` is treated as inside the namespace and gets `readlink()`-dereferenced during reverse mapping. Suggest requiring the byte after the prefix to be '\0' or '/', and `lstat()` before `readlink()` so only real shortening symlinks are followed.

jserv (commented):
Review of the sysroot path-translation consolidation. Findings below are inline on the relevant lines. Focus was correctness and the never-follow / no-host-path-leak invariants the PR sets out to enforce.

Out of scope but worth noting: in sys_fchownat the post-chown verification stat uses `fstatat(dir_ref.fd, ...)` while the before-stat and the chown use `host_dirfd` (`path_translation_dirfd`). These only diverge on a `/dev/shm` redirect, where the path is absolute and dirfd is ignored, so it is benign today. This code predates this PR; flagging it because the fakeroot race check would inspect the wrong descriptor if the dirfd helper ever changes.

henrybear327:
Ongoing rewrite of side car in progress. Closing this PR in favor of the rewrite attempt.

PR #245, opened 2026-07-26
Add vcpu run loop hooks

### LINKED ISSUE #244: vcpu_run_loop has no embedder progress callback for UI/event pumps
### Description

`vcpu_run_loop()` currently owns guest execution until the guest exits,
times out, or hits a fatal condition. This works well for normal command-line
guest programs, but it is difficult for embedders that need to service host-side
event sources while the guest remains alive.

For example, an embedder may need to periodically run a host event pump,
flush pending host-side callbacks, or service synthetic devices from the same
thread that owns the vCPU. Since `vcpu_run_loop()` does not expose a progress
or idle callback, the embedder cannot safely do this while the guest is running.

Running the vCPU loop on a worker thread is not a reliable workaround because
some host hypervisor backends require vCPU operations to remain on the thread
that created/owns the vCPU.

### Current Behavior

The public loop shape is effectively:

```c
int vcpu_run_loop(hv_vcpu_t vcpu,
                  hv_vcpu_exit_t *vexit,
                  guest_t *g,
                  bool verbose,
                  int timeout_sec,
                  int *wait_status);
```

Once called, control does not return to the embedder until guest termination or
timeout. The embedder has no opportunity to perform periodic host-side work
between guest exits, syscalls, HVC handling, signal checks, or timer wakeups.

### Expected Behavior

`vcpu_run_loop()` should optionally allow an embedder to run a small callback
from inside the loop on the same host thread as the vCPU.

The callback should be optional and preserve current behavior when unset.

Possible API shape:

```c
typedef bool (*vcpu_embedder_tick_fn)(guest_t *g, void *opaque);

typedef struct vcpu_run_hooks {
    vcpu_embedder_tick_fn tick;
    void *opaque;
} vcpu_run_hooks_t;

int vcpu_run_loop_ex(hv_vcpu_t vcpu,
                     hv_vcpu_exit_t *vexit,
                     guest_t *g,
                     bool verbose,
                     int timeout_sec,
                     int *wait_status,
                     const vcpu_run_hooks_t *hooks);
```

The existing `vcpu_run_loop()` could remain as a compatibility wrapper around
`vcpu_run_loop_ex(..., NULL)`.

### Suggested Semantics

The callback could run at safe progress points such as:

- after handling a guest exit
- after syscall/HVC dispatch completes
- after signal/timer checks
- before sleeping or returning to `hv_vcpu_run()`

The callback should:

- run on the same host thread as `hv_vcpu_run()`
- be optional
- avoid changing behavior for existing callers
- return whether it performed work, if that is useful for scheduling decisions

### Why This Is Needed

Some embedders need to keep host-side state machines alive while guest code is
still running. Without a callback, those embedders must choose between:

- blocking forever inside `vcpu_run_loop()`
- using a timeout and repeatedly re-entering the loop
- moving the vCPU loop to another thread, which can conflict with hypervisor
  thread-affinity constraints
- patching elfuse locally

An optional in-loop callback would provide a cleaner and more general embedding
interface without affecting normal Linux binary execution.

### Compatibility

This can be implemented without changing current users by keeping the existing
`vcpu_run_loop()` signature and adding a new extended API or optional hooks
structure.

doanbaotrung:
Add an optional hook API for embedders that need to run host-side work before re-entering hv_vcpu_run on the vCPU owner thread.

Keep the existing vcpu_run_loop signature as a compatibility wrapper around the extended entry point with no hooks installed.

Add a host-side compile test that pins the public signatures and verifies the hook structure can be initialized. Run it from the normal host test paths used by check and check-sanitizer.

Fix #244

---
## Summary by cubic
Add `vcpu_run_loop_with_hooks` to let embedders run host work before each `hv_vcpu_run` re-entry. Keep `vcpu_run_loop` as a wrapper and add a host compile test.

- **New Features**
  - New `vcpu_run_loop_with_hooks(..., const vcpu_run_hooks_t *hooks)` and `vcpu_run_hooks_t` with an optional `tick`.
  - `tick` runs on the vCPU owner thread before `hv_vcpu_run()`, checked only at host boundaries; nonzero return stops the loop.
  - `vcpu_run_loop` remains as a no-hook wrapper for API compatibility.
  - Added host compile test `test-vcpu-run-hooks-host`; wired into `check`, `check-sanitizer`, and `make test-vcpu-run-hooks-host`.

<sup>Written for commit e19686a88fd1677c8dfc6dd0b7cb11754b4f1dbc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/245?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/proc.c
+++ b/src/syscall/proc.c
@@ -2962,12 +2962,13 @@ static const hv_sys_reg_t hvc4_sysregs[] = {
  * Both modes check proc_exit_group_requested so the main thread also reacts to
  * exit_group called by a worker.
  */
-int vcpu_run_loop(hv_vcpu_t vcpu,
-                  hv_vcpu_exit_t *vexit,
-                  guest_t *g,
-                  bool verbose,
-                  int timeout_sec,
-                  int *wait_status_out)
+int vcpu_run_loop_with_hooks(hv_vcpu_t vcpu,
+                             hv_vcpu_exit_t *vexit,
+                             guest_t *g,
+                             bool verbose,
+                             int timeout_sec,
+                             int *wait_status_out,
+                             const vcpu_run_hooks_t *hooks)
 {
     int exit_code = 0;
     (void) signal_take_termination_wait_status();
@@ -2997,6 +2998,17 @@ int vcpu_run_loop(hv_vcpu_t vcpu,
         if (proc_exit_group_requested()) {
             exit_code = proc_exit_group_code();
             break;
+        }
+        if (hooks && hooks->tick) {
+            int tick_ret = hooks->tick(g, hooks->opaque);
+            if (tick_ret != 0) {
+                exit_code = tick_ret;
+                break;
+            }
+            if (proc_exit_group_requested()) {
+                exit_code = proc_exit_group_code();
+                break;
+            }
         }
 
         if (verbose) {
@@ -4117,4 +4129,15 @@ int vcpu_run_loop(hv_vcpu_t vcpu,
             signal_status != 0 ? signal_status : (exit_code & 0xff) << 8;
     }
     return exit_code;
+}
+
+int vcpu_run_loop(hv_vcpu_t vcpu,
+                  hv_vcpu_exit_t *vexit,
+                  guest_t *g,
+                  bool verbose,
+                  int timeout_sec,
+                  int *wait_status_out)
+{
+    return vcpu_run_loop_with_hooks(vcpu, vexit, g, verbose, timeout_sec,
+                                    wait_status_out, NULL);
 }
--- a/src/syscall/proc.h
+++ b/src/syscall/proc.h
@@ -498,6 +498,34 @@ int proc_exit_group_requested(void);
  */
 void proc_request_hvc6_yield(void);
 
+/* Optional embedder hook. Called on the vCPU owner thread at host boundaries
+ * (before each hv_vcpu_run() entry or re-entry).
+ *
+ * Stop semantics:
+ * Return nonzero to stop the loop. A nonzero tick return is propagated
+ * and returned as the run loop's exit code.
+ *
+ * Cooperative-only timing:
+ * The tick is only checked at host boundaries, not during guest execution.
+ * A guest spinning entirely in EL0 without VM-exiting will never trigger
+ * the tick, meaning a tick-initiated hook cannot interrupt a runaway guest.
+ * To force interruption of a runaway guest, use hv_vcpus_exit().
+ *
+ * Concurrency and lifetime:
+ * When a single vcpu_run_hooks_t and its opaque data are shared across
+ * multiple vCPUs, the tick function will be invoked concurrently from
+ * multiple host threads and must be thread-safe. The run loop does not keep
+ * a copy of the hooks structure and dereferences it every iteration; the
+ * caller must ensure the hooks structure and opaque lifetime extend until
+ * all run loops return.
+ */
+typedef int (*vcpu_run_loop_tick_fn)(guest_t *g, void *opaque);
+
+typedef struct vcpu_run_hooks {
+    vcpu_run_loop_tick_fn tick;
+    void *opaque;
+} vcpu_run_hooks_t;
+
 /* Run the vCPU execution loop.
  *
  * Returns the exit code.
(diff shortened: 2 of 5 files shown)


src/syscall/proc.c
```diff
@@ -2998,6 +2999,8 @@ int vcpu_run_loop(hv_vcpu_t vcpu,
             exit_code = proc_exit_group_code();
             break;
         }
+        if (hooks && hooks->tick && hooks->tick(g, hooks->opaque))
```
jserv: A nonzero tick() return breaks the loop while exit_code is still 0 (it is only assigned on the HVC #0 / crash / exit_group paths). A hook that stops the loop for a real reason is therefore indistinguishable from a clean guest exit 0, and wait_status_out also comes back 0. There is no channel for the hook to communicate a result. Either propagate the tick's return as the exit code, or document that a hook-initiated stop always yields exit code 0.

Separately: proc_exit_group_requested() is checked just above the tick but not re-checked after it. If a hook drives shutdown yet returns 0, the loop re-enters hv_vcpu_run() once more.

src/syscall/proc.h
```diff
@@ -498,6 +498,16 @@ int proc_exit_group_requested(void);
  */
 void proc_request_hvc6_yield(void);
 
+/* Optional embedder hook. Called on the vCPU owner thread before each
+ * hv_vcpu_run() re-entry. Return nonzero to stop the loop.
+ */
+typedef int (*vcpu_run_loop_tick_fn)(guest_t *g, void *opaque);
```
jserv, line 522: The hook contract is underspecified in three ways:

- Stop semantics: state what the function returns when tick returns nonzero (currently exit code 0, indistinguishable from guest exit 0).
- Concurrency and lifetime: when one vcpu_run_hooks_t/opaque is shared across worker vCPUs, tick is called concurrently from multiple threads, so it must be thread-safe. The implementation keeps no copy and dereferences hooks/opaque every iteration, so the caller must keep both alive until the function returns.
- Cooperative-only timing: the tick fires only at the top of the loop (host round-trips), not after hv_vcpu_run() returns. A worker (timeout_sec == 0, no alarm()) spinning entirely in EL0 never reaches the top, so a hook cannot interrupt a runaway guest -- only exit_group's hv_vcpus_exit() kick can. "before each hv_vcpu_run() re-entry" reads as periodic; clarify it is checked at host boundaries only.

tests/test-vcpu-run-hooks-host.c
```diff
@@ -0,0 +1,55 @@
...
+}
+
+int main(void)
+{
+    vcpu_run_hooks_t hooks = {
+        .tick = tick,
+        .opaque = (void *) 0x1,
+    };
+
+    if (hooks.tick == NULL || hooks.opaque == NULL) {
```
jserv, line 49: The runtime half of this test is tautological: given the initializer, hooks.tick/hooks.opaque are never NULL, so the if can't fire and main always reaches PASS; the tick function is defined but never called. The real value is the two _Static_assert's, which fire at compile time -- building the object is what matters, not running the binary, yet mk/tests.mk prints a "unit test" banner and executes it, implying runtime validation that does not exist. The new executable behavior (invoke tick, break on nonzero) has no coverage since no in-tree caller passes non-NULL hooks. Either make it a compile-only target, or have main actually call tick(g, opaque) and assert on both the non-NULL and NULL-opaque short-circuit results.

jserv (commented):
Review of the vCPU run-loop hook API. Three points below, anchored inline.

doanbaotrung:
Here is a summary of my update:
1. **Loop Exit & Shutdown Logic**: Modified the run loop in `proc.c` to propagate non-zero `tick` hook return values as the run loop's `exit_code`, and re-check for group exit requests immediately following a `tick` execution.
2. **Hook Contract Documentation**: Documented the stop semantics, concurrency/lifetime rules, and cooperative-only timing details in the comments of `proc.c`.
3. **Unit Tests**: Updated `test-vcpu-run-hooks-host.c` to verify the hook's execution and assert behavior under different combinations of `NULL`/non-`NULL` guest and opaque arguments.
4. **Verification**: Ran all host unit tests and verified that they compile and execute successfully.

jserv:

PR #254, opened 2026-07-31
syscall sysroot-prefix symlink targets

### LINKED ISSUE #253: proc_resolve_sysroot_path_flags() silently returns an unprefixed guest path
# `proc_resolve_sysroot_path_flags()` silently returns an unprefixed guest path when a symlink's target can't be verified to exist, corrupting `stat()`-family syscalls outside the small hardcoded allowlist

## Environment
- Host: macOS 26.5.2, Darwin Kernel Version 25.5.0 (arm64, Apple Silicon)
- Guest: Ubuntu 26.04 (aarch64) rootfs, `apt-get install` with a multi-package transaction

## Summary

`stat()`/`fstatat()` (follow-symlinks mode) on a valid, existing symlink inside the guest rootfs can return `ENOENT`, even though the symlink is genuinely present on the host filesystem at the correctly-translated sysroot path. This happens specifically for symlinks whose **target** is itself an untranslated guest-absolute path (i.e. the normal, correct way symlinks are stored), located under a guest directory that isn't on a small hardcoded allowlist of "always sysroot-prefix regardless" paths (`/usr/`, `/var/`, `/etc/`, `/bin/`, `/sbin/`, `/lib/`, `/lib64/`, `/opt/`, `/boot/`, `/srv/`, `/root/`, `/home/`). **`/tmp/` is not on that list**, which is what surfaced this.

The user-visible symptom: real-world package managers (apt/dpkg observed directly) that stage multiple files via symlinks into a scratch directory before processing them in a batch fail unpredictably with "file not found" errors on files that are demonstrably present, because *some* of the batch's `stat()` calls get silently misdirected to the real host path instead of the sysroot path.

## Root cause

`src/syscall/proc-state.c`, `proc_resolve_sysroot_path_flags()`:

```c
static const char *proc_resolve_sysroot_path_flags(const char *path,
                                                   char *buf,
                                                   size_t bufsz,
                                                   bool follow_final)
{
    char sr[LINUX_PATH_MAX];
    if (!proc_sysroot_snapshot(sr, sizeof(sr)) || !path || path[0] != '/')
        return path;
    if (bufsz == 0) {
        errno = ENAMETOOLONG;
        return NULL;
    }

    int n = snprintf(buf, bufsz, "%s%s", sr, path);
    ...
    bool full_path_truncated = (size_t) n >= bufsz;
    if (!full_path_truncated && sysroot_path_exists(buf, follow_final)) {
        if (!sysroot_path_is_contained(buf, sr, follow_final)) {
            errno = ELOOP;
            return NULL;
        }
        return buf;   // correct: sysroot-prefixed path
    }

    if (full_path_truncated) {
        errno = ENAMETOOLONG;
        return NULL;
    }

    /* Prevent escaping guest system paths to macOS host paths, which leads
     * to host contamination and permission failures (e.g. SIP/EPERM).
     */
    char norm_path[LINUX_PATH_MAX];
    bool has_norm =
        lexical_normalize_absolute_path(norm_path, path, sizeof(norm_path));
    if (is_guest_system_path(has_norm ? norm_path : path))
        return buf;

    return path;   // <-- returns the RAW GUEST PATH, unprefixed
}
```

Walkthrough of the failure, confirmed via a temporary instrumented build:

1. `buf` is correctly built as `sysroot + path` (e.g. `<sysroot>/tmp/apt-dpkg-install-XXXXX/foo.deb`).
2. `sysroot_path_exists(buf, follow_final=true)` is called with `follow_final=true` for a normal (non-`AT_SYMLINK_NOFOLLOW`) `stat()`. This follows the symlink to verify the **target** exists, not just the symlink file itself.
3. The target of such a staging symlink is (correctly) stored as a raw, untranslated guest-absolute path — e.g. `/var/cache/apt/archives/foo.deb` — because `sys_symlinkat()` stores exactly what the guest wrote (`src/syscall/fs.c:sys_symlinkat`), which is the right behavior: guest-side `readlink()` and guest-side re-resolution both need to see the guest's own view of the target.
4. `sysroot_path_exists()`'s follow-symlink existence check, however, resolves that target **natively as a real host path** — since it wasn't translated — and finds nothing at `/var/cache/apt/archives/foo.deb` on the real host. The existence check spuriously fails, even though the symlink and its (correctly, guest-relatively resolvable) target both genuinely exist under the sysroot.
5. Falling through, `is_guest_system_path()` is checked as a fallback — but it only recognizes a short hardcoded list of top-level directories, and `/tmp` isn't among them.
6. The function returns the **raw, unprefixed guest path** as if it were already a valid host path.
7. The caller (`stat_at_path()` in `src/syscall/fs-stat.c`) then calls the real `fstatat()` on that raw path, which resolves against the real host filesystem root — not the sysroot — and returns `ENOENT`.

Confirmed via a live instrumented trace on the exact same symlink: the `AT_SYMLINK_NOFOLLOW` variant (`follow_final=false`, used for e.g. `lstat()`) correctly resolves and returns the sysroot-prefixed path and the subsequent `fstatat()` succeeds; the follow-symlinks variant (`follow_final=true`, used for plain `stat()`) returns the bare unprefixed guest path and the subsequent `fstatat()` fails with `ENOENT` on the identical symlink, moments apart.

## Reproduction

1. Set up any Debian/Ubuntu aarch64 rootfs under elfuse with a populated apt cache (`/var/cache/apt/archives/`).
2. Run `apt-get install -y <multiple packages>` for a package set with several dependencies (a single-package install does not reliably trigger it; a batch of ~4+ packages usually does, though which specific package in the batch fails is not consistent across otherwise-identical runs against the same starting state).
3. Observe `dpkg: error: cannot stat pathname '/tmp/apt-dpkg-install-XXXXXX/NN-<package>.deb': No such file or directory` even though:
   - the referenced `.deb` is present and complete in `/var/cache/apt/archives/`,
   - the symlink at that exact path is confirmed to exist on the real host disk (verified via a plain host-side `ls -la` on the underlying sysroot directory, independent of elfuse) at the moment of failure.

A minimal, generic repro outside of apt: from inside the guest, `symlink("/some/absolute/guest/path", "/tmp/mylink")` where the target does not happen to also exist verbatim on the real host filesystem, then `stat("/tmp/mylink", &st)` (not `lstat`) — expect `ENOENT` despite the symlink existing.

**Note on nondeterminism**: across otherwise-identical apt transactions (same package set, same starting `dpkg` state, cache fully warm), *which* file in the batch trips this varies between runs, and single-package installs of the same packages individually always succeed. We haven't fully characterized why some symlinks in a batch hit the existence check's follow-symlink path and others don't — it may relate to `stat`-cache warmth from an earlier pass (e.g. apt's own pre-flight listing), but this is unconfirmed.

## Suggested direction (not implemented — flagging for the maintainer)

The existence check conflates two different questions: "is this path, once translated, safe/contained under the sysroot" (a real security concern, per the function's own comment about SIP/EPERM host contamination) vs. "does the fully-resolved target exist" (which requires guest-aware target resolution, not a native host follow). A `stat()`-style existence probe with `follow_final=true` on a symlink needs to resolve the symlink's target through the same guest→host translation as everything else, not through the host kernel's native symlink following — otherwise any symlink whose target is a valid guest path but not also a coincidentally-identical real host path hits this fallback. The `is_guest_system_path()` allowlist papering over this for `/usr`, `/var`, `/etc`, etc. suggests this has been hit before in some form; extending the allowlist to include `/tmp` would mask this specific symptom without addressing the underlying conflation for other paths not on the list.

doanbaotrung:
sys_symlinkat() stored an absolute guest target verbatim, so any later native follow (open() stat() etc.) resolved it against the real host root instead of --sysroot. Guest tools that symlink into scratch paths with absolute targets -- e.g. dpkg staging a .deb via a symlink into /var/cache/apt/archives -- hit a spurious ENOENT since the host kernel's own symlink resolution never passes through elfuse's path translation.

Store <sysroot><target> on disk for absolute targets instead so native follows land inside the sysroot like every other absolute host path access already does. sys_readlinkat() strips that same prefix back off before returning the target to the guest so readlink() semantics are unchanged.

Relative targets and symlinks not created through sys_symlinkat() (pre-staged fixtures symlinks predating --sysroot) are untouched.

Fix #253

---
## Summary by cubic
Rewrites absolute symlink targets created under --sysroot to a relative path from the link’s directory so native follows stay inside the sysroot and the tree remains relocatable. Fixes ENOENTs from native follows of absolute links (e.g., dpkg staging into /var/cache/apt/archives).

- **Bug Fixes**
  - In `sys_symlinkat()`: when the new link is inside the sysroot and the target is absolute, normalize “..” (clamped at “/”) and store a relative target from the link’s directory so native follows can’t escape.
  - `readlink()` unchanged; the guest sees the relative form. Relative targets unchanged. Links created outside the sysroot and pre-staged symlinks are left as-is. If a rewrite is required but can’t be formed (e.g., too long), we fail with `errno` rather than writing an escaping target.

<sup>Written for commit 78c69be76e3ed817328788a2f56a1b46a3590f06. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/254?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/fs.c
+++ b/src/syscall/fs.c
@@ -2209,6 +2209,113 @@ int64_t sys_mknodat(guest_t *g, int dirfd, uint64_t path_gva, int mode, int dev)
     return -LINUX_ENOSYS;
 }
 
+typedef enum {
+    SYMLINK_TARGET_VERBATIM,  /* store the guest's target unchanged */
+    SYMLINK_TARGET_REWRITTEN, /* store the composed relative path */
+    SYMLINK_TARGET_FAILED,    /* must rewrite but cannot; errno is set */
+} symlink_target_rewrite_t;
+
+/* Absolute host path of the new symlink itself.
+ *
+ * path_translate_at() only rewrites absolute guest paths; a relative linkpath
+ * is handed back untouched because the resolvers have no dirfd context (see
+ * the comment in path_translate_at). Measuring depth against that bare name
+ * would silently skip the rewrite and leave an escaping absolute target on
+ * disk, so rebuild the location from the dirfd first.
+ */
+static bool symlink_host_location(int dirfd,
+                                  const host_fd_ref_t *dir_ref,
+                                  const char *host_path,
+                                  char *out,
+                                  size_t outsz)
+{
+    if (host_path[0] == '/') {
+        if (str_copy_trunc(out, host_path, outsz) >= outsz) {
+            errno = ENAMETOOLONG;
+            return false;
+        }
+        return true;
+    }
+
+    char base[LINUX_PATH_MAX];
+    if (dirfd == LINUX_AT_FDCWD) {
+        if (!getcwd(base, sizeof(base)))
+            return false;
+    } else if (dir_ref->fd < 0 || fcntl(dir_ref->fd, F_GETPATH, base) < 0) {
+        return false;
+    }
+
+    int n = snprintf(out, outsz, "%s/%s", base, host_path);
+    if (n < 0 || (size_t) n >= outsz) {
+        errno = ENAMETOOLONG;
+        return false;
+    }
+    return true;
+}
+
+/* Express absolute guest path @target as a path relative to the directory that
+ * holds the new symlink, so a native follow stays inside the sysroot. ".." is
+ * normalized away first, clamped at "/" the way a chroot resolves it, so the
+ * result can never climb above the sysroot.
+ *
+ * VERBATIM covers the cases where no rewrite is owed: a relative target, no
+ * sysroot configured, or a symlink created outside the sysroot, where the
+ * guest is deliberately working against host paths. Anything that must be
+ * rewritten but cannot is FAILED rather than VERBATIM -- falling back to the
+ * literal target there would hand back the very escape this closes.
+ */
+static symlink_target_rewrite_t symlink_rewrite_target(const char *target,
+                                                       int dirfd,
+                                                       const host_fd_ref_t *ref,
+                                                       const char *host_path,
+                                                       char *out,
+                                                       size_t outsz)
+{
+    char sr[LINUX_PATH_MAX], norm[LINUX_PATH_MAX], link_host[LINUX_PATH_MAX];
+    if (target[0] != '/' || !proc_sysroot_snapshot(sr, sizeof(sr)))
+        return SYMLINK_TARGET_VERBATIM;
+
+    if (!symlink_host_location(dirfd, ref, host_path, link_host,
+                               sizeof(link_host)))
+        return SYMLINK_TARGET_FAILED;
+
+    size_t sr_len = strlen(sr);
+    if (strncmp(link_host, sr, sr_len) != 0 || link_host[sr_len] != '/')
+        return SYMLINK_TARGET_VERBATIM;
+
+    if (path_openat2_normalize_in_root(target, norm, sizeof(norm)) != 0)
+        return SYMLINK_TARGET_FAILED;
+
+    /* One ".." per directory between the sysroot and the symlink itself.
+     * Counting components rather than separators keeps repeated or trailing
+     * slashes ("/a//b/link") from inflating the depth and walking out above
+     * the sysroot -- host_path is a plain concatenation with no normalization.
+     */
+    size_t components = 0;
+    for (const char *p = link_host + sr_len; *p;) {
+        while (*p == '/')
+            p++;
+        if (!*p)
+            break;
+        components++;
+        while (*p && *p != '/')
+            p++;
+    }
+    size_t depth = components ? components - 1 : 0;
+
+    size_t norm_len = strlen(norm);
+    if (depth * 3 + norm_len + 1 > outsz) {
+        errno = ENAMETOOLONG;
+        return SYMLINK_TARGET_FAILED;
+    }
+
+    char *w = out;
+    for (size_t i = 0; i < depth; i++, w += 3)
+        memcpy(w, "../", 3);
+    memcpy(w, norm, norm_len + 1);
+    return SYMLINK_TARGET_REWRITTEN;
+}
+
 int64_t sys_symlinkat(guest_t *g,
                       uint64_t target_gva,
                       int dirfd,
@@ -2231,9 +2338,36 @@ int64_t sys_symlinkat(guest_t *g,
     if (host_dirfd_ref_open(dirfd, &dir_ref) < 0)
         return -LINUX_EBADF;
 
+    /* An absolute target is stored verbatim, but the host kernel resolves it
+     * against the real host root -- not --sysroot -- whenever anything follows
+     * the symlink natively, so the follow escapes the guest tree. Store it
+     * relative to the symlink's own directory instead: it stays inside the
+     * sysroot and survives the tree being moved, where "<sysroot><target>"
+     * would strand every such link.
+     *
+     * Two divergences the guest can see: readlink() reports the rewritten
+     * form, since nothing on disk tells a rewritten target from a relative one
+     * the guest wrote; and an absolute target loses the host fallback a direct
+     * open() would get, the choice being made once at creation rather than at
+     * follow time.
+     */
+    const char *host_target = target;
+    char rel_target[LINUX_PATH_MAX];
+    switch (symlink_rewrite_target(target, dirfd, &dir_ref, tx.host_path,
+                                   rel_target, sizeof(rel_target))) {
+    case SYMLINK_TARGET_REWRITTEN:
+        host_target = rel_target;
+        break;
+    case SYMLINK_TARGET_FAILED:
+        host_fd_ref_close(&dir_ref);
+        return linux_errno();
+    case SYMLINK_TARGET_VERBATIM:
+        break;
+    }
+
     /* Resolve linkpath (the new symlink location) through sysroot */
-    if (symlinkat(target, path_translation_dirfd(&tx, &dir_ref), tx.host_path) <
-        0) {
+    if (symlinkat(host_target, path_translation_dirfd(&tx, &dir_ref),
+                  tx.host_path) < 0) {
         host_fd_ref_close(&dir_ref);
         return linux_errno();
     }


src/syscall/fs.c
```diff
@@ -2231,9 +2231,54 @@ int64_t sys_symlinkat(guest_t *g,
...
+     * "<sysroot><target>" instead would stamp absolute host paths across the
+     * guest filesystem and strand every one of them as soon as the tree is
+     * moved or copied. ".." is normalized away first, clamped at "/" the way
+     * a chroot resolves it, so the result can never climb above the sysroot.
+     *
+     * readlink() is deliberately left alone. The guest observes the relative
+     * form, which resolves to the same file from inside the guest, and no
+     * reverse mapping is needed -- there is no way to tell a rewritten target
+     * from one a guest wrote itself once it is on disk.
+     */
```
jserv: Shorten the comments slightly.

src/syscall/fs.c
```diff
@@ -2231,9 +2231,54 @@ int64_t sys_symlinkat(guest_t *g,
...
+    const char *host_target = target;
+    char rel_target[LINUX_PATH_MAX];
+    char sr[LINUX_PATH_MAX], norm[LINUX_PATH_MAX];
+    if (target[0] == '/' && proc_sysroot_snapshot(sr, sizeof(sr)) &&
+        path_openat2_normalize_in_root(target, norm, sizeof(norm)) == 0) {
+        size_t sr_len = strlen(sr);
+        /* Only when the new symlink itself lands inside the sysroot. A
+         * host-fallback location has no sysroot-relative depth to walk up.
+         */
+        if (!strncmp(tx.host_path, sr, sr_len) && tx.host_path[sr_len] == '/') {
```
jserv: Replace with early break.

doanbaotrung:
Updated base on comments

src/syscall/fs.c
```diff
@@ -2231,9 +2277,25 @@ int64_t sys_symlinkat(guest_t *g,
...
+     * sees a spurious ENOENT. Store it relative to the symlink's own directory
+     * instead, which walks back down inside the sysroot and keeps the tree
+     * relocatable; baking in "<sysroot><target>" would strand every such link
+     * the moment the tree moves. readlink() is left alone: the guest sees the
+     * relative form, which resolves to the same file, and a rewritten target is
+     * indistinguishable from one the guest wrote itself.
+     */
+    const char *host_target = target;
+    char rel_target[LINUX_PATH_MAX];
+    if (symlink_target_relative_to_sysroot(target, tx.host_path, rel_target,
```
jserv: Relative linkpath keeps the escape this PR sets out to close. The rewrite only fires when tx.host_path is sysroot-prefixed, but for a relative linkpath path_resolve_sysroot_create_path hands back the bare relative name (see the comment at path.c:244-256), so the strncmp(host_path, sr, ...) check fails, the helper returns false, and the absolute target is stored verbatim. symlinkat("/bin/sh", dirfd_in_sysroot, "x") still writes /bin/sh and a native follow escapes to the host root. Reconstruct the absolute host location for relative linkpaths before calling the helper.

src/syscall/fs.c
```diff
@@ -2209,6 +2209,52 @@ int64_t sys_mknodat(guest_t *g, int dirfd, uint64_t path_gva, int mode, int dev)
...
+    size_t sr_len = strlen(sr);
+    if (strncmp(host_path, sr, sr_len) != 0 || host_path[sr_len] != '/')
+        return false;
+
+    /* One ".." per directory between the sysroot and the symlink itself. The
+     * trailing-slash guard keeps a "/dir/" spelling from counting a level that
+     * is not there.
+     */
+    size_t depth = 0;
+    for (const char *p = host_path + sr_len + 1; *p; p++) {
```
jserv: Depth overcounts on consecutive slashes. host_path is snprintf("%s%s", sr, guest_linkpath) with no slash normalization, so a linkpath like /a//b/link arrives with // embedded. This loop counts a level per slash, so // inflates depth, emits an extra ../, and the follow climbs above the sysroot. Skip runs of slashes: if (*p == '/' && p[1] != '\0' && p[1] != '/').

src/syscall/fs.c
```diff
@@ -2209,6 +2209,52 @@ int64_t sys_mknodat(guest_t *g, int dirfd, uint64_t path_gva, int mode, int dev)
...
+        if (*p == '/' && p[1] != '\0')
+            depth++;
+    }
+
+    size_t norm_len = strlen(norm);
+    if (depth * 3 + norm_len + 1 > outsz)
```
jserv: Overflow silently reopens the escape. When the composed path does not fit, the helper returns false and the caller stores the absolute target verbatim, turning a bounds failure back into the sysroot escape. An absolute target that must be rewritten but cannot fit should fail with -ENAMETOOLONG, not fall through. That needs a tri-state result (no-rewrite / rewritten / hard-fail) rather than a bool.

src/syscall/fs.c
```diff
@@ -2231,9 +2277,25 @@ int64_t sys_symlinkat(guest_t *g,
     if (host_dirfd_ref_open(dirfd, &dir_ref) < 0)
         return -LINUX_EBADF;
 
+    /* An absolute target is stored verbatim, but the host kernel resolves it
+     * against the real host root -- not --sysroot -- whenever anything follows
+     * the symlink natively, so the follow escapes the guest tree and the caller
+     * sees a spurious ENOENT. Store it relative to the symlink's own directory
+     * instead, which walks back down inside the sysroot and keeps the tree
+     * relocatable; baking in "<sysroot><target>" would strand every such link
+     * the moment the tree moves. readlink() is left alone: the guest sees the
```
jserv: The "indistinguishable from one the guest wrote itself" claim does not hold. symlinkat("/a/b", ..., "l"); readlink("l") now returns ../a/b, not the bytes the guest supplied, so a guest that reads back its own symlink sees a different target. Either translate the relative form back to the original absolute target in sys_readlinkat (fs.c:1901), or drop this claim and document/test the intentional divergence.

src/syscall/fs.c
```diff
@@ -2231,9 +2277,25 @@ int64_t sys_symlinkat(guest_t *g,
     if (host_dirfd_ref_open(dirfd, &dir_ref) < 0)
         return -LINUX_EBADF;
 
+    /* An absolute target is stored verbatim, but the host kernel resolves it
```
jserv, line 2341: The rewrite discards host-fallback for symlink targets. A direct guest open("/etc/foo") can fall back to the host literal when the path is absent in the sysroot (proc_resolve_sysroot_create_path returns the raw path for non-/tmp, non-system paths), but a symlink to the same absolute path now always resolves inside the sysroot and never gets that fallback. Confirm this asymmetry is intended.

jserv (commented):
Review notes on the absolute-symlink rewrite. Five inline points below. Separately: only fs.c changed, so there is no regression test. Worth covering absolute target + absolute linkpath, absolute target + relative linkpath (inline 1), a linkpath with consecutive slashes (inline 2), a too-long rewrite (inline 3), and a readlink round-trip (inline 4).

doanbaotrung:
1. Relative linkpath escape — **confirmed and fixed**. symlinkat("/var/cache/target", dirfd, "rel-link") really did write /var/cache/target verbatim. Added symlink_host_location(), which rebuilds the absolute host path from the dirfd (F_GETPATH, or getcwd() for AT_FDCWD) before depth is measured. Now writes ../var/cache/target.

2. Depth overcount — **confirmed and fixed**. /d1//d2/link produced ../../../ for a depth-2 link. I did not use the suggested p[1] != '/' guard: it fixes interior runs but still overcounts a leading // (<sr>//a/link). Counting components instead is correct for every slash arrangement, including trailing ones. Now ../../.

3. Overflow reopening the escape — **fixed**. Replaced the bool with a tri-state (VERBATIM / REWRITTEN / FAILED); overflow sets ENAMETOOLONG and the caller returns linux_errno(). Verified: a 4091-char target gives errno 36 and creates no link. symlink_host_location() failures fail closed the same way.

4. Overclaiming comment — **dropped**. You'd already chosen the relative-target trade-off deliberately, so I documented the divergence honestly rather than translating back (which is impossible from on-disk state — a rewritten target is indistinguishable from a guest-written relative one, so undoing it would corrupt the latter).

5. Host-fallback asymmetry — **intended, now documented**. A symlink is persistent and followed later by the host kernel with no translation, so the decision must be made once at creation, before it's knowable whether the target will exist by follow time — dpkg routinely links to files it unpacks later. Per-call fallback can't be replicated; resolving inside the sysroot is the deterministic choice.

jserv:

PR #255, opened 2026-07-31
Resolve /tmp through the sysroot on lookups

henrybear327:
Under `--sysroot`, guest paths in `/tmp`, `/var/tmp`, and `~/.ccache` are
redirected into the sysroot when a file is created there, so guest temp files
avoid collisions on the host's case-insensitive `/tmp`. Lookups were not
redirected: a path absent from the sysroot falls back to the host, so a `/tmp`
entry that exists on the host but not the sysroot was visible to `stat` and
`open` yet missing to `unlink`, `rmdir`, and `rename`, which translate through
the create resolver. The result is an entry the guest can see and read but
cannot remove. This is long-standing, not a recent regression.

### Reproduction steps

The issue was discovered with setting up CI pipeline for the `oci/setup` branch. Across runs,
an attempt was made to `rm` the old tmp files from other tests, and we ran into errors.

Create `/tmp/leftover/report.txt` on the host, and make sure that the sysroot does not have it. 
Inside the guest: `rm` lists the file, then fails to delete it despite `-f`, because the delete
resolves into the sysroot where it does not exist.

### Change

Redirect `/tmp`, `/var/tmp`, and ccache in the lookup resolver as well, so the
guest's `/tmp` is backed by the sysroot for every operation, matching the
create side. A host entry the guest never created there is no longer visible;
entries the guest does create stay readable and removable. Host fallback is
unchanged for all other paths.

This does change reads: such host `/tmp` entries go from visible-but-unremovable
to invisible, which is the behavior the create-side redirect already implied.

---
## Summary by cubic
Fixes inconsistent temp path handling under `--sysroot` by resolving lookups for `/tmp`, `/var/tmp`, and any `.ccache` directory through the sysroot and excluding them (and guest system dirs) from host fallback. Lookup, listings, and removal now agree, preventing `rm -rf` failures.

- **Bug Fixes**
  - Resolve lookups for temp roots in the sysroot (including the roots themselves), matching create-time redirects; `/tmp` listings no longer show host entries; guest system dirs stay sysroot-only; host fallback unchanged elsewhere.
  - Add `test-sysroot-tmp-remove` to `make check` to verify unlink/rmdir/rename, listing consistency, and a guest-created file round-trip; the harness confirms host `/tmp` is untouched and writes land in the sysroot, and moves the TOCTOU host-escape assertion out of the guest test into the harness.

- **Refactors**
  - Share one predicate for temp roots across resolvers (`is_sysroot_backed_temp_path`), and export `path_prefix_match` for reuse.

<sup>Written for commit 24581394d2f660c8fb83df6741d1aec33cf7d4e7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/255?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -17,7 +17,7 @@ Supported user-facing options:
 | `-V`, `--version` | Print the build version and exit |
 | `-v`, `--verbose` | Enable syscall-level and loader diagnostics |
 | `-t`, `--timeout N` | Per-iteration vCPU watchdog, in seconds (default `10`, `0` disables) |
-| `--sysroot PATH` | Resolve guest absolute paths under `PATH` first |
+| `--sysroot PATH` | Resolve guest absolute paths under `PATH`, falling back to the host for paths it does not hold |
 | `--create-sysroot PATH` | Provision a case-sensitive APFS sparsebundle mounted at `PATH`, then use it as the sysroot |
 | `--no-rosetta` | Disable the x86_64-via-Rosetta translator (also `ELFUSE_NO_ROSETTA=1`) |
 | `--gdb PORT` | Listen for a GDB RSP client on `PORT` (aarch64 guests only) |
@@ -56,7 +56,9 @@ anything else that inspects `$?`.
 ### Worked Examples
 
 The guest reads and writes the host filesystem directly (no overlay,
-no volume mount), so file arguments are just file arguments.
+no volume mount), so file arguments are just file arguments. Under
+`--sysroot` the temporary directories are the exception; see
+[Dynamic Linking And Sysroots](#dynamic-linking-and-sysroots).
 
 Run a Linux static `jq` against a host JSON file:
 
@@ -138,8 +140,8 @@ the SHA-256-keyed translations.
 
 Dynamic Linux guests need a sysroot that contains the expected interpreter and
 shared libraries. `elfuse` reads `PT_INTERP`, loads the requested interpreter
-from the supplied sysroot, and redirects guest absolute-path opens to that tree
-before falling back to the host filesystem.
+from the supplied sysroot, and redirects guest absolute-path opens to that tree,
+falling back to the host filesystem for paths the sysroot does not hold.
 
 Example:
 
@@ -155,6 +157,13 @@ Practical notes:
 
 - The sysroot is consulted only for guest absolute paths; relative paths still
   resolve from the guest working directory.
+- `/tmp`, `/var/tmp`, and any `.ccache` directory are backed by the sysroot
+  alone. A guest's temporary files go there so they cannot collide on a
+  case-insensitive host `/tmp`, and every operation on those paths, including
+  `stat`, `open`, and directory listings, addresses the same place. Host files
+  under those directories are therefore invisible to the guest, and a guest
+  program or interpreter stored there cannot be loaded; pass one from anywhere
+  else.
 - The sysroot setting is preserved across guest `fork` and `execve`, so spawned
   children see the same view of the filesystem.
 - On case-insensitive macOS volumes, `elfuse` maintains per-directory
--- a/src/syscall/proc-state.c
+++ b/src/syscall/proc-state.c
@@ -518,8 +518,11 @@ static bool sysroot_path_exists(const char *resolved_path, bool follow_final)
 /* Resolve an absolute guest path against --sysroot. This keeps absolute guest
  * filesystem syscalls inside the sysroot when the target exists there, and
  * otherwise falls back to the literal host path so apps can still reach host
- * resources such as /tmp or /etc/resolv.conf. Containment via realpath() is
- * enforced only when the path actually resolves under sysroot, to prevent
+ * resources such as /etc/resolv.conf. The temp roots
+ * (is_sysroot_backed_temp_path) and the guest system directories
+ * (is_guest_system_path) are excluded from that fallback and stay on the
+ * sysroot spelling whether or not they exist there. Containment via realpath()
+ * is enforced only when the path actually resolves under sysroot, to prevent
  * symlink escape from a tree the caller intended to stay inside.
  */
 static bool lexical_normalize_absolute_path(char *dest,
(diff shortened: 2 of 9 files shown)


henrybear327:
Just realized that this is also targetting fixing https://github.com/sysprog21/elfuse/issues/253

docs/usage.md
```diff
@@ -17,7 +17,7 @@ Supported user-facing options:
 | `-V`, `--version` | Print the build version and exit |
 | `-v`, `--verbose` | Enable syscall-level and loader diagnostics |
 | `-t`, `--timeout N` | Per-iteration vCPU watchdog, in seconds (default `10`, `0` disables) |
-| `--sysroot PATH` | Resolve guest absolute paths under `PATH` first |
+| `--sysroot PATH` | Resolve guest absolute paths under `PATH`, ahead of the host |
```
jserv: Clarify "ahead of the host."

tests/test-sysroot-tmp-remove.c
```diff
@@ -0,0 +1,177 @@
...
+
+#include "test-harness.h"
+
+int passes = 0, fails = 0;
+
+/* True when the entry at @path is not visible to the guest at all. */
+static bool invisible(const char *path)
+{
+    struct stat st;
+    return stat(path, &st) != 0;
```
jserv: invisible() treats every stat() failure as "not visible", so a real resolver error (ELOOP, EACCES, ENAMETOOLONG) folds into a pass on the unlink/rmdir/rename assertions that guard on `invisible(p) || op == 0`. It also follows symlinks (stat) while the recursive-remove contract this test models uses lstat (line 108), so the two visibility notions disagree within one test. Count only ENOENT/ENOTDIR as invisible, and switch to lstat for consistency.

tests/test-sysroot-tmp-remove.c
```diff
@@ -0,0 +1,177 @@
...
+ * through a second open is what proves the redirect sends both operations to
+ * the same file, which a create alone does not.
+ */
+static bool round_trips(const char *path)
+{
+    int fd = open(path, O_CREAT | O_WRONLY | O_TRUNC, 0644);
+    if (fd < 0)
+        return false;
+    bool ok = write(fd, "k", 1) == 1;
+    if (close(fd) != 0 || !ok)
```
jserv: `close(fd) != 0 || !ok` leans on short-circuit evaluation order for the descriptor close. Splitting the close onto its own line reads clearer and keeps it unconditional.

tests/test-sysroot-tmp-remove.c
```diff
@@ -0,0 +1,177 @@
...
+ * cannot disagree with anything. Reports the first offending name through
+ * @offender.
+ */
+static bool listing_resolves(const char *dir,
+                             char *offender,
+                             size_t offender_sz)
+{
+    DIR *d = opendir(dir);
+    if (!d)
+        return true;
```
jserv: listing_resolves() returns true on any opendir() failure. A missing sysroot /tmp (ENOENT) is legitimately consistent, but ELOOP/EACCES signal a resolver bug and would still pass. The "a /tmp listing resolves" assertion is also weak: readdir and lstat share one resolver, so it cannot observe the readdir-vs-unlink disagreement the regression is about, and with an empty sysroot /tmp the opendir fails and the check passes vacuously. Distinguish ENOENT from other errno, or strengthen it to attempt a create-resolved op on each listed name. The unlink/rmdir/rename assertions below are the genuine regression catchers.

src/syscall/proc-state.c
```diff
@@ -609,6 +612,35 @@ static bool is_guest_system_path(const char *path)
...
+ * readable but not removable. The roots match as directories too, not only as
+ * prefixes of their children, so a listing cannot enumerate host entries whose
+ * children resolve into the sysroot.
+ *
+ * /var/tmp also lands in is_guest_system_path(), which claims all of /var for
+ * macOS SIP reasons; it belongs here on its own terms as a temp root, so the
+ * two rules stay independent. ccache matches by component because HOME may be a
+ * macOS home directory (/Users/<user>/.ccache) that no other rule covers.
+ */
+static bool is_sysroot_backed_temp_path(const char *path)
```
jserv, line 630: is_sysroot_backed_temp_path() no longer byte-matches the old create-side test in three ways: bare /tmp and /var/tmp roots now match (old required a trailing slash); .ccache matches as a final leaf (old strstr required "/.ccache/"); and .ccache matches at any depth and even when it names a regular file. Each widens only toward the sysroot, so there is no correctness or security problem, but the divergence from the old strncmp/strstr create semantics is worth a one-line note that it is deliberate.

jserv (commented):
Inline notes on the /tmp-lookup redirect. The core change checks out: buf is populated before the temp/system-path early return, the path_prefix_match header matches its definition, /var/tmp overlapping both classifiers is harmless (both OR to buf), and the matcher only ever widens toward the sysroot, so no host-fallback path is newly exposed. The remaining notes are test hardening plus one documentation point.

src/syscall/proc-state.c
henrybear327: Added a comment

tests/test-sysroot-tmp-remove.c
henrybear327: improved the error handling conditions

tests/test-sysroot-tmp-remove.c
henrybear327: inproved the error handling conditions

docs/usage.md
henrybear327: rephrased.

jserv:

PR #256, opened 2026-07-31
Redesign sidecar using byte-exact guest filenames without a sidecar index

henrybear327:
**The design is being refined as the testing and simplification of the code are carried out continuously**

---

A Linux guest names files by exact bytes. The default APFS volume matches names
case- and normalization-blind, so `Foo` and `foo` cannot coexist and a wrong-case
lookup succeeds where Linux owes `ENOENT`.

The previous answer, `src/syscall/sidecar.c` (2217 lines), stored such a name
under an opaque random token (`.ef_` plus 64 bits from `arc4random`, retried on
`EEXIST`) and recorded what the token meant in a per-directory
`.elfuse_case_index` file. The mapping lived outside the name, and everything
below follows from that one decision.

### A name was two objects, so no mutation was atomic

Creating, removing, or renaming a name had to touch the token and the index
file, which no host syscall can do together. `renameat` therefore ran a
hand-rolled two-phase commit: clone both directory indices as rollback
snapshots, write the index or indices to disk, call `renameat`, and on failure
write the saved snapshots back. The comment conceded the gap: "A failed
rollback write is the best-effort case." A crash between the index write and
the host rename left a mapping pointing at a token that had moved or did not
exist, and nothing repaired it on the next mount. Cross-directory rename took
two index locks in a fixed order, with a `swapped` flag deciding which went
first, so the ordinary rename path carried deadlock-avoidance logic.

### It forced serialization that the filesystem does not need

Two lock layers, because POSIX advisory locks are per-process and every vCPU
thread would share one: a process-wide `pthread_mutex_t` across all vCPU
threads, plus an `fcntl` write lock on a `.elfuse_case_index.lock` sentinel to
serialize against other elfuse processes on the same sysroot. Two guests
creating unrelated files in one directory contended, and the process mutex
serialized name creation across every directory at once.

### The index was advisory, so anything else desynchronized it

`fcntl` locks bind only the processes that take them. Any writer that is not
elfuse (`tar`, `rsync`, `git`, Finder) could add or delete entries without
touching the index, leaving rows describing tokens that were gone and files
with no row at all. In the other direction the tokens are opaque, so a sysroot
was unreadable and unmaintainable outside elfuse, and copying one preserved
the mapping only if the index files came along intact.

### Reads paid for it, and the cache was itself a correctness surface

The walker visited every parent directory of every translated path, probing
each for an index. That cost dominated startup, measured at "openat 61% of
`getent`'s 7.5 ms warm path", so it needed a 64-slot open-addressed cache
keyed on `(dev, ino, mtime, ctime)`, last writer wins on collision. Ordering
then became load-bearing: the publish path had to invalidate the cache
*before* the rename, or a concurrent walker could cache "no index here" and
miss one another process had just published.

### Leftovers

Three reserved names per directory (index, `.tmp`, `.lock`) had to be filtered
out of every listing and refused as guest paths. Each mutation parsed the whole
index into a heap row array and deep-cloned it for rollback, so a create in a
large directory cost work proportional to the directory. And the sidecar
resolved paths independently of `path_translate_at`, so two walks disagreed
about when a guest path falls through to the host.

## Approach

Decide a name's on-disk spelling from the name alone. Nothing is persisted, so
nothing can desynchronize, and no lock is needed because there is no second
object to keep consistent.

- **Codec** (`src/syscall/casefold.c`): a name containing an uppercase ASCII
  letter, a byte above `0x7F`, or an escape-shaped prefix is stored as
  `.ef=<payload>`. What is left literal is lowercase ASCII, a fixed point of
  every transformation the volume applies, so two literal names cannot collide
  however the folding table changes. The payload has two tiers because the
  per-name limit is 255 UTF-16 code units rather than bytes: hex to 125 bytes,
  then twelve bits per symbol from 4096 CJK ideographs at U+4E00, a block that
  neither folds nor normalizes. Guest names keep all 255 bytes Linux allows.
- **Walk** (`src/syscall/casefold-walk.c`): resolve a path one component at a
  time, asking the volume for each name as stored (`getattrlistat`) and
  comparing bytes, the only way to separate "exists as spelled" from "exists
  under a spelling that folded onto it". Absence is three answers, not one:
  unclaimed falls through to the host, a slot held by a differently-spelled
  sibling is `ENOENT`, and a non-directory component is `ENOTDIR`
  (`path_resolution(7)`).
- **One resolver**: `proc_resolve_sysroot_path_flags` and `path_translate_at`
  fold into that single walk, which also backs the `openat2`
  `RESOLVE_NO_SYMLINKS` and `RESOLVE_NO_XDEV` walkers. Directory entries and
  the cwd decode on the way back out.
- **Sidecar deleted**: index, parsers, both locks, the snapshots, both caches,
  and its five private syscall handlers, which now get the ordinary `/proc`,
  FUSE, and `/dev/shm` handling. Colliding creates never contend, because the
  spelling is a function of the name. Two elfuse processes sharing a sysroot
  need no coordination, because no shared mutable state remains.

## Prior art

The same question ("the host filesystem cannot hold this name") has three
standard answers, and this branch moves elfuse from the worst to the best of
them.

- **Ask the filesystem.** Cygwin took this route: `managed` mounts, its
  encoding scheme, were removed in 1.7 in favor of real NTFS case sensitivity
  via the `obcaseinsensitive` kernel flag, with per-mount `posix=[0|1]`.
  elfuse's equivalent is `--create-sysroot`, which builds a case-sensitive APFS
  sparsebundle, and Docker Desktop and Lima sidestep it entirely with a Linux
  VM over ext4. It is the cleanest answer when available, which is why the
  check suite re-runs the whole name suite on that volume as its oracle. It is
  not always available: the sparsebundle is opt-in and a user may point
  `--sysroot` at any directory.
- **Derive the stored name from the guest name.** Where Cygwin still must
  encode, for characters Windows disallows, it transposes them into the Unicode
  private use area by adding `0xf000`: a pure per-character function with no
  side table. gocryptfs, EncFS, and eCryptfs are the same family, deriving the
  on-disk name from the plaintext name and a key. This branch is that answer.
- **Keep a mapping.** The sidecar was here, and so is CryFS, which stores names
  inside encrypted blocks. Even gocryptfs falls back to it, but only at the
  boundary the derivation cannot cross: when an encrypted name would exceed
  255 bytes it writes a `gocryptfs.longname.<hash>.name` support file. That is
  precisely what elfuse's second tier avoids. Packing twelve bits per CJK
  symbol keeps the derived property across the full 255-byte guest range, so
  there is no length at which a side file reappears.

## Compatibility

**A sysroot written by the old encoding must be recreated.** Its `.ef_<token>`
entries and `.elfuse_case_index` files decode to nothing and surface under
their literal host names. `docs/usage.md` says so and gives the recipe.

## Tests

Two host unit lanes (codec, resolver) link exactly the code under test.
`tests/casefold-vectors.h` freezes the on-disk spellings in both directions,
making a format change a deliberate migration. Guest lanes cover edge shapes
(one-character sysroot, no sysroot, chdir), i18n pairs the volume folds, name
and path length budgets, host-staged spellings, a corpus read back by guest
name, and concurrency (a fork race run ten times in `check`, plus a manual
soak).

Two lanes are oracles rather than expectations:

- `check-name-caseexact` re-runs the name suite on a case-sensitive APFS
  sparsebundle, where the volume itself enforces what the tests assert. It is
  what surfaced the `ENOTDIR` and vanished-path fixes.
- `test-sysroot-path-matrix` enumerates addressing mode x operation x path
  shape x name class and holds each cell to `path_resolution(7)`: absolute,
  cwd-relative, and dirfd-relative spellings of one file must agree on result,
  errno, and object. On first run it caught two bugs nobody predicted. It runs
  on the folding tmpdir, the byte-exact volume, and the qemu reference kernel.

The matrix gains `ELFUSE_SKIP` as the inverse of `QEMU_SKIP`, with
`.ci/check-matrix-lists.sh` rejecting a label naming no registered test or
sitting in both lists.

---
## Summary by cubic
Redesigns filename handling with a stateless `.ef=` codec and a case‑exact path resolver, replacing the sidecar so Linux byte‑exact names work on case‑folding APFS without locks or per‑dir files. All path surfaces resolve and report guest bytes (symlink targets, exec, inotify, pathname `AF_UNIX`, directory listings, cwd); temp roots (`/tmp`, `/var/tmp`, `~/.ccache`) are consistently redirected through the sysroot for creates and lookups. `..` now clamps at the guest root, and `openat2(RESOLVE_NO_SYMLINKS)` handles deep link‑free paths correctly.

- **Bug Fixes**
  - Symlinks: follow targets in the guest namespace; absolute targets resolve against the sysroot; `openat2(RESOLVE_NO_SYMLINKS)` returns `ELOOP` when a link is present.
  - Exec: translate dirfd‑relative paths; resolve `PT_INTERP` through the same walk with `/lib/<basename>` fallback; `/proc/self/exe` and `/proc/self/fd/N` report guest bytes; shm probes honor `O_NOFOLLOW`.
  - AF_UNIX (pathname) sockets: bind/connect/send* resolve through the sysroot with case‑exact names; over‑long paths shorten via a private namespace symlink with correct cleanup across fork/exit; getsockname/getpeername/accept/recv* decode guest paths and fix msg name lengths.
  - Inotify: watches translate through the resolver; event names decode to guest bytes; `/dev/shm` leaves are watched without following.
  - Path correctness: clean `ENAMETOOLONG` at the host path ceiling (no truncation); trailing `/` keeps `ENOTDIR`; `--sysroot /` parent split fixed; directory entry decoding is scoped to sysroot‑owned directories (host fallback and no‑sysroot lists stay literal); `openat2(RESOLVE_NO_XDEV)` re‑checks mount class after open; `..` is clamped at the guest root for absolute and dirfd‑relative paths; `RESOLVE_NO_SYMLINKS` no longer caps components, so deep link‑free paths resolve.
  - Cwd: `getcwd` and `/proc/self/cwd` report guest bytes so the path is chdir‑able.

- **Migration**
  - Sysroots written by the old sidecar format must be recreated. Existing `.ef_<token>` and `.elfuse_case_index` files appear as literal host entries. See docs/filenames.md and docs/usage.md.

<sup>Written for commit c65bf64c3437f892c0226e44e3fc97193080edb4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/256?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src/syscall/path.c
```diff
@@ -403,6 +553,49 @@ int sys_path_has_symlink(guest_fd_t dirfd, const char *path)
...
+                errno = ELOOP;
+                rc = -1;
+                goto out;
+            }
+            scan = sysroot_buf;
+        }
+    }
+
     while (path_next_component(&scan, &comp, &len)) {
         if (++walk_count > MAXSYMLINKS) {
```
jserv: This counter increments once per path component, but any symlink already returns ELOOP at the S_ISLNK check twenty lines down, so the walk never follows a link. The only thing walk_count > MAXSYMLINKS rejects is a legitimate symlink-free path deeper than 40 components, which RESOLVE_NO_SYMLINKS is supposed to resolve. Drop the counter, or bump it only when a link is actually crossed.

src/syscall/proc-state.c
```diff
@@ -643,6 +664,90 @@ static bool is_sysroot_backed_temp_path(const char *path)
...
+    int which = 0;
+
+    for (int depth = 0;; depth++) {
+        casefold_verdict_t verdict;
+        char target[LINUX_PATH_MAX];
+        const char *rest;
+        char *next;
+        ssize_t n;
+
+        if (depth > MAXSYMLINKS) {
```
jserv: Off-by-one against Linux. depth starts at 0 and one link is followed per iteration, so depths 0..40 follow 41 links before this fires. Linux caps at 40, and the other two walkers here use ++count > MAXSYMLINKS. Use depth >= MAXSYMLINKS so a 41-link chain gets ELOOP like everywhere else.

src/syscall/proc-state.c
```diff
@@ -656,25 +761,102 @@ static const char *proc_resolve_sysroot_path_flags(const char *path,
...
+     * the host spelling, and a concatenation plus one probe answers both.
+     */
+    bool present;
+    bool folded = false;
+    bool followed_link = false;
+    char followed[LINUX_PATH_MAX];
+    if (casefold_active()) {
+        casefold_walk_t walk;
+        casefold_verdict_t verdict =
+            resolve_through_links(sr, path, follow_final, buf, bufsz, &walk,
```
jserv: resolve_through_links gets the raw guest path, and the walk appends '..' literally (casefold-walk.c:429) on the stated precondition that absolute paths are already collapsed. This resolver never collapses them. A guest path like /../etc/hostname becomes sysroot/../etc/hostname on the host, escapes the sysroot lexically, and the containment check turns that into ELOOP where Linux clamps '..' at / and opens sysroot/etc/hostname. Run the absolute path through lexical_normalize_absolute_path before seeding the walk. (Fail-closed, and the byte-exact branch shares the shape, so confirm scope.)

src/syscall/casefold-walk.c
```diff
@@ -0,0 +1,507 @@
...
+
+            /* Read here, not before the test: obj_type only sits at this
+             * offset because name_ref precedes it, and it does so only when
+             * the name was returned. With the name withheld the field would
+             * be read one attrreference_t past where the volume wrote it, and
+             * a garbage VLNK sends the walk chasing a link that is not there.
+             */
+            if (is_link && (attr_buf.returned.commonattr & ATTR_CMN_OBJTYPE))
+                *is_link = attr_buf.obj_type == VLNK;
+            if (!strcmp(stored, leaf))
```
jserv, line 214: stored points into the getattrlistat reply via attr_dataoffset and goes straight into strcmp with no bound on attr_length and no guaranteed NUL. APFS behaves, but --sysroot can point at an SMB or NFS mount that need not. A name that fills the buffer with no terminator over-reads adjacent stack. Validate name_ref offset+length against the returned buffer and require a NUL before comparing.

src/syscall/path.c
```diff
@@ -280,10 +371,59 @@ int path_translate_at(guest_fd_t dirfd,
...
 }
 
-int path_translate_dirent_name(guest_fd_t dirfd,
+bool path_dirent_dir_holds_escapes(host_fd_t host_dirfd)
+{
+    char sr[LINUX_PATH_MAX], dirpath[LINUX_PATH_MAX];
+
+    if (!casefold_active() || !proc_sysroot_snapshot(sr, sizeof(sr)))
+        return false;
+    if (fcntl(host_dirfd, F_GETPATH, dirpath) < 0)
```
jserv, line 424: Returning true when F_GETPATH fails means an fd whose sysroot membership cannot be established has its directory entries decoded as guest names. A host-backed directory that cannot be named back but happens to hold a foreign file literally named .ef=61 then shows the guest a phantom 'a' that no open can resolve. This decode question should fail closed the other way: return false unless the directory is proven inside the sysroot.

src/syscall/net-absock.c
```diff
@@ -283,19 +535,53 @@ void absock_bind_rollback(int idx)
...
+    /* The shortening links are untracked, so no table owns them and the
+     * process that minted the namespace retires them on its way out. Sweep
+     * only symlinks: the abstract-socket backing files sharing this directory
+     * belong to whichever process bound them, including children that outlive
+     * the owner, and each unlinks its own above. Unlinking those too would
+     * destroy a socket a live child still has bound. A live child that loses
+     * a shortening link only degrades to the raw host link path, never to a
+     * missing socket.
+     */
+    if ((uint64_t) getpid() == absock_get_namespace_id()) {
```
jserv: Only the namespace owner sweeps shortening symlinks, but any forked child can mint one via absock_shorten_path after the owner has exited. That child is not the owner, skips the sweep, and its rmdir at line 586 fails with the link still present, so the link and the directory leak. The mirror case also bites: when the owner exits first it unlinks a shortening link a live child is still resolving connect() through. Track shortening links per process and unlink your own in every process's cleanup, regardless of ownership.

src/syscall/net-absock.c
```diff
@@ -196,6 +230,224 @@ static int absock_build_sun(const char *fs_path,
...
+    }
+    absock_encode_name((const uint8_t *) host_path,
+                       (uint32_t) strlen(host_path), out, out_sz);
+    /* Create-first, never unlink a matching link: absock_lock is
+     * per-process, so an unconditional unlink could yank a forked sibling's
+     * just-created link between its shorten and its bind(2). The encoded
+     * name is derived from the target, so an existing link with the same
+     * name almost always already points at the right place.
+     */
+    if (symlink(host_path, out) < 0) {
```
jserv: The shortening name comes from absock_encode_name, which for a path too long to hex-encode (every path that reaches here) keeps a 20-byte literal prefix plus a 32-bit FNV hash. Two distinct targets that share the first 20 bytes and collide on the hash derive the same link name; the block below then unlinks and repoints a link whose readlink target does not match, so one socket's live link is aimed at another's target, and with the lock dropped before the caller's connect() a peer reaches the wrong socket. Widen the digest (64-bit, or a truncated SHA-256 of the full path).

jserv (commented):
Correctness pass over the casefold redesign. Seven findings inline; each is on a line this PR introduces or newly exercises. No blockers to the direction, which is a clear win over the sidecar.

jserv:

henrybear327:
This branch contains an additional change that I was about to split out as a separate PR for review, but since the PR is now merge I will document it down here. 

The first 2 commits of this PR is to fix a problem that was discovered on the main branch (the error line `ls: /..: Symbolic link loop`):
```
➜  elfuse git:(oci_on_main) ✗ ./build/elfuse-oci run --entrypoint /bin/sh alpine:3 -c 'ls -la /'
ls: /..: Symbolic link loop
total 0
drwxr-xr-x   19 501      dialout        608 Aug  5 07:10 .
drwxr-xr-x   84 501      dialout       2688 Aug  5 07:10 bin
drwxr-xr-x    2 501      dialout         64 Jul 21 20:52 dev
drwxr-xr-x   38 501      dialout       1216 Aug  5 07:10 etc
drwxr-xr-x    2 501      dialout         64 Jul 21 20:52 home
drwxr-xr-x    8 501      dialout        256 Aug  5 07:10 lib
drwxr-xr-x    5 501      dialout        160 Aug  5 07:10 media
drwxr-xr-x    2 501      dialout         64 Jul 21 20:52 mnt
drwxr-xr-x    2 501      dialout         64 Jul 21 20:52 opt
dr-xr-xr-x    3 root     root             0 Jan  1  1970 proc
drwx------    2 501      dialout         64 Jul 21 20:52 root
drwxr-xr-x    3 501      dialout         96 Aug  5 07:10 run
drwxr-xr-x   64 501      dialout       2048 Aug  5 07:10 sbin
drwxr-xr-x    2 501      dialout         64 Jul 21 20:52 srv
drwxr-xr-x    2 501      dialout         64 Jul 21 20:52 sys
drwxrwxrwt    2 501      dialout         64 Jul 21 20:52 tmp
drwxr-xr-x    7 501      dialout        224 Aug  5 07:10 usr
drwxr-xr-x   13 501      dialout        416 Aug  5 07:10 var
```

jserv:
> This branch contains an additional change that I was about to split out as a separate PR for review, but since the PR is now merge I will document it down here.

Next time, do incremental changes for better review.

PR #259, opened 2026-08-01
Add Linux-compatible /proc/self/smaps emulation

### LINKED ISSUE #258: Running redis-server under elfuse fails because /proc/self/smaps is unavailable
## Background

I tried to run `redis-server` as a Linux guest workload under elfuse. Redis failed during startup and exited intentionally:

```text
31 Jul 2026 03:43:51.806 # Failed to test the kernel for a bug that could lead to data corruption during background save. Your system could be affected, please report this error.
31 Jul 2026 03:43:51.806 # Redis will now exit to prevent data corruption. Note that it is possible to suppress this warning by setting the following config: ignore-warnings ARM64-COW-BUG
```

## Root cause

Redis performs its ARM64 copy-on-write safety check by reading `/proc/self/smaps`. elfuse currently exposes `/proc/self/maps`, but does not provide a Linux-compatible `/proc/self/smaps` (and the corresponding `/proc/<pid>/smaps` path), so Redis cannot inspect the mapping and dirty-page information it expects. Redis treats a failed check as unsafe and terminates to avoid possible data corruption during background saves.

This makes Redis unusable under elfuse even when the underlying system is not affected by the reported kernel bug.

## Expected behavior

elfuse should provide a parseable, Linux-shaped `/proc/self/smaps` implementation, with the same VMA view and snapshot semantics as `/proc/self/maps`. The implementation should expose the fields required by Redis's COW check and derive the available values from elfuse's tracked guest VMAs and fork state. Values that cannot be observed precisely by the emulator should be documented as coarse compatibility values rather than presented as host page-accounting data.

## Proposed scope

- Generate `/proc/self/smaps` and `/proc/<pid>/smaps` from the tracked guest VMA list.
- Emit Linux-compatible VMA headers and the standard smaps fields, including `VmFlags`.
- Preserve fork-related COW information sufficiently for Redis's ARM64 safety check.
- Keep `/proc/self/maps` and `/proc/self/smaps` consistent and ensure output is not truncated for larger VMA sets.
- Add regression coverage that parses the synthetic smaps output and exercises the fork/COW accounting path.

`smaps_rollup` and exact host page-residency accounting are out of scope unless a consumer requires them.

## Acceptance criteria

1. `redis-server` starts successfully when run as a Linux workload under elfuse.
2. Redis no longer reports the failed ARM64-COW-BUG kernel test solely because `/proc/self/smaps` is missing or malformed.
3. Guest reads of `/proc/self/smaps` and `/proc/<pid>/smaps` return complete, parseable records for all tracked VMAs.
4. Existing procfs and `/proc/self/maps` behavior remains compatible.
5. Automated tests cover normal mappings, forked children, writable private anonymous mappings, and large VMA snapshots.

sunxiaoguang:
Fixes #258

## Summary

Running `redis-server` under elfuse currently fails during Redis's ARM64 copy-on-write safety check because Redis reads `/proc/self/smaps`, which elfuse does not provide with Linux-compatible contents. Redis treats the failed check as unsafe and exits to protect background saves.

This PR adds synthetic `smaps` support so Redis can inspect guest VMAs and complete startup under elfuse.

## Implementation

- Generate `/proc/self/smaps` and `/proc/<pid>/smaps` from the tracked guest VMA snapshot.
- Emit Linux-shaped VMA headers and the standard smaps fields, including `VmFlags`.
- Preserve fork-aware COW information with a coarse `Shared_Dirty` compatibility signal for writable private anonymous mappings in fork children.
- Share VMA collection and formatting logic with `/proc/self/maps`.
- Replace the fixed maps-entry/output limits with checked growable array and string-builder utilities.
- Add parser and regression coverage for smaps layout, field completeness, fork behavior, and larger VMA snapshots.
- Document the synthetic smaps semantics and intentional limitations; exact host page residency and `smaps_rollup` remain out of scope.

## Testing

- `make test-dynamic-array-host test-string-builder-host` — passed.
- `git diff --check` — passed.
- Guest `test-proc-smap` compilation was attempted, but the local environment does not have `aarch64-linux-gnu-gcc`.

## Compatibility notes

The synthetic values are derived from elfuse's tracked guest VMAs and fork state. Fields requiring host kernel page accounting are emitted as stable compatibility values and should not be interpreted as precise memory profiling data.

---
## Summary by cubic
Adds Linux-compatible `/proc/self/smaps` and `/proc/<pid>/smaps` so `redis-server` passes its ARM64 COW safety check and starts under elfuse. Also standardizes FD-limit handling and adds descriptor-exhaustion regressions to `make check`.

- **New Features**
  - Emit Linux-shaped smaps (headers, standard fields, `VmFlags`) for self and PIDs; reuse `/proc/self/maps` VMA collection; append then sort+merge once; cap by guest limits.
  - Track per-VMA fork state and stable `vma_id`; keep `Shared_Dirty`/`Pss_Dirty` consistent and exclude post-fork mappings; propagate via fork IPC and memory ops.
  - Add `dynamic-array` and `string-builder` utilities with host tests; add a smaps parser and fork/mremap regressions.
  - Introduce `elfuse-limits.h` for shared FD limits (`FD_TABLE_SIZE`, `HOST_FD_RESERVE`, `HOST_NOFILE_MIN`); test runners derive limits via `tests/test-config.sh`; `tests/manifest.txt` supports `host_nofile=elfuse-minimum`.

- **Bug Fixes**
  - Harden fork-aware smaps and `mremap`: preserve VMA lineage across fork/restore; make region splits/moves transactional; flush shared files before removes; avoid flushing writable aliases for read-only `MAP_SHARED`; cover `PROT_NONE`.
  - Harden `mmap`/`mremap` failure paths: reserve FDs before teardown; keep `EMFILE` failures side-effect free; preserve smaps accounting and errno; preserve failed `mmap` reservation addresses; add descriptor- and table-exhaustion regressions; run the EMFILE regression in `make check`.
  - Use unique `/dev/shm` fixture names to avoid collisions in concurrent test runs.

<sup>Written for commit 69aace6e53d4382831a4ac0cfc2bc0ea020898f1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/259?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jserv (changes requested):
Run `make indent` before committing.

sunxiaoguang:
> `make indent`

Thanks for the review. I’ve addressed all other review comments as well, including the dynamic-array and string-builder initialization/aliasing issues and the `/proc/*/maps`/`smaps` overlap and performance issues. `make
  indent` now completes successfully. Please take another look when you have a chance.

jserv (commented):
A few defensive and semantic notes on the smaps/maps generation. None block the Redis CoW goal.

src/runtime/procemu.c
```diff
@@ -2330,144 +2428,291 @@ static int pty_open_master(int linux_flags)
...
         uint64_t start = r->start & ~0xFFFULL;
-        uint64_t end = (r->end + 0xFFF) & ~0xFFFULL;
-
-        if (nentries > 0 && entries[nentries - 1].end == start &&
-            entries[nentries - 1].prot == r->prot &&
-            entries[nentries - 1].flags == r->flags &&
-            entries[nentries - 1].offset == r->offset &&
-            !strcmp(entries[nentries - 1].name, r->name)) {
-            entries[nentries - 1].end = end;
+        uint64_t end = (r->end + 0xFFFULL) & ~0xFFFULL;
```
jserv: Page round-up `(r->end + 0xFFFULL) & ~0xFFFULL` wraps to 0 when `r->end` is within 0xFFF of UINT64_MAX. That makes the later `end <= start` test true and silently drops or corrupts the mapping. Same pattern at lines 206 and 213 for shadow start/end. Guest user VAs are 48-bit today so this is not reachable in practice, but a saturating `align_up_page()` helper would remove the silent wrap.

src/runtime/procemu.c
```diff
@@ -95,57 +103,145 @@ typedef struct {
...
+            continue;
+        if (live_start >= shadow_end)
+            break;
+
+        if (live_start > cursor) {
+            uint64_t gap_end =
+                live_start < shadow_end ? live_start : shadow_end;
+            uint64_t offset = shadow->offset;
+            uint64_t delta = cursor - shadow_start;
+            if (delta <= UINT64_MAX - offset)
```
jserv: When `delta > UINT64_MAX - offset`, the code leaves `offset` unchanged and still appends the gap (same at line 238), so the emitted VMA reports a stale, repeated file offset instead of failing. Returning -1 with `errno = EOVERFLOW` would surface the condition rather than producing a silently wrong offset. Reachable only for offsets near UINT64_MAX.

src/runtime/procemu.c
```diff
@@ -2330,144 +2428,291 @@ static int pty_open_master(int linux_flags)
...
+        qsort(maps_entries_data(&entries), maps_entries_count(&entries),
+              sizeof(maps_entry_t), maps_entries_compare_start);
+    maps_entries_merge_adjacent(&entries);
+    result = (int) maps_entries_count(&entries);
+
+out_unlock:
+    pthread_mutex_unlock(&mmap_lock);
+    if (result < 0) {
+        maps_entries_destroy(&entries);
+        errno = ENOMEM;
```
jserv: This cleanup path flattens every failure to `errno = ENOMEM`, so a genuine EOVERFLOW or EINVAL from the snapshot builders reaches the guest as ENOMEM. `proc_finish_maps_output()` already does the save/restore pattern; mirroring it here preserves the real errno for the intercepted open.

src/runtime/procemu.c
```diff
@@ -2330,144 +2428,291 @@ static int pty_open_master(int linux_flags)
...
+    if (nentries < 0)
+        return -1;
+
+    string_builder_t builder = {0};
+    size_t initial_capacity = (size_t) nentries * 768;
+    int result = -1;
+    if (string_builder_init(&builder, initial_capacity) < 0)
+        goto out;
+
+    bool fork_child = proc_get_ppid() != 0;
```
jserv: `fork_child` is derived only from ppid, so `Shared_Dirty` (line 2686) reports the full VMA size for every writable private-anonymous VMA in a child, including mappings created after the fork, which are not part of the CoW snapshot. The docs already frame this as a coarse compatibility signal, and it errs in the conservative direction for a fork-safety check, so this is a confirm-the-tradeoff note. A per-region inherited-at-fork flag would make it precise if that ever matters.

jserv (changes requested):
After responding to @cubic-dev-ai , squash commits and enforce rules described by https://cbea.ms/git-commit/ .

sunxiaoguang:
> After responding to @cubic-dev-ai , squash commits and enforce rules described by https://cbea.ms/git-commit/ .

Sure, it's getting complicated when it comes to fork-safe. I'm working on it and trying to implement related book keepings.

sunxiaoguang:
I guess there are still corner cases and ways to improve the fork-safe memory statistics. However it's getting too complicated, maybe we can stop digging deeper and start reviewing at this point. Please take another look when you have a chance, Thanks. @jserv @Max042004

src/runtime/procemu.c
```diff
@@ -2330,144 +2462,295 @@ static int pty_open_master(int linux_flags)
...
+#undef APPEND_VMFLAG
+        vmflags[vmflags_len] = '\0';
+
+        if (string_builder_appendf(
+                &builder,
+                "%.*s\n"
+                "Size: %llu kB\n"
+                "KernelPageSize: 4 kB\n"
+                "MMUPageSize: 4 kB\n"
+                "Rss: 0 kB\n"
```
jserv: This block reports Rss: 0 kB but Shared_Dirty: <size> kB for a fork-inherited writable private-anon VMA (shared_dirty_kb == size_kb at line 2683). Linux guarantees Rss is at least the sum of the clean and dirty page counts, so Rss: 0 alongside a non-zero Shared_Dirty is internally contradictory. A field-summing consumer tolerates it, but a validating parser sees a malformed block. Set Rss (and Pss) to shared_dirty_kb so the residency total is consistent with the dirty count it reports.

src/runtime/procemu.c
```diff
@@ -2330,144 +2462,295 @@ static int pty_open_master(int linux_flags)
...
+                "AnonHugePages: 0 kB\n"
+                "ShmemPmdMapped: 0 kB\n"
+                "FilePmdMapped: 0 kB\n"
+                "Shared_Hugetlb: 0 kB\n"
+                "Private_Hugetlb: 0 kB\n"
+                "Swap: 0 kB\n"
+                "SwapPss: 0 kB\n"
+                "Locked: 0 kB\n"
+                "THPeligible: 0\n"
+                "ProtectionKey: 0\n"
```
jserv: ProtectionKey is an x86 pkeys field; arm64 kernels do not emit it in smaps at all. Since the point here is a Linux-arm64-shaped block, a schema-exact parser treats this line as unexpected. Drop ProtectionKey on this view (KSM and Pss_Dirty are also kernel-version-gated fields). Separately, an empty VmFlags renders as "VmFlags:\n" at line 2741, where Linux prints "VmFlags: " with a trailing space; the space is only added per token, so emit it unconditionally.

src/syscall/mem.c
```diff
@@ -1099,8 +1487,12 @@ static int capture_region_snapshots(guest_t *g,
                                     region_snapshot_t *snaps,
                                     int max_snaps)
 {
-    split_regions_at_boundary(g, start);
-    split_regions_at_boundary(g, end);
+    int split_err = split_regions_at_boundary(g, start);
```
jserv, line 1632: capture_region_snapshots is not failure-atomic. split_regions_at_boundary(start) commits its split in place (memmove + nregions++, plus a dup'd backing fd, lines 490-504); if the following split at end then fails with ENOMEM (region table full) or a dup failure, this returns the error with the start split already committed. The caller aborts the syscall, but the region table is left carrying a spurious boundary at start and holding one extra dup'd fd. This is not corruption (the two halves are metadata-equivalent) and not a leak (the fd is owned by the new region), but the operation reports failure after having mutated state, and the ignored guest_region_remove() return downstream at mem.c:3470 is only safe because these boundaries were pre-split here. Preflight/reserve both boundary splits before committing either, or roll back the start split when the end split fails.

jserv (commented):
Correctness pass over the smaps emulation. Three findings inline. The new dynamic-array and string-builder utilities are careful (overflow-guarded growth, realloc-NULL handled, NUL kept), and the fork IPC carries the new VMA fields via the existing whole-region memcpy with the MAGIC bump, so no serialization gap there.

Two items left out of line, for the record:

- mem.c:3770 (MAP_SHARED mremap): possible write-loss when the source is a MAP_SHARED file mapping with no live overlay. The aligned/overlay case is fine, since tearing the overlay down flushes dirty pages back to the file; the concern is only the non-overlay fallback, and it looks pre-existing rather than introduced here. Unresolved either way. Worth a targeted test (dirty a misaligned MAP_SHARED file mapping, mremap it, assert the file reflects the writes) before any change.

- next_vma_id not serialized across fork: the child keeps next_vma_id == 0 while inherited regions carry large ids. Not a correctness bug, allocate_vma_id linear-scans live regions and never hands back an in-use id, so collisions cannot happen; the only cost is O(nregions) work per allocation in a fork child and a field documented as "last allocated" that is silently wrong there. Optional: set it to max(region.vma_id)+1 in the child where inherited_at_fork is stamped.

jserv:
> I guess there are still corner cases and ways to improve the fork-safe memory statistics.

Yes, that is exactly the direction of my current research. Beyond addressing corner cases in fork-safe memory statistics, the broader goal is to consolidate a kernel-less approach that provides a Linux system call compatibility layer while rigorously validating memory safety and futex correctness. The challenge is not only functional compatibility but also ensuring that the concurrency semantics remain equivalent to those expected by Linux applications.

sunxiaoguang:
Thanks, let me check the comments.

jserv:
I defer to @Max042004 and @henrybear327 for confirmation.

sunxiaoguang:
> I defer to @Max042004 and @henrybear327 for confirmation.

Thank you for all your help.

jserv (changes requested):

henrybear327:
Add OCI image Redis workflow https://github.com/sysprog21/elfuse/pull/270

src/core/guest.c
```diff
@@ -2217,6 +2321,17 @@ void guest_region_remove(guest_t *g, uint64_t start, uint64_t end)
...
     g->nregions = out + tail;
+    if (reserved_backing_fd >= 0)
+        close(reserved_backing_fd);
+    return 0;
+}
+
+int guest_region_remove(guest_t *g, uint64_t start, uint64_t end)
+{
+    int reserved_backing_fd = -1;
+    if (guest_region_remove_prepare(g, start, end, &reserved_backing_fd) < 0)
```
henrybear327, line 2383: P1: Fixed mmap and mremap operations can leave overlapping or stale region metadata when descriptor exhaustion makes the new removal preflight fail, because existing callers ignore guest_region_remove()'s return value. The callers should use the prepare/reserved API and abort before mutating mappings when reservation fails.

https://github.com/sysprog21/elfuse/pull/270/changes#r3723128567

src/string-builder.c
```diff
@@ -0,0 +1,244 @@
...
+    char sizing_sink;
+
+    if (builder == NULL || format == NULL)
+        return string_builder_invalid();
+
+    saved_errno = errno;
+    va_start(arguments, format);
+
+    va_copy(sizing, arguments);
+    errno = 0;
```
henrybear327: P2: Printf-style %m formatting loses the caller's error and emits Success whenever errno is nonzero because both sizing and rendering passes run with errno = 0. Preserve the saved errno while invoking vsnprintf, then normalize errno only when handling a failed pass.

https://github.com/sysprog21/elfuse/pull/270/changes#r3723128655

docs/internals.md
```diff
@@ -768,6 +794,54 @@ under `/proc`, `/dev`, and a few Linux-expected compatibility files:
...
+Anonymous, KSM, LazyFree, AnonHugePages, ShmemPmdMapped, FilePmdMapped,
+Shared_Hugetlb, Private_Hugetlb, Swap, SwapPss, Locked, THPeligible,
+VmFlags
+```
+
+`Size` is the VMA length in KiB. `KernelPageSize` and `MMUPageSize` are
+reported as 4 KiB. In a fork child, writable private anonymous VMAs that
+existed in the parent's CoW snapshot report their full VMA size for
+`Shared_Dirty`, `Rss`, and `Pss`, keeping those coarse counters internally
+consistent. Newly-created VMAs are excluded from that signal. Every other
```
henrybear327: P3: The smaps section says fork-child writable private anonymous VMAs report their full VMA size for Shared_Dirty, Rss, and Pss, but the code also sets Pss_Dirty to the same value (pss_dirty_kb = shared_dirty_kb in proc_open_self_smaps). Since the doc states every other numeric counter is a stable zero, a reader would infer Pss_Dirty stays 0 in that signal, which is wrong. Extend the enumeration to include Pss_Dirty.

https://github.com/sysprog21/elfuse/pull/270/changes#r3723128676

sunxiaoguang:
> Rebase latest `main` branch and resolve conflicts.

Sure, I have rebased the PR with the main branch

mk/tests.mk
```diff
@@ -40,6 +43,11 @@ test-hello: $(ELFUSE_BIN) $(TEST_HELLO_DEP)
 	@printf "$(BLUE)▸ Running$(RESET) test-hello\n"
 	$(ELFUSE_BIN) $(TEST_DIR)/test-hello
 
+## Run the libc-based file-backed region removal EMFILE regression probe
```
Max042004, line 57: `tests/test-mremap-tail-emfile.c` is absent from the `check:`
recipe, from `tests/test-matrix.sh`, and from `tests/manifest.txt`.

tests/test-mremap-tail-emfile.c
```diff
@@ -0,0 +1,235 @@
...
+    void *result = mremap(base, SPAN, 3 * SPAN, 0);
+    if (result == MAP_FAILED || result != base)
+        child_fail(10);
+
+    base[SPAN] = 'T';
+    base[2 * SPAN] = 'R';
+    base[SPAN + PAGE_SIZE] = 'S';
+    base[SPAN + 3 * PAGE_SIZE] = 'M';
+
+    /* Exhaust both descriptor tables, then leave one slot available. */
```
Max042004, line 55: `FD_TABLE_SIZE` is 1024, host `RLIMIT_NOFILE` is `FD_TABLE_SIZE + HOST_FD_RESERVE` (256). A guest that exhausts its own fd table cannot starve elfuse's internal `dup()`.

To actually exercise it, the test would have to lower the host `RLIMIT_NOFILE`.

Max042004 (changes requested):

Max042004:
Inside `munmap_guest_range()`, `cleanup_overlays_in_range()` (`mem.c:4196`)
splits at host-page boundaries — transiently borrowing one region slot — two
stages before `guest_region_remove_reserved()` (`mem.c:4224`) returns the slots
the call frees. With `split_regions_at_boundary()` now returning
`-LINUX_ENOMEM` on a full table instead of logging and continuing
(`mem.c:548`), that borrow aborts the whole `munmap`: step 18's
`munmap([0x200034000-0x202034000), 32 MiB)` fails at `nregions=4096` even
though it fully covers ~4000 filler regions and would have left the table at
~90 (the only error logged is the split refusing the straddling region
`[0x202030000-0x202041000)` at `host_end`). On Linux a whole-range `munmap`
never needs a new VMA and always succeeds — measured in the same VM at
`max_map_count=1000` — so step 18's assertion is right and the elfuse path is
what needs adjusting; the host-page split is an elfuse implementation detail
with no Linux counterpart and should not be able to fail an operation that only
reduces the mapping count (e.g. skip the boundary splits when no region in the
window has a live overlay).

sunxiaoguang:
Follow-up for commit `d7db64a` (`Harden mmap failure handling`):

The review findings are addressed as follows:

- Fixed `mmap`/`mremap` descriptor-exhaustion atomicity by reserving the backing descriptors with `guest_region_remove_prepare()` before overlay teardown, page-table changes, or mapping destruction. The commit consumes the reservation through `guest_region_remove_reserved()` and closes it on every failure path, so `EMFILE` leaves the existing mappings and region metadata unchanged.
- Preserved the caller's `errno` during both `string_builder_appendf()` `vsnprintf` passes, including `%m`, and added host coverage for the formatted errno expansion.
- Updated the synthetic smaps documentation to include `Pss_Dirty` in the fork-inherited dirty-accounting signal.
- Added `test-mremap-tail-emfile` to the manifest and test matrix, lowered the host `RLIMIT_NOFILE` to exercise elfuse's descriptor reserve, and skipped the test in QEMU because it validates elfuse-specific bookkeeping.
- Changed `cleanup_overlays_in_range()` to avoid boundary splits when the range contains no live overlay, so a full region table cannot make a reducing-only `munmap` fail. The new test covers this case.
- Removed the unrelated `PREBUILT_TEST_BINARIES`/`prebuilt=skip` compatibility layer; the repository does not use such a corpus.

jserv (changes requested):
Rebase latest `main` branch and resolve conflicts, as #272 and #277 introduced machine-checked proofs.

sunxiaoguang:
Rebased with latest main and passed all the checks.

jserv:
Thank @sunxiaoguang for contributing!

PR #262, opened 2026-08-02
Report AF_UNIX datagram peer close as EOF

### LINKED ISSUE #260: Rust programs abort when spawning a subprocess: recvfrom returns ECONNRESET where Linux returns EOF
# Rust programs abort when spawning a subprocess: recvfrom returns ECONNRESET where Linux returns EOF

## Environment

- elfuse: `58e0cf4`
- Host: macOS 26.6, Darwin 25.6.0, arm64 (Apple Silicon)
- Guest: Ubuntu 26.04 aarch64 rootfs, glibc

## Summary

Any Rust program running in the guest aborts as soon as it spawns a child
process. Rust's standard library reports an error that cannot occur on Linux
for this operation, so its `unwrap`-free error path panics and the process
dies with SIGABRT:

```
thread 'main' (1) panicked at library/std/src/sys/process/unix/unix.rs:155:21:
the CLOEXEC pipe failed: Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" }
fatal runtime error: failed to initiate panic, aborting
```

Linux errno 104 is `ECONNRESET`. After a successful `execve` in the child, the
parent's read of the exec-status channel must return EOF (0); it can never
return `ECONNRESET` on Linux, so std treats it as unreachable and aborts.

This bites hard on modern distributions. Ubuntu 25.10+ ships Rust coreutils
(uutils) as `/usr/bin`, and desktop stacks now delegate image decoding to Rust
helpers that spawn sandboxed loader processes, so a GTK application can die
simply by rendering an icon.

## Reproduction

Any Rust binary that spawns a child reproduces it. Ubuntu's uutils build is
convenient because it is already present in a stock rootfs:

```
elfuse --sysroot <ubuntu-rootfs> /usr/lib/cargo/bin/coreutils.real/timeout 5 /bin/true
```

Observed: panics as above, every run.
Expected: exits 0 silently.

Useful control from the same binary set — `env` execs directly, with no
exec-status channel, and succeeds:

```
elfuse --sysroot <ubuntu-rootfs> /usr/lib/cargo/bin/coreutils.real/env /bin/true
```

## What is known about the cause

The failing errno is produced inside `sys_recvfrom()` (`src/syscall/net.c:886`).
Instrumenting `linux_errno()` — the single point every failing syscall passes
through — logs exactly one hit immediately before the panic:

```
DIAG errno=54 caller=sys_recvfrom
```

macOS errno 54 is `ECONNRESET`, which `linux_errno()` translates to Linux 104,
the value Rust then reports. `sys_recvfrom()` forwards the host error verbatim
via `linux_errno()` with no EOF special-casing.

A `--verbose` trace shows the fork and exec themselves are fine:

```
clone(flags=0x1200011, stack=0x0, ptid=0x0, tls=0x0, ctid=0x2003f80f0)
fork-child: pid=2 ppid=1
clone: sending 25 backing fds for 43 regions
execve("/bin/true")
execve: loaded .../bin/true, entry=0xff0001c640
```

`0x1200011` is `CLONE_CHILD_SETTID | CLONE_CHILD_CLEARTID | SIGCHLD`, i.e. a
plain fork, and the child's `execve` completes successfully. So the failure is
purely in the parent's read of the status channel afterwards.

## What has been ruled out

These were each tested and are *not* the cause:

- **Not a general socketpair EOF divergence.** A C program using the same shape
  as Rust's spawn — `socketpair(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC)`, fork,
  child `execv`s, parent `recv()`s — correctly returns 0 (EOF) under elfuse.
- **Not a pipe problem.** The equivalent `pipe2(O_CLOEXEC)` + fork + exec +
  `read()` also returns 0 correctly, both single-threaded and from 8 concurrent
  worker threads.
- **Not multithreading.** The panic occurs on `thread 'main'`.
- **Not `O_CLOEXEC` bookkeeping.** `sys_pipe2()` records the flag in the fd
  table as expected.

So plain, hand-written versions of the pattern work; something specific to how
Rust's std sets up or reads this channel does not. I was not able to pin down
that difference, and would rather report the reproducible failure than guess at
a mechanism the C controls above already contradict.

## Impact

Every Rust program in the guest that spawns a subprocess aborts. On a
distribution that ships Rust coreutils this reaches ordinary shell usage, and
it takes down GUI applications indirectly through Rust image-loader helpers.
The failure is an abort rather than an error return, so callers cannot recover.

## Suggested direction

Worth checking whether `sys_recvfrom()` should translate `ECONNRESET` on a
connected stream socket into a 0-length (EOF) return, as Linux does when a peer
closes cleanly. That matches the documented divergence between the two
platforms on `SOCK_STREAM` receive after peer close. The C control above shows
the simple case already behaves correctly, so any fix should be validated
against the Rust reproducer rather than a hand-written equivalent.

jserv:
macOS has no AF_UNIX SOCK_SEQPACKET, so sys_socketpair substitutes SOCK_DGRAM. A macOS AF_UNIX datagram socket reports ECONNRESET (errno 54 -> Linux 104) on recv/read once the peer closes, whereas Linux never surfaces ECONNRESET on an AF_UNIX datagram receive -- SEQPACKET returns a clean EOF (0). Rust std uses an AF_UNIX SOCK_SEQPACKET socketpair as its exec CLOEXEC status channel and treats a recv error there as unreachable, so every Rust program that spawns a subprocess aborted with SIGABRT ("the CLOEXEC pipe failed: ... ConnectionReset"). On Ubuntu 25.10, whose coreutils are the Rust uutils build, this reached ordinary shell usage.

recv_eof_or_errno() folds the substitute's peer-close ECONNRESET into a clean EOF. It scopes the translation by probing the pinned host fd (getsockopt SO_TYPE + getsockname) and firing only for AF_UNIX SOCK_DGRAM, so a genuine reset on the AF_UNIX SOCK_STREAM substitute (socket()/accept SEQPACKET) or an INET UDP ICMP-unreachable ECONNRESET is left untouched. Probing the pinned host fd on the rare error tail is race-free against guest fd reuse and off the socket hot path. It is applied at every receive site: sys_recvfrom, both sys_recvmsg paths, sys_recvmmsg, sys_read, sys_readv, and the vCPU-loop fast-path read. The recvfrom/recvmsg/recvmmsg EOF paths also write back the empty addrlen/msg_namelen/msg_controllen/msg_flags a real 0-length receive would report.

Close #260

---
## Summary by cubic
Treat AF_UNIX SEQPACKET-over-DGRAM peer close as EOF on macOS to match Linux, fixing Rust subprocess aborts caused by ECONNRESET. Applies to all recv/read paths, clears address/control metadata, and stops recvmmsg(vlen>1) at EOF to avoid hangs.

- **Bug Fixes**
  - Added `recv_eof_or_errno()` to fold host ECONNRESET into EOF only for AF_UNIX `SOCK_DGRAM` when the guest cached type is `SOCK_SEQPACKET` (via `SO_TYPE` + `getsockname`), preserving errno across probes.
  - Wired into `sys_recvfrom`, both `sys_recvmsg` paths, `sys_recvmmsg` (records one empty message, then ends the batch), `sys_read`, `sys_readv`, and the vCPU fast-path read.
  - Preserves real resets for AF_UNIX `SOCK_STREAM`, genuine AF_UNIX `SOCK_DGRAM`, and INET UDP (ICMP unreachable).
  - On EOF, writes back empty `addrlen`/`msg_namelen`/`msg_controllen`/`msg_flags` to match Linux.
  - Added tests: EOF across `recv`/`read`/`readv`/`recvfrom`/`recvmsg`/`recvmmsg`, no fold for AF_UNIX `SOCK_DGRAM`, and `recvmmsg(vlen>1)` does not hang.

<sup>Written for commit 1e4274326cfc9455eebe39efd629ec9c7fe2455e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/262?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/io.c
+++ b/src/syscall/io.c
@@ -46,6 +46,7 @@
 #include "syscall/io.h"
 #include "syscall/net.h"
 #include "syscall/net-identity.h"
+#include "syscall/net-sockopt.h"
 #include "syscall/poll.h"
 #include "syscall/proc.h"
 #include "syscall/signal.h"
@@ -468,6 +469,7 @@ static int64_t rosetta_vz_ioctl(guest_t *g, uint64_t request, uint64_t arg)
         static const char fake_sock_path[] = ROSETTAD_SOCKET_PATH;
         memcpy(&caps[ROSETTA_CAPS_SOCKET_PATH], fake_sock_path,
                sizeof(fake_sock_path));
+
         /* Snapshot the caps binary path under the rosetta path lock so a
          * concurrent execve cannot tear the string between length probe and
          * copy. Inline buffer matches the cap exactly; the snapshot helper
@@ -615,6 +617,7 @@ static uint32_t mac_oflag_to_linux(tcflag_t mf)
 #define LINUX_PARODD 0x0200
 #define LINUX_HUPCL 0x0400
 #define LINUX_CLOCAL 0x0800
+
 /* LINUX_CBAUD 0x0000100f and LINUX_CBAUDEX 0x00001000 encode baud in c_cflag;
  * macOS uses dedicated speed fields, so termios translation ignores CBAUD on
  * translation. TCGETS2/TCSETS2 use BOTHER to signal numeric c_ispeed/c_ospeed.
@@ -650,6 +653,7 @@ static speed_t linux_cbaud_to_speed(uint32_t cbaud)
 static tcflag_t linux_cflag_to_mac(uint32_t lf)
 {
     tcflag_t mf = 0;
+
     /* CSIZE: Linux CS5=0x00, CS6=0x10, CS7=0x20, CS8=0x30
      *        macOS CS5=0x00, CS6=0x100, CS7=0x200, CS8=0x300
      */
@@ -1088,8 +1092,9 @@ int64_t sys_read(guest_t *g, int fd, uint64_t buf_gva, uint64_t count)
     }
 
     ssize_t ret = read(host_ref.fd, buf, count);
+    int64_t result = ret < 0 ? recv_eof_or_errno(host_ref.fd, fd) : ret;
     host_fd_ref_close(&host_ref);
-    return ret < 0 ? linux_errno() : ret;
+    return result;
 }
 
 int64_t sys_pread64(guest_t *g,
@@ -1212,6 +1217,7 @@ static int64_t build_host_iov(guest_t *g,
             free(heap);
             return -LINUX_EFAULT;
         }
+
         /* Cap to contiguous permitted bytes. When the guest iov entry spans a
          * non-contiguous boundary (different mapping or permission), zero every
          * subsequent host iov length so the host readv/writev returns a
@@ -1386,6 +1392,7 @@ int64_t sys_readv(guest_t *g, int fd, uint64_t iov_gva, int iovcnt)
             return err;
         int64_t ret = urandom_fill_iov(fd, host_iov.iov, iovcnt);
         host_iov_free(&host_iov);
+
         /* Mirror sys_read's slow-path refill so a readv consumer that drains
          * the shim ring leaves it ready for the next call, instead of forcing
          * every subsequent EL1 fast-path attempt back through HVC until some
@@ -1398,6 +1405,7 @@ int64_t sys_readv(guest_t *g, int fd, uint64_t iov_gva, int iovcnt)
         type == FD_INOTIFY) {
         if (iovcnt <= 0)
             return -LINUX_EINVAL;
+
         /* Use guest_read for the iov array since guest_ptr alone is unsafe if
          * the array spans a 2MiB block boundary.
          */
@@ -1444,7 +1452,7 @@ int64_t sys_readv(guest_t *g, int fd, uint64_t iov_gva, int iovcnt)
     }
 
     ssize_t ret = readv(host_ref.fd, host_iov.iov, iovcnt);
-    int64_t result = ret < 0 ? linux_errno() : ret;
+    int64_t result = ret < 0 ? recv_eof_or_errno(host_ref.fd, fd) : ret;
     host_iov_free(&host_iov);
     host_fd_ref_close(&host_ref);
     return result;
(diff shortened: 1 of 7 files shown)


doanbaotrung:
Reviewed and tested `d109016` on macOS arm64 (Apple silicon). Looks correct — no blocking issues, approve from me.

## Verified

- **Rust `process::spawn` reproducer** (Rust-uutils `timeout` from an Ubuntu 25.10 arm64 sysroot): 10/10 pass, was 0/10 before the change.
- **`tests/test-socket`**: 18/18 pass, including the new 16 / 17 / 18.
- **`tests/test-matrix.sh elfuse-aarch64`**: 243 passed / 1 failed. The one failure is `test-hello`, which I can't build locally (no `aarch64-none-elf-as`) — unrelated to this change.
- **The fold doesn't mask real errors**: a failed exec still gives `rc=127` + "No such file or directory", child exit status still propagates (`rc=1`), and a real timeout still gives `rc=124`.
- **End-to-end**: a D-Bus-activated GTK application in the same sysroot now gets through its Rust-based subprocess spawns and completes full toolkit startup, which it never did before this change. It later fails for an unrelated reason on my side, but the `ECONNRESET` abort is gone.

## On the design

The three-part predicate in `recv_eof_or_errno()` looks tight:

- `ECONNRESET` short-circuits first, so the probes stay off the hot path.
- `errno` is saved and restored around them.
- Test 17 pins down that genuine `AF_UNIX` datagram sockets still surface `ECONNRESET` rather than being folded.

Coverage across `read` / `readv` / `recv` / `recvfrom` / `recvmsg` / `recvmmsg` plus the `syscall_dispatch` fast path looks complete, and the fast path correctly probes before releasing the pinned host fd.

I also checked whether `EXPECTED_BASELINES` in `test-matrix.sh` needed bumping — it doesn't, 238 is a floor and the run clears it.

PR #266, opened 2026-08-03
Fix Unix98 pty allocation under a sysroot

### LINKED ISSUE #263: `posix_openpt()` fails: `statfs()` does not identify devpts, and `f_type` is not translated to Linux magics
# `posix_openpt()` fails: `statfs()` does not identify devpts, and `f_type` is not translated to Linux magics

## Summary

Opening `/dev/ptmx` works, but glibc's `posix_openpt()` still fails with `ENOENT`.

Before returning the master it just opened, glibc verifies that devpts is actually mounted. Under elfuse that check cannot pass, so glibc closes the working master and reports failure. The net effect is that **Unix98 PTY allocation does not work for any glibc program**.

## Root cause

glibc, `sysdeps/unix/sysv/linux/getpt.c`:

```c
fd = open("/dev/ptmx", oflag);
if (fd != -1) {
    struct statfs fsbuf;
    if ((statfs("/dev/pts", &fsbuf) == 0 && fsbuf.f_type == DEVPTS_SUPER_MAGIC)
     || (statfs("/dev",     &fsbuf) == 0 && fsbuf.f_type == DEVFS_SUPER_MAGIC))
        return fd;          /* devpts confirmed */
    close(fd);              /* <-- elfuse lands here */
    ...
}
```

Two things prevent that from succeeding:

1. `statfs("/dev/pts")` returns `ENOENT`. `/dev/pts/N` is virtualized for `open()` (`src/runtime/procemu.c`), but the path is not resolvable by `statfs`.
2. `statfs()` does not translate `f_type`. `src/syscall/fs-stat.c:152` assigns it straight across:

   ```c
   lin->f_type = mac->f_type;
   ```

   macOS `f_type` is a small filesystem-type index, whereas Linux `f_type` is a superblock magic. Observed values are `0x13` for `/dev` and `0x1a` for `/tmp`, which match nothing on Linux.

Translation is already done elsewhere in the same file — `0x9fa0` (`PROC_SUPER_MAGIC`) at line 375 and `0x01021994` (`TMPFS_MAGIC`) at line 454 — so this looks like devpts simply not being covered by the existing pattern.

## Reproducer

```c
/* aarch64-linux-gnu-gcc -static -o repro-devpts repro-devpts.c */
#define _GNU_SOURCE
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/vfs.h>
#include <unistd.h>

static void probe(const char *path)
{
    struct statfs b;
    if (statfs(path, &b) != 0) {
        printf("  statfs(%-9s) FAILED errno=%d (%s)\n", path, errno,
               strerror(errno));
        return;
    }
    printf("  statfs(%-9s) ok, f_type=0x%lx\n", path,
           (unsigned long) b.f_type);
}

int main(void)
{
    int fd = open("/dev/ptmx", O_RDWR | O_NOCTTY);
    printf("open(\"/dev/ptmx\") = %d%s\n", fd,
           fd < 0 ? strerror(errno) : "  <- works");
    if (fd >= 0)
        close(fd);

    printf("what glibc's posix_openpt() checks:\n");
    probe("/dev/pts");
    probe("/dev");
    printf("  needs /dev/pts f_type==0x1cd1, or /dev f_type==0x1373\n");

    int pt = posix_openpt(O_RDWR | O_NOCTTY);
    printf("posix_openpt() = %d%s\n", pt,
           pt < 0 ? strerror(errno) : "  <- ok");
    return pt < 0;
}
```

### Actual

```
open("/dev/ptmx") = 3  <- works
what glibc's posix_openpt() checks:
  statfs(/dev/pts ) FAILED errno=2 (No such file or directory)
  statfs(/dev     ) ok, f_type=0x13
  needs /dev/pts f_type==0x1cd1, or /dev f_type==0x1373
posix_openpt() = -1No such file or directory
```

### Expected

`posix_openpt()` returns a valid fd, as it does on Linux.

## Syscall trace

`--verbose` shows the master being opened and then immediately discarded:

```
syscall 56@... (openat "/dev/ptmx")   -> 5      # master opened
syscall 43@... (statfs)               -> -2     # ENOENT on /dev/pts
syscall 43@... (statfs)               -> 0      # /dev, wrong f_type
syscall 57@... (close 5)              -> 0      # glibc gives the master back
```

## Impact

Any glibc program that allocates a PTY through the Unix98 API fails: VTE-based terminal emulators, `script(1)`, `expect`, `tmux`, `screen`, `sshd`, and anything using `openpty()`/`forkpty()`.

A VTE-based terminal emulator surfaces it as:

```
Failed to open PTY: Inappropriate ioctl for device
```

## Notes

- **No workaround from the filesystem side.** Creating a real `/dev/pts` directory in the sysroot makes `statfs` succeed but it still reports the host `f_type` (`0x1a`), so `posix_openpt()` still fails. The fix has to be in the `statfs` translation.
- Programs that call `open("/dev/ptmx")` directly are unaffected — only the glibc wrapper's validation step fails.
- musl's `posix_openpt()` does not perform this check, so musl-linked programs are likely unaffected. Not verified.
- Reproduced on `d109016` (branch `recvfrom`), macOS arm64. `src/syscall/fs-stat.c` is untouched by that branch, so this is pre-existing rather than a regression.

doanbaotrung:
glibc's posix_openpt() opens /dev/ptmx and then confirms devpts is mounted before handing the master back. statfs did not identify /dev/pts, so glibc closed a working master and every glibc pty consumer failed -- terminal emulators, script(1), expect, tmux, screen, sshd, anything using openpty() or forkpty().

Report devpts from statfs and fstatfs, following the per-filesystem pattern already used for proc and tmpfs.

That exposed two more links in the same chain. grantpt(3) chmods the slave when it thinks the mode is wrong, and /dev/pts/N is virtualized for open and stat but not for chmod, so the call fell through to the host and failed with ENOENT. glibc then fell back to the pt_chown helper, which no modern distro ships. Intercept chmod and chown on a live slave instead, where the mode and owner come from the pty layer.

The synthesized slave stat also reported the host file owner rather than the guest uid, which is why grantpt wanted to chown at all. Report the opening uid and the tty group, matching what devpts gives.

tests/test-devpts.c covers the statfs magic, posix_openpt, fstatfs agreement, the grantpt sequence, a master/slave round trip and a negative control that an ordinary path is not devpts. The round trip runs in raw mode under an alarm so a stuck pty fails the test rather than hanging the run.

Fix #263

---
## Summary by cubic
Fixes Unix98 pty allocation under a sysroot by identifying `devpts` via `statfs`, enforcing canonical `/dev/pts/N` names, and handling `chmod`/`chown` on virtual pty slaves. glibc tools using `posix_openpt()`/`openpty()` (tmux, screen, sshd, etc.) work again.

- **Bug Fixes**
  - `statfs` on `/dev/pts` and live, canonical `/dev/pts/N` now returns `DEVPTS_SUPER_MAGIC`; `/dev/ptmx` is not spoofed; no `fstatfs` spoof for master fds.
  - Strict slave path parsing: unallocated or non-canonical names (leading zeros, sign/space, junk) return `ENOENT`, matching Linux.
  - Virtual `/dev/pts/N` `stat`: uid is the opener’s uid, gid `tty` (5), mode `0620`; `fchmodat` is accepted (no-op); `fchownat` accepts no-op only and rejects ownership changes with `EPERM`.

- **Tests**
  - Added `tests/test-devpts.c` covering `statfs` for `devpts`, the `grantpt` sequence, canonical path checks and negative controls, a master/slave round trip with timeout, and `chmod`/`chown` semantics; wired into `tests/test-matrix.sh`.

<sup>Written for commit f26e57fa13a537f46391dd4b74337248ffa4d2e4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/266?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/runtime/procemu.c
+++ b/src/runtime/procemu.c
@@ -1589,6 +1589,45 @@ static void proc_task_collect_cb(thread_entry_t *t, void *arg)
  */
 #define PTY_KEEPALIVE_MAX 256
 #define PTY_KEEPALIVE_FREE (-1)
+/* Group that owns pty slaves. Linux distributions mount devpts with gid=5
+ * ("tty") and glibc's grantpt(3) looks that group up before deciding whether
+ * the slave needs chowning.
+ */
+#define PTY_SLAVE_TTY_GID 5u
+
+/* Parse the N out of "/dev/pts/N". Returns false for the directory itself, a
+ * missing or non-numeric tail, or trailing garbage.
+ *
+ * Deliberately stricter than strtoul, which would take leading whitespace, a
+ * "+" sign and leading zeros. devpts dentries are decimal and canonical, so
+ * Linux answers ENOENT for "/dev/pts/016" even while slave 16 is open, and
+ * accepting the alias here would let one live slave answer under many names --
+ * for its stat, its statfs identity, and for whether chmod and chown are
+ * intercepted at all.
+ */
+static bool pty_slave_num_from_path(const char *path, uint32_t *out)
+{
+    if (!path || strncmp(path, "/dev/pts/", 9) != 0)
+        return false;
+    const char *digits = path + 9;
+    if (!*digits)
+        return false;
+    /* "0" is the only name that may start with a zero. */
+    if (digits[0] == '0' && digits[1] != '\0')
+        return false;
+
+    unsigned long n = 0;
+    for (const char *d = digits; *d; d++) {
+        if (*d < '0' || *d > '9')
+            return false;
+        if (n > (UINT32_MAX - (unsigned long) (*d - '0')) / 10)
+            return false;
+        n = n * 10 + (unsigned long) (*d - '0');
+    }
+    if (out)
+        *out = (uint32_t) n;
+    return true;
+}
 /* PTY_SLAVE_PATH_MAX lives in procemu.h so this table and the fork-IPC payload
  * (proc_pty_ipc_entry_t) cannot drift apart.
  */
@@ -1952,6 +1991,16 @@ static int pty_lookup_slave_path(uint32_t linux_pts_num,
     memcpy(out, pty_keepalive_table[hit].slave_path, len + 1);
     pthread_mutex_unlock(&pty_keepalive_lock);
     return 0;
+}
+
+bool proc_pty_slave_stat(const char *path, struct stat *out)
+{
+    if (!path || strncmp(path, "/dev/pts/", 9) != 0 || !path[9])
+        return false;
+    struct stat st;
+    if (proc_intercept_stat(path, out ? out : &st) != 0)
+        return false;
+    return true;
 }
 
 static int pty_open_slave(uint32_t linux_pts_num, int linux_flags)
@@ -2816,14 +2865,8 @@ int proc_intercept_open(const guest_t *g,
      * Linux devpts behavior for an unallocated slave number.
      */
     if (!strncmp(path, "/dev/pts/", 9)) {
-        const char *digits = path + 9;
-        if (!*digits) {
-            errno = ENOENT;
-            return -1;
-        }
-        char *endp;
-        unsigned long n = strtoul(digits, &endp, 10);
-        if (endp == digits || *endp != '\0' || n > UINT32_MAX) {
+        uint32_t n;
+        if (!pty_slave_num_from_path(path, &n)) {
             errno = ENOENT;
             return -1;
         }
@@ -3539,14 +3582,8 @@ int proc_intercept_stat(const char *path, struct stat *st)
         return 0;
     }
     if (!strncmp(path, "/dev/pts/", 9)) {
-        const char *digits = path + 9;
-        if (!*digits) {
-            errno = ENOENT;
-            return -1;
-        }
-        char *endp;
-        unsigned long n = strtoul(digits, &endp, 10);
-        if (endp == digits || *endp != '\0' || n > UINT32_MAX) {
+        uint32_t n;
+        if (!pty_slave_num_from_path(path, &n)) {
             errno = ENOENT;
             return -1;
         }
@@ -3567,8 +3604,15 @@ int proc_intercept_stat(const char *path, struct stat *st)
         memset(st, 0, sizeof(*st));
         st->st_mode = S_IFCHR | 0620;
         st->st_nlink = 1;
-        st->st_uid = host_st.st_uid;
-        st->st_gid = host_st.st_gid;
+        /* devpts gives the slave to whoever opened the master, group tty --
+         * that is what grantpt(3) expects to find. Reporting the host owner
+         * instead makes glibc see a foreign uid, try to chown the slave, fail,
+         * and fall back to exec'ing the pt_chown helper, which does not exist
+         * on a modern distro: grantpt then fails with ENOENT and no pty can be
+         * allocated even though the master opened fine.
+         */
+        st->st_uid = (uid_t) proc_get_uid();
+        st->st_gid = (gid_t) PTY_SLAVE_TTY_GID;
         /* macOS dev_t = (major << 24) | minor; the fs-stat translation layer
          * (mac_to_linux_dev) re-encodes that into Linux's split major/minor
          * layout, so storing 136 in the macOS-major slot makes glibc's
--- a/src/runtime/procemu.h
+++ b/src/runtime/procemu.h
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
 #include <sys/mount.h>
@@ -44,6 +45,17 @@ int proc_intercept_readlink(const char *path, char *buf, size_t bufsiz);
  * (fall through to real fstatat).
  */
 int proc_intercept_stat(const char *path, struct stat *mac_st);
+
+/* True when PATH names a live Unix98 pty slave (/dev/pts/N whose master is
+ * open), filling out with the same synthesized stat proc_intercept_stat would
+ * return. Such a path has no host backing, so metadata operations on it must be
+ * answered here rather than passed through to the host filesystem.
+ *
+ * Answering and stat'ing together keeps callers that need both from walking the
+ * locked pty table twice, which would leave a window where the slave goes away
+ * between the test and the stat. out may be NULL for a pure test.
+ */
+bool proc_pty_slave_stat(const char *path, struct stat *out);
 
 /* Intercept writes to synthetic proc files that need stateful behavior.
  * Returns 1 if handled (with *written_out set), 0 if not intercepted, or -1 on
--- a/src/syscall/fs-stat.c
+++ b/src/syscall/fs-stat.c
@@ -369,6 +369,67 @@ static bool statfs_path_is_proc(const char *path)
     return !strncmp(path, "/proc", 5) && (path[5] == '\0' || path[5] == '/');
 }
 
+/* /dev/pts itself and the pty slaves under it. /dev/ptmx is the multiplexer
+ * that hands out those slaves; Linux reports devpts for a master fd too.
+ */
+/* How statfs should answer for a path under the virtual devpts mount. */
+typedef enum {
+    DEVPTS_UNRELATED = 0, /* not under /dev/pts; carry on */
+    DEVPTS_MOUNT,         /* the mount point or a live slave: synthesize */
+    DEVPTS_ABSENT,        /* under /dev/pts but no such slave: ENOENT */
+} devpts_class_t;
+
+static devpts_class_t statfs_devpts_class(const char *path)
+{
+    if (!path || strncmp(path, "/dev/pts", 8) != 0)
+        return DEVPTS_UNRELATED;
+    if (path[8] != '\0' && path[8] != '/')
+        return DEVPTS_UNRELATED; /* "/dev/ptsfoo" is an ordinary name */
+
+    const char *tail = path + 8;
+    while (*tail == '/')
+        tail++;
+    if (!*tail)
+        return DEVPTS_MOUNT; /* "/dev/pts", "/dev/pts/", "/dev/pts//" */
+
+    /* The mount point exists for as long as the pty layer does. A particular
+     * slave does not: Linux answers ENOENT for an unallocated or malformed
+     * /dev/pts/N. Report that rather than falling through to the host, so the
+     * answer cannot depend on whether the sysroot happens to carry a file of
+     * the same name -- the devpts mount shadows whatever is underneath it, and
+     * the stat and open intercepts already treat the directory that way.
+     *
+     * Only the canonical spelling of a slave resolves, matching those
+     * intercepts: "/dev/pts/0" is the dentry, "/dev/pts/00" and "/dev/pts/./0"
+     * are not. Non-canonical spellings land here as ENOENT rather than
+     * resolving, which diverges from a real kernel for the "." form and is
+     * consistent across every /dev/pts intercept.
+     *
+     * /dev/ptmx is deliberately not claimed. Which filesystem backs it depends
+     * on whether it resolves to /dev/pts/ptmx or to the devtmpfs node, the same
+     * ambiguity that keeps sys_fstatfs from answering for a master fd, and
+     * nothing asks: glibc's getpt statfs's /dev/pts and /dev, never /dev/ptmx.
+     */
+    return proc_pty_slave_stat(path, NULL) ? DEVPTS_MOUNT : DEVPTS_ABSENT;
+}
+
+/* devpts is virtualized rather than host-backed, so answer synthetically.
+ * Matches what Linux reports for a devpts mount: no blocks, no inodes.
+ */
+static void fill_devpts_statfs(linux_statfs_t *lin)
+{
+    memset(lin, 0, sizeof(*lin));
+    lin->f_type = 0x1cd1; /* DEVPTS_SUPER_MAGIC */
+    lin->f_bsize = 4096;
+    lin->f_blocks = 0;
+    lin->f_bfree = 0;
+    lin->f_bavail = 0;
+    lin->f_files = 0;
+    lin->f_ffree = 0;
+    lin->f_namelen = 255;
+    lin->f_frsize = 4096;
+}
+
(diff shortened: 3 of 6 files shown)


src/syscall/fs-stat.c
```diff
@@ -480,6 +529,11 @@ int64_t sys_statfs(guest_t *g, uint64_t path_gva, uint64_t buf_gva)
 
 int64_t sys_fstatfs(guest_t *g, int fd, uint64_t buf_gva)
 {
+    /* Deliberately no devpts case for a pty master fd: Linux answers from
```
jserv, line 561: Issue #263 reports two bugs and this fixes one of them. `f_type` is still passed straight through from the macOS `struct statfs`, so the guest sees macOS mount-type indices rather than Linux superblock magics. Measured on this branch:

```
fstatfs(master) f_type=0x13
fstatfs(slave)  f_type=0x13
statfs("/")     f_type=0x1a
```

None of those are Linux magics. That is the second half of what #263 describes (`lin->f_type = mac->f_type`). Worth saying in the PR body whether it is deliberately out of scope, because as it stands the issue is only half closed and the untranslated value is now load-bearing for test 7 in the new test file.

tests/test-devpts.c
```diff
@@ -0,0 +1,240 @@
...
+    if (master < 0) {
+        printf("FAIL (%m)\n");
+        failures++;
+        /* Everything below needs the master; report what we have. */
+        printf("test-devpts: %s\n", failures ? "FAILED" : "all tests passed");
+        return failures != 0;
+    }
+    printf("PASS (fd=%d)\n", master);
+
+    /* No assertion on fstatfs(master): Linux answers from whatever filesystem
```
jserv, line 66: This rationale is about the master fd, and it is sound there. It does not cover the slave fd, which is unambiguous: on Linux `fstatfs` on an open `/dev/pts/N` reports devpts, because the slave really does live on that mount. Measured here it reports `0x13`, the macOS type index.

That gap has a real caller. glibc up to 2.32 has `grantpt` short-circuit on a devpts `fstatfs`, and other code sniffs the fs of a tty the same way. Either handle the slave fd in `sys_fstatfs` (the fd already carries the pty identity) or extend this comment to say the slave case is knowingly left alone.

src/syscall/fs-stat.c
```diff
@@ -369,6 +369,41 @@ static bool statfs_path_is_proc(const char *path)
...
+static bool statfs_path_is_devpts(const char *path)
+{
+    if (!path || path[0] != '/')
+        return false;
+    /* The multiplexer and the mount point exist for as long as the pty layer
+     * does. A particular slave does not: Linux answers ENOENT for an
+     * unallocated or malformed /dev/pts/N, so only a live slave may claim
+     * devpts here -- everything else falls through and misses on the host.
+     */
+    if (!strcmp(path, "/dev/ptmx") || !strcmp(path, "/dev/pts") ||
```
jserv: Claiming devpts for `/dev/ptmx` here contradicts the reasoning this same PR gives at `sys_fstatfs`: that the filesystem behind `/dev/ptmx` is devpts only when it resolves to `/dev/pts/ptmx`, and devtmpfs or tmpfs otherwise. If there is no single correct value for the fd, there is none for the pathname either.

It is also not needed. glibc's `getpt.c` statfs's `/dev/pts` and `/dev`, never `/dev/ptmx`. Suggest dropping `/dev/ptmx` from this predicate, or, if something does need it, say what and drop the fstatfs rationale instead so the two agree.

src/runtime/procemu.c
```diff
@@ -1589,6 +1589,30 @@ static void proc_task_collect_cb(thread_entry_t *t, void *arg)
...
+ */
+static bool pty_slave_num_from_path(const char *path, uint32_t *out)
+{
+    if (!path || strncmp(path, "/dev/pts/", 9) != 0)
+        return false;
+    const char *digits = path + 9;
+    if (!*digits)
+        return false;
+    char *endp;
+    unsigned long n = strtoul(digits, &endp, 10);
```
jserv: `strtoul` is too permissive for a devpts dentry name. It skips leading whitespace, accepts a `+` sign, and accepts leading zeros, so with slave 16 live:

```
statfs("/dev/pts/016") -> 0x1cd1
statfs("/dev/pts/+16") -> 0x1cd1
statfs("/dev/pts/ 16") -> 0x1cd1
open("/dev/pts/016")   -> 4  (a working slave fd)
```

Linux answers ENOENT for all three. The open and stat laxness is pre-existing, this PR only moved it, but it now also decides devpts identity and whether chmod and chown are intercepted, so it is worth tightening while the parser is being extracted anyway. Requiring ASCII digits only, with no leading zero unless the whole name is `"0"`, is a few lines and removes the aliasing.

A case for test 5 would fall out of it: `/dev/pts/0<live N>` currently resolves where it should not.

src/runtime/procemu.c
```diff
@@ -1954,6 +1978,15 @@ static int pty_lookup_slave_path(uint32_t linux_pts_num,
     return 0;
 }
 
+bool proc_path_is_pty_slave(const char *path)
```
jserv: This duplicates machinery that is already here. `path_might_use_stat_intercept()` in `src/syscall/path.c:91` already matches `/dev/pts`, `/dev/pts/` and `/dev/pts/N`, and `proc_intercept_stat()` already returns 0 only for a live slave (it does the same `pty_lookup_slave_path` call). `sys_faccessat` uses exactly that pair for this purpose already.

So the three new call sites can be `path_might_use_stat_intercept(p) && proc_intercept_stat(p, &st) == 0`, and this export plus the header declaration can go. `sys_fchownat` in particular calls this helper and then `proc_intercept_stat` right after, so it walks the locked pty table twice and opens a window between the two where the slave can go away.

src/syscall/fs.c
```diff
@@ -2860,6 +2874,25 @@ int64_t sys_fchownat(guest_t *g,
...
 
+    /* A pty slave has no host file to chown -- its owner comes from the pty
+     * layer, and grantpt(3) chowns before it chmods. Accept only a request
+     * that leaves the reported owner alone, which is all grantpt asks for
+     * since it chowns the slave to its own caller. Anything else would have to
+     * be remembered to be observable, and an unprivileged process cannot give
+     * a Linux pty slave away either, so refuse rather than report a success
+     * the next stat() contradicts.
+     */
+    if (proc_path_is_pty_slave(tx.intercept_path)) {
```
jserv: This branch has no known caller once the stat fix in this same PR lands. glibc's `grantpt` chowns only when `st.st_uid != __getuid()`, and reporting `proc_get_uid()` from the synthesized stat makes that false by construction. musl's `grantpt` is a no-op. Verified on this branch: `grantpt` returns 0 and never chowns.

What it does buy is a divergence. A guest running as root gets EPERM where Linux would let the chown through, which is the `login` and `sshd` case for handing a tty to a user. Suggest deleting the branch and letting the pass-through fail as it does today, or, if it stays, saying in the comment that it exists to keep a future non-matching uid honest rather than to serve `grantpt`.

src/syscall/fs.c
```diff
@@ -2709,6 +2709,20 @@ int64_t sys_fchmodat(guest_t *g,
...
+     * which is what grantpt(3) does when it decides the slave's mode needs
+     * adjusting: it would then fall back to the pt_chown helper and fail.
+     *
+     * The requested mode is accepted but not retained, so a later stat still
+     * reports the 0620 the pty layer synthesizes. grantpt only ever asks for
+     * the owner-access bits it is about to hand out, so nothing observes the
+     * difference; keeping it would need per-slave state that also has to cross
+     * the fork-IPC boundary.
+     */
+    if (proc_path_is_pty_slave(tx.intercept_path))
```
jserv: This is the branch that actually carries the fix, and nothing tests it. Test 6 covers chown only, and under musl (what the matrix runs) `grantpt` is a no-op, so neither is reached there.

It is reached on glibc, and reliably so. `grantpt` computes the mode it wants from `getgrnam("tty")`. On a macOS host that group is gid 4, while `PTY_SLAVE_TTY_GID` here is 5, so `st.st_gid == gid` is false, the wanted mode comes out 0600 against the synthesized 0620, and `grantpt` chmods every time. Remove this branch and `grantpt` fails.

Two asks. Add a chmod case to `tests/test-devpts.c` so the branch cannot be deleted silently. And consider whether the comment should mention that the gid mismatch against the host group database is what makes the chmod unconditional, since that is not obvious and it is the reason 0620 is never accepted as-is.

tests/test-matrix.sh
```diff
@@ -629,6 +629,7 @@ run_unit_tests()
 
     printf "\nSyscall coverage\n"
     test_check "$runner" "test-file-ops" "0 failed" "$bindir/test-file-ops"
+    test_check "$runner" "test-devpts" "PASS|0 failed" "$bindir/test-devpts"
```
jserv: `PASS` matches every per-step line the test prints, not just the summary, so this pattern is satisfied by step 1 alone. The rc check in `evaluate_result` is what actually catches a failure today, which makes the pattern decoration rather than a gate.

Anchoring it on the final line is the same amount of typing: `"all tests passed -- PASS"`. Every neighbouring entry uses `"0 failed"` for the same reason, so this would also stop being the one line in the block with a two-alternative pattern.

src/runtime/procemu.h
```diff
@@ -45,6 +45,12 @@ int proc_intercept_readlink(const char *path, char *buf, size_t bufsiz);
  */
 int proc_intercept_stat(const char *path, struct stat *mac_st);
 
+/* True when PATH names a live Unix98 pty slave (/dev/pts/N whose master is
+ * open). Such a path has no host backing, so metadata operations on it must be
+ * answered here rather than passed through to the host filesystem.
+ */
+bool proc_path_is_pty_slave(const char *path);
```
jserv: This is the first `bool` in the header, and the header does not include `<stdbool.h>`. It compiles today only because `core/guest.h` happens to pull it in. Add the include, or drop the declaration entirely per the duplication note on `procemu.c`.

jserv (changes requested):
The fix is real and I verified it: built both branches with the aarch64-linux-gnu toolchain (glibc 2.28, which does have the `statfs` check in `getpt.c` and the old chmod/chown `grantpt`). `tests/test-devpts` fails at steps 1 and 2 on `main` and passes all 7 steps on this branch. Locking is right, `pty_lookup_slave_path` takes `pty_keepalive_lock`. `tx.intercept_path` is the correct selector under `--sysroot`, `host_path` would be wrong. The new test is picked up by the existing `mk/config.mk` wildcard.

What follows is from a probe run under this branch:

```
statfs(/dev/pts/016) -> 0x1cd1     statfs(/dev/pts/+16) -> 0x1cd1
statfs(/dev/pts/ 16) -> 0x1cd1     open(/dev/pts/016)   -> 4 OPENED
chmod(slave,0600)=0, stat after   -> mode=0620
fstatfs(master) f_type=0x13        fstatfs(slave) f_type=0x13
fstatat(ptsdir,"16") mode=0444 rdev=0x0 NOT a chardev
```

Two notes I am not asking you to fix here. Relative access through a `/dev/pts` dirfd resolves to a 0444 scratch stub rather than the pty (last probe line), but that is pre-existing on `main` and not a regression from this PR. And the matrix runs unit tests under musl, where `grantpt` is a no-op, so the glibc path this PR exists for is not exercised in CI.

doanbaotrung:
**Strict pts parser** (`procemu.c:1598`) — ASCII digits only, no leading zero unless the name is `"0"`, with overflow folded into the loop. Your three aliases now behave: step 5 grew `/dev/pts/0N`, `/dev/pts/+N` and `/dev/pts/ N` built from the live slave, and against the pre-review tip it reproduces your probe exactly — all three resolved with `f_type=0x1cd1`.

**Duplicated machinery** — collapsed, though not quite as suggested. `path_might_use_stat_intercept() && proc_intercept_stat()` matches `/proc`, `/dev/shm`, `/dev/fuse` and sysfs too, so at `statfs_path_is_devpts` that pair would have made `statfs("/proc/self")` report devpts. Instead the export became `proc_pty_slave_stat(path, out)` — one definition, one table walk, answering and stat'ing together. That closes the `sys_fchownat` window you spotted, and since `out` may be NULL the two pure tests read as tests.

**`procemu.h` stdbool** — the declaration stayed, so `<stdbool.h>` is now included rather than borrowed from `core/guest.h`.

**`/dev/ptmx` in the statfs predicate** — dropped. You're right that it contradicts the `sys_fstatfs` reasoning and that `getpt` never asks; the comment now says why it's absent.

**chown branch** — kept, comment rewritten. It no longer claims to serve `grantpt` (the synthesized uid makes glibc's test false by construction); it says the branch exists so a request naming another owner is refused rather than silently lost, and it names the divergence — a privileged guest doing the `login`/`sshd` handoff gets EPERM where Linux would allow it. I left the behavior alone because reporting success without retaining the owner is the worse lie, and changing it would be untested behavior in a PR about `posix_openpt`.

**chmod coverage** — new step 7 chmods the slave directly and asserts the follow-up stat still reports 0620, with the gid-4-vs-5 explanation you asked for. Calling `chmod` itself rather than leaning on `grantpt` is deliberate: it covers the branch whatever libc the lane uses.

**Matrix pattern** — now `"all tests passed -- PASS"`.

My local `CROSS_COMPILE` is glibc, not musl (`strings build/test-devpts` shows GLIBC_TUNABLES), so the glibc path *is* what I've been exercising locally. And deleting the chmod branch here does not just lose coverage — step 3 fails outright with `grantpt: No such file or directory`, so the branch is load-bearing in this lane rather than merely reached.

**Left alone stated in the code:**

- `fstatfs` on the *slave* fd still reports the macOS type. Documented at the master-fd rationale in the test rather than fixed, because `sys_fstatfs` passes `f_type` through untranslated for every filesystem — patching the slave alone is a spot fix on the same mapping gap, and `grantpt` is unaffected since the fd it checks is the master.

PR #267, opened 2026-08-03
Ignore setuid from owners the guest cannot mean

### LINKED ISSUE #264: setuid-on-exec grants the host uid, which is outside the guest's uid namespace
# setuid-on-exec grants the host uid, which is outside the guest's uid namespace

## Summary

When a guest process `execve`s a setuid binary inside a `--sysroot`, elfuse sets the
new euid to the **host** file owner. On macOS the sysroot's files are owned by the
invoking macOS user (e.g. uid 502), so the guest ends up with an euid that does not
exist in its own uid namespace — neither root (0) nor the guest user (1000).

## Reproducer

```c
/* suidtest.c — aarch64-linux-gnu-gcc -static -o suidtest suidtest.c */
#define _GNU_SOURCE
#include <stdio.h>
#include <sys/stat.h>
#include <unistd.h>
    struct stat st;
    if (argc > 1 && stat(argv[1], &st) == 0)
        printf("guest stat(%s): st_uid=%u mode=%04o\n", argv[1],
               (unsigned) st.st_uid, st.st_mode & 07777);
    printf("uid=%u euid=%u\n", (unsigned) getuid(), (unsigned) geteuid());
    return 0;
}
```

```sh
cp suidtest "$SYSROOT/tmp/suidtest"
chmod u+s "$SYSROOT/tmp/suidtest"          # owner is the macOS user, e.g. uid 502

# must be a *guest* execve -- the initial program load does not take this path
./elfuse --sysroot "$SYSROOT" /bin/sh -c "/tmp/suidtest /tmp/suidtest"
```

### Actual

```
guest stat(/tmp/suidtest): st_uid=502 mode=4755
uid=1000 euid=502
```

### Expected

Something representable in the guest. On Linux a setuid binary owned by root yields
`euid=0`; here the owner is a host uid with no meaning inside the guest, so the
result is neither root nor the guest user.

## Cause

`src/syscall/exec.c` computes the new euid straight from a raw host `fstat`:

```c
bool have_exec_st = (fstat(exec_fd, &exec_st) == 0);
...
if (have_exec_st && !exec_is_script && S_ISREG(exec_st.st_mode)) {
    if (exec_st.st_mode & S_ISUID) {
        new_euid = (uint32_t) exec_st.st_uid;   /* host uid, unmapped */
    }
```

Host file ownership is never translated into the guest's uid namespace, so whatever
uid happens to own the file on the host becomes the guest's effective uid.

## Impact

- Any setuid binary in a sysroot yields a meaningless euid. Privilege checks against
  `euid == 0` fail, and ownership checks against guest uids fail too.
- The granted uid is whatever host uid owns the file — on a shared or restored
  sysroot that need not be the invoking user.
- Correctly, elfuse already ignores setuid on `#!` scripts (matching
  `fs/exec.c:bprm_fill_uid`); this is only about regular files.

## Suggested fix

Either map host ownership into the guest namespace before computing setuid, or
ignore setuid entirely when the owner is not representable in the guest. Note
`src/syscall/fs-stat.c` applies `chown_overlay_apply()` to guest-facing stat results
while `exec.c` does not consult it at all, so any mapping should be applied
consistently in both places.

doanbaotrung:
execve took the new effective ID straight from a raw host fstat. A sysroot is an ordinary tree owned by whoever unpacked it and no host IDs are mapped into the guest, so a setuid binary left the process at an ID that exists nowhere in the guest -- neither root nor its own. Privilege checks against euid 0 failed, ownership comparisons against guest IDs failed, and the ID granted followed whoever owned the tree.

Read ownership through chown_overlay_apply, the way fs-stat.c reports it, so exec and stat cannot disagree about who owns a file. Honour the bit only for root or the caller's own ID and leave the ID untouched otherwise, matching what Linux does for a setuid binary on a nosuid mount rather than failing the exec.

Root stays reachable because a guest chown is recorded in the overlay, which is how an emulated-root guest marks a helper setuid-root.

Fix #264

---
## Summary by cubic
Prevent exec from switching to unmapped host IDs by ignoring setuid/setgid owners the guest cannot mean and by checking exec permissions against guest-visible ownership for the main binary and any shebang interpreter. Set‑id only applies when physical and guest-visible owners agree and the owner is root or the caller; set‑id scripts and interpreters do not elevate; otherwise euid/egid stay unchanged.

- **Bug Fixes**
  - Judge exec permission via `chown_overlay_apply` for the main binary and any shebang interpreter so exec and stat agree.
  - Grant setuid/setgid only if physical and guest-visible IDs match and are 0 or the caller’s uid/gid; ignore others (nosuid-like). Root is not pulled down by host-owned setuid binaries; interpreter elevation is disabled.
  - Add `test-setuid-exec` and `test-setuid-exec-fakeroot`; include in `tests/test-matrix.sh` and skip on QEMU; update `elfuse-aarch64` baseline to 241.

<sup>Written for commit 52f882dba207305c3fadf7b702cf1f36702865e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/267?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/exec.c
+++ b/src/syscall/exec.c
@@ -35,6 +35,7 @@
 #include "runtime/futex.h"
 
 #include "syscall/abi.h"
+#include "syscall/chown-overlay.h"
 #include "syscall/exec.h"
 #include "syscall/fuse.h"
 #include "syscall/internal.h"
@@ -369,6 +370,45 @@ static bool exec_matches_fakeroot_target(const struct stat *st)
     return marked_st.st_dev == st->st_dev && marked_st.st_ino == st->st_ino;
 }
 
+/* Is ID something the guest can actually mean?
+ *
+ * A sysroot is an ordinary directory tree owned by whoever unpacked it, and
+ * elfuse maps no host IDs into the guest, so most files report the invoking
+ * macOS user (e.g. 501) -- an ID that exists nowhere in the guest's own
+ * /etc/passwd. Honouring setuid on such a file would leave the process at an
+ * effective ID that is neither root nor its own, failing both `euid == 0`
+ * privilege checks and ownership comparisons against guest IDs, and the
+ * granted ID would follow whoever happens to own the tree rather than anything
+ * the guest chose.
+ *
+ * Accept only root and the caller's own ID, and only when both views of
+ * ownership agree on it.
+ *
+ * physical is what the host reports; seen is what the guest's own stat reports,
+ * after the virtual chown overlay. Elevating needs both because each view alone
+ * fails in a different direction. Reading the set-id owner through the overlay
+ * would make root self-service, since the overlay takes any unprivileged
+ * guest's chown -- chown its own file to 0, set the bit, exec. Reading it only
+ * from the host would let a file the guest's own stat says belongs to someone
+ * else still elevate, because a physically root-owned file keeps elevating
+ * after the guest chowns it away.
+ *
+ * Requiring agreement grants nothing new: root still means a file the host
+ * really owns as root, and a guest chown can only ever withdraw an elevation,
+ * never create one. The caller's own ID makes the bit a no-op, as on Linux.
+ * Anything else leaves the ID untouched and the program simply runs
+ * unprivileged -- the outcome Linux gives for a setuid binary on a nosuid
+ * mount, and never an exec failure.
+ */
+static bool exec_id_may_elevate(uint32_t physical,
+                                uint32_t seen,
+                                uint32_t self_id)
+{
+    if (physical != seen)
+        return false;
+    return physical == 0 || physical == self_id;
+}
+
 static int check_exec_permission(const struct stat *st)
 {
     uint32_t uid = proc_get_euid();
@@ -549,8 +589,22 @@ int64_t sys_execve(hv_vcpu_t vcpu,
         err = linux_errno();
         goto fail;
     }
+    /* Judge access by the ownership the guest sees: fs-stat.c reports every
+     * guest stat through the virtual chown overlay, so the physical owner would
+     * refuse a file the guest's own stat says it may execute. Trusting a
+     * guest-writable overlay here -- chown_result records the intended owner on
+     * any EPERM, unchecked -- is a deliberate trade. A guest can chown a file
+     * into its own name to pass a check the physical mode refuses, and gains
+     * nothing: the host open must still succeed on the physical file.
+     *
+     * exec_st itself stays physical, and the set-id decision below takes both:
+     * it grants only where the two agree, so the overlay can withdraw an
+     * elevation but never conjure one.
+     */
+    struct stat exec_st_seen = exec_st;
+    chown_overlay_apply(&exec_st_seen);
 
-    err = check_exec_permission(&exec_st);
+    err = check_exec_permission(&exec_st_seen);
     if (err < 0) {
         goto fail;
     }
@@ -710,6 +764,7 @@ int64_t sys_execve(hv_vcpu_t vcpu,
             err = linux_errno();
             goto fail;
         }
+        chown_overlay_apply(&interp_st);
         err = check_exec_permission(&interp_st);
         if (err < 0) {
             goto fail;
@@ -721,20 +776,32 @@ int64_t sys_execve(hv_vcpu_t vcpu,
         goto fail;
     }
 
-    /* Compute setuid/setgid from the directly-executed file, matching Linux
-     * kernel behaviour (fs/exec.c bprm_fill_uid).  Scripts are
-     * deliberately excluded: the kernel ignores setuid/setgid on shebang
-     * scripts to prevent privilege escalation via interpreter manipulation.
-     * S_ISGID is only effective when the group-execute bit is also set,
-     * matching the kernel's mandatory-locking vs setgid distinction.
+    /* Compute setuid/setgid from the directly-executed file (fs/exec.c
+     * bprm_fill_uid). A set-id bit on the shebang script itself is ignored, as
+     * on Linux, so the interpreter named in the script cannot be manipulated
+     * into carrying the script's privilege. S_ISGID is only effective when the
+     * group-execute bit is also set, matching the kernel's mandatory-locking vs
+     * setgid distinction.
+     *
+     * Where this deliberately stops short of Linux: the kernel derives file
+     * creds from the file it finally executes, so a set-id *interpreter* named
+     * in a "#!" line does elevate there. exec_st is the directly-executed file
+     * and is never refreshed after the shebang loop, so it does not here.
+     * Elevating through an interpreter is unsupported, not accidentally lost.
      */
     uint32_t new_euid = proc_get_euid();
     uint32_t new_egid = proc_get_egid();
     if (have_exec_st && !exec_is_script && S_ISREG(exec_st.st_mode)) {
-        if (exec_st.st_mode & S_ISUID) {
+        if ((exec_st.st_mode & S_ISUID) &&
+            exec_id_may_elevate((uint32_t) exec_st.st_uid,
+                                (uint32_t) exec_st_seen.st_uid,
+                                proc_get_uid())) {
             new_euid = (uint32_t) exec_st.st_uid;
         }
-        if ((exec_st.st_mode & S_ISGID) && (exec_st.st_mode & S_IXGRP)) {
+        if ((exec_st.st_mode & S_ISGID) && (exec_st.st_mode & S_IXGRP) &&
+            exec_id_may_elevate((uint32_t) exec_st.st_gid,
+                                (uint32_t) exec_st_seen.st_gid,
+                                proc_get_gid())) {
             new_egid = (uint32_t) exec_st.st_gid;
         }
     }
@@ -844,6 +911,7 @@ int64_t sys_execve(hv_vcpu_t vcpu,
             err = linux_errno();
             goto fail;
         }
+        chown_overlay_apply(&interp_st);
 
         err = check_exec_permission(&interp_st);
         if (err < 0) {
--- a/tests/test-matrix.sh
+++ b/tests/test-matrix.sh
@@ -277,6 +277,8 @@ QEMU_SKIP="
     test-credentials
     test-credentials-fakeroot
     test-fakeroot-exec
+    test-setuid-exec
+    test-setuid-exec-fakeroot
     test-sched-policy
     test-rseq
     test-tier-a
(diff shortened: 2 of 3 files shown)


tests/test-setuid-exec.c
```diff
@@ -0,0 +1,231 @@
...
+        return child_main(argv[2]);
+
+    int failures = 0;
+    unsigned self_uid = (unsigned) getuid();
+    unsigned self_euid = (unsigned) geteuid();
+
+    printf("test-setuid-exec: 1. build setuid helper... ");
+    if (build_setuid_helper() != 0) {
+        printf("FAIL (could not build helper: %m)\n");
+        return 1;
```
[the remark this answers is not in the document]
doanbaotrung: already address

src/syscall/exec.c
```diff
@@ -681,10 +722,14 @@ int64_t sys_execve(hv_vcpu_t vcpu,
...
+        if ((exec_st.st_mode & S_ISUID) &&
+            exec_id_is_guest_meaningful((uint32_t) exec_st.st_uid,
+                                        proc_get_uid())) {
             new_euid = (uint32_t) exec_st.st_uid;
         }
-        if ((exec_st.st_mode & S_ISGID) && (exec_st.st_mode & S_IXGRP)) {
+        if ((exec_st.st_mode & S_ISGID) && (exec_st.st_mode & S_IXGRP) &&
+            exec_id_is_guest_meaningful((uint32_t) exec_st.st_gid,
+                                        proc_get_gid())) {
             new_egid = (uint32_t) exec_st.st_gid;
```
jserv, line 805: Three things about this block.

First, the commit message contradicts the code. Its closing paragraph says "Root stays reachable because a guest chown is recorded in the overlay, which is how an emulated-root guest marks a helper setuid-root." The set-id owner is read from the physical stat here, so an overlay-recorded owner of 0 never elevates. The comment at line 548 says exactly that, and test case 3 asserts it. Root is reachable only through a file the host really owns as root. This lands in git history, so worth fixing before merge.

Second, ignoring the bit is fail-open in the drop direction, and the commit message does not mention it. A caller already at euid 0 (`--fakeroot`, or after a legitimate setuid-root exec) that execs a setuid binary owned by the host user previously dropped to that owner and now keeps euid 0. That is what Linux does for a nosuid mount, so it is defensible, but the message frames the change only as "never an exec failure". Please state the other half. The tempting fix, widening the predicate when the caller is already root, would reintroduce what this PR fixes: a root guest would then take euid 501 from a sysroot file, which is the unmapped-host-uid outcome of #264.

Third, the `!exec_is_script` gate and the comment above it. The comment claims the current behavior matches `bprm_fill_uid`, which is half right. Linux ignores set-id on the script, but `exec_binprm` replaces `bprm->file` with the interpreter, and the final binfmt's `begin_new_exec` runs `bprm_creds_from_file` on that file, so a setuid interpreter named in a `#!` line does elevate on Linux. Either derive set-id from a physical `fstat(exec_fd)` taken after the shebang loop, or correct the comment to say the interpreter case is deliberately unsupported. Pre-existing, but this PR rewrites the block and the comment.

tests/test-matrix.sh
```diff
@@ -765,6 +765,8 @@ run_unit_tests()
     printf "\nCredential/identity emulation\n"
     test_rc "$runner" "test-credentials" 0 "$bindir/test-credentials"
     test_rc "$runner" "test-credentials-fakeroot" 0 --fakeroot "$bindir/test-credentials"
+    test_check "$runner" "test-setuid-exec" "PASS|0 failed" \
```
jserv: This pattern never adds anything. Step 1 prints `PASS (owner=...)` unconditionally, so the first alternative always matches, and `0 failed` is never printed at all: the success line is `all tests passed -- PASS` and a failing run prints `3 failed`. The only thing catching a failure here is the exit-code gate at line 487.

Use `"all tests passed"` so the regex is a real assertion.

tests/test-setuid-exec.c
```diff
@@ -0,0 +1,267 @@
...
+            failures++;
+        } else if (staged == 0) {
+            printf("SKIP (chown to %u refused: %m)\n", FOREIGN_UID);
+        } else {
+            int rc = run_helper(self_euid);
+            if (rc < 0 || rc == 127) {
+                printf("FAIL (helper did not run)\n");
+                failures++;
+            } else if (rc != 0) {
+                printf("FAIL (euid moved to the foreign owner)\n");
```
jserv: Case 2 detects something other than what it says. What makes it pass is that the physical owner (501) is foreign to guest uid 1000; the staged 4242 is virtual and never reaches the set-id path, which reads the physical stat. Running it against pre-PR `exec.c` prints `child: euid=501 expected=1000`, not 4242.

Two consequences. The message at line 193 misnames the owner it moved to, and the `SKIP` here gives up on a check that would still be meaningful: the physical host owner already stages "an owner the guest cannot mean" whether or not the chown was recorded, so this branch could run the helper anyway instead of skipping.

tests/test-setuid-exec.c
```diff
@@ -0,0 +1,267 @@
...
+}
+
+/* An unprivileged caller cannot really change a file's owner: a real kernel
+ * answers EPERM. So a chown that an unprivileged caller got away with proves
+ * the ownership is virtual, which is the state this test needs to reason about.
+ * A privileged caller cannot tell the two apart and skips those checks.
+ */
+static int caller_is_privileged(void)
+{
+    return geteuid() == 0;
```
jserv, line 162: With `euid == 0` cases 2 and 3 skip, case 4 becomes trivial, and the test still prints `all tests passed` and exits 0. The qemu lane runs as genuine root, so it is a vacuous green there.

Either add `test-setuid-exec` to `QEMU_SKIP` in `test-matrix.sh` with a rationale comment (the file is overlay-specific, in the same spirit as the existing `test-credentials` entry), or count executed checks and refuse to report success when none of the interesting ones ran.

jserv (commented):
Built the branch and ran the new test both ways. Against pre-PR `exec.c` all three substantive cases fail (`child: euid=501 expected=1000` twice, and the owner-only exec is refused with 127); after the patch all four pass. So the test is a real regression guard, not decoration. It also compiles clean under `-Wall -Wextra`, and no Makefile change is needed since `mk/config.mk:40` picks it up through `wildcard tests/*.c`.

Other things I checked and found clean: `exec_st` comes from `fstat(exec_fd, ...)` on the already-opened image, so there is no TOCTOU between the check and the load; both interpreter sites apply the overlay before checking; `chown_overlay_apply` takes its own rwlock and neither new call site holds a conflicting one; `exec_st` is never overlay-applied in place.

Two notes that are not line-anchored.

The overlay is trusted for the permission check but not for the set-id owner, and the argument for distrusting it applies to both. `chown_result` (`src/syscall/fs.c:2741`) records on EPERM with no ownership test, so a guest can chown a file it does not own into its own name and gain exec permission the physical mode denies. There is no real capability gain, since the host open gates readability first, but the comment could say in one line that the asymmetry is a deliberate stat-consistency tradeoff rather than an oversight.

The mapping gap in #264 is untouched. `src/syscall/fs-stat.c:52` passes `st_uid` through raw, so a sysroot file the guest genuinely owns still reports 501 while the guest is 1000, and the new overlay-applied check only helps files the guest explicitly chowned. That is a reasonable first step, but worth saying in the PR so #264 is not closed as fully solved.

One test suggestion beyond the inline comments: `test-matrix.sh` already runs `test-credentials` in both modes, and a `--fakeroot` lane for this test would cover the privileged-caller path that is currently unexercised.

tests/test-setuid-exec.c
```diff
@@ -0,0 +1,267 @@
...
+        return child_main(argv[2]);
+
+    int failures = 0;
+    unsigned self_uid = (unsigned) getuid();
+    unsigned self_euid = (unsigned) geteuid();
+
+    printf("test-setuid-exec: 1. build setuid helper... ");
+    if (build_setuid_helper() != 0) {
+        printf("FAIL (could not build helper: %m)\n");
+        return 1;
```
jserv, line 182: This path returns without the cleanup `unlink`, and every failure inside `build_setuid_helper` after `mkstemp` leaves the file behind in `/tmp`. Use `goto cleanup`.

doanbaotrung:
**Drop-direction fail-open** — real, and it's now the assertion the privileged lane makes rather than just prose: under `--fakeroot`, a setuid file owned by host uid 502 leaves euid at 0. Stated in the message, including why widening the predicate for root callers would reintroduce the unmapped-host-ID outcome.

**`!exec_is_script` comment** — corrected at `exec.c:715`: the script's own set-id bit is ignored as on Linux, and elevating through a set-id *interpreter* is called out as unsupported rather than accidentally dropped. I did not implement it — deriving set-id from a post-shebang `fstat` adds an elevation path, which is not this PR's business.

**Matrix regex** — `"PASS|0 failed"` matched step 1's unconditional `PASS (owner=...)`, so it asserted nothing. Now `"all tests passed"`.

**Case 2 was detecting the wrong thing** — right: the physical 502 is what the set-id path reads; 4242 is virtual and never reaches it. It no longer skips (the physical owner already stages a foreign owner whether or not the chown records), the message names both owners, and the header says which one does the work.

**Vacuous green at euid 0** — took both remedies rather than either. `test-setuid-exec` and the new `test-setuid-exec-fakeroot` are in `QEMU_SKIP` with a rationale, *and* a `checks_run` counter makes the run fail if no check exercised an exec. Combined with the `--fakeroot` lane you suggested, the privileged path is now a real check instead of two skips.

**Helper leak** — `build_setuid_helper` has one cleanup path; every failure after `mkstemp` unlinks.

**Overlay asymmetry** — the one-line justification you asked for is in the access comment: a guest can chown a file into its own name and pass a check the physical mode refuses, it gains no capability because the host open gates the physical file first, and the alternative is an execve contradicting the guest's own stat.

Verified both directions : unprivileged fails all three substantive checks (`euid=502 expected=1000` twice, exec refused with 127), `--fakeroot` fails its one (`euid=502 expected=0`). Post-patch both lanes pass.

**On #264** — agreed, and I changed the trailer to `Addresses #264` so it doesn't auto-close, since `fs-stat.c:52` still passes `st_uid` through raw and a genuinely guest-owned sysroot file still reports 502. Switch it back to `Fix #264` if you'd rather close it and track the mapping separately.

jserv (changes requested):

doanbaotrung:
Updated

PR #268, opened 2026-08-03
Let a marked exec enter fakeroot mode

### LINKED ISSUE #265: No way for a guest-initiated exec to enter fakeroot
# No way for a guest-initiated exec to enter fakeroot

## Summary

`ELFUSE_FAKEROOT` / `--fakeroot` can only be decided **before the guest starts**.
There is no way for a program running inside the guest to elevate, which makes a
`sudo`-style workflow impossible without running the entire session as root.

## Why the existing knobs don't cover it

- `ELFUSE_FAKEROOT` is read in `main()` (`src/main.c`), i.e. once per *host* process
  at startup.
- A guest `execve` replaces the guest image **inside the same host process**, so
  `main()` never runs again and the setting cannot change.
- A guest `fork` does spawn a new host process, but `src/runtime/forkipc.c` passes
  the parent's **host** `environ` plus an explicit `--fakeroot` copied from the
  parent's current state. The guest's own environment is never consulted, so
  `export ELFUSE_FAKEROOT=1` inside the guest has no effect.

Net effect: fakeroot is all-or-nothing for the lifetime of a guest tree.

## Use case

An embedder wants the guest to run unprivileged (uid 1000) by default, but needs a
`sudo`-equivalent so package managers work on demand. Today the only options are
running the whole session as root, or intercepting `sudo` on the host command line —
which only sees the top-level program, not a `sudo` typed into a shell inside the
guest.

A setuid helper is not a workaround either: it cannot reach `euid=0`, because the
host owner is not root and the chown overlay that could record root ownership is
in-memory only, so it does not survive into a later process. See the related report
on setuid-on-exec granting the unmapped host uid.

## Possible approaches

1. An opt-in env var naming an executable that enters fakeroot, e.g.
   `ELFUSE_FAKEROOT_EXEC=/path/to/helper` — when `execve` resolves to exactly that
   path, enable fakeroot for the new image. Off unless set, so no default-on
   privilege hole, and it reuses `proc_set_fakeroot_enabled()`.
2. Fix the setuid-on-exec uid mapping *and* persist the chown overlay, so a helper
   recorded as root-owned and setuid grants `euid=0` through ordinary setuid
   semantics. More faithful, but a larger change.

Approach 1 is far smaller; approach 2 needs no new configuration surface.

doanbaotrung:
Fakeroot could only be armed before the first guest image ran, via
--fakeroot or ELFUSE_FAKEROOT, so nothing a guest did afterwards
could raise privilege. A guest shell had no equivalent of sudo: the
one command that needs root forced the whole session to run as root.

Add ELFUSE_FAKEROOT_EXEC, naming a single executable whose exec
crosses into fakeroot. The decision is made on file identity, not on
the pathname: execve resolves the configured path the same way it
resolves the target, and compares st_dev/st_ino against the fstat of
the descriptor it already opened. A name compare would decide on one
path and execute another, since the guest and host spellings of a
sysroot file differ, translation collapses symlinks and "..", and a
writable parent lets the guest swap the leaf after the compare.
Identity also makes execve and execveat reach one decision, because
both arrive with the image already open.

The target is resolved per call rather than cached, so replacing the
marked executable takes effect, and because --sysroot is not
established when the environment is parsed. Anything unresolvable
fails closed. A script is not excluded the way a setuid script is:
that rule distrusts the interpreter line of a file the kernel never
vetted, while here the embedder named one file out of band.

The transition is committed alongside the setuid credential commit,
past every failure path, so an exec that never happens leaves the
caller unprivileged. It sets root ids and the process-wide gate,
mirroring proc_identity_init under --fakeroot, and reaches fork
children through the --fakeroot argv forkipc derives from that gate.
Nothing clears the gate, so this elevates the process tree from
there on, which the docs now say outright. The pair runs inside the
ATTN_BIT_CRED bracket the setuid family already uses, so sibling
getuid fast paths cannot read pre-exec ids out of the shim cache.

A malformed value is rejected at startup rather than ignored, since
a privilege boundary that fails quietly is worse than one that
refuses to start. With the variable unset, the default, no exec ever
elevates.

tests/test-fakeroot-exec.c re-execs itself as the marked path and
covers the unprivileged start, the elevated exec, root surviving a
fork into a fresh host process, another spelling of the same file
elevating too, execveat reaching the same verdict, and a different
file not elevating. The escape hatch is elfuse-only, so the qemu
lane skips it.

Fix #265

---
## Summary by cubic
Let a marked exec opt into fakeroot so a guest can start unprivileged and elevate one command, like sudo. Fixes #265.

- **New Features**
  - Added `ELFUSE_FAKEROOT_EXEC`: absolute path of the one executable whose exec enters fakeroot; matched by file identity, not by name (works with guest/host paths, symlinks, and `execveat`); unset is a no‑op; malformed values fail at startup.
  - Decision happens before shebang resolution, so a marked script elevates on its own identity, not `/bin/sh`.
  - Transition commits only on successful exec; sets uid/gid 0 and the process‑wide fakeroot gate. Elevation is never dropped (applies to the process tree); the exec’ing parent keeps its ids.
  - Updated `--help` and docs to describe `ELFUSE_NO_ROSETTA`, `ELFUSE_FAKEROOT`, and `ELFUSE_FAKEROOT_EXEC`. Added `tests/test-fakeroot-exec.c` covering unprivileged start, marked exec reaching root, root surviving a fork, elevation staying in the exec’d process, other spellings/`execveat` elevating, and a copy not elevating; skipped in the QEMU lane.

<sup>Written for commit 10767a09e2922c3159ca283fe3f003fcfe41a2da. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/268?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -314,7 +314,17 @@ int main(int argc, char **argv)
                 "  --gdb PORT              Listen for GDB Remote Serial "
                 "Protocol on PORT\n"
                 "  --gdb-stop-on-entry     Halt before the first guest "
-                "instruction\n");
+                "instruction\n"
+                "\n"
+                "Environment:\n"
+                "  ELFUSE_NO_ROSETTA=1     Same as --no-rosetta\n"
+                "  ELFUSE_FAKEROOT=1       Same as --fakeroot\n"
+                "  ELFUSE_FAKEROOT_EXEC    Absolute path of the one executable "
+                "whose exec enters fakeroot mode, letting a guest start "
+                "unprivileged and raise privilege later; matched by file "
+                "identity, and never dropped afterwards, so that image and "
+                "everything it execs or forks stay root (unset: no exec ever "
+                "elevates)\n");
             return 0;
         }
     }
@@ -416,6 +426,27 @@ int main(int argc, char **argv)
             fakeroot = true;
     }
     proc_set_fakeroot_enabled(fakeroot);
+
+    /* Opt-in sudo-style transition: name one executable whose exec enters
+     * fakeroot, so a guest can start unprivileged and raise privilege later.
+     * Nothing happens unless the embedder sets this.
+     *
+     * A malformed value is fatal, like every other rejected option in the loop
+     * above. Starting anyway would leave a guest whose marked command silently
+     * never elevates, and a privilege boundary that fails quietly is worse than
+     * one that refuses to start. Arming only records the path; whether it names
+     * an existing file is decided per exec, so a missing file is not an error
+     * here.
+     */
+    const char *fakeroot_exec_env = getenv("ELFUSE_FAKEROOT_EXEC");
+    if (fakeroot_exec_env && *fakeroot_exec_env &&
+        !proc_set_fakeroot_exec_path(fakeroot_exec_env)) {
+        log_error(
+            "ELFUSE_FAKEROOT_EXEC must be an absolute path shorter than %d "
+            "bytes",
+            LINUX_PATH_MAX);
+        return 1;
+    }
 
     /* Top-level processes establish the capacity; fork helpers normally inherit
      * it, but recheck before receiving the parent's FD table. Guest
--- a/src/syscall/exec.c
+++ b/src/syscall/exec.c
@@ -334,6 +334,41 @@ static int read_string_array(guest_t *g,
     return count;
 }
 
+/* True when st -- the fstat of the image execve actually opened -- is the file
+ * ELFUSE_FAKEROOT_EXEC names.
+ *
+ * Matching on (st_dev, st_ino) rather than on the pathname is what makes the
+ * hatch safe to hand a guest-supplied string. A name compare would decide on
+ * one path and execute another: the guest spelling and the host spelling of a
+ * sysroot file differ, translation collapses symlinks and ".." on the way to
+ * the file, and a writable parent directory lets the guest swap the leaf
+ * between the compare and the open. Identity answers "is this that file" about
+ * the descriptor already open, so every spelling that reaches the marked
+ * executable elevates and nothing else does.
+ *
+ * Resolved per call rather than cached at startup so that replacing the marked
+ * executable takes effect, and because --sysroot is not established yet when
+ * the environment is parsed. Fails closed on anything unresolvable, including a
+ * fuse-materialized image, whose private temp copy has an identity of its own.
+ */
+static bool exec_matches_fakeroot_target(const struct stat *st)
+{
+    const char *marked = proc_fakeroot_exec_path();
+    if (!marked || !st)
+        return false;
+
+    path_translation_t tx;
+    if (path_translate_at(LINUX_AT_FDCWD, marked, PATH_TR_NONE, &tx) < 0)
+        return false;
+    if (tx.fuse_path)
+        return false;
+
+    struct stat marked_st;
+    if (stat(tx.host_path, &marked_st) != 0)
+        return false;
+    return marked_st.st_dev == st->st_dev && marked_st.st_ino == st->st_ino;
+}
+
 static int check_exec_permission(const struct stat *st)
 {
     uint32_t uid = proc_get_euid();
@@ -455,12 +490,14 @@ int64_t sys_execve(hv_vcpu_t vcpu,
      * /proc filesystem.
      */
     if (!strcmp(path, "/proc/self/exe")) {
-        const char *exe = proc_get_elf_path();
-        if (!exe) {
+        /* Snapshot rather than proc_get_elf_path(): that pointer is shared
+         * mutable state, and a sibling execve republishing it would tear the
+         * string this copy -- and the fakeroot decision below -- reads.
+         */
+        if (!proc_elf_path_snapshot(path, sizeof(path))) {
             err = -LINUX_ENOENT;
             goto fail;
         }
-        str_copy_trunc(path, exe, sizeof(path));
         log_debug("execve resolved to \"%s\"", path);
     }
 
@@ -517,6 +554,19 @@ int64_t sys_execve(hv_vcpu_t vcpu,
     if (err < 0) {
         goto fail;
     }
+
+    /* Decide the fakeroot transition from the directly-executed file, before
+     * the shebang loop repoints path_host at an interpreter: a marked wrapper
+     * script has to elevate on its own identity, not on /bin/sh's.
+     *
+     * Unlike the setuid rule below, a script is not excluded. That rule exists
+     * because any file on the system can carry a setuid bit, so the kernel
+     * cannot trust the interpreter line of one it never vetted. Here the
+     * embedder named exactly one file out of band; its shebang line is as much
+     * the embedder's choice as its ELF contents would be, and a marked dynamic
+     * binary would trust guest-reachable shared libraries just the same.
+     */
+    bool enter_fakeroot = exec_matches_fakeroot_target(&exec_st);
 
     while (true) {
         char interp_start[256];
@@ -820,9 +870,42 @@ int64_t sys_execve(hv_vcpu_t vcpu,
      */
     /* Commit credentials right before the Point of No Return.
      * Saved UID/GID are refreshed from the final effective IDs.
+     *
+     * The fakeroot transition lands here, past every failure path, so an exec
+     * that never happens leaves the current image unprivileged. It mirrors what
+     * --fakeroot gives a process at startup (proc_identity_init): root IDs plus
+     * the process-wide gate, and it reaches fork children through the
+     * --fakeroot argv forkipc already derives from that gate. Nothing clears
+     * the gate afterwards, so this elevates the whole process tree from here
+     * on, not just the image being loaded.
+     *
+     * elfuse never tears sibling guest threads down at exec, so a sibling that
+     * outlives this call keeps the credentials committed here for the rest of
+     * the process lifetime -- a multithreaded guest that execs the marked
+     * binary from one thread hands root to every thread that was already
+     * running. The gate is published after the IDs because no permission check
+     * grants on the gate alone: proc-identity.c pairs it with "emu_euid == 0 ||
+     * fakeroot", and sys_getgroups and capget require both. Publishing it last
+     * can therefore only narrow the window, never open one.
+     *
+     * The ATTN_BIT_CRED bracket is the same protocol the setuid family uses in
+     * syscall.c: without it the shim's EL1 identity cache keeps answering
+     * sibling getuid fast paths with pre-exec IDs until
+     * exec_republish_shim_globals_or_die runs, well past guest_reset.
      */
-    proc_set_ids(proc_get_uid(), new_euid, new_euid, proc_get_gid(), new_egid,
-                 new_egid);
+    uint32_t new_uid = proc_get_uid();
+    uint32_t new_gid = proc_get_gid();
+    if (enter_fakeroot) {
+        new_uid = new_euid = 0;
+        new_gid = new_egid = 0;
+    }
+    shim_globals_attn_or(g, ATTN_BIT_CRED);
+    proc_set_ids(new_uid, new_euid, new_euid, new_gid, new_egid, new_egid);
+    if (enter_fakeroot)
+        proc_set_fakeroot_enabled(true);
+    shim_globals_publish_creds(g, proc_get_uid(), proc_get_euid(),
+                               proc_get_gid(), proc_get_egid());
+    shim_globals_attn_and(g, ~ATTN_BIT_CRED);
 
     if (0) {
     fail:
--- a/src/syscall/proc.h
+++ b/src/syscall/proc.h
@@ -108,6 +108,38 @@ bool proc_rosetta_active(void);
 void proc_set_fakeroot_enabled(bool enabled);
 bool proc_fakeroot_enabled(void);
 
+/* Opt-in escape hatch letting a guest-initiated exec enter fakeroot mode.
+ * Without it fakeroot can only be turned on before the first image runs
+ * (--fakeroot / ELFUSE_FAKEROOT), so a guest shell has no way to raise
+ * privilege for a single command the way sudo does on Linux.
+ *
+ * path names the one executable allowed to make that transition, as an absolute
+ * path resolved the way the guest resolves paths: under --sysroot the sysroot
+ * spelling and the host spelling of one file both name that file, because the
+ * match is on file identity rather than on the string (see
+ * proc_fakeroot_exec_path). Returns false without arming anything for NULL, an
+ * empty string, a relative path, or one that does not fit.
+ *
+ * Startup only. The stored path is read without a lock by every vCPU thread
+ * that reaches execve, so this must be called before any of them exists;
+ * calling it later is a data race.
+ */
+bool proc_set_fakeroot_exec_path(const char *path);
+
+/* The configured fakeroot exec path, or NULL when none was armed.
+ *
+ * The caller resolves and stats it to compare against the file it actually
+ * opened, so any spelling that reaches that file elevates and a spelling that
+ * reaches a different file does not. What elevates is therefore an inode, not a
+ * name: replacing the file at the configured path replaces what elevates, and
+ * moving the marked executable away disarms the hatch.
+ *
+ * The elevation this feeds is not per-command. Nothing ever clears the fakeroot
+ * gate, so the marked image, everything it execs afterwards, and everything it
+ * forks all stay root -- the same shape as a sudo child, not a one-shot.
+ */
+const char *proc_fakeroot_exec_path(void);
+
 /* Store the guest command line for /proc/self/cmdline emulation. argv is a
  * NULL-terminated array of strings.
  */
--- a/tests/test-fakeroot-exec.c
+++ b/tests/test-fakeroot-exec.c
@@ -1,0 +1,298 @@
+/*
+ * ELFUSE_FAKEROOT_EXEC opt-in privilege transition
+ *
+ * Copyright 2026 elfuse contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Fakeroot can otherwise only be armed before the first guest image runs, so a
+ * guest shell has no way to raise privilege for one command the way sudo does
+ * on Linux. ELFUSE_FAKEROOT_EXEC names a single executable whose exec enters
+ * fakeroot: exec it and the new image runs as root; exec anything else and the
+ * caller stays unprivileged.
+ *
+ * The test re-execs itself, so the matrix points ELFUSE_FAKEROOT_EXEC at this
+ * binary. It checks the initial image is unprivileged, that the marked exec
+ * lands at uid/gid 0, that root survives the fork into a fresh host process,
+ * that another spelling of the same file elevates too (the match is on file
+ * identity, not on the pathname), that execveat reaches the same decision, and
+ * that a copy of the same program at a different path does not elevate.
+ *
+ * Not covered here: a sysroot lane, where the guest and host spellings of the
+ * marked file genuinely differ; a marked path that is a shebang script; and
+ * what a sibling *thread* observes when another thread execs the marked binary,
+ * which needs a same-process test rather than the fork-based ones below.
+ */
+
+#include <errno.h>
+#include <fcntl.h>
+#include <limits.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/syscall.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#include "test-harness.h"
+
+#ifndef AT_EMPTY_PATH
+#define AT_EMPTY_PATH 0x1000
+#endif
+
+int passes = 0, fails = 0;
+
+extern char **environ;
+
+/* Child modes report through the exit status: 0 when the observed credentials
+ * match what the mode expects, 1 otherwise. The parent turns that into a check.
+ */
+#define CHILD_OK 0
+#define CHILD_BAD 1
+
+static int all_ids_are(uid_t want)
+{
+    return getuid() == want && geteuid() == want && getgid() == want &&
+           getegid() == want;
+}
+
+/* True when none of the four ids is root. The group ids matter as much as the
+ * user ids here: the contract for an unmarked exec is that credentials are
+ * untouched, not merely that uid stayed non-zero.
+ */
+static int no_id_is_root(void)
+{
+    return getuid() != 0 && geteuid() != 0 && getgid() != 0 && getegid() != 0;
+}
+
+static void report_ids(const char *what)
+{
+    fprintf(stderr, "%s: uid=%d euid=%d gid=%d egid=%d\n", what, (int) getuid(),
+            (int) geteuid(), (int) getgid(), (int) getegid());
+}
+
+/* Exec'd as the ELFUSE_FAKEROOT_EXEC path: must be root, and a fork from here
+ * must land in a host process that is still root.
+ */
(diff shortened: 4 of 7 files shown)


src/syscall/exec.c
```diff
@@ -464,6 +464,15 @@ int64_t sys_execve(hv_vcpu_t vcpu,
...
     }
 
+    /* Decide the fakeroot transition from the directly-executed file, before
+     * shebang resolution repoints path_host at an interpreter: a wrapper script
+     * must elevate on its own path, not on /bin/sh. Either namespace matches so
+     * the embedder can name the executable the way it knows it -- the guest
+     * path, or the host path it translates to under --sysroot. The compare is
+     * exact, so any other spelling of the same file simply does not elevate.
+     */
+    bool enter_fakeroot = proc_path_is_fakeroot_exec(path);
```
jserv: Accepting "either namespace" lets a guest-namespace path spoof a host-namespace configuration, and the result runs as root.

With `--sysroot $SR` and `ELFUSE_FAKEROOT_EXEC=$SR/bin/prog` (the host spelling, which `proc.h:117` explicitly invites), a guest `execve("$SR/bin/prog")` matches this raw compare, but `path_translate_at` then prefixes the sysroot again and opens `$SR$SR/bin/prog`. That is a path the guest can create for itself, and the file found there is what executes with uid/gid 0. The string that is matched is not the string that is opened.

Underneath that is the more general problem: the decision carries no file identity at all. There is no dev/ino check, no realpath, and `exec_open_image` only sets `O_NOFOLLOW` on the shm path, so a guest-controlled symlink at the final component, a guest-controlled directory component, or an unlink-and-replace in a writable parent each yield root. In a `--sysroot` setup the guest usually can write there.

`exec_fd` is already opened TOCTOU-safe a few lines below, so the fix is cheap: resolve and `stat` the configured path once at startup, record `st_dev`/`st_ino`, drop the dual-namespace match, and gate the transition on `fstat(exec_fd)` matching that pair rather than on the pathname.

src/syscall/exec.c
```diff
@@ -464,6 +464,15 @@ int64_t sys_execve(hv_vcpu_t vcpu,
         log_debug("execve resolved to \"%s\"", path);
     }
 
+    /* Decide the fakeroot transition from the directly-executed file, before
+     * shebang resolution repoints path_host at an interpreter: a wrapper script
+     * must elevate on its own path, not on /bin/sh. Either namespace matches so
+     * the embedder can name the executable the way it knows it -- the guest
+     * path, or the host path it translates to under --sysroot. The compare is
+     * exact, so any other spelling of the same file simply does not elevate.
```
jserv: `enter_fakeroot` ignores `exec_is_script`, which contradicts the rule enforced 200 lines below at `exec.c:689-692`: "Scripts are deliberately excluded: the kernel ignores setuid/setgid on shebang scripts to prevent privilege escalation via interpreter manipulation."

If the marked path is a script, the interpreter is whatever its `#!` line names, resolved through `path_translate_at` with no further check, and the whole shebang chain up to `ELF_SHEBANG_MAX_DEPTH` runs as root. The wrapper-script use case in the commit message is exactly this shape, so the exposure is not hypothetical.

Either clear `enter_fakeroot` when the shebang loop sets `exec_is_script`, matching the setuid rule it sits next to, or say in the comment why a script is trusted here when a setuid script is not.

src/syscall/exec.c
```diff
@@ -477,6 +486,12 @@ int64_t sys_execve(hv_vcpu_t vcpu,
                 goto fail;
             path_host_temp = true;
         } else {
+            /* Test tx.host_path, not the path_host that survives this block:
+             * for a fuse image that is a private materialization temp name,
+             * which no configured path could ever equal.
+             */
+            if (proc_path_is_fakeroot_exec(tx.host_path))
```
jserv: Under `--sysroot` this arm is alias-permissive, which is the opposite of what the comment three lines up promises.

`path_translate_at` runs the joined path through `realpath()` (`src/syscall/path.c:782`) whenever the final component is followed, so `tx.host_path` arrives with symlinks and `.`/`..` components already collapsed. Any spelling that resolves to the configured host path therefore elevates, including a symlink the guest plants.

Without a sysroot the opposite holds: `path_translate_at` leaves `tx.host_path` pointing at the input string (`src/syscall/path.c:171`), so this check is byte-for-byte the same compare as line 474 and adds nothing. The check is either redundant or alias-permissive depending on configuration, and never behaves as the "exact" contract describes.

src/syscall/exec.c
```diff
@@ -477,6 +486,12 @@ int64_t sys_execve(hv_vcpu_t vcpu,
                 goto fail;
             path_host_temp = true;
         } else {
+            /* Test tx.host_path, not the path_host that survives this block:
```
jserv: The comment justifies a distinction that does not exist at this location. This is the non-fuse `else` branch, where `path_host` is about to become a byte-for-byte copy of `tx.host_path`, so testing one or the other is identical here. The fuse case being described is handled by the check simply being absent from the other branch.

Suggest moving the rationale into the `tx.fuse_path` branch as a note on why no host-path check runs there, and dropping "not the path_host that survives this block".

src/syscall/exec.c
```diff
@@ -820,9 +835,28 @@ int64_t sys_execve(hv_vcpu_t vcpu,
...
+     * that never happens leaves the current image unprivileged. It mirrors what
+     * --fakeroot gives a process at startup (proc_identity_init): root IDs plus
+     * the process-wide gate, and it reaches fork children through the
+     * --fakeroot argv forkipc already derives from that gate.
+     *
+     * elfuse does not tear sibling guest threads down at exec the way Linux
+     * does, so this commit is observable to them for the rest of the call. The
+     * gate is published last for that reason: it widens nothing a sibling could
+     * not already do through the root IDs written just before it, since every
+     * fakeroot check is an "emu_euid == 0 || fakeroot" pair.
```
jserv: The load-bearing claim in this comment is false. Two fakeroot checks are ANDs, not the "`emu_euid == 0 || fakeroot`" pair asserted here:

- `src/syscall/sys.c:481` (`sys_getgroups`): `proc_fakeroot_enabled() && proc_get_euid() == 0`
- `src/syscall/syscall.c:1793` (capget): same shape

The gate-last ordering does stay safe for those, since they fail closed while the gate is still off, but for a different reason than the one written down. That matters because the comment exists to tell the next person these two lines can be ordered this way, and it will justify the reordering that is not safe.

Suggest restating the invariant as what it actually is: no check grants on the gate alone, so publishing the gate after the ids can only narrow the window. Or make the pair atomic and delete the argument.

jserv (commented):
Review of the `ELFUSE_FAKEROOT_EXEC` transition, focused on the privilege boundary.

The headline concern is that the elevation decision is a pathname string compare with no file identity behind it, and that it accepts two namespaces at once. Details are inline. Findings that anchor outside the diff are listed here.

`src/syscall/syscall.c:2143-2207` (`sc_execveat`): execve and execveat disagree about which spelling elevates, for the same file. With a dirfd or `AT_EMPTY_PATH`, `sc_execveat` resolves through `F_GETPATH` and hands `sys_execve` a canonical host path, which makes `host_path` non-NULL and skips the `if (!host_path)` block entirely. Only the raw compare runs, and only against a macOS-canonicalized host path. So a guest-path configuration never matches through that form of execveat, and a host-path configuration never matches a path reached through a symlinked component via execve. Whichever identity rule you settle on, both entry points should reach it through one resolution.

`src/syscall/exec.c:463`: the `/proc/self/exe` resolution copies out of `proc_get_elf_path()`, which `src/syscall/proc.h:71-77` documents as "safe only for boolean tests; callers that consume the string must use `proc_elf_path_snapshot`". That copied string now feeds the privilege decision at line 474, so a concurrent `proc_set_elf_path` from a sibling execve can tear it. The contract violation predates this branch; this branch is what makes it security-relevant. Suggest switching to `proc_elf_path_snapshot(path, sizeof(path))` and treating a false return as `-ENOENT`.

`tests/test-matrix.sh:1244`: the `elfuse-aarch64` baseline stays at 238 while the branch adds a passing `test_rc`, so the real count is 239. The gate is a minimum and will not fail, but the comment at line 1210 asks for the bump in the same commit that changes behavior.

`docs/usage.md:22`: the option table documents `ELFUSE_NO_ROSETTA` next to its flag, but a new privilege-boundary control is documented only in `--help`. Worth adding `ELFUSE_FAKEROOT_EXEC` (and the pre-existing, also-missing `--fakeroot`) with the matching rule and the "elevation is never dropped" caveat spelled out.

Not raised as findings, for the record: the ordering in `main.c` is right, the env parse at line 433 sits before the `--fork-child` dispatch at line 458 so fork children do re-arm the path, and `posix_spawn` passes `environ` so the variable survives the hop. The new image also gets correct credentials, since `exec_republish_shim_globals_or_die` republishes after `guest_reset`.

src/syscall/exec.c
```diff
@@ -820,9 +835,28 @@ int64_t sys_execve(hv_vcpu_t vcpu,
...
+     *
+     * The fakeroot transition lands here, past every failure path, so an exec
+     * that never happens leaves the current image unprivileged. It mirrors what
+     * --fakeroot gives a process at startup (proc_identity_init): root IDs plus
+     * the process-wide gate, and it reaches fork children through the
+     * --fakeroot argv forkipc already derives from that gate.
+     *
+     * elfuse does not tear sibling guest threads down at exec the way Linux
+     * does, so this commit is observable to them for the rest of the call. The
+     * gate is published last for that reason: it widens nothing a sibling could
```
jserv: "observable to them for the rest of the call" understates the lifetime. `sys_execve` never tears sibling threads down (there is no `thread_` or `hv_vcpus_exit` call anywhere in this file), so a sibling that survives the exec holds root for the rest of the process lifetime, not the rest of the syscall.

Worth saying plainly, along with what it means for a multithreaded guest that execs the marked binary from one thread while others keep running.

src/syscall/exec.c
```diff
@@ -820,9 +835,28 @@ int64_t sys_execve(hv_vcpu_t vcpu,
...
      */
-    proc_set_ids(proc_get_uid(), new_euid, new_euid, proc_get_gid(), new_egid,
-                 new_egid);
+    uint32_t new_uid = proc_get_uid();
+    uint32_t new_gid = proc_get_gid();
+    if (enter_fakeroot) {
+        new_uid = new_euid = 0;
+        new_gid = new_egid = 0;
+    }
+    proc_set_ids(new_uid, new_euid, new_euid, new_gid, new_egid, new_egid);
```
jserv, line 903: This credential commit does not use the `CRED_BRACKETED` protocol that `src/syscall/syscall.c:192-211` established for exactly this situation, so the shim's EL1 identity cache keeps serving pre-exec ids to sibling vCPUs until `exec_republish_shim_globals_or_die` runs, well past `guest_reset`. A sibling calling `getuid` on the inline fast path gets the old value while the SVC path returns 0.

The window skews unprivileged, so nothing is granted early, and the pre-existing setuid commit had the same gap. This branch widens it from euid-only to all six ids. Wrapping the `proc_set_ids` plus `proc_set_fakeroot_enabled` pair in the same `ATTN_BIT_CRED` bracket forces the fast path to the host for the duration.

src/syscall/exec.c
```diff
@@ -820,9 +835,28 @@ int64_t sys_execve(hv_vcpu_t vcpu,
...
+    if (enter_fakeroot)
+        proc_set_fakeroot_enabled(true);
```
jserv, line 905: The gate is a one-way latch with no clear path anywhere, which makes "raise privilege per command" not quite what the code does.

Once this fires, `uid_is_permitted` (`src/syscall/proc-identity.c:139`) returns true unconditionally, so a process that drops privilege with `setuid(1000)` can take it straight back with `setuid(0)`. Every later execve of an unmarked binary also keeps uid 0, because `new_uid = proc_get_uid()` is already 0 by then, and every fork inherits through the `--fakeroot` argv.

That may well be the intended sudo-like shape, since a real sudo child also keeps root for its descendants. The gap is that nothing says so. Decide whether the gate is per-image or per-process-tree, and if it persists, document it in `proc.h` and stop describing it as per-command.

src/syscall/proc.h
```diff
@@ -108,6 +108,28 @@ bool proc_rosetta_active(void);
...
+ * (--fakeroot / ELFUSE_FAKEROOT), so a guest shell has no way to raise
+ * privilege for a single command the way sudo does on Linux.
+ *
+ * path names the one executable allowed to make that transition, as an
+ * absolute path in either namespace: the guest path, or the host path it
+ * translates to under --sysroot. NULL, an empty string, a relative path, or one
+ * that does not fit clears it -- as does never calling this, which is the
+ * default. Nothing is elevated unless the embedder opts in.
+ */
+void proc_set_fakeroot_exec_path(const char *path);
```
jserv: The header contract omits the restriction the implementation depends on. `src/syscall/proc-identity.c:45` says "Written once during startup, before any guest code runs... No lock", but this declaration reads like an ordinary runtime API. A caller who follows the header and calls it after startup introduces an unsynchronized write to `fakeroot_exec_path` racing every vCPU thread's `strcmp`.

Worth stating here that the setter is startup-only, before any vCPU thread exists, and that calling it later is a data race.

src/syscall/proc.h
```diff
@@ -108,6 +108,28 @@ bool proc_rosetta_active(void);
...
+
+/* True when path is the configured fakeroot exec and one was configured.
+ * Always false when no path is set, so callers need no extra gate.
+ *
+ * The compare is exact. Another spelling of the same file -- relative, or
```
jserv: "The compare is exact" does not hold for the host-path arm under `--sysroot`, where `path_translate_at` hands back a `realpath()`-resolved string (see the note on `exec.c:493`). Relative spellings and `..` do fail to match as documented, but symlinks and `.` components elevate. Since this paragraph is the contract the tests are written against, it should describe the resolved-path behavior rather than the exact-string ideal.

src/main.c
```diff
@@ -417,6 +426,19 @@ int main(int argc, char **argv)
...
+    /* Opt-in sudo-style transition: name one executable whose exec enters
+     * fakeroot, so the guest can start unprivileged and raise privilege per
+     * command. Nothing happens unless the embedder sets this.
+     */
+    const char *fakeroot_exec_env = getenv("ELFUSE_FAKEROOT_EXEC");
+    if (fakeroot_exec_env && *fakeroot_exec_env) {
+        proc_set_fakeroot_exec_path(fakeroot_exec_env);
+        if (!proc_path_is_fakeroot_exec(fakeroot_exec_env))
+            log_error(
+                "ELFUSE_FAKEROOT_EXEC ignored: need an absolute path that "
```
jserv: Three things about this rejection path:

Rejection is inferred by comparing the string back rather than reported. `proc_set_fakeroot_exec_path` knows exactly why it refused; making it return `bool` and checking that is both clearer and immune to the setter ever normalizing what it stores, which would silently break this round-trip.

A rejected value is logged but not fatal. Every other invalid input in this option loop returns 1 (`main.c:352`, `371`, `394`). A mistyped `ELFUSE_FAKEROOT_EXEC` instead leaves a running guest whose marked command quietly never elevates, which is a confusing failure for a privilege-boundary setting. Worth an explicit decision either way, stated in the comment.

The message says "fits in PATH_MAX" but the bound is `LINUX_PATH_MAX`.

src/main.c
```diff
@@ -314,7 +314,16 @@ int main(int argc, char **argv)
...
                 "  --gdb-stop-on-entry     Halt before the first guest "
-                "instruction\n");
+                "instruction\n"
+                "\n"
+                "Environment:\n"
+                "  ELFUSE_NO_ROSETTA=1     Same as --no-rosetta\n"
+                "  ELFUSE_FAKEROOT=1       Same as --fakeroot\n"
+                "  ELFUSE_FAKEROOT_EXEC    Absolute path of the one executable "
+                "whose exec enters fakeroot mode, so a guest can start "
+                "unprivileged and raise privilege per command; the guest must "
```
jserv: "raise privilege per command" is not what the code does. The gate is never cleared, so the marked exec elevates that process plus everything it later execs or forks, with no way back down. Suggest saying that outright here, since `--help` is where most people will meet this feature.

tests/test-fakeroot-exec.c
```diff
@@ -0,0 +1,247 @@
...
+    printf("ELFUSE_FAKEROOT_EXEC tests (marked=%s)\n", marked);
+
+    TEST("initial image unprivileged");
+    EXPECT_TRUE(no_id_is_root(), "started as root");
+
+    TEST("exec of marked path is root");
+    EXPECT_EQ(run_child(marked, "expect-root"), CHILD_OK,
+              "marked exec did not reach root");
+
+    TEST("exec'ing parent keeps its ids");
```
jserv: This check cannot fail. `run_child` forks first, and a guest fork in elfuse is a separate host process via `posix_spawn` (`src/runtime/forkipc.c:1613`), so the exec that elevates happens in a different address space with its own copy of every identity global. The parent's ids are untouched no matter what `sys_execve` does.

It is also the check that would have exercised the sibling-observability question the `exec.c:845-849` comment is written to defend, and it does not. A same-process version would: spawn a pthread, have the main thread exec the marked path, and assert what the surviving sibling observes.

tests/test-fakeroot-exec.c
```diff
@@ -0,0 +1,247 @@
...
+    EXPECT_TRUE(no_id_is_root(), "parent gained root");
+
+    /* The match is an exact string compare, so another spelling of the very
+     * same file must not elevate. Pins the fail-closed half of that contract.
+     */
+    char spelled[512];
```
jserv: `spelled[512]` truncates a marked path longer than 508 bytes, and `LINUX_PATH_MAX` is 4096. After truncation the exec fails with ENOENT, `run_child` returns 127, and the test reports "non-exact path gained root" for a path that was never opened.

The `if (base)` guard below is dead: `proc_set_fakeroot_exec_path` rejects anything not starting with `/`, so `strrchr` always finds a slash.

Suggest sizing at `LINUX_PATH_MAX`, checking the `snprintf` return for truncation, and distinguishing an exec failure (127) from an elevation failure in the message.

Separately, this case only proves what it claims when no sysroot is configured. With one, `realpath()` collapses the `/.` and the host-path arm matches, so the assertion inverts.

tests/test-fakeroot-exec.c
```diff
@@ -0,0 +1,247 @@
...
+
+int main(int argc, char **argv)
+{
+    if (argc > 1 && !strcmp(argv[1], "expect-root"))
+        return mode_expect_root();
+    if (argc > 1 && !strcmp(argv[1], "expect-unprivileged"))
+        return mode_expect_unprivileged();
+
+    const char *marked = getenv("ELFUSE_FAKEROOT_EXEC");
+    if (!marked || !*marked) {
```
jserv, line 240: An unset `ELFUSE_FAKEROOT_EXEC` returns 0 with no assertions and no `SUMMARY`, so a harness that stops exporting the variable turns the whole file into a silent pass rather than a visible skip. `ALLOW_MISSING_BINARIES` defaults to strict for this class of failure; this reopens it inside the binary. Since the matrix always sets it, the unset case could reasonably be a `FAIL` when run from the matrix.

tests/test-fakeroot-exec.c
```diff
@@ -0,0 +1,247 @@
...
+ * Copyright 2026 elfuse contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Fakeroot can otherwise only be armed before the first guest image runs, so a
+ * guest shell has no way to raise privilege for one command the way sudo does
+ * on Linux. ELFUSE_FAKEROOT_EXEC names a single executable whose exec enters
+ * fakeroot: exec it and the new image runs as root; exec anything else and the
+ * caller stays unprivileged.
+ *
+ * The test re-execs itself, so the matrix points ELFUSE_FAKEROOT_EXEC at this
```
jserv, line 13: The header claims the test covers the contract, but the host-path half of it has never run. The matrix invokes this without `--sysroot`, and with no sysroot `path_translate_at` leaves `tx.host_path` pointing at the input string (`src/syscall/path.c:171`), so the second check at `exec.c:494` is the same compare as line 474 and has never executed as a distinct path.

Also uncovered: execveat, a marked path that is a shebang script, and an unmarked exec from an already-elevated image (the case that shows the gate never clears).

tests/test-matrix.sh
```diff
@@ -765,6 +769,11 @@ run_unit_tests()
     printf "\nCredential/identity emulation\n"
     test_rc "$runner" "test-credentials" 0 "$bindir/test-credentials"
     test_rc "$runner" "test-credentials-fakeroot" 0 --fakeroot "$bindir/test-credentials"
+    # Arm the opt-in transition on the test binary itself: it re-execs its own
+    # path to cross into fakeroot, and a copy of itself to prove the negative.
+    export ELFUSE_FAKEROOT_EXEC="$bindir/test-fakeroot-exec"
```
jserv: A sysroot lane would be worth adding here, since the host-path arm of the match is unreachable without one and it is also the configuration that reproduces the spoof described on `exec.c:474`.

Minor: the `export` is only undone after `test_rc` returns, so an interrupted run leaks it into later shell state. `env ELFUSE_FAKEROOT_EXEC=... test_rc ...` or a subshell avoids that.

doanbaotrung:
Updated as follow

### Let a marked exec enter fakeroot mode

Fakeroot could only be armed before the first guest image ran, via
--fakeroot or ELFUSE_FAKEROOT, so nothing a guest did afterwards
could raise privilege. A guest shell had no equivalent of sudo: the
one command that needs root forced the whole session to run as root.

Add ELFUSE_FAKEROOT_EXEC, naming a single executable whose exec
crosses into fakeroot. The decision is made on file identity, not on
the pathname: execve resolves the configured path the same way it
resolves the target, and compares st_dev/st_ino against the fstat of
the descriptor it already opened. A name compare would decide on one
path and execute another, since the guest and host spellings of a
sysroot file differ, translation collapses symlinks and "..", and a
writable parent lets the guest swap the leaf after the compare.
Identity also makes execve and execveat reach one decision, because
both arrive with the image already open.

The target is resolved per call rather than cached, so replacing the
marked executable takes effect, and because --sysroot is not
established when the environment is parsed. Anything unresolvable
fails closed. A script is not excluded the way a setuid script is:
that rule distrusts the interpreter line of a file the kernel never
vetted, while here the embedder named one file out of band.

The transition is committed alongside the setuid credential commit,
past every failure path, so an exec that never happens leaves the
caller unprivileged. It sets root ids and the process-wide gate,
mirroring proc_identity_init under --fakeroot, and reaches fork
children through the --fakeroot argv forkipc derives from that gate.
Nothing clears the gate, so this elevates the process tree from
there on, which the docs now say outright. The pair runs inside the
ATTN_BIT_CRED bracket the setuid family already uses, so sibling
getuid fast paths cannot read pre-exec ids out of the shim cache.

A malformed value is rejected at startup rather than ignored, since
a privilege boundary that fails quietly is worse than one that
refuses to start. With the variable unset, the default, no exec ever
elevates.

tests/test-fakeroot-exec.c re-execs itself as the marked path and
covers the unprivileged start, the elevated exec, root surviving a
fork into a fresh host process, another spelling of the same file
elevating too, execveat reaching the same verdict, and a different
file not elevating. The escape hatch is elfuse-only, so the qemu
lane skips it.

PR #269, opened 2026-08-05
Implement TIOCPKT packet mode on pty masters

doanbaotrung:
libvte switches a pty master into packet mode immediately after posix_openpt and treats any failure as fatal, so every VTE terminal -- gnome-terminal, xfce4-terminal, tilix -- dies at startup with "Failed to open PTY: Inappropriate ioctl for device". That errno is elfuse's own: TIOCPKT had no case and fell through to the default arm, which answers ENOTTY. Everything else in the sequence already worked, so the failure read as a pty allocation problem rather than one absent request.

macOS numbers TIOCPKT differently but takes the same int flag through a pointer, and its TIOCPKT_* status bits match Linux value for value, so the packet stream needs no translation on the read path. Adopt a master received through SCM_RIGHTS before the host ioctl, the way TIOCSWINSZ does, since such an fd never passed the /dev/ptmx open intercept.

test-pty asserts the effect rather than the return code: with packet mode on, a master read is prefixed with a status byte and ordinary data carries TIOCPKT_DATA. It skips status-only packets first, since setting raw mode on the slave emits TIOCPKT_NOSTOP ahead of any data, and bounds the wait with poll so a missing packet fails the test instead of blocking it.

---
## Summary by cubic
Add support for `TIOCPKT` packet mode on PTY masters so VTE-based terminals (gnome-terminal, xfce4-terminal, tilix) start correctly. Fixes ENOTTY from ioctl(`TIOCPKT`) right after `posix_openpt`.

- **Bug Fixes**
  - Define `LINUX_TIOCPKT` and pass the int flag pointer to host `TIOCPKT`; status bits match, so no translation.
  - Adopt masters received via SCM_RIGHTS before the ioctl (like `TIOCSWINSZ`) to cover fds that skipped `/dev/ptmx`.
  - Extend `test-pty`: enable packet mode, assert status-byte prefix and `TIOCPKT_DATA`, skip control packets, reassemble split payloads, bound reads with `poll`, and verify ioctl on an adopted master.

<sup>Written for commit fc1b590d5200ad1a72961ab70d92c06edc114c17. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/269?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/test-pty.c
+++ b/tests/test-pty.c
@@ -14,6 +14,8 @@
  *   4. /dev/pts/N open + stat intercept and the slave fd's window size
  *   5. /dev/pts/N open in a forked child after the child has already closed
  *      its master (foot/sshd/openssh sftp-server pattern)
+ *   6. TIOCPKT packet mode on the master, which libvte turns on right after
+ *      posix_openpt and treats as fatal on failure
  *
  * The test stays self-contained on the syscall surface (no libutil/openpty), so
  * it runs the same way under elfuse-aarch64, qemu-aarch64, and any future
@@ -24,6 +26,8 @@
 #include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <poll.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -38,6 +42,13 @@
 #include <unistd.h>
 
 #include "test-harness.h"
+
+/* Linux spells the ordinary-data packet type this way; the
+ * value is 0 and matches macOS.
+ */
+#ifndef TIOCPKT_DATA
+#define TIOCPKT_DATA 0
+#endif
 
 #ifndef TIOCSWINSZ
 #define TIOCSWINSZ 0x5414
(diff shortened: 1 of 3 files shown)


tests/test-pty.c
```diff
@@ -581,6 +585,98 @@ int main(void)
...
+                            wrote != (ssize_t) (sizeof(payload) - 1) ||
+                            poll(&pfd, 1, 2000) <= 0)
+                            break;
+                        ssize_t got = read(pkt_master, buf, sizeof(buf));
+                        if (got <= 0)
+                            break;
+                        if (got == 1)
+                            continue; /* status-only packet */
+                        saw_data =
+                            buf[0] == 0 && got == (ssize_t) sizeof(payload) &&
```
jserv: This assumes the first data packet carries the whole payload, and on a mismatch it breaks rather than reading again, so the retry budget above never gets used for the case it exists for. A pty may legally split a 3-byte write across reads: the packet stream `\0p` then `\0kt` is valid and would fail here. tests/test-devpts.c:141 already takes the other position for the same reason.

Accumulate instead: on a data packet require `buf[0] == TIOCPKT_DATA`, append `buf + 1` through `buf + got - 1` to a scratch buffer, and keep looping until the accumulated bytes equal the payload or the poll budget runs out.

tests/test-pty.c
```diff
@@ -581,6 +585,98 @@ int main(void)
...
+                     * poll window below and then blame the read.
+                     */
+                    struct termios tio;
+                    bool raw_ok = tcgetattr(pkt_slave, &tio) == 0;
+                    if (raw_ok) {
+                        cfmakeraw(&tio);
+                        raw_ok = tcsetattr(pkt_slave, TCSANOW, &tio) == 0;
+                    }
+                    static const char payload[] = "pkt";
+                    ssize_t wrote =
```
jserv: Same short-transfer point on the write side. tests/test-devpts.c:147 loops until the payload is fully written; this takes it in one call. Loop until `sizeof(payload) - 1` bytes are out or `write` returns <= 0.

tests/test-pty.c
```diff
@@ -581,6 +585,98 @@ int main(void)
...
+                     * test rather than block it.
+                     */
+                    char buf[32];
+                    bool saw_data = false;
+                    for (int attempt = 0; !saw_data && attempt < 8; attempt++) {
+                        struct pollfd pfd = {
+                            .fd = pkt_master,
+                            .events = POLLIN,
+                        };
+                        if (!raw_ok ||
```
jserv: `raw_ok` and `wrote` are both loop invariant, so re-evaluating them per attempt only costs a reader a second look to confirm they cannot change.

More to the point, a short or failed write lands on the EXPECT_TRUE below with `raw_ok` true, so it reports "master read is not a TIOCPKT_DATA packet" and blames packet mode for a setup failure. Hoist both checks above the loop and give the write case its own message.

jserv (commented):
The ioctl side checks out. 0x5420 is the right Linux number, macOS TIOCPKT is _IOW('t', 112, int) and takes the same int through a pointer, and the TIOCPKT_* status bits are identical value for value (0/1/2/4/8/16/32/64), so the claim that the packet stream needs no translation on the read path holds. There is nothing to write back: the request consumes the flag rather than producing one. host_fd_ref_close is balanced on both exits, the -LINUX_EFAULT before the host call and linux_errno() after it match the neighboring terminal cases.

The proc_pty_master_adopt call is load-bearing rather than cosmetic, and worth keeping the comment on: src/runtime/procemu.c:1556 documents that a bare macOS master answers ENOTTY to tty ioctls until something has opened the corresponding slave once, so an SCM_RIGHTS master with no keepalive would fail in exactly the way this PR is fixing.

Everything below is on the test.

PR #278, opened 2026-08-07
Report a pty hangup on every path that can observe one

### LINKED ISSUE #274: Pty master hangup is not reported on every path
Follow-up to the pty master hangup support. The mechanism works -- a
master reports POLLHUP through poll and EIO through read once the guest
closes its last slave -- but two paths are not covered, and each leaves
a client waiting forever.

- [ ] epoll_wait does not report EPOLLHUP for a hung-up master. poll
      answers from proc_pty_master_hung_up(); epoll_wait was never
      wired to it, so any epoll-based client is unaffected by the fix.
      foot is one: it never exits when its shell does, leaving the
      window open after "exit".

- [ ] Slaves opened through TIOCGPTPEER are never registered with
      proc_pty_note_guest_slave(), which is called only from the
      /dev/pts/N open intercept. For a master whose only slave came
      that way, guest_slave_seen stays false and the hangup is never
      reported. TIOCGPTPEER has been the recommended way to get the
      peer since Linux 4.13. The docstring in src/runtime/procemu.h
      also claims both open paths report through it, which does not
      match the code.

Reproducer for the first, aarch64 guest under --sysroot:

    m = posix_openpt(O_RDWR | O_NOCTTY);
    grantpt(m); unlockpt(m);
    s = open(ptsname(m), O_RDWR | O_NOCTTY);
    close(s);

    ep = epoll_create1(0);
    epoll_ctl(ep, EPOLL_CTL_ADD, m, &(struct epoll_event){.events = EPOLLIN});
    epoll_wait(ep, &ev, 1, 2000);   /* returns 0; poll() returns POLLHUP */

The host cannot supply the event either way: elfuse holds a keepalive
slave open for the master's whole life (see the pty side-table header in
src/runtime/procemu.c), so macOS never hangs the master up.

doanbaotrung:
Report a pty hangup on every path that can observe one

epoll_wait never reported EPOLLHUP for a pty master whose slaves had all
closed, and slaves obtained through TIOCGPTPEER were never registered, so
the master's hangup accounting missed them. Wire both, and re-verify the
fd generation a registration pinned at ADD/MOD so a close-and-reuse in the
lookup window cannot charge the hangup to an unrelated file.

Drop proc_pty_any_master_hung_up(). It was meant as a cheap gate that let
epoll skip the registration sweep, but it latches: a master that hangs up
while the guest still holds its fd keeps the gate true for the process's
lifetime, and every epoll_pwait then paid for a full FD_TABLE_SIZE sweep.
Track per-instance active and pty-master counts instead, so the scan is
bounded by what the instance actually registered and skipped entirely when
it holds no pty master.

Delete kqueue events whose guest fd is no longer registered. A
level-triggered event for a deregistered fd is re-reported on every call
and translates to zero epoll events, which spins the caller.

Keep an epoll registration alive while another fd still refers to the same
open file description, matching Linux, and identify that description by
ofd_id at the close chokepoint.

Return ready events in preference to EINTR. Linux ep_poll() checks
ep_events_available() and jumps to send_events before it consults
signal_pending, so EINTR is only the answer for a wait that produced
nothing. Discarding a ready fd to report EINTR loses it in practice: kqueue
re-reports it, but the same pending signal is still there, so the caller is
handed EINTR forever and never drains the fd. A terminal with an unhandled
SIGCHLD spun at 100% CPU this way without ever drawing.

Fix #274

---
## Summary by cubic
Make pty hangups visible on every path and across processes. poll, epoll, read, and readv now report hangups reliably; epoll stamps EPOLLHUP immediately and reads return EIO after draining pending data.

- **Bug Fixes**
  - Per-pty shared counters via named shm (fresh on new host pty, join on dup/SCM_RIGHTS/fork-restore; fallback to per-process). Release this process’s remaining slaves at teardown; parent credits inherited slaves before siblings run; child adopts inherited slaves on init.
  - Track slaves opened via TIOCGPTPEER (only if the master generation matches) and mirror dup/dup2 of pty slaves so aliases are counted.
  - Pin fd generations across read/poll/epoll/ioctl to avoid close+reuse races; expose `fd_current_generation`; resolve host fd and generation atomically.
  - Epoll: stamp EPOLLHUP immediately (no waiting), merge with EPOLLIN when data is pending, bound scan using per-instance active and pty-master counts, honor EPOLLONESHOT, verify registration generation to avoid DEL/ADD races, keep registrations alive while another fd refers to the same open file description (track ofd_id), drop kqueue events for deregistered fds, and return ready events in preference to EINTR.
  - read and readv on hung-up masters return EIO after draining queued data.

<sup>Written for commit 26de80c6a4dbe712f848fc2ec3fd696d0a265f79. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/278?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -36,7 +36,8 @@
 #include "core/sysroot.h"
 
 #include "runtime/forkipc.h"
-#include "runtime/futex.h" /* futex_interrupt_request */
+#include "runtime/futex.h"   /* futex_interrupt_request */
+#include "runtime/procemu.h" /* proc_pty_release_process_slaves */
 #include "runtime/proctitle.h"
 #include "runtime/thread.h"
 
@@ -762,6 +763,14 @@ cleanup:
      * temp ELF, which the post-prepare error paths and a Rosetta guest's
      * teardown previously leaked.
      */
+    /* Give back any pty slaves this process still holds before the fd table
+     * goes away. The guest's stdio slaves are closed by the kernel, not by the
+     * guest, so they never pass through the per-fd close hook; a master in
+     * another process would otherwise wait forever for a hangup that this exit
+     * should have produced.
+     */
+    proc_pty_release_process_slaves();
+
     cleanup_main_resources(&g, guest_initialized, &sysroot_mount,
                            have_host_cwd ? host_cwd : NULL, guest_argv,
                            guest_argc, elf_path, sysroot_path);
--- a/src/runtime/forkipc.c
+++ b/src/runtime/forkipc.c
@@ -36,6 +36,7 @@
 #include "runtime/forkipc.h"
 #include "runtime/fork-state.h"
 #include "runtime/futex.h"
+#include "runtime/procemu.h"
 
 #include "syscall/linux-wire.h"
 #include "syscall/chown-overlay.h"
@@ -294,15 +295,34 @@ int fork_child_main(int ipc_fd,
         return 1;
     }
 
+    /* Both the fd table and the keepalives are in place, which is what this
+     * needs to recognize the slave fds inherited from the parent and put them
+     * back on the books. Without it the parent's close of its own copy makes
+     * the pty look hung up while this child still holds a live slave.
+     */
+    proc_pty_adopt_inherited_slaves();
+
     signal_state_t sig;
     if (fork_ipc_recv_process_state(ipc_fd, &g, &sig) < 0) {
         log_error("fork-child: failed to receive process state");
+        /* Give back the credit the parent added for this child before
+         * bailing: nothing here reaches the teardown that normally
+         * returns it, and a child that never runs must not leave the
+         * pty looking busy to the master the parent still holds.
+         */
+        proc_pty_release_process_slaves();
         guest_destroy(&g);
         return 1;
     }
 
     if (chown_overlay_recv(ipc_fd) < 0) {
         log_error("fork-child: failed to receive chown overlay");
+        /* Give back the credit the parent added for this child before
+         * bailing: nothing here reaches the teardown that normally
+         * returns it, and a child that never runs must not leave the
+         * pty looking busy to the master the parent still holds.
+         */
+        proc_pty_release_process_slaves();
         guest_destroy(&g);
         return 1;
     }
@@ -317,6 +337,12 @@ int fork_child_main(int ipc_fd,
             0 ||
         admission_ready != 1) {
         log_error("fork-child: parent did not commit child admission");
+        /* Give back the credit the parent added for this child before
+         * bailing: nothing here reaches the teardown that normally
+         * returns it, and a child that never runs must not leave the
+         * pty looking busy to the master the parent still holds.
+         */
+        proc_pty_release_process_slaves();
         guest_destroy(&g);
         return 1;
     }
@@ -371,6 +397,12 @@ int fork_child_main(int ipc_fd,
      * the single-threaded child at this point).
      */
     if (shim_globals_install_per_vcpu(vcpu, &g, hdr.child_pid) < 0) {
+        /* Give back the credit the parent added for this child before
+         * bailing: nothing here reaches the teardown that normally
+         * returns it, and a child that never runs must not leave the
+         * pty looking busy to the master the parent still holds.
+         */
+        proc_pty_release_process_slaves();
         guest_destroy(&g);
         return 1;
     }
@@ -498,6 +530,14 @@ int fork_child_main(int ipc_fd,
         vcpu_run_loop(vcpu, vexit, &g, verbose, timeout_sec, &wait_status);
 
     proc_process_exit(wait_status);
+
+    /* Same reason as the main-process path: this child is where a forked shell
+     * runs, and its stdio slaves are closed by the kernel rather than by the
+     * guest, so they never reach the per-fd close hook. The parent still holds
+     * the master and is waiting for exactly this hangup.
+     */
+    proc_pty_release_process_slaves();
+
     guest_destroy(&g);
     return exit_code;
 }
@@ -1910,6 +1950,18 @@ int64_t sys_clone(hv_vcpu_t vcpu,
         log_error("clone: failed to release admitted child");
         goto fail_snapshot;
     }
+
+    /* The child's inherited slave copies are live from the moment fork returns,
+     * so they go on the shared count here rather than in the child's own init,
+     * which the parent's close of its copy can beat.
+     *
+     * Deliberately after the last failure exit: every goto above unwinds a
+     * child that will never run, and a credit added before them would never be
+     * given back -- the pty would then look permanently busy and its master
+     * would stop reporting hangups for good. Still before siblings resume, so
+     * no guest code can close a slave in between.
+     */
+    proc_pty_fork_parent_note_inherited();
 
     /* The process-state payload includes the SCM_RIGHTS handoff for region
      * backing fds. Keep siblings quiesced until that send completes so a
--- a/src/runtime/procemu.c
+++ b/src/runtime/procemu.c
@@ -39,6 +39,7 @@
 #include <errno.h>
 #include <dirent.h>
 #include <pthread.h>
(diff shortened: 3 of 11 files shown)


jserv (changes requested):

doanbaotrung:
Updated

Max042004:
Wiring epoll to `proc_pty_master_hung_up`, registering TIOCGPTPEER slaves, and
pinning fd generations are all right, and the epoll cleanups (dropping the
latching gate, deleting kqueue events for deregistered fds, preferring ready
events over EINTR) are real fixes. The new cross-process accounting is where
the problems are: several independent paths drive `shared->slave_count` to zero
while slaves are open, or leave it permanently above zero.

### Shared slave accounting

**1. Record and release can target different slots, and the decrement is
silently dropped.** `pty_guest_slave_record_locked` (procemu.c:2353) and
`pty_guest_slave_release_locked` (procemu.c:2304) both credit the *lowest-index*
slot whose path matches, but the winner of that scan changes whenever a
registration lands in a freed lower slot. One process, no fork:

```
m1 = posix_openpt()        slot0 registered
m2 = dup(m1)               slot1 registered
close(m1)                  slot0 cleared, slave_path[0] = '\0'
s  = open("/dev/pts/N")    record -> slot1: count 1, shared->slave_count 1, seen 1
m3 = dup(m2)               register -> lowest empty slot is slot0, count 0
close(s)                   release -> slot0, `guest_slave_count > 0` is false,
                           then break: neither count is decremented
```

`shared->slave_count` is stuck at 1 with no live slave, so `m2`/`m3` never
report `POLLHUP`/`EPOLLHUP`/`EIO` — issue #274 itself, reintroduced. The
mirror case is as bad: when the newly-lowest slot does hold a count, the
release consumes a credit belonging to a different alias, and the local counts
that `clear_slot` / `release_process_slaves` hand back at teardown stop
matching reality.

**2. Closing one master alias hands back slaves another alias still owns.**
`pty_keepalive_clear_slot_locked` (procemu.c:1893) now passes the slot's
`guest_slave_count` to `pty_shared_detach`. Since every alias shares one
segment whose `seen` is sticky, `m = open(ptmx); m2 = dup(m); s = open(slave);
close(m)` drops the count to 0 with `seen == 1`, and `m2` reports a hangup with
`s` still open. Before this commit the counters were per-slot and this could
not happen.

**3. A fork child's in-flight attach can create a disjoint segment.**
`pty_shared_detach` (procemu.c:1854) `shm_unlink()`s as soon as *its own*
refcount hits zero, but the child attaches later, in a `posix_spawn`'d process.
The `openpty(3)` startup this PR targets hits the window directly:

```
m = posix_openpt(); s = open(slave); fork();
close(m)  in the parent   -> refs 1 -> 0 -> shm_unlink(name)
  ... child still in proc_pty_restore_keepalive ...
child attach (fresh=false) -> the name no longer resolves, so
                              shm_open(O_CREAT|O_EXCL) SUCCEEDS: a new, empty segment
child adopt                -> sets seen = 1 on it, count stays 0 (bump_shared = false)
```

The child's restored master reports a hangup at shell startup with two slave
references live; the original segment is orphaned at `count = 1, refs = 0`.
Nothing holds a reference on the child's behalf between
`proc_pty_snapshot_keepalive` and the child's restore.

**4. `slave_count` can go negative.** Same mechanism as 3: the child's bail
(`proc_pty_release_process_slaves`) subtracts a local count from a segment that
never received it. `pty_slot_hung_up_locked` testing `<= 0` masks the symptom
but not the damage — a negative count needs an extra slave open just to reach
zero, so later genuine hangups are missed.

**5. Fork credit skips master-less slots.**
`proc_pty_fork_parent_note_inherited` (procemu.c:2435) still requires
`master_host_fd != PTY_KEEPALIVE_FREE`, while record/release were deliberately
relaxed to path matching. A fork-restored shell that dropped its master copy
credits nothing for its own forks, and `proc_pty_snapshot_keepalive` skips the
same slots, so the grandchild is booked by nobody.

**6. The child's early bail returns credit the parent never added.** The parent
credits at forkipc.c:1964, *after* every send; the child's bails at
forkipc.c:313 and :325 subtract unconditionally — precisely the case where those
bails fire.

**7. A child that cannot attach never returns its credit.**
`proc_pty_release_process_slaves` (procemu.c:2493) skips slots whose `shared` is
NULL, so if `pty_shared_attach` returned NULL in the child the parent's `+N`
stays in the segment for the life of the host pty. Note the converse is not a
fix either: returning the count without knowing where it was added underflows a
partially-attached process. The credit needs to record its destination.

**8. Two sources of truth for the inherited set.**
`proc_pty_adopt_inherited_slaves` (procemu.c:2560) re-derives it by scanning the
fd table for a matching `st_rdev`; the parent counted
`pty_guest_slave_table` entries. Any live slave the table never learned about
(TIOCGPTPEER when the generation re-check at io.c:2410 fails, SCM_RIGHTS,
`PTY_GUEST_SLAVE_MAX` exhaustion) is visible to one and not the other. The
underlying gap is broader than fork: a single process holding one unrecorded
slave reports a false hangup as soon as the recorded ones drain.

**9. `shm_unlink` by name can delete a live successor.** procemu.c:1854 unlinks
the name without checking it still resolves to the object it mapped, so the last
holder of a retired segment destroys the segment a recycled pty minor created
under the same name. This is the same `(name, generation)` identity problem the
PR solves correctly for fds.

### epoll

**10. `reg->pty_master` is snapshotted at ADD/MOD** (poll.c:1114) and
`pty_master_count` gates the whole scan, so a master adopted after registration
(SCM_RIGHTS, or before the first TIOCGPTN) never gets `EPOLLHUP`. `sys_ppoll`
has no such narrowing — the two paths disagree.

**11. `epoll_note_fd_closed` returns early for every instance** when any fd
shares the closing `ofd_id` (poll.c:961), so `regs[closed_fd]` keeps
`active`/`pty_master` with a generation that can never resolve. A master dup'd
before the original closes never delivers its hangup, and both counters stay
inflated.

**12. A finite timeout cannot notice a mid-wait hangup.** The re-probe is gated
on `!has_timeout` (poll.c:1467), so `epoll_wait(..., 5000)` sleeps out the full
deadline. Any libuv/Node loop computes a finite timeout. The new test only
covers a hangup that already exists on entry.

**13. The wakeup-pipe registration was deleted from `sys_epoll_pwait`**
(poll.c:1404) with no replacement, leaving the 200 ms re-arm as the only
interruption path. `sys_ppoll` and `pselect` kept theirs.

**14. `futex_interrupt_consume()` no longer runs when events are ready.**
The new `nready <= 0` guard (poll.c:1455) splits a short-circuit chain whose
middle term is a consuming CAS, not a query — contrary to the comment above it.
The flag survives to the next `futex_wait` and surfaces as a spurious `EINTR`.

**15. The hangup pre-probe is unlocked and generation-unverified**
(poll.c:1412) while the stamping pass re-checks under `inst->lock`, so
`epoll_wait(-1)` can return 0 — which Linux never does.

**16. The copy-out is wrapped in a `for (i = 0; i < nout; i++)` loop with an
unused induction variable** (poll.c:1603), writing the same buffer `nout` times.

**17. `HUP_SCAN_BATCH`** (poll.c:1302) is over-built for at most 128 candidates,
and `seen` accumulating across batches while `active_count` is re-read can skip
later slots.

### Coverage

**18. `select`/`pselect` is the one readiness path still unaware of pty
hangups.** `sys_pselect6` (poll.c:341) goes straight to the host, where the
keepalive slave stops macOS reporting anything, so a select-based reader wedges
exactly as `readv` did before this change. Worth either covering it or
narrowing the commit message's "every path that can observe one".

PR #282, opened 2026-08-11
Take AT_EXECFN from the execve filename not argv[0]

doanbaotrung:
Take AT_EXECFN from the execve filename, not argv[0]

build_linux_stack derived AT_EXECFN from argv[0]. Linux takes it from
bprm-&gt;filename, and the two diverge in two ways elfuse reproduces:
execve(path, ["altname"], ...) reports path, and under binfmt_misc the
prepended interpreter is not the program the guest asked to run.

Under rosetta argv[0] is the translator, so every translated process
reported "rosetta" as its own program name. Guests that identify
themselves through auxv rather than argv[0] broke on it: rust-coreutils
dispatches its multi-call applet from AT_EXECFN, so every coreutils
invocation in an x86_64 Ubuntu 26.04 rootfs died with

  coreutils: unknown program 'rosetta'

taking down any apt/dpkg run that shelled out to ln, cat or rm.

Pass the filename explicitly and copy it onto the stack as its own
string, as fs/binfmt_elf.c does, rather than pointing at an argv slot.
The rosetta argv layout is free to change (see the preserving-form note
in rosetta.c) without silently re-breaking this. All three stack-build
paths supply the guest-visible spelling they already publish through
/proc/self/exe: initial bootstrap, rosetta execve re-bootstrap, and
native execve re-bootstrap. The last of these also fixes the native case
of execve(path, ["altname"], ...), which reported the alternate name.

---
## Summary by cubic
Set AT_EXECFN to the execve filename across native and rosetta so processes report the target binary, not the translator, matching Linux semantics. This restores rust-coreutils multi-call dispatch and unblocks apt/dpkg on translated guests.

- **Bug Fixes**
  - Pass the execve path into `build_linux_stack` and copy it onto the stack as its own string for AT_EXECFN; fall back to argv[0] only if no filename is provided.
  - Apply to initial bootstrap, rosetta execve re-bootstrap, and native execve (including execve(path, ["altname"], ...)); keeps AT_EXECFN consistent with /proc/self/exe.

<sup>Written for commit e22f298c421cf0b8b62a19f4edaf1fb7337d6eb4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/282?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/core/stack.c
+++ b/src/core/stack.c
@@ -140,6 +140,7 @@ uint64_t build_linux_stack(guest_t *g,
                            uint64_t interp_base,
                            uint64_t vdso_base,
                            int execfd,
+                           const char *execfn,
                            linux_stack_auxv_t *auxv_out)
 {
     /* Linux initial stack layout (growing from high to low):
@@ -218,6 +219,31 @@ uint64_t build_linux_stack(guest_t *g,
     uint64_t platform_ptr = str_ptr;
     str_err |= write_str(g, platform_ptr, "aarch64");
 
+    /* AT_EXECFN: the filename handed to execve, copied onto the stack as its
+     * own string exactly as fs/binfmt_elf.c does.
+     *
+     * The kernel takes this from bprm->filename, not from argv[0], and the two
+     * diverge in two ways elfuse reproduces: execve(path, "altname", ...)
+     * reports path, and under binfmt_misc the interpreter rosetta.c prepends to
+     * argv is not the program the guest asked to run. Taking the string from
+     * the caller rather than from an argv index keeps that contract out of the
+     * argv layout, which differs between the native and rosetta forms and is
+     * free to change again (see the preserving-form note in rosetta.c).
+     *
+     * Guests that identify themselves through auxv rather than argv[0] depend
+     * on getting this right: rust-coreutils dispatches its multi-call applet
+     * from AT_EXECFN, so a leaked interpreter path makes every applet abort
+     * with "unknown program 'rosetta'".
+     */
+    uint64_t execfn_ptr = 0;
+    if (execfn) {
+        size_t execfn_len = strlen(execfn) + 1;
+        if (!stack_take(&str_ptr, stack_floor, execfn_len))
+            return 0;
+        execfn_ptr = str_ptr;
+        str_err |= write_str(g, execfn_ptr, execfn);
+    }
+
     /* Dynamically allocate pointer arrays to avoid stack buffer overflow with
      * large argument or environment lists. calloc(0, ...) is
      * implementation-defined, so always allocate at least one slot. The extra
@@ -251,8 +277,11 @@ uint64_t build_linux_stack(guest_t *g,
         str_err |= write_str(g, str_ptr, argv[i]);
     }
 
-    /* AT_EXECFN: pointer to argv[0] string (write it near the top) */
-    uint64_t execfn_ptr = (argc > 0) ? arg_ptrs[0] : 0;
+    /* Callers with no filename to report keep the historical argv[0] spelling
+     * rather than an AT_EXECFN of 0, which no Linux process ever sees.
+     */
+    if (!execfn_ptr && argc > 0)
+        execfn_ptr = arg_ptrs[0];
 
     /* Phase 2: Build the structured part of the stack. Align str_ptr down to 16
      * bytes first.
--- a/docs/internals.md
+++ b/docs/internals.md
@@ -981,7 +981,10 @@ How it works:
 2. The interpreter is loaded as `ET_DYN` at `g->interp_base` (computed
    dynamically: 60 GiB for 36-bit IPA, 1020 GiB for 40-bit IPA).
 3. `build_linux_stack()` passes `AT_BASE` (interpreter load address) and
-   `AT_EXECFN` (`argv[0]`) in the auxiliary vector.
+   `AT_EXECFN` (the execve filename, supplied by the caller) in the auxiliary
+   vector. Linux takes `AT_EXECFN` from `bprm->filename`, so it stays the
+   program the guest asked for even when `argv[0]` is an alternate name or the
+   rosetta binfmt_misc argv prepends the translator.
 4. The entry point becomes `interp_entry + load_base`; the dynamic linker
    takes over from there.
 5. Guest absolute paths reach the host through `path_translate_at()`
--- a/src/core/bootstrap.c
+++ b/src/core/bootstrap.c
@@ -627,7 +627,7 @@ int guest_bootstrap_prepare(guest_t *g,
     boot->stack_pointer = build_linux_stack(
         g, g->stack_top, stack_argc, stack_argv, (const char **) environ,
         stack_elf, stack_elf_load_base, stack_interp_base, native_vdso,
-        rosetta_execfd, &auxv);
+        rosetta_execfd, elf_guest_path, &auxv);
     if (boot->stack_pointer == 0) {
         log_error("failed to build initial stack");
         free(rosetta_argv);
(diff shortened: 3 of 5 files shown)


src/core/stack.c
```diff
@@ -251,8 +251,20 @@ uint64_t build_linux_stack(guest_t *g,
...
+     *
+     * Normally that is argv[0]. Under rosetta argv[0] is the translator
+     * (ROSETTA_PATH) and argv[1] is the real target, matching the binfmt_misc
+     * argv rosetta.c builds. Linux sets AT_EXECFN from the filename passed to
+     * execve, not from the interpreter, so report argv[1] there: guests that
+     * identify themselves through auxv (rust-coreutils multi-call dispatch)
+     * otherwise read back "rosetta" as their own program name and abort.
+     */
+    uint64_t execfn_ptr = 0;
+    if (g->is_rosetta && argc > 1)
```
jserv: The fix is right, but the exec filename is inferred from argv shape rather than passed in. It holds today only because `rosetta_finalize` owns both ends of the contract, and that coupling is implicit: the comment at src/core/rosetta.c:424 already contemplates switching to the 4-slot preserving form (`[rosetta, binary, original_argv[0], original_argv[1..]]`) if a login-shell or `execve(path, "altname", ...)` workload ever surfaces. If that switch happens, the target moves to slot 1 or 2 depending on the form and this line silently points at the wrong string again, with no compile-time or runtime signal. Threading an explicit `const char *execfn` into `build_linux_stack` from the two callers that already know the path would make the contract checkable at the call site instead of inferred from indices.

That parameter would also close the matching gap on the native side, which still reports `argv[0]`. Linux takes AT_EXECFN from the filename handed to execve, so `execve("/bin/ls", ["altname"], ...)` on a native aarch64 guest still reports `altname` where the kernel reports `/bin/ls`. Pre-existing and out of scope for this PR, but after this change the rosetta branch is the accurate one and the native branch is not, so the asymmetry is worth recording somewhere.

jserv (commented):
Reviewed the AT_EXECFN change on three independent passes. The change is correct and I found no blocker.

Verified as non-issues:

- `argc > 1` can never be false on a rosetta stack build. `rosetta_finalize` computes `rosetta_argc = (guest_argc > 0) ? guest_argc + 1 : 2` (src/core/rosetta.c:442), so slot 1 always exists even when the caller supplied no argv. The fallback branch is unreachable from the rosetta side rather than a silent re-introduction of the bug.
- `arg_ptrs[1]` is in bounds under that guard, and `total_entries` is unaffected since AT_EXECFN still contributes exactly one entry.
- `g->is_rosetta` is set before both `build_linux_stack` call sites: `bootstrap.c:421` for the initial boot, and `exec.c:1179` before `guest_reset` for the re-bootstrap. `guest_clear_rosetta_state` only runs when leaving rosetta, so there is no window where the flag is stale at stack-build time.
- No host path leaks into guest auxv. `rosetta_argv[1]` is `binary_guest_path`, not `elf_host_path` or the materialized temp, so a FUSE-backed binary still reports its guest spelling.

Two residual gaps that this PR does not introduce but that sit on the same path, filed here rather than inline because they are outside the diff:

1. `src/syscall/syscall.c:2218` resolves execveat targets through `fcntl(F_GETPATH)`, which resolves symlinks. That resolved path becomes `sys_execve`'s `path`, then `rosetta_argv[1]`, then AT_EXECFN. A multi-call binary reached through a symlinked applet name (`/usr/bin/ln` pointing at `coreutils`) would report `coreutils` on that route, which is the same failure this PR fixes for plain execve. Plain `execve` is unaffected because `path` keeps the guest's literal string. Worth knowing whether any of the apt/dpkg workloads reach exec through `execveat`/`fexecve`, or whether this stays theoretical.

2. `src/syscall/exec.c:779` overwrites `path` with the interpreter before the rosetta re-bootstrap, so a shebang script under rosetta reports the x86_64 interpreter in AT_EXECFN. Linux keeps the original script filename there: binfmt_script prepends the interpreter to argv but leaves `bprm->exec` pointing at the script. This is consistent with what `/proc/self/exe` reports today, so changing it would mean deciding those two surfaces are allowed to disagree, which they do on real Linux.

doanbaotrung:
<html><head></head><body>
<h2>What changed</h2>
<p><code>build_linux_stack()</code> now takes an explicit <code>const char *execfn</code> and writes it onto the stack as its own string, the way <code>fs/binfmt_elf.c</code> does, instead of deriving <code>AT_EXECFN</code> from an argv index. The argv-shape coupling is gone: if <code>rosetta_finalize</code> ever switches to the 4-slot preserving form, nothing in <code>stack.c</code> needs to know.</p>
<p>Three call sites, not two — your list missed <code>src/syscall/exec.c:1548</code>, the native execve re-bootstrap. The compiler caught it:</p>
<ul>
<li><code>bootstrap.c:627</code> (initial boot, native + rosetta) → <code>elf_guest_path</code></li>
<li><code>bootstrap.c:884</code> (rosetta execve re-bootstrap) → <code>elf_guest_path</code></li>
<li><code>exec.c:1548</code> (native execve re-bootstrap) → <code>path</code>, the same guest-visible spelling published to <code>proc_set_elf_path()</code> twenty lines later</li>
</ul>
<p>That third one closes the native asymmetry you flagged: <code>execve("/bin/ls", ["altname"], …)</code> now reports <code>/bin/ls</code>, and <code>AT_EXECFN</code> agrees with <code>/proc/self/exe</code> on all paths rather than only the rosetta one. <code>execfn == NULL</code> still falls back to <code>argv[0]</code> so no caller can produce an <code>AT_EXECFN</code> of 0.</p>
<h2>Your question about <code>execveat</code></h2>
<p>Not theoretical-only by construction, but theoretical for these workloads. Grepping <code>syscall 281@</code> across the verbose traces I captured of the failing <code>apt-get install</code> (<code>apt-get → dpkg → dpkg-deb → tar → rm → dpkg-split</code>, 32 <code>execve</code> calls) and of the standalone dpkg runs: <strong>zero</strong> <code>execveat</code> calls, all 32 execs go through plain <code>execve(221)</code>. So the <code>F_GETPATH</code> symlink-resolution route at <code>syscall.c:2218</code> is unreached by apt/dpkg; the multi-call-via-symlinked-applet case would need a caller that actually uses <code>execveat</code>/<code>fexecve</code>.</p>
<h2>Gap 2 (shebang) left alone</h2>
<p>I passed <code>path</code> at <code>exec.c:1548</code> deliberately, which preserves today's behaviour: <code>AT_EXECFN</code> reports the interpreter for a shebang script, matching what <code>/proc/self/exe</code> says. Changing it means letting those two surfaces disagree — which, as you note, is what real Linux does. That's a decision about which surface to make accurate, not a bug this PR should settle.</p>
<h2>Verification</h2>

  |  
-- | --
x64 rosetta, initial bootstrap | coreutils.real echo hi → hi
x64 rosetta, execve re-bootstrap | ln --version \| head -1 → ln (uutils coreutils) 0.8.0
arm64 native | same, no regression
make test-sysroot-procfs-exec | 6 passed, 0 failed
make test-proctitle-low-stack | PASS (stack=1024 KiB) — relevant since the execfn string adds to the string area
tests/test-rosetta-execfd.sh | 2 passed, 1 skipped (no aarch64 cross-compiler)

<h2>Updated commit message</h2>
<pre><code>Take AT_EXECFN from the execve filename, not argv[0]

build_linux_stack derived AT_EXECFN from argv[0]. Linux takes it from
bprm-&gt;filename, and the two diverge in two ways elfuse reproduces:
execve(path, ["altname"], ...) reports path, and under binfmt_misc the
prepended interpreter is not the program the guest asked to run.

Under rosetta argv[0] is the translator, so every translated process
reported "rosetta" as its own program name. Guests that identify
themselves through auxv rather than argv[0] broke on it: rust-coreutils
dispatches its multi-call applet from AT_EXECFN, so every coreutils
invocation in an x86_64 Ubuntu 26.04 rootfs died with

  coreutils: unknown program 'rosetta'

taking down any apt/dpkg run that shelled out to ln, cat or rm.

Pass the filename explicitly and copy it onto the stack as its own
string, as fs/binfmt_elf.c does, rather than pointing at an argv slot.
The rosetta argv layout is free to change (see the preserving-form note
in rosetta.c) without silently re-breaking this. All three stack-build
paths supply the guest-visible spelling they already publish through
/proc/self/exe: initial bootstrap, rosetta execve re-bootstrap, and
native execve re-bootstrap. The last of these also fixes the native case
of execve(path, ["altname"], ...), which reported the alternate name.
</code></pre></body></html>

PR #283, opened 2026-08-11
Improve test harness code introduced during sidecar migration

henrybear327:
Follow up improvement effort to improve code introduce in https://github.com/sysprog21/elfuse/pull/256 

Mainly focusing on improving the test harness makefile, and refactoring the helper functions.

---
## Summary by cubic
Modernizes the test harness by generating check lanes with canned make rules, centralizing shared helpers, and adding header dependency tracking for guest tests. This reduces duplication, fixes rebuild correctness, and removes stale sidecar references.

- **Refactors**
  - Generate check and sanitizer lanes via canned recipes in `mk/tests.mk` (shared scratch setup, absock leak check, banners preserved; single '+' run line so make -n expands the lane).
  - Track shared header deps for cross-compiled guest tests via `CROSS_TEST_CFLAGS` (-MMD/-MP; driver injects -MQ for the -o target).
  - Consolidate helpers into `tests/test-util.h` and `tests/host-test-util.h` (dirfd file I/O, dir scans, UNIX socket helpers with cleanup on failure, host pass/fail summary; unify openat2 ABI via `linux-openat2.h`); update tests to use them.
  - Fold the no-sysroot variant into `test-sysroot-outside-names` via a mode arg; remove `tests/test-nosysroot-literal-names.c`.
  - Remove stale sidecar mentions from docs and test headers; clarify resolver behavior (clamp '..' only at guest root; case-insensitive sysroot may store escaped leaves).

- **Bug Fixes**
  - Bounds-check the stored-name probe reply and zero the buffer; accept SMB/NFS attr layouts; mirror the check in host-only lanes.

<sup>Written for commit 83aa927cf51493cfe7286925dd2b2933a8e63b7f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/283?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mk/tests.mk
```diff
@@ -90,147 +90,118 @@ check-tsan: clean
...
+@$(BUILD_DIR)/$(1)
+endef
+
+# The recursive line carries '+' because make decides what still runs under
+# -n by looking for a literal $(MAKE) in the unexpanded recipe line, which a
+# canned recipe hides (make manual 9.3). Without it, "make -n check" prints
+# the sub-make instead of expanding the lane it would run.
+define run-lane
+@printf "\n$(BLUE)━━━ $(2) ━━━$(RESET)\n"
++@$(MAKE) --no-print-directory $(1)
```
jserv: The `+` does more than keep this one line live under `-n`. Once a `+` line appears in a canned recipe's expansion, every *later* line of that same expansion is executed for real rather than echoed. So `make -n check` runs the banner `printf` for every lane after the first: 4 banners are actually written on this branch, 0 on the base. Reproduced on GNU make 3.81 (the one macOS ships) and 4.4.1:

```make
define lane
@echo BANNER $(1)
+@$(MAKE) --no-print-directory sub$(1)
endef
define all_lanes
$(call lane,1)
@echo MID
$(call lane,2)
endef
top:
	$(all_lanes)
	@echo TAIL
```

`make -n top` echoes `BANNER 1` but executes `MID`, `BANNER 2` and everything else inside the expansion; `TAIL`, which sits outside it, stays dry.

Harmless right now because the only other lines in `CHECK_SHARED_LANES` are `printf`s, but that define is exactly where lanes get added, so the next line appended after a `run-lane` will run during a dry build. Either collapse a lane into one `+`-marked line (`+@printf "..." ; $(MAKE) --no-print-directory $(1)`) so all banners behave alike, or say so in the comment above.

tests/host-test-util.h
```diff
@@ -81,8 +143,26 @@ static inline const char *disk_name(const char *dir, const char *name)
         return NULL;
     if (!(buf.returned.commonattr & ATTR_CMN_NAME))
         return NULL;
-    snprintf(out, sizeof(out), "%s",
-             (const char *) &buf.name_ref + buf.name_ref.attr_dataoffset);
+
+    /* attr_dataoffset is signed; a negative one points before the reference,
+     * outside anything the kernel wrote for this attribute.
+     */
+    if (buf.name_ref.attr_dataoffset <= 0 || buf.name_ref.attr_length == 0)
```
jserv, line 164: `buf.name_ref` is read before anything establishes that the reply is long enough to hold it, so this is not equivalent to the production copy it mirrors: `casefold_attr_stored_name` rejects that case up front with `ref_off > usable || usable - ref_off < sizeof(ref)` (src/syscall/casefold-walk.c:126). `buf` has no initializer, so under the exact premise this check was added for (a volume that sets ATTR_CMN_NAME in `returned` without filling the reference the way APFS does) a `buf.length` below 32 leaves both fields reading uninitialized stack. If those bytes happen to land in range (dataoffset 1..7, a small `attr_length`, a NUL inside it) the bounds check below passes and the function returns a name assembled from bytes the volume never wrote.

Two lines: initialize `buf`, and hoist `usable`/`ref_off` above this test so the reference itself is checked against the reply first.

tests/host-test-util.h
```diff
@@ -16,12 +16,68 @@
...
+                                    char *out,
+                                    size_t outsz)
+{
+    const char *base = argv1 ? argv1 : getenv("TMPDIR");
+
+    if (!base || base[0] == '\0')
+        base = "/tmp";
+    snprintf(out, outsz, "%s/%s-XXXXXX", base, tag);
+    if (!mkdtemp(out)) {
+        fprintf(stderr, "%s: cannot create a scratch directory in %s: %s\n",
```
jserv, line 77: The diagnostic now names the scratch tag instead of the binary: `test-casefold-host` failing to create a scratch directory prints `elfuse-casefold: cannot create a scratch directory in ...`, where it used to print the lane name. In a CI log that line is the only thing pointing at which test died. Passing the lane name (or `argv[0]`) as a separate argument keeps it.

mk/tests.mk
```diff
@@ -90,147 +90,118 @@ check-tsan: clean
...
+define run-host-unit
+@printf "\n$(BLUE)━━━ $(2) ━━━$(RESET)\n"
+@$(BUILD_DIR)/$(1)
+endef
+
+# The recursive line carries '+' because make decides what still runs under
+# -n by looking for a literal $(MAKE) in the unexpanded recipe line, which a
+# canned recipe hides (make manual 9.3). Without it, "make -n check" prints
+# the sub-make instead of expanding the lane it would run.
+define run-lane
```
jserv, line 112: Banner text arrives through `$(call)`, so a comma anywhere in it splits into `$(3)` and the banner silently truncates at the comma. No banner has one today, and this refactor exists to make lane-list edits safe, so it is worth a line in the comment above (or hand the banner in through a variable).

tests/test-case-collision.c
```diff
@@ -328,8 +298,8 @@ int main(void)
         } else if (st_src.st_ino != st_alias.st_ino || st_src.st_nlink < 2) {
             FAIL("colliding hardlinks do not share inode");
         } else if (unlink(src) < 0 || stat(alias, &st_alias) < 0 ||
-                   dir_has_entry(base, "hardlink") != 0 ||
-                   dir_has_entry(base, "HARDLINK") != 1) {
+                   dir_contains(base, "hardlink") ||
```
jserv, line 301: The local `dir_has_entry` returned -1 for a directory it could not open, and the old assertions compared against `1`/`0`, so an `opendir` failure failed the test. `dir_contains` collapses that into `false`, which makes the negative half of each assertion pass when the directory cannot be read. Every call site here is still covered by the positive half of the same expression, so nothing is broken today; flagging it because this is now the shared helper and the next negative-only use of it will not notice.

jserv (commented):
Verified the three load-bearing claims before commenting: `make -n check` emits the same 46 banners in the same order as the base (byte-identical), `check-sanitizer` expands to the same 13 lanes it did before, the guest `.d` files land in `build/` and a `touch tests/test-util.h` really does re-trigger the cross rule, and both modes of the folded lane pass (5 assertions without a sysroot, 7 with). The consolidation itself reads faithfully: the `write_at`/`content_at`/`create_file`/`bind_*` replacements keep the same semantics, and the nosysroot leg gained an assertion rather than losing one. Notes below, most important first.

PR #284, opened 2026-08-11
Remove dead code and simplify code introduced during sidecar migration

henrybear327:
Follow up improvement effort to improve code introduce in https://github.com/sysprog21/elfuse/pull/256 

---
## Summary by cubic
Removes sidecar-era paths and cuts per-lookup work in casefold and sysroot resolution. Old: each component copied prefixes, re-derived `realpath()` and object types, and rebuilt resolver outputs; new: probe components in place, reuse `ATTR_CMN_OBJTYPE`, skip redundant `realpath()` and concatenations, and treat fully typed walks as containment proofs.

- Probe components in place: append candidates to `out` and restore its terminator. `out` doubles as probe scratch; do not overlap `out` with `guest_path` or `base_host_prefix`.
- Reuse object types from the probe: record `leaf_type_known`, `leaf_is_link`, and `all_types_known`. Run `fstatat`/`faccessat` only when types are unknown. `host_component_spelling` exposes link-ness to the NO_XDEV walker.
- Skip the sysroot containment recheck for a fully typed walk; keep it for listing-answered components, dot components, and byte-exact paths.
- Canonicalize the sysroot once and read `_Atomic` fold/canonical flags. Avoid per-call `realpath()` and skip sysroot snapshots for relative paths.
- Remove unread resolver outputs and duplicate sysroot+lookup concatenations; rely on the seed buffer.
- Make the walk report optional in `casefold_resolve_at`; pass NULL when only the verdict or `out` is needed.
- Move the UTF-8/UTF-16 budget oracles into tests; drop `casefold_utf16_units` from `casefold.h`.

- Fixes
  - Withdraw type info on `readdir` fallback so callers do not skip required `fstatat` checks.
  - Bounds-check `getattrlistat` stored-name replies; prevent an overlong append from leaving a trailing `/`.
  - Make the escape fallback strict: on broken preconditions, return an error rather than guess a spelling.

<sup>Written for commit 2c1a26c4072072f4fb84b7c9f4c45896f419c366. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/284?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


henrybear327:
https://github.com/sysprog21/elfuse/pull/290 fixes the issue raised in https://github.com/sysprog21/elfuse/actions/runs/31543781893/job/93951954868?pr=284

PR #298, opened 2026-08-15
Add `--env` and `--clear-env` launch flag support

henrybear327:
In preparation for the OCI support, we need to support the ENV flag. 

---
## Summary by cubic
Adds --env and --clear-env flags to control the guest environment for OCI workflows. Previously the guest always inherited the host environment; now you can clear it and set/import variables, while default behavior remains inherit when neither flag is used.

- Implements guest_env_build with docker run -e semantics: KEY=VALUE replaces or appends; bare KEY imports the host value (even with --clear-env); rejects empty variable names; drops malformed base entries and duplicate names to avoid stale copies.
- main collects repeated --env tokens and builds the vector before proctitle updates; rejects invalid entries early; -- still ends elfuse option parsing. Launch builds an env only when a flag asks for one.
- Extends launch_args_t with envp; elfuse_launch uses envp when provided or falls back to the host environ when NULL. Caller retains ownership.
- Adds tests: host-side cross product oracle for the merge, guest env-dump, and launch-flag lanes validating initial stack and /proc/self/environ. Adds strv_free for argv/env cleanup.
- Updates usage docs to describe import/replace rules and --clear-env.

**Migration for embedders calling elfuse_launch**
- Set the new envp field. Set envp = NULL to preserve old behavior (inherit host environment), or pass a malloc’d NULL-terminated vector and free it after launch with strv_free.

<sup>Written for commit 90d407b5909396316057b973d5563c63466adffb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/298?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/core/guest-env.h
+++ b/src/core/guest-env.h
@@ -1,0 +1,44 @@
+/*
+ * Guest environment vector construction for the launch flags
+ *
+ * Copyright 2026 elfuse contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * elfuse builds the guest's "KEY=VALUE" array, which build_linux_stack copies
+ * onto the initial guest stack, from the host environment plus the --env /
+ * --clear-env flags, following `docker run -e` so an OCI front end can hand a
+ * guest exactly the environment an image config asks for. The host
+ * environment arrives as a parameter rather than from environ, so tests can
+ * hand guest_env_build arbitrary base vectors, malformed entries included.
+ */
+
+#pragma once
+
+#include <stdbool.h>
+
+/* Build the guest environment vector from @host_env and the --env overrides.
+ *
+ * @host_env (NULL-terminated) fills two roles: the base the overrides merge
+ * into (unless @clear_env), and the source a bare "KEY" override imports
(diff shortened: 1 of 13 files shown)


src/core/guest-env.h
```diff
@@ -0,0 +1,45 @@
...
+ * name; @host_env entries violating that are dropped, or a later override
+ * would append beside the entry it meant to replace. On allocation failure
+ * or an empty override name (as setenv(3)) returns -1, logged, with the
+ * outputs untouched.
+ */
+int guest_env_build(char *const *host_env,
+                    char *const *overrides,
+                    int n_overrides,
+                    bool clear_env,
+                    char ***out_envp,
```
jserv: It is confusing to declare as `char ***out_envp`.

PR #300, opened 2026-08-16
Report one memory total from sysinfo and meminfo

xalestar:
sysinfo(2) caps totalram at a hardcoded 4094595072 bytes while /proc/meminfo reports sysctl(HW_MEMSIZE), so the two surfaces describe different machines: 3998628 kB against 25165824 kB on a 24 GiB host.

A guest that reads both then subtracts one from the other. busybox `free` takes total and free from sysinfo and buff/cache from /proc/meminfo and computes `used = total - free - cached`; a Cached larger than the whole of totalram wraps that subtraction, and `free` prints the total and a used of 18014398505005547 run together into a single 24-digit field:

```
              total        used        free      shared  buff/cache   available
Mem:        399862818014398505005547       35065           0     8440000     8660688
```

To the question in the review: the ceiling is cosmetic. Nothing outside `sys.c` reads it, no test pins it, and no guest is known to choke on a large MemTotal, so this deletes it rather than keeping the machinery that services it. Both surfaces take the total from `sys_guest_ram_bytes()`, which is HW_MEMSIZE. That retires the proportional scaling the ceiling required and the 64-bit overflow inside it, and a guest sizing a heap or a job count from MemTotal reads the machine it runs on rather than a sixth of it.

Free is reported once as well, since busybox subtracts sysinfo's freeram rather than MemFree and the two were separate host_statistics64 samples that drift against each other. `sys_guest_ram_free()` serves both from the snapshot sysinfo reports.

Deleting the cap does not retire the clamp. The Mach counters do not partition: a purgeable page is also counted in inactive, so free plus inactive plus purgeable can exceed physical memory whatever the total is derived from. `proc_open_meminfo()` clamps Cached against what MemFree leaves, which is the invariant the guest's subtraction actually needs. Cached gives up the precision rather than MemFree, since MemFree has to keep matching freeram.

The test asserts the agreement rather than either number, since the totals are host-dependent, and it runs in test-matrix against a real kernel where the two surfaces come from one si_meminfo() and agree by construction.

With the change in place the reproducer above reports

```
              total        used        free      shared  buff/cache   available
Mem:       25165824    16527232      282896           0     8355696     8638592
```

and 25165824 - 282896 - 8355696 = 16527232, which is the used column.

The new assertion fails without the fix rather than passing vacuously. The same test binary run against a build of `main` gives

```
sysinfo agrees with /proc/meminfo FAIL: sysinfo and /proc/meminfo disagree
```

Verified on an Apple Silicon host (Darwin 25.6, 24 GiB), on a branch rebased onto `main` at 3541d18. make check, make lint and the .ci/ format, newline and security checks all pass. The three test-matrix modes stay within their baselines with zero failures: elfuse-aarch64 252 passed, qemu-aarch64 231 passed, elfuse-x86_64 78 passed. test-sysinfo passes in both the elfuse-aarch64 and qemu-aarch64 columns, so the assertion holds against a real kernel as well.

### DIFF
--- a/src/runtime/procemu.c
+++ b/src/runtime/procemu.c
@@ -1995,31 +1995,37 @@ out:
     return proc_finish_maps_output(result, &entries, &builder);
 }
 
-/* Emit /proc/meminfo from host sysctl (HW_MEMSIZE) plus mach vm_statistics64,
- * approximating the Linux fields macOS does not expose.
+/* Emit /proc/meminfo from the guest-visible memory figures plus mach
+ * vm_statistics64, approximating the Linux fields macOS does not expose.
+ *
+ * MemTotal and MemFree come from sys_guest_ram_bytes()/sys_guest_ram_free()
+ * rather than from a host sample taken here, so this file and sysinfo(2) read
+ * one snapshot rather than two: a guest reading both back to back, as busybox
+ * free does, sees figures that were measured together. They used to describe
+ * two different machines outright, never mind two instants -- meminfo reported
+ * the host's real memory while sysinfo capped it at a hardcoded 4GiB -- and
+ * that guest underflowed. busybox free takes total and free from sysinfo and
+ * buff/cache from here, so a Cached larger than the sysinfo total made
+ * used = total - free - cached wrap into a 24-digit figure.
  *
  * Returns a host fd, or -1. Split out of proc_intercept_open to keep that
  * dispatcher readable.
  */
 static int proc_open_meminfo(void)
 {
-    int64_t physmem = 0;
-    size_t sz = sizeof(physmem);
-    int mib[2] = {CTL_HW, HW_MEMSIZE};
-    sysctl(mib, 2, &physmem, &sz, NULL, 0);
-    uint64_t total_kb = (uint64_t) physmem / 1024;
+    uint64_t total_kb = sys_guest_ram_bytes() / 1024;
+    uint64_t free_kb = sys_guest_ram_free() / 1024;
 
-    /* Query host vm_statistics for accurate free/active/inactive. Falls back to
+    /* Query host vm_statistics for accurate active/inactive. Falls back to
      * approximations if the mach call fails.
      */
-    uint64_t free_kb, avail_kb, buffers_kb, cached_kb;
+    uint64_t avail_kb, buffers_kb, cached_kb;
     vm_statistics64_data_t vm_stat = {0};
     mach_msg_type_number_t count = HOST_VM_INFO64_COUNT;
     uint64_t page_size = 4096;
     if (host_statistics64(mach_host_self(), HOST_VM_INFO64,
                           (host_info64_t) &vm_stat, &count) == KERN_SUCCESS) {
         host_page_size(mach_host_self(), (vm_size_t *) &page_size);
-        free_kb = (uint64_t) vm_stat.free_count * page_size / 1024;
         uint64_t inactive_kb =
             (uint64_t) vm_stat.inactive_count * page_size / 1024;
         uint64_t purgeable_kb =
@@ -2031,18 +2037,30 @@ static int proc_open_meminfo(void)
         cached_kb = inactive_kb + purgeable_kb;
         buffers_kb = 0; /* macOS does not expose buffer cache separately */
     } else {
-        free_kb = total_kb / 2;
         avail_kb = total_kb * 3 / 4;
         buffers_kb = total_kb / 20;
         cached_kb = total_kb / 4;
     }
 
-    /* Saturating subtraction. On macOS free + cached (inactive + purgeable) can
-     * exceed physical total, which would unsigned-underflow these derived
-     * fields into absurd values.
+    /* Cached has to fit in what MemFree leaves, not merely be smaller than the
+     * total. The Mach counters do not partition: a purgeable page is also
+     * counted in inactive, so free + inactive + purgeable can exceed physical
+     * memory however the total is derived. A guest computing
+     * used = total - free - cached wraps on that excess, which is the failure
+     * reporting one total is meant to end. Cached gives up the precision rather
+     * than MemFree, since MemFree is what sysinfo reports as freeram and the
+     * two have to keep matching. free_kb needs no clamp of its own: it comes
+     * from sys_guest_ram_free(), which is already held below the total.
+     */
+    if (cached_kb > total_kb - free_kb)
+        cached_kb = total_kb - free_kb;
+
+    /* Active is now an exact remainder -- the clamp above makes free + cached
+     * fit inside the total. AnonPages still saturates, because Buffers is not
+     * part of that clamp and the fallback branch above sets it non-zero.
      */
     uint64_t fc_kb = free_kb + cached_kb;
-    uint64_t active_kb = total_kb > fc_kb ? total_kb - fc_kb : 0;
+    uint64_t active_kb = total_kb - fc_kb;
     uint64_t anon_kb =
         total_kb > fc_kb + buffers_kb ? total_kb - fc_kb - buffers_kb : 0;
 
--- a/src/syscall/sys.c
+++ b/src/syscall/sys.c
@@ -47,6 +47,14 @@ static const linux_utsname_t cached_uname = {
 };
 static const uint8_t cached_affinity_mask[256] = {1}, zero_block[256] = {0};
 
+/* Physical memory to report when sysctl(HW_MEMSIZE) fails. Reporting zero is
+ * not an option: a guest computing used = total - free - cached against a zero
+ * total wraps, which is the failure sys_guest_ram_bytes() exists to prevent.
+ * The figure is one captured from a 4 GiB VZ VM, so the fallback describes a
+ * machine that plausibly exists.
+ */
+#define GUEST_RAM_FALLBACK_BYTES 4094595072ULL
+
 /* sysinfo cache.
  *
  * Process-scoped by intent: the cache mirrors the host's view (totalram from
@@ -59,7 +67,7 @@ static const uint8_t cached_affinity_mask[256] = {1}, zero_block[256] = {0};
 static pthread_once_t sysinfo_once = PTHREAD_ONCE_INIT;
 static pthread_rwlock_t sysinfo_lock = PTHREAD_RWLOCK_INITIALIZER;
 static time_t cached_boottime_sec = 0;
-static uint64_t cached_totalram = 0, cached_real_memsize = 0;
+static uint64_t cached_totalram = 0;
 static uint64_t cached_page_size = 0;
 static mach_port_t cached_host_port = MACH_PORT_NULL;
 static time_t cached_sysinfo_sec = -1;
@@ -103,12 +111,9 @@ static void sysinfo_init_cached_host_state(void)
     uint64_t memsize = 0;
     size_t ms_len = sizeof(memsize);
     int mib_mem[2] = {CTL_HW, HW_MEMSIZE};
-    if (sysctl(mib_mem, 2, &memsize, &ms_len, NULL, 0) == 0) {
-        const uint64_t vm_ram_cap =
-            4094595072ULL; /* totalram cap captured from a 4 GiB VZ VM */
-        cached_real_memsize = memsize;
-        cached_totalram = (memsize > vm_ram_cap) ? vm_ram_cap : memsize;
-    }
+    if (sysctl(mib_mem, 2, &memsize, &ms_len, NULL, 0) != 0 || memsize == 0)
+        memsize = GUEST_RAM_FALLBACK_BYTES;
+    cached_totalram = memsize;
 
     long page_size = sysconf(_SC_PAGESIZE);
     if (page_size > 0)
@@ -127,24 +132,18 @@ static void sysinfo_refresh_cached_locked(time_t now_sec)
     if (cached_boottime_sec != 0)
         cached_sysinfo.uptime = now_sec - cached_boottime_sec;
 
-    /* Free RAM from vm_statistics64. Scale proportionally if totalram is
-     * capped.
+    /* Free RAM from vm_statistics64, held below the total unconditionally so
+     * that every reader of the pair -- here and /proc/meminfo -- can subtract
+     * one from the other without checking.
      */
     vm_statistics64_data_t vmstat = {0};
     mach_msg_type_number_t count = HOST_VM_INFO64_COUNT;
     if (cached_host_port != MACH_PORT_NULL &&
         host_statistics64(cached_host_port, HOST_VM_INFO64,
                           (host_info64_t) &vmstat, &count) == KERN_SUCCESS) {
-        uint64_t real_free = (uint64_t) vmstat.free_count * cached_page_size;
-        if (cached_real_memsize > 0 &&
-            cached_real_memsize > cached_sysinfo.totalram) {
-            uint64_t scaled_free = real_free;
-            scaled_free *= cached_sysinfo.totalram;
-            scaled_free /= cached_real_memsize;
-            cached_sysinfo.freeram = scaled_free;
-        } else {
-            cached_sysinfo.freeram = real_free;
-        }
+        uint64_t free_bytes = (uint64_t) vmstat.free_count * cached_page_size;
+        cached_sysinfo.freeram =
+            free_bytes > cached_totalram ? cached_totalram : free_bytes;
     }
 
     /* Load averages (x 65536 for fixed-point). */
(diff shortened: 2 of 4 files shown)


src/syscall/sys.c
```diff
@@ -118,6 +125,30 @@ static void sysinfo_init_cached_host_state(void)
...
+    /* Scale in KiB rather than bytes. A byte-unit "host_bytes *
+     * cached_totalram" overflows 64 bits once host_bytes passes ~4.2 GiB
+     * against a 4GiB cap, which host free memory reaches routinely; dividing
+     * both operands by 1024 first keeps the product far inside range (a
+     * 2^64-byte host is 2^54 KiB) at a cost of under 1 KiB of precision.
+     */
+    uint64_t real_kb = cached_real_memsize / 1024;
+    if (real_kb == 0)
+        return host_bytes;
+    return (host_bytes / 1024) * (cached_totalram / 1024) / real_kb * 1024;
```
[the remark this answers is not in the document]
xalestar: Addressed in `7bef252`.

Valid, and this was the real gap -- thanks. Agreeing on the total does not by itself stop the wrap, because the Mach counters do not partition: a purgeable page is also counted in inactive, so free + inactive + purgeable can exceed HW_MEMSIZE and the scaling here carries that excess through proportionally. `proc_open_meminfo()` says as much already, in the comment above the saturating `active_kb`/`anon_kb` subtraction, but that saturation only protects the fields elfuse derives -- `Cached` is emitted raw, and it is `Cached` the guest subtracts.

Fixed by clamping the pair in `proc_open_meminfo()` before formatting rather than in the scaling helper, since the invariant belongs to the emitted file and the helper has no view of the other components:

```c
if (free_kb > total_kb)
    free_kb = total_kb;
if (cached_kb > total_kb - free_kb)
    cached_kb = total_kb - free_kb;
```

`Cached` gives up the precision rather than `MemFree`, because `MemFree` has to keep matching what `sysinfo` reports as `freeram`.

This host was not in the overlap state (`vm_stat` put free + inactive + purgeable at 47% of `hw.memsize`), so I forced it with a scratch build multiplying `inactive_kb` by 10. Without the clamp the original 24-digit figure returns and `test-sysinfo` fails; with it, `MemFree: 111078` and `Cached: 3887550` sum to exactly `MemTotal: 3998628` and `free` reports `used` as 0.

src/runtime/procemu.c
```diff
@@ -2019,11 +2023,16 @@ static int proc_open_meminfo(void)
     if (host_statistics64(mach_host_self(), HOST_VM_INFO64,
                           (host_info64_t) &vm_stat, &count) == KERN_SUCCESS) {
         host_page_size(mach_host_self(), (vm_size_t *) &page_size);
-        free_kb = (uint64_t) vm_stat.free_count * page_size / 1024;
+        free_kb =
+            sys_guest_ram_scale((uint64_t) vm_stat.free_count * page_size) /
+            1024;
         uint64_t inactive_kb =
-            (uint64_t) vm_stat.inactive_count * page_size / 1024;
+            sys_guest_ram_scale((uint64_t) vm_stat.inactive_count * page_size) /
```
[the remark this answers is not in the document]
xalestar: Addressed in `7bef252`.

Valid, same root cause as the finding on the scaling helper, and fixed in the same place -- here, where the components are all in scope.

`proc_open_meminfo()` now clamps `free_kb` to the total and `cached_kb` to what is left after it, immediately after the mach branch and its fallback so both paths are covered. That makes `MemFree + Cached <= MemTotal` hold whatever the counters overlap by, which is the invariant busybox's `total - free - cached` actually depends on. Clamping `cached_kb` rather than normalising all three keeps `MemFree` equal to the `freeram` `sysinfo` reports, which is the agreement this PR is about.

Verified against a forced overlap (scratch build with `inactive_kb` multiplied by 10): without the clamp the 24-digit `used` returns, with it the pair sums to exactly `MemTotal`.

src/syscall/sys.c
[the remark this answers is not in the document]
xalestar: Addressed in `7bef252`.

Correct arithmetic, declining the guard.

The bound is real but unreachable: the second operand is `GUEST_RAM_CAP_BYTES` in KiB, a compile-time constant near 4e6, so the product only leaves 64 bits once `cached_real_memsize` itself is around 4 PB. The largest Mac Apple currently ships has 512 GB, four orders of magnitude short, and `HW_MEMSIZE` is the host's own physical memory rather than anything a guest can influence.

Rather than add a branch that cannot be taken, I documented the bound in the comment above the expression, so the next reader can see it was computed rather than overlooked. Happy to swap in an overflow-safe mul/div if you would rather have the guard than the note.

jserv (changes requested):

src/runtime/procemu.c
```diff
@@ -2037,6 +2046,20 @@ static int proc_open_meminfo(void)
...
+     * HW_MEMSIZE, and scaling preserves that excess proportionally. A guest
+     * computing used = total - free - cached wraps on it, which is the same
+     * failure reporting one total is meant to end. Give up precision on
+     * cached, not on free, since free is what sysinfo reports as freeram and
+     * the two have to keep matching.
+     */
+    if (free_kb > total_kb)
+        free_kb = total_kb;
+    if (cached_kb > total_kb - free_kb)
+        cached_kb = total_kb - free_kb;
```
jserv, line 2056: This clamps `meminfo.MemFree + meminfo.Cached` against `MemTotal`, but busybox `free` computes `used = sysinfo.totalram - sysinfo.freeram - meminfo.Cached`. It never reads `MemFree`.

`freeram` and `MemFree` are two independent `host_statistics64` samples: sysinfo caches for up to a second (sys.c:177), meminfo samples fresh at open (procemu.c:2026). Measured drift on an 8 GiB host reached 15 MB in both directions across 8 reads in one process. When this clamp fires, the margin it leaves is exactly `Active`, which is 0 by construction, so any `freeram > MemFree` gap wraps busybox again.

Report one free as well as one total: export the guest-visible free the same way as the total, and use it both for `MemFree` and as the clamp denominator, so the two surfaces come from a single snapshot.

src/runtime/procemu.c
```diff
@@ -2037,6 +2046,20 @@ static int proc_open_meminfo(void)
...
+
     /* Saturating subtraction. On macOS free + cached (inactive + purgeable) can
```
jserv: The two clamps above now guarantee `free + cached <= total`, and the fallback branch gives `0.8 * total`, so this subtraction can no longer saturate. The comment describes a state the code two lines up just made impossible. Worth rewording, otherwise the next reader assumes the clamps are not load-bearing.

src/syscall/sys.c
```diff
@@ -106,9 +113,9 @@ static void sysinfo_init_cached_host_state(void)
     size_t ms_len = sizeof(memsize);
     int mib_mem[2] = {CTL_HW, HW_MEMSIZE};
     if (sysctl(mib_mem, 2, &memsize, &ms_len, NULL, 0) == 0) {
-        const uint64_t vm_ram_cap = 4094595072ULL; /* Match Lima VZ 4GiB VM */
         cached_real_memsize = memsize;
-        cached_totalram = (memsize > vm_ram_cap) ? vm_ram_cap : memsize;
+        cached_totalram =
+            (memsize > GUEST_RAM_CAP_BYTES) ? GUEST_RAM_CAP_BYTES : memsize;
```
jserv: If the sysctl fails, `cached_totalram` and `cached_real_memsize` both stay 0 and `pthread_once` never retries. sysinfo then reports `totalram` 0 with `freeram` at several GiB, since `sys_guest_ram_scale()` takes its identity branch, and /proc/meminfo reports `MemTotal: 0 kB`. That is the original wrap, restored. The new test passes vacuously on that path: `0 == 0` and `0 + 0 <= 0`.

One line keeps every surface describing the same plausible machine:

```c
if (cached_totalram == 0)
    cached_real_memsize = cached_totalram = GUEST_RAM_CAP_BYTES;
```

src/syscall/sys.c
```diff
@@ -118,6 +125,37 @@ static void sysinfo_init_cached_host_state(void)
...
+     * memory figure, so it cannot exceed cached_real_memsize, and the other
+     * operand is GUEST_RAM_CAP_BYTES in KiB, a constant near 4e6. The product
+     * therefore stays inside 64 bits until the host itself holds about 4 PB.
+     * That is not guarded, on the grounds that a machine which reaches it has
+     * left every other assumption in this file behind too.
+     */
+    uint64_t real_kb = cached_real_memsize / 1024;
+    if (real_kb == 0)
+        return host_bytes;
+    return (host_bytes / 1024) * (cached_totalram / 1024) / real_kb * 1024;
```
jserv: sys.h documents this as keeping derived values below the total, but both identity returns (line 138 and line 155) pass `host_bytes` through unchanged, so the postcondition holds only under the undocumented precondition `host_bytes <= cached_real_memsize`. Clamping the result makes it unconditional:

```c
uint64_t out = (host_bytes / 1024) * (cached_totalram / 1024) / real_kb * 1024;
return out > cached_totalram ? cached_totalram : out;
```

with the same clamp on the early returns. That also makes `if (free_kb > total_kb)` at procemu.c:2058 provably dead and removable.

Separately, "under 1 KiB of precision" understates it: three truncating divisions compose, and `/ real_kb * 1024` alone loses up to 1023 bytes, so the worst case is nearer 2 KiB.

tests/test-sysinfo.c
```diff
@@ -33,6 +33,50 @@ int main(void)
...
+     * computes used = total - free - cached. While elfuse capped sysinfo's
+     * total but let /proc/meminfo report the host's real memory, Cached
+     * exceeded MemTotal and that subtraction wrapped, printing a 24-digit
+     * "used". The wrap happens in the guest, past any saturating arithmetic
+     * elfuse does on its own side, so the two surfaces have to agree here.
+     */
+    TEST("sysinfo agrees with /proc/meminfo");
+    {
+        struct sysinfo si;
+        FILE *f = fopen("/proc/meminfo", "r");
```
jserv: `f` leaks if `sysinfo()` fails: it is opened before the check, and `FAIL()` skips the `fclose`. Move the `fopen` below the sysinfo check.

tests/test-sysinfo.c
```diff
@@ -33,6 +33,50 @@ int main(void)
...
+            fclose(f);
+
+            unsigned long long si_total_kb =
+                (unsigned long long) si.totalram * si.mem_unit / 1024;
+
+            if (seen != 7)
+                FAIL("MemTotal/MemFree/Cached missing from /proc/meminfo");
+            else
+                EXPECT_TRUE(
+                    si_total_kb == total_kb && free_kb + cached_kb <= total_kb,
```
jserv: Two gaps. The assertion uses `meminfo.MemFree`, but the computation that broke is `sysinfo.freeram + meminfo.Cached <= sysinfo.totalram`; adding that check is what would catch the drift noted on procemu.c:2061. And on a host with at most 4 GiB of RAM the cap never binds, so the pre-fix code passes this test too. Either note that the assertion is host-dependent, or drive the totals through an injectable model.

jserv (commented):
Reviewed the arithmetic, the init path, the emitted field set and the test.

What holds up: the underlying bug is real and the test catches it. Building HEAD~1 on an 8 GiB host gives `MemTotal: 8388608 kB` against a sysinfo `totalram` of 3998628 kB, and the new assertion fails. The 64-bit overflow this retires in `sysinfo_refresh_cached_locked()` was real too: `scaled_free *= 4094595072` wraps once host free RAM passes 4.19 GiB, which is routine.

Checked and not problems, so nobody re-runs them: `MemAvailable == MemFree + Cached` exactly in both the clamped and unclamped cases, so no Linux invariant breaks; no emitted field exceeds `MemTotal` and none can underflow; there is no division by zero, since `total_kb` is never a divisor; the sscanf formats tolerate the kernel's whitespace padding and `SwapCached` cannot false-match `Cached`; the KiB product in `sys_guest_ram_scale()` stays inside 64 bits until the host holds about 4 PiB; and calling the new helpers under the sysinfo write lock is not an inversion, since they take no lock.

One question above all the inline comments: what is the cap for? `GUEST_RAM_CAP_BYTES` has been there since the initial import, nothing in the tree references it outside `sys.c`, and no test pins it. Everything here exists to service it, including the overflow it introduced. It also has a cost the description does not weigh: a JVM sizing its heap, or a build sizing its parallelism, from `MemTotal` under-commits by 6x on a 24 GiB host. If some guest is known to choke on a large `MemTotal`, say so and the machinery is justified. If it is cosmetic, deleting the cap is a much shorter path to the same agreement.

src/runtime/procemu.c
xalestar: Correct, and it was the wrong pair to clamp. MemFree and freeram were separate `host_statistics64` samples, so the margin the clamp left could not cover the drift between them.

`sys_guest_ram_free()` exports the guest-visible free the same way as the total: it returns freeram out of the cached sysinfo snapshot, and `proc_open_meminfo()` uses it both for MemFree and as the clamp denominator. `sys_sysinfo()` reads that same snapshot through a shared `sysinfo_snapshot()`, so the two surfaces come from one refresh rather than two samples.

One run on this host reports `sysinfo freeram=325696 kB` and `MemFree: 325696 kB`, and busybox `free` gives 25165824 - 282896 - 8355696 = 16527232 for used.

Two of the properties you checked and listed so nobody re-runs them are touched by this, so both are re-checked. `MemAvailable == MemFree + Cached` still holds exactly: the Cached clamp can only fire when free plus inactive plus purgeable exceeds the total, which is the same condition that caps MemAvailable at the total, so the two fire together and both sides land on MemTotal. The no-inversion argument changes rather than survives: `sys_guest_ram_free()` does take `sysinfo_lock`, but `sysinfo_refresh_cached_locked()` no longer calls any helper that locks, and `proc_intercept_open()` runs before `fd_alloc()` with no lock held, so the lock stays a leaf on both paths.

Two costs worth naming. MemFree is up to a second stale now, since it comes from the sysinfo cache instead of a fresh sample; that staleness is what buys the agreement. And a guest whose two reads straddle a cache refresh can still see freeram from one snapshot against Cached from the next. Making that impossible would mean freezing one of two independently timed reads of a moving figure, which a real kernel does not do either; the snapshot removes the drift for adjacent reads, which is what busybox performs.

src/runtime/procemu.c
xalestar: Reworded. Active is an exact remainder once Cached is clamped, so its saturation is gone. AnonPages keeps its saturation, because Buffers sits outside the clamp and the fallback branch sets it to total/20.

src/syscall/sys.c
xalestar: The zero path goes with the cap. `sysinfo_init_cached_host_state()` substitutes `GUEST_RAM_FALLBACK_BYTES` when the sysctl fails or reports zero, so `cached_totalram` is never zero and every surface still describes a machine that exists.

src/syscall/sys.c
xalestar: Both retired with the helper. `sys_guest_ram_scale()` is deleted, so there is no postcondition to make unconditional and no composed truncation to quantify.

tests/test-sysinfo.c
xalestar: Fixed. The `fopen` moved below the sysinfo check, so the failure path returns before a FILE is opened.

tests/test-sysinfo.c
xalestar: The assertion is stated in sysinfo's terms now: `si.freeram + Cached <= si.totalram`, alongside `si.totalram == MemTotal`. Exact equality of MemFree and freeram is deliberately not asserted, since under a real kernel the two are separate samples of a moving figure.

Host-dependence is noted in the comment rather than modelled away, since the totals are whatever the machine has. It is not vacuous here: the same test binary run against a build of `main` fails the first assertion on this 24 GiB host. On a host with 4 GiB or less it would pass either way, which the comment says.

xalestar:
The cap is cosmetic, so it is deleted rather than justified. Nothing outside `sys.c` reads it, no test pins it, and I have no guest that chokes on a large MemTotal -- a 24 GiB host reporting 3998628 kB was the only thing it bought. Both surfaces take the total from `sys_guest_ram_bytes()`, which is HW_MEMSIZE, and the scaling helper and its overflow go with it.

The clamp stays, because it does not depend on the cap: the Mach counters double-count purgeable pages inside inactive, so free plus inactive plus purgeable can exceed physical memory whatever the total is derived from.

Rebased onto `main` at 3541d18.

PR #302, opened 2026-08-16
Add pull and inspect OCI capabilities

henrybear327:
We will introduce the go toolchain in this PR, as we rely on the existing libraries to perform the job.

---
## Summary by cubic
Adds `elfuse-oci`, a standalone Go CLI to pull and inspect OCI images into a spec-compliant local store, keeping `elfuse` OCI-agnostic. New behavior: pins images by reference and platform in a durable, lock-serialized store; default platform is linux/arm64, non-linux is rejected.

- Store: OCI image-layout plus `refs.json` (ref→manifest digest keyed by `os/arch[/variant]`). Metadata writes are durable (temp+fsync+rename+fsync dir). An exclusive `.lock` flock serializes `index.json` and pin updates; corrupt `refs.json`/`index.json` fail closed. Index descriptors record platform. Blob writes and fsync happen before index/pin and heal missing/truncated blobs.
- Pull: Uses `github.com/google/go-containerregistry`; supports `--platform` and `--timeout` (bounds the whole pull). Validates the pulled config matches the requested platform (single-manifest refs cannot silently mismatch). Progress prints to stderr. Credentials come from the default keychain, time-bounded with an actionable error that suggests `DOCKER_CONFIG` for anonymous pulls.
- Inspect: Prints a human summary or `--json` the raw config blob (not re-marshaled). Write errors propagate. Includes `help` and `version`.
- Build/CI: `make all` builds `elfuse-oci` only when `GO` is on PATH; `make oci-test` runs package tests. Adds a Linux CI job to vet and test the CLI. No impact to `elfuse` on Go-less hosts.
- Defaults: Store is `$ELFUSE_OCI_STORE` or `~/.local/share/elfuse/oci`.

**Rollout**
- Install Go to build/test `elfuse-oci` (`make oci-test`). Existing `elfuse` workflows remain unchanged.
- Use `build/elfuse-oci pull [--store DIR] [--platform] [--timeout DUR] <ref>` and `build/elfuse-oci inspect [--store DIR] [--platform] [--json] <ref>`. Set `$ELFUSE_OCI_STORE` to override the store location.

<sup>Written for commit 0f591a47cc92b6a611fa4b350ebeb44c8b29f6de. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/302?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


cmd/elfuse-oci/pull.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/google/go-containerregistry/pkg/v1"
+)
+
+func platformOption(cf commonFlags) crane.Option {
+	p := v1.Platform{
+		OS:           cf.platform.OS,
+		Architecture: cf.platform.Arch,
+		Variant:      cf.platform.Variant,
+	}
+	return crane.WithPlatform(&p)
```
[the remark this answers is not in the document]
henrybear327: Good call. Will reject non-Linux OS pulls in `parsePlatform`. We don't reject by ARCH since Rosetta hosts amd64 guests.

cmd/elfuse-oci/flock.go
```diff
@@ -0,0 +1,98 @@
...
+			if errors.Is(err, syscall.ENOENT) {
+				continue // unlinked concurrently; retry on the recreated file
+			}
+			return nil, fmt.Errorf("lock %s: %w", path, err)
+		}
+		if err := syscall.Fstat(int(f.Fd()), &fdSt); err != nil {
+			f.Close()
+			return nil, fmt.Errorf("lock %s: %w", path, err)
+		}
+		if pathSt.Dev == fdSt.Dev && pathSt.Ino == fdSt.Ino {
```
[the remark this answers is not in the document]
henrybear327: Nothing in this PR removes the store .lock or any tree containing it; the identity check targets the acquisition-time unlink race only, and the header comment assigns guard/order contracts to each consumer. The follow-up PR that does sweep removable trees serializes removal with its own bundle locks and will be reviewed there.

Makefile
```diff
@@ -132,6 +141,27 @@ elfuse: $(ELFUSE_BIN)
...
+elfuse-oci: $(OCI_BIN)
+
+# Go unit tests for the OCI image CLI (offline).
+.PHONY: oci-test
+oci-test:
+	$(Q)$(GO) test -race ./cmd/elfuse-oci/
+
+# rm -f first: `go build -o` follows an existing symlink at the output path,
+# so a stale build/elfuse-oci symlink would clobber build/elfuse.
+$(OCI_BIN): go.mod go.sum $(OCI_SRCS) $(VERSION_DEPS) | $(BUILD_DIR)
```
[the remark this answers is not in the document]
henrybear327: The rule already lists $(VERSION_DEPS) (.git/HEAD, .git/index, mk/config.mk), the same mechanism build/version.h uses for the C binary, so git-state version changes do rebuild. A command-line VERSION= override skipping the rebuild is a pre-existing repo-wide property shared with version.h, out of scope here.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,414 @@
...
+	d, err := img.Digest()
+	if err != nil {
+		return "", fmt.Errorf("store: compute manifest digest: %w", err)
+	}
+	h, err := v1.NewHash(d.String())
+	if err != nil {
+		return "", err
+	}
+	err = s.withLock(func() error {
+		present, err := s.hasImageLocked(h)
```
jserv, line 328: Presence is decided from `index.json` alone, so a store damaged by a crash can never heal.

Remove one layer blob and re-pull: it fails with `store: sync image blobs: open .../blobs/sha256/<hex>: no such file or directory`, and it fails identically on every later pull, because `appendImageLocked` is skipped and the blob is never refetched. The ref is stuck permanently.

Zero-filling that same blob at its original size is worse: the re-pull succeeds and pins the corrupt content, since ggcr's size check in `layout/write.go:216` only rewrites a blob whose size differs.

Make presence mean "descriptor in `index.json` and every blob it names present at the right size", so a damaged image falls back to `appendImageLocked`.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,414 @@
...
+	im, err := ii.IndexManifest()
+	if err != nil {
+		return err
+	}
+	im.Manifests = append(im.Manifests, *desc)
+	b, err := json.MarshalIndent(im, "", "  ")
+	if err != nil {
+		return err
+	}
+	if err := writeFileDurable(s.indexPath(), b, 0o644); err != nil {
```
jserv, line 429: `index.json` commits durably here, but the layer blobs it names are fsynced later, in `addImage`'s `syncImageBlobs` call at line 308. `WriteImage` above leaves them rename-atomic and nothing more, so the index that names them commits first.

A crash in this window is exactly what produces the unrecoverable store described above. The ordering the design comment claims holds for the pin, but it is inverted for the index.

Move the blob fsync ahead of the index write: call `syncImageBlobs` at the end of `appendImageLocked`, before `writeFileDurable(s.indexPath(), ...)`.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,414 @@
...
+func (s *store) addImage(ref string, img v1.Image) (string, error) {
+	d, err := img.Digest()
+	if err != nil {
+		return "", fmt.Errorf("store: compute manifest digest: %w", err)
+	}
+	h, err := v1.NewHash(d.String())
+	if err != nil {
+		return "", err
+	}
+	err = s.withLock(func() error {
```
jserv, line 327: `crane.Pull` returns a lazy image, so the layer download happens inside this lock, at `s.path.WriteImage` on line 357. The exclusive store flock is held for the whole network transfer, not just the index and pin update the lock comment describes.

One slow or hung registry therefore blocks every other `elfuse-oci` process, including a first-run `openStore` bootstrap. There is no context or deadline on the pull either, so "slow" has no upper bound.

Blobs are content-addressed, so write them unlocked and take the lock only around `hasImageLocked`, the index write, and the pin. Pass `crane.WithContext` with a timeout while you are here.

cmd/elfuse-oci/pull.go
```diff
@@ -0,0 +1,47 @@
...
+	// Progress goes to stderr, keeping stdout reserved for command output
+	// that callers capture. Printed before the pull so a slow or
+	// credential-blocked registry is not silent.
+	fmt.Fprintf(os.Stderr, "Pulling %s...\n", ref)
+
+	// Wrap the ambient keychain so a wedged credential helper fails fast with
+	// an explanation instead of hanging the pull; see timedKeychain.
+	keychain := crane.WithAuthFromKeychain(
+		timedKeychain{authn.DefaultKeychain, credResolveTimeout})
+	img, err := crane.Pull(ref, platformOption(cf), keychain)
```
jserv: `crane.WithPlatform` applies only when the ref resolves to an index. In ggcr v0.21.7, `remote/descriptor.go:108` returns a plain `OCIManifestSchema1` or `DockerManifestSchema2` unchecked, so a single-manifest ref pins whatever OS and arch the registry serves.

The parse-time linux check in `parsePlatform` does not cover that path, and `docs/oci-design.md:39` states it does.

Read `img.ConfigFile()` after the pull and reject when `cfg.OS` or `cfg.Architecture` disagrees with `cf.platform`.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,414 @@
...
+		}
+		if !present {
+			if err := s.appendImageLocked(img); err != nil {
+				return err
+			}
+		}
+		if err := s.syncImageBlobs(img); err != nil {
+			return fmt.Errorf("store: sync image blobs: %w", err)
+		}
+		return s.pinLocked(ref, d.String())
```
jserv: The pin is keyed by ref alone, so `pull --platform linux/amd64 alpine:3` silently replaces the `linux/arm64` pin for the same ref.

`index.json` cannot disambiguate the two manifests either: `partial.Descriptor` on line 360 leaves `Platform` nil. The store cannot hold both variants that `--platform` invites users to pull, which matters given the stated Rosetta amd64 case.

Key the pin on ref plus platform (or make the value a struct carrying digest and platform), and set `desc.Platform` from the resolved config before appending.

Makefile
```diff
@@ -132,6 +141,27 @@ elfuse: $(ELFUSE_BIN)
...
+# so it also builds under Linux for spec-conformance / interop CI.
+OCI_BIN := $(BUILD_DIR)/elfuse-oci
+OCI_SRCS := $(shell find cmd/elfuse-oci -type f -name '*.go' ! -name '*_test.go' 2>/dev/null)
+
+.PHONY: elfuse-oci
+elfuse-oci: $(OCI_BIN)
+
+# Go unit tests for the OCI image CLI (offline).
+.PHONY: oci-test
+oci-test:
```
jserv, line 155: Nothing runs this. `oci-test` is absent from `check` (`mk/tests.mk:204`) and from every workflow, and no job builds `build/elfuse-oci`, so the new Go package and its test suite are never compiled or exercised in CI. A regression here lands green.

Add `oci-test` to `check` behind `HAVE_GO`, or add a Linux job running `go vet ./cmd/elfuse-oci/ && make oci-test`. The package is pure Go, so a Linux runner covers it.

go.mod
```diff
@@ -0,0 +1,17 @@
+module github.com/sysprog21/elfuse
+
+go 1.26.4
```
jserv: `go 1.26.4` is patch-exact, while `Makefile:112` only tests that some `go` is on PATH. A host with 1.25 either silently downloads a toolchain over the network or, under `GOTOOLCHAIN=local`, hard-fails the build that `HAVE_GO` was meant to protect.

Use the language version actually needed (`go 1.26`, or lower), and if a floor matters, check `go version` in `HAVE_GO` rather than mere presence.

cmd/elfuse-oci/flock.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"errors"
+	"fmt"
+	"os"
+	"syscall"
+)
+
+// errCacheBusy reports that a non-blocking acquisition lost to a live
+// holder, so the caller must not remove what the lock guards.
+var errCacheBusy = errors.New("in use by a live run")
```
jserv: `errCacheBusy`, the `LOCK_NB` branch, and the 16-round unlink-race retry have no caller: `store.lock` only ever passes `LOCK_EX`, and nothing in the tree removes `.lock`.

That is a third of the file guarding a sweeper that does not exist, and none of it is reachable from a test. Keep the EINTR retry and drop the rest until a remove or gc command actually needs it.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,414 @@
...
+}
+
+// appendImageLocked writes img's blobs and appends its manifest
+// descriptor to index.json; the caller holds the store lock. Config,
+// manifest, and index.json go through writeFileDurable, not the layout
+// package's AppendImage, which writes all three in place: a crash there
+// leaves a partial blob a later pull accepts as complete, or a truncated
+// index the store fails closed on forever. Layer blobs stay with
+// WriteImage: its temp-and-rename write is atomic, and the pinned
+// go-containerregistry version skips existing blobs without a size
```
jserv: The stated reason does not hold for the pinned version. `layout/write.go:216` in v0.21.7 does size-check an existing blob (`s.Size() == size || size == -1`) before skipping it. The durable copies stand because their size matches, not because the check is absent.

Worth correcting, since that same size check is what makes finding on line 299 narrower than it would otherwise be.

jserv (commented):
Reviewed the store, lock, durability, platform, and build wiring paths end to end against the branch. `go vet`, `gofmt -l`, and `go test -race ./cmd/elfuse-oci/` are all clean on go1.26.6.

The design decisions in the commit message mostly hold, with one cluster that does not: the crash-recovery story. Findings 1 and 2 are the same root cause seen from two ends, and I reproduced both against the real store rather than reasoning about them. The rest are independent.

Comments are inline.

jserv (changes requested):
Rebase latest `main` branch as #313 installs Git hooks and a CI gate to enforce commit‑message rules and staged checks.

henrybear327:
As discussed offline, the goal is just to leverage OCI image to build sysroot, no to support OCI commands for managing the images, and manage the lifecycle. 

Closing this PR, in favor of a new rewrite, where the go logic code should be in ~1000 LoC.

PR #305, opened 2026-08-17
Dispatch write, writev, and readv on netlink fds

xalestar:
`busybox ip addr` under elfuse exits 1 on its first rtnetlink request:

```
$ build/elfuse --sysroot externals/test-fixtures/rootfs \
      externals/test-fixtures/rootfs/bin/busybox ip addr
ip: write error: Bad file descriptor
```

The verbose trace has the socket, the bind and the getsockname reaching the netlink handlers, and the request itself failing:

```
syscall 198 socket(16, 3, 0)    -> 3
syscall 200 bind(3, ...)        -> 0
syscall 204 getsockname(3, ...) -> 0
syscall 64  write(3, buf, 0x14) -> -9
```

`netlink_socket()` registers `pipefd[0]` as the host fd so that poll and epoll have something to wait on. `sys_write()` and `sys_writev()` carry no `FD_NETLINK` case, so the request reaches that pipe end, which is open read-only, and macOS returns EBADF. `sys_readv()` has the same gap for `iovcnt > 1`; a single entry already delegates to `sys_read()`.

Linux accepts `write()` on a netlink socket as `sendto()` with no explicit destination, and busybox's libiproute sends every request that way. glibc `getifaddrs()` uses `sendto` and `recvmsg`, both already routed, which is why the existing coverage passes over the gap.

Scope: all eight spellings (`sendto`, `sendmsg`, `write`, `writev`, `recvfrom`, `recvmsg`, `read`, `readv`) reach one gather and one scatter, so a request split across iov entries is not truncated, a receive is not capped by its first entry, and `nl_complete_span()` bounds all four receive spellings alike.

`tests/test-netlink.c` drives each of the eight against a `NETLINK_ROUTE` socket and then the end-to-end `getifaddrs()` path that originally regressed. Every receive polls before it reads, so a regression fails an assertion instead of blocking the run, and the assertions hold for a real kernel too and run in the qemu column of the matrix.

## Verification

Verified on macOS 26.6.1 (build 25G76), SDK 26.5, Apple M5 Pro, on a branch based on `upstream/main` at 3541d18.

`make test-matrix` stays within baseline in all three modes with zero failures: elfuse-aarch64 252 passed, qemu-aarch64 231 passed, elfuse-x86_64 78 passed. `test-netlink` is OK in the qemu-aarch64 column as well as the elfuse-aarch64 one, so all 22 assertions hold against a real 6.18 kernel.

`make lint`, `.ci/check-format.sh`, `.ci/check-newline.sh`, `.ci/check-security.sh` and `.ci/check-matrix-lists.sh` pass. `.ci/check-cppcheck.sh` needs `mapfile`, so it is left to the CI runner rather than this host's bash 3.2.

`make check` passes the unit suite and reports 81 of 84 in the BusyBox smoke, with one failure that is this host and not the branch: `nslookup` times out because no DNS server is reachable from here, and it fails the same way with the branch's changes backed out.

The new assertions fail without the change rather than passing vacuously. Against the branch's own sources backed out, the four that cover `sendmsg` and `recvmsg`:

```
FAIL: sendmsg() gathers a split request (errno=22 Invalid argument)
FAIL: recvmsg() fills past the first entry (errno=22 Invalid argument)
FAIL: the scattered recvmsg() bytes parse as RTM_NEWLINK (errno=22 Invalid argument)
FAIL: sendmsg() of an empty iovec fails with ENODATA (errno=22 Invalid argument)
```

With only the `recvmsg` scatter removed, so the two receive assertions stand on their own rather than on the send above them:

```
FAIL: recvmsg() fills past the first entry (errno=61 No data available)
FAIL: the scattered recvmsg() bytes parse as RTM_NEWLINK (errno=61 No data available)
```

Removing the `sys_write` and `sys_writev` dispatch:

```
FAIL: write(RTM_GETLINK) accepts the request (errno=9 Bad file descriptor)
FAIL: write() request produces an RTM_NEWLINK dump (errno=9 Bad file descriptor)
FAIL: write() of zero bytes fails with ENODATA (errno=0 Success)
FAIL: writev() of zero bytes returns 0 (errno=9 Bad file descriptor)
FAIL: writev(RTM_GETLINK) accepts a split request (errno=9 Bad file descriptor)
FAIL: writev() gathers the split request into one dump (errno=9 Bad file descriptor)
```

Removing the `sys_readv` dispatch leaves the readv drain empty, and the recvmsg pair after it fails on the buffer that drain would have emptied:

```
FAIL: readv() fills past the first entry (errno=61 No data available)
FAIL: the scattered readv() bytes parse as RTM_NEWLINK (errno=61 No data available)
FAIL: recvmsg() fills past the first entry (errno=61 No data available)
FAIL: the scattered recvmsg() bytes parse as RTM_NEWLINK (errno=61 No data available)
```

### DIFF
--- a/src/syscall/io.c
+++ b/src/syscall/io.c
@@ -1005,6 +1005,15 @@ int64_t sys_write(guest_t *g, int fd, uint64_t buf_gva, uint64_t count)
     if (type == FD_EVENTFD)
         return eventfd_write(fd, g, buf_gva, count);
 
+    /* Linux accepts write() on a netlink socket as sendto() with no explicit
+     * destination, and iproute2-style senders use it in place of sendto. The
+     * host fd behind a netlink guest fd is the read end of the readiness pipe,
+     * so falling through would write to a read-only pipe end and report EBADF
+     * for a request the emulation can answer.
+     */
+    if (type == FD_NETLINK)
+        return netlink_send(fd, g, buf_gva, count);
+
     host_fd_ref_t host_ref;
     int64_t err = host_fd_ref_open_checked(fd, &host_ref, true);
     if (err < 0)
@@ -1411,6 +1420,8 @@ int64_t sys_readv(guest_t *g, int fd, uint64_t iov_gva, int iovcnt)
 {
     if (iovcnt == 0)
         return vec_zero_iovcnt(fd, false, false);
+    if (fd_get_type(fd) == FD_NETLINK)
+        return netlink_readv(fd, g, iov_gva, iovcnt);
     if (iovcnt == 1) {
         linux_iovec_t giov;
         int64_t err = single_guest_iov(g, iov_gva, &giov);
@@ -1564,6 +1575,11 @@ int64_t sys_writev(guest_t *g, int fd, uint64_t iov_gva, int iovcnt)
 {
     if (iovcnt == 0)
         return vec_zero_iovcnt(fd, true, false);
+    /* Ahead of the single-entry shortcut: a one-entry writev of nothing reports
+     * 0, while the write(2) the shortcut would reach reports ENODATA.
+     */
+    if (fd_get_type(fd) == FD_NETLINK)
+        return netlink_writev(fd, g, iov_gva, iovcnt);
     if (iovcnt == 1) {
         linux_iovec_t giov;
         int64_t err = single_guest_iov(g, iov_gva, &giov);
@@ -1577,7 +1593,8 @@ int64_t sys_writev(guest_t *g, int fd, uint64_t iov_gva, int iovcnt)
      * iovs) because the data is at giov.iov_base which is only giov.iov_len
      * bytes. eventfd expects exactly 8 bytes.
      */
-    if (fd_get_type(fd) == FD_EVENTFD) {
+    int wtype = fd_get_type(fd);
+    if (wtype == FD_EVENTFD) {
         if (iovcnt <= 0)
             return -LINUX_EINVAL;
         linux_iovec_t giov;
--- a/src/syscall/netlink.c
+++ b/src/syscall/netlink.c
@@ -112,6 +112,13 @@ typedef struct {
 #define MAX_NETLINK_FDS 16
 #define NETLINK_BUF_SIZE 8192
 
+/* Ceiling on one staged request. Linux bounds a send by sk_sndbuf and reports
+ * EMSGSIZE past it; this is that refusal against a fixed buffer. The largest
+ * request nl_process_request() reads is an RTM_GETLINK carrying an IFLA_IFNAME
+ * filter, an nlmsghdr and an ifinfomsg and an attribute holding IFNAMSIZ.
+ */
+#define NETLINK_REQ_MAX 512
+
 typedef struct {
     bool in_use;
     int guest_fd;                  /* Guest fd number */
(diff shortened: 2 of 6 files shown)


src/syscall/io.c
```diff
@@ -1005,6 +1005,15 @@ int64_t sys_write(guest_t *g, int fd, uint64_t buf_gva, uint64_t count)
...
         return eventfd_write(fd, g, buf_gva, count);
 
+    /* Linux accepts write() on a netlink socket as sendto() with no explicit
+     * destination, and iproute2-style senders use it in place of sendto. The
+     * host fd behind a netlink guest fd is the read end of the readiness pipe,
+     * so falling through would write to a read-only pipe end and report EBADF
+     * for a request the emulation can answer.
+     */
+    if (type == FD_NETLINK)
+        return netlink_send(fd, g, buf_gva, count);
```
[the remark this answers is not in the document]
xalestar: Measured on Linux 6.18 under qemu-aarch64: `write(fd, buf, 0)` returns -1/ENODATA, not 0. A zero-length `sendto` does the same, while a `writev` of zero total returns 0, since `do_readv_writev` returns before the socket is reached. The dispatch reproduces all three and `tests/test-netlink.c` asserts them in both the elfuse and qemu columns.

src/syscall/netlink.c
```diff
@@ -692,31 +702,62 @@ int64_t netlink_send(int guest_fd, guest_t *g, uint64_t buf_gva, uint64_t len)
...
-    uint8_t req[512];
-    size_t rlen = (len < sizeof(req)) ? len : sizeof(req);
-    if (guest_read(g, buf_gva, req, rlen) < 0) {
-        result = -LINUX_EFAULT;
+    if (total < (uint64_t) NLMSG_HDRLEN) {
+        result = (int64_t) total;
         goto out;
     }
 
+    uint8_t req[512];
```
jserv: The gather stops at 512 bytes but the call reports `total` as accepted. Entries past the cut are never read, so their `iov_base` is never validated: a `writev` with a 600-byte first entry and an unmapped second entry returns `600 + len` instead of EFAULT, and the request itself is silently truncated before `nl_process_request()` parses it. Either read every entry and reject past the cap, or return only the bytes actually consumed.

src/syscall/netlink.c
```diff
@@ -997,21 +1053,36 @@ int64_t netlink_read(int guest_fd, guest_t *g, uint64_t buf_gva, uint64_t count)
...
+    for (int i = 0; i < iovcnt && ns->buf_pos < ns->buf_len; i++) {
+        size_t avail = ns->buf_len - ns->buf_pos;
+        size_t to_copy =
+            (avail < iov[i].iov_len) ? avail : (size_t) iov[i].iov_len;
+        if (to_copy == 0)
+            continue;
+        if (guest_write(g, iov[i].iov_base, ns->buf + ns->buf_pos, to_copy) <
+            0) {
+            pthread_mutex_unlock(&nl_lock);
+            /* Bytes already placed are transferred; reporting EFAULT over them
```
jserv: `guest_write()` copies chunk by chunk and can fail after placing part of this entry (`guest_copy()`, src/core/guest.c:1735). Those bytes land in guest memory but are counted in neither `done` nor `buf_pos`, so the invariant this comment states holds only at whole-entry granularity; Linux counts what `copy_to_iter()` placed. A `guest_write` variant reporting bytes copied would close the gap.

src/syscall/netlink.c
```diff
@@ -681,8 +681,18 @@ int64_t netlink_sendmsg(int guest_fd, guest_t *g, uint64_t msg_gva, int flags)
     return result;
 }
 
-/* sendto(2) on a netlink socket: a flat request buffer (no msghdr). */
-int64_t netlink_send(int guest_fd, guest_t *g, uint64_t buf_gva, uint64_t len)
+/* sendto(2), write(2) and writev(2) on a netlink socket.
```
jserv: `netlink_sendmsg()` is now the odd one out: it still reads only `msg_iov[0]`, still returns EINVAL for `msg_iovlen == 0` and for a first entry under `NLMSG_HDRLEN`, and reports `iov.iov_len` rather than a gathered total. Staging the entries and calling `netlink_send_iov()` deletes that copy and makes all four send paths agree.

src/syscall/netlink.c
```diff
@@ -976,7 +1017,13 @@ int64_t netlink_recvmsg(int guest_fd, guest_t *g, uint64_t msg_gva, int flags)
     return (int64_t) msg_end;
 }
 
-int64_t netlink_read(int guest_fd, guest_t *g, uint64_t buf_gva, uint64_t count)
+/* read(2) and readv(2) on a netlink socket. The response fills every entry in
```
jserv: Same asymmetry on the receive side: `netlink_recvmsg()` fills only `msg_iov[0]`, so a caller that splits its receive buffer gets a short read where `readv` now succeeds. Note also that both recv paths apply `nl_complete_span()` and `readv` does not, so `readv` can hand out a trailing message split across two calls while `recvmsg` never will.

jserv (commented):
Dispatching write/writev/readv on netlink fds is the right fix, and the tests fail without it. Four notes: the 512-byte gather cap reports bytes it never read or validated; the partial-copy comment in `netlink_readv` overstates what the code accounts for; and `netlink_sendmsg`/`netlink_recvmsg` still take the first iov entry only, which the two new helpers would fix.

Also, the PR description carries the commit message twice, and the first copy still says "the three paths take the first iov entry" and cites `netlink_sendmsg()` as the rule, which contradicts the gather that was implemented.

src/syscall/netlink.c
xalestar: The gathered total is checked against the staging buffer before anything is read, so the count reported is bytes that were read and parsed. Past `NETLINK_REQ_MAX` the send is EMSGSIZE, the errno Linux reports past `sk_sndbuf`; inside it every entry is read, so an unmapped entry anywhere in the vector is EFAULT. Measured under elfuse: a `writev` of a valid entry plus an unmapped one is EFAULT, and a 600-byte `write` is EMSGSIZE.

I took the reject option over the short-count one because the truncation is the harder half to notice: a short write on a request the parser then answers truncated looks like backpressure, where a refusal does not. The ceiling is the emulation's own and a real kernel's sits far higher, so this is a refusal rather than a match.

src/syscall/netlink.c
xalestar: `guest_write_partial()` is the `guest_write` variant that reports bytes copied, built on the `guest_host_copy_partial()` from c141b58. `netlink_recv_iov()` advances both `done` and `ns->buf_pos` by what landed, so a copy that faults inside an entry is counted the way `copy_to_iter()` counts it, exact to the one-page race bound `guest_host_copy_partial()` documents. The comment states that rather than the whole-entry version.

src/syscall/netlink.c
xalestar: `netlink_sendmsg()` stages `msg_iov` and calls `netlink_send_iov()`, so it reports the gathered total. An empty vector is ENODATA, since `___sys_sendmsg` hands it to the socket rather than answering it above, and a gathered length under `NLMSG_HDRLEN` transfers: no EINVAL on either. `tests/test-netlink.c` asserts a split request through `sendmsg` and the empty-vector ENODATA, both in the elfuse and the qemu-aarch64 columns.

src/syscall/netlink.c
xalestar: `netlink_recvmsg()` stages `msg_iov` and shares `netlink_recv_iov()` with `recvfrom`, `read` and `readv`, so `nl_complete_span()` bounds all four and no receive hands out a message split across two calls. `netlink.c` stages its own iovec now instead of `io.c` doing it, which is what lets the eight spellings reach one gather and one scatter.

`test-netlink` asserts a 16-byte first entry followed by the rest, and it fails when only the `recvmsg` scatter is removed:

```
FAIL: recvmsg() fills past the first entry (errno=61 No data available)
FAIL: the scattered recvmsg() bytes parse as RTM_NEWLINK (errno=61 No data available)
```

PR #307, opened 2026-08-17
mmap/munmap EL1 fastpath

### LINKED ISSUE #165: sys_mmap: eager zero-fill on MAP_ANONYMOUS makes sparse mappings ~5x slower than Linux

maxliu0:
The primary goal is to reduce VM exits for `mmap` and `munmap`.

### Architecture

截圖 2026-08-18 凌晨2 24 31

Anonymous mappings are lazy: `sys_mmap` only records the region; page-table creation, host memory commit, and zeroing of reused backing all happen at first touch.

Each vCPU owns a private contiguous virtual address arena. When the guest traps into EL1 for `mmap`/`munmap`, EL1 serves the request directly out of its own arena and never leaves EL1. The host's job is to keep those arenas supplied with VA and to reconcile their effects into host-side region/PTE state, both done lazily under `mmap_lock` rather than synchronously per call.

## mmap fast path

EL1 serves exactly one shape:

```
mmap(NULL, len, PROT_READ|PROT_WRITE,
     MAP_PRIVATE|MAP_ANONYMOUS[|MAP_NORESERVE], fd, off)
```

file-backed mapping is excluded. The reason is that EL1 is a freestanding, no-syscall context that cannot read a file or install a page-cache overlay itself — that work is inherently the host's.

The host prepares a contiguous VA region per vCPU. Each vCPU is the sole producer of its own 32-entry publication ring and bump cursor, and publishes the extent into its ring. The host installs the corresponding PTEs.

Arena refill is dynamic, not fixed-size. Any host path that already holds `mmap_lock` — a syscall, a page fault, or a natural VM exit — opportunistically tops up the current thread's arena once its remaining drops under that recent high-water mark. Arena exhaustion, an unsupported shape, a full ring, or a stale arena generation each bail to the ordinary host `mmap` slow path.

## munmap fast path

EL1 validates that the target range belongs to a live arena, clears the covering page-table entries directly, and issues a broadcast TLBI, all without leaving EL1

The address is unusable to the guest the instant EL1 returns. Only after clearing the PTEs and completing the TLBI does EL1 publish a retirement record — address, length, and the arena generation it came from — to a second, separate 32-entry SPSC ring. The host has not been involved yet at that point; its bookkeeping is reconciled lazily by `mmap_fastpath_drain_locked()`, which walks every vCPU's retire ring and commits the deferred cleanup. That drain runs on every natural VM exit as well as on every syscall or fault that needs the lock.

Cross-thread ordering matters here, and A and B are often touching the *same* address, not two unrelated ones: EL1's fast `munmap` scans every vCPU's arena, not just its own, because handing a fresh allocation to another thread to free is the common case. If the host were to drain B's teardown of address X before it has drained A's publication of that same X, the removal would be a no-op (X isn't in the region table yet) and the publication drain right after would insert X as live — reviving, in host bookkeeping, an address EL1 already invalidated. The drain avoids this by snapshotting every vCPU's retire-ring tail first, draining every mmap publication next, and only then consuming retirements up to that snapshot, so within one drain pass a publication a retirement depends on is always applied before the retirement removes it:

```
 vCPU A                                  vCPU B
 ───────                                 ───────
 mmap(len) -> X
 EL1 bump-allocates X
 ring.push(publish X)  ──┐
                         │   guest hands X to B
                         │   (shared pointer / queue)
                         └──────────────────────┐
                                                 ▼
                                        munmap(X)
                                        EL1 clears PTE(X), TLBI
                                        retire.push(X, arena=A)

 ════════════════════ host: mmap_fastpath_drain_locked() ════════════════════
  1. snapshot every retire-ring tail          <- B's entry is already in view
  2. drain ALL mmap publications              <- installs A's region X
  3. consume retirements up to that snapshot  <- removes region X

  region-table never shows X as live after EL1 has already invalidated it.

  (wrong order, for contrast: consume B's retirement first -> no region X
  exists yet, so removal is a no-op; drain A's publication after -> X gets
  inserted and marked live, even though EL1 killed it already.)
```

EL1 self-throttles rather than pinging the host mid-flight. If a producer's retire ring is within one slot of full, `munmap` bails out of the fast path entirely and falls through to the ordinary syscall trap, which drains the ring as a side effect of taking `mmap_lock`. Separately, once a vCPU's own unconsumed retired bytes cross a 256 MiB soft threshold, it sets an advisory `cleanup_requested` flag; this never forces an HVC, it only marks that real cleanup work is waiting for the next natural drain.

### behavior while fork

Before the fork snapshot is taken, the parent-side handler acquires `mmap_lock` through the fork variant of the acquire path, drains every vCPU's publication and retire ring, and waits for any in-flight lazy materialization to finish. It then revokes every per-vCPU arena descriptor. Because sibling vCPUs are already quiesced for the fork snapshot, this revocation cannot race a live EL1 producer.

The child does not inherit the parent's arena or ring layout. The child's main thread begins with a fresh, empty arena rather than a stale one aliased to the parent's.

Any `clone()` call that publishes a stack for the new thread goes through this too. Any anonymous fast-path allocation that has become a live thread stack must remain reachable only through the ordinary stack-lifetime bookkeeping, not through EL1's arena-generation check, so revoking first guarantees no later EL1 fast `munmap` can tear down a live stack range by still recognizing it as belonging to a live arena. Only a `clone()` with a null `child_stack` skips this.

### behavior while exec

`sys_execve()` takes `mmap_lock` immediately before the point of no return. That both closes the EL1 producer gate and drains every vCPU's mmap and munmap rings, so `g->regions[]` and PTE state agree with each other before the address space they describe is destroyed. `guest_reset()` then zeroes the entire shim-data page, arena, both rings.

`exec` does not explicitly do arena's control block comes back. The replacement image's first eligible slow-path `mmap` by calling the refill path and turns the fast path back.

### Frama-C proof coverage

The guest-influenced arena sizing and capacity arithmetic is pulled out of `src/syscall/mem.c` into `src/proved/mmap-fastpath.h` specifically so it can be proved.

Four functions are proved, every input treated as fully guest-influenced (`request_len` is the guest's own mmap length; the history window is built from a sequence of guest-chosen lengths), because a slip here either wedges the allocator by undersizing an arena forever, or lets `mmap_fastpath_request_fits` accept a request that runs past `arena_limit`:

- `mmap_fastpath_request_fits`: whether a `len`-byte request still fits before `limit`, covering the zero-length, sub-block, and block-aligned cases. Proved to never answer "fits" when the aligned start would actually run past `limit`.
- `mmap_fastpath_pow2_clamped`: rounds a target size up to the nearest power of two inside `[MIN, MAX]`.
- `mmap_fastpath_window_max`: the largest of the last 16 registered mapping sizes. Proved as an upper bound over the whole history window.
- `mmap_fastpath_arena_size`: the target arena size given recent history and the request about to be served. Proved to always land in `[MIN, MAX]`, with `-wp-rte` separately closing both multiplication-overflow guards and the division-by-zero case.

92 of 93 discharged goals (the 93rd, `align_up_ok_ensures_rejects_only_on_wrap`, is a pre-existing `align.h` goal that also fails standalone under `make verify-align` in this environment, not something this change introduced).

### Benchmark results
mmap-isolated-a-mmap
mmap-isolated-a-munmap
mmap-isolated-b-materialized
mmap-isolated-c-dirty

**A. One timed mmap/munmap fast-path pair per fresh process**

All values are nanoseconds. Speedup is `OrbStack / elfuse`; values above 1
mean elfuse is faster.

| Size | elfuse mmap | OrbStack mmap | mmap speedup | elfuse munmap | OrbStack munmap | munmap speedup |
| 4 KiB | 50.8 | 266.7 | 5.25x | 47.1 | 268.8 | 5.71x |
| 16 KiB | 52.7 | 245.7 | 4.66x | 51.7 | 251.3 | 4.86x |
| 64 KiB | 51.4 | 281.3 | 5.47x | 51.4 | 294.1 | 5.72x |
| 256 KiB | 57.4 | 275.6 | 4.80x | 55.9 | 313.0 | 5.60x |
| 1 MiB | 54.7 | 259.3 | 4.74x | 54.3 | 359.5 | 6.62x |
| 2 MiB | 52.0 | 394.6 | 7.59x | 54.8 | 267.5 | 4.88x |
| 8 MiB | 52.4 | 256.6 | 4.90x | 52.4 | 282.4 | 5.39x |
| 64 MiB | 54.7 | 248.4 | 4.54x | 56.8 | 335.5 | 5.91x |
| 256 MiB | 63.7 | 313.7 | 4.92x | 57.1 | 574.8 | 10.07x |
| 1 GiB | 59.0 | 262.9 | 4.46x | 52.1 | 834.0 | 16.01x |
| 4 GiB | 59.2 | 273.5 | 4.62x | 53.2 | 904.9 | 17.01x |
| 16 GiB | 57.6 | 273.9 | 4.76x | 52.1 | 856.1 | 16.43x |
| 32 GiB | 60.7 | 250.5 | 4.13x | 54.0 | 973.4 | 18.03x |

## B. munmap after materializing one 4 KiB page

Fifteen fresh processes are used per size. Values are outer medians of the
per-process medians.

| Size | elfuse | OrbStack | speedup |
|---|---:|---:|---:|
| 4 KiB | 357.0 | 608.3 | 1.70x |
| 16 KiB | 442.1 | 568.3 | 1.29x |
| 64 KiB | 384.9 | 622.7 | 1.62x |
| 256 KiB | 388.3 | 621.0 | 1.60x |
| 1 MiB | 430.7 | 730.4 | 1.70x |
| 2 MiB | 430.7 | 1429.2 | 3.32x |
| 8 MiB | 422.8 | 1245.5 | 2.95x |
| 64 MiB | 397.9 | 1366.5 | 3.43x |
| 256 MiB | 400.3 | 1591.6 | 3.98x |
| 1 GiB | 441.1 | 2481.6 | 5.63x |
| 4 GiB | 409.7 | 2436.4 | 5.95x |
| 16 GiB | 298.3 | 2376.5 | 7.97x |
| 32 GiB | 295.7 | 2290.2 | 7.75x |

## C. munmap after dirtying every 4 KiB page

One fresh process is used per size. Each process performs one warmup, then the
listed number of timed operations. The table reports the in-process
distribution.

| Size | elfuse p50 | elfuse p95 | elfuse max | OrbStack p50 | OrbStack p95 | OrbStack max | p50 speedup |
|---|---:|---:|---:|---:|---:|---:|---:|
| 4 KiB | 241.3 | 532.9 | 6032.9 | 573.5 | 1031.9 | 6073.5 | 2.38x |
| 16 KiB | 324.6 | 616.3 | 10032.9 | 532.8 | 1976.5 | 8741.1 | 1.64x |
| 64 KiB | 282.9 | 532.9 | 8241.3 | 906.9 | 2142.3 | 17490.2 | 3.21x |
| 256 KiB | 282.9 | 574.6 | 8866.3 | 1948.4 | 3429.7 | 9448.4 | 6.89x |
| 1 MiB | 324.6 | 574.6 | 3491.3 | 5531.7 | 7152.6 | 19240.1 | 17.04x |
| 2 MiB | 282.9 | 532.9 | 7241.3 | 11052.6 | 15208.8 | 20156.7 | 39.07x |
| 8 MiB | 324.6 | 712.1 | 2032.9 | 51448.4 | 61521.3 | 65115.1 | 158.50x |
| 64 MiB | 657.7 | 1895.2 | 158116.0 | 423532.0 | 483825.7 | 676823.6 | 643.96x |
| 256 MiB | 1137.1 | 1395.4 | 1574.6 | 1812344.2 | 2048150.5 | 2105948.4 | 1593.83x |
| 1 GiB | 1523.0 | 3639.7 | 4106.3 | 7029220.3 | 7461557.8 | 7482907.8 | 4615.38x |

`munmap` latency has a step right at the 2 MiB boundary. Under 2 MiB, which walks and clears one 4 KiB L3 leaf at a time.

Arena size is clamped to `MMAP_FAST_ARENA_MIN = 64 MiB` and `MMAP_FAST_ARENA_MAX = 32 GiB` per vCPU.

`tests/bench-mmap-isolated all` to reproduce mmap/munmap-vs-size numbers

Closes #165

### DIFF
--- a/src/core/shim-mmap.c
+++ b/src/core/shim-mmap.c
@@ -1,0 +1,221 @@
+/*
+ * EL1 mmap syscall fast path.
+ *
+ * Copyright 2026 elfuse contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * This file is compiled freestanding and linked into the shim image.  It may
+ * not call host code or a C runtime: the only state it consumes is the saved
+ * EL0 register frame, TPIDR_EL1's shim-data mapping, and the shared mmap
+ * control protocol.  Keep architecture-only operations in the small helper
+ * below; the allocator policy remains ordinary C.
+ */
+
+#include "core/shim-mmap.h"
+
+#include <stdatomic.h>
+#include <stdint.h>
+
+#include "core/mmap-fastpath.h"
+#include "core/shim-globals.h"
+
+#define EL1_PAGE_SIZE 0x1000ULL
+#define EL1_BLOCK_SIZE 0x200000ULL
+#define EL1_SHIM_DATA_SIZE 0x200000ULL
+
+#define EL1_SYS_MMAP 222ULL
+
+#define EL1_PROT_READ 1ULL
+#define EL1_PROT_WRITE 2ULL
+#define EL1_PROT_RW (EL1_PROT_READ | EL1_PROT_WRITE)
+
+#define EL1_MAP_PRIVATE 0x02ULL
+#define EL1_MAP_ANONYMOUS 0x20ULL
+#define EL1_MAP_NORESERVE 0x4000ULL
+
+/* The bump path promises 2 MiB-aligned starts at or above this request size,
+ * so the host can back them with L2 blocks. */
+#define EL1_ALIGN_THRESHOLD EL1_BLOCK_SIZE
+
+typedef struct {
+    uint8_t *shim_data;
+    shim_mmap_control_t *control;
+} el1_mmap_context_t;
+
+_Static_assert(EL1_SAVED_GPRS * sizeof(uint64_t) == 248,
+               "saved GPR frame layout changed");
+_Static_assert(sizeof(shim_mmap_entry_t) == 24,
+               "EL1 mmap publication ABI changed");
+
+static inline uint64_t el1_read_tpidr(void)
+{
+    uint64_t value;
+    __asm__ volatile("mrs %0, tpidr_el1" : "=r"(value));
+    return value;
+}
+
+static inline bool el1_add_overflow(uint64_t left,
+                                    uint64_t right,
+                                    uint64_t *result)
+{
+    return __builtin_add_overflow(left, right, result);
+}
+
+static bool el1_page_round(uint64_t length, uint64_t *rounded)
+{
+    uint64_t value;
+    if (length == 0 || el1_add_overflow(length, EL1_PAGE_SIZE - 1, &value))
+        return false;
+    *rounded = value & ~(EL1_PAGE_SIZE - 1);
+    return true;
+}
+
+/* SP_EL1 stack tops are shim_data_end - slot*4KiB and controls are
+ * shim_data + 0x20000 + slot*2KiB, so the saved frame's page locates the
+ * control without consuming another system register. */
+static el1_mmap_context_t el1_context(uint64_t *saved_gprs)
+{
+    uintptr_t shim_data = (uintptr_t) el1_read_tpidr();
+    uintptr_t frame = (uintptr_t) saved_gprs;
+    uintptr_t stack_top =
+        (frame + EL1_PAGE_SIZE - 1) & ~(uintptr_t) (EL1_PAGE_SIZE - 1);
+    uintptr_t slot_bytes = shim_data + EL1_SHIM_DATA_SIZE - stack_top;
+    unsigned slot = (unsigned) (slot_bytes / EL1_PAGE_SIZE);
+    return (el1_mmap_context_t) {
+        .shim_data = (uint8_t *) shim_data,
+        .control = (shim_mmap_control_t *) (shim_data + SHIM_MMAP_CONTROL_BASE +
+                                            (uintptr_t) slot *
+                                                SHIM_MMAP_CONTROL_STRIDE),
+    };
+}
+
+static inline bool el1_stats_enabled(const el1_mmap_context_t *context)
+{
+    return context->shim_data[SHIM_GLOBALS_OFF_STATS_EN] != 0;
+}
+
+/* Each vCPU is the sole writer of its own counter slots. */
+static inline void el1_counter_increment(_Atomic uint64_t *counter)
+{
+    uint64_t value = atomic_load_explicit(counter, memory_order_relaxed);
+    atomic_store_explicit(counter, value + 1, memory_order_relaxed);
+}
+
+static inline void el1_mmap_counter(const el1_mmap_context_t *context,
+                                    unsigned counter)
+{
+    if (el1_stats_enabled(context))
+        el1_counter_increment(&context->control->counters[counter]);
+}
+
+static inline void el1_attention_counter(const el1_mmap_context_t *context)
+{
+    if (!el1_stats_enabled(context))
+        return;
+    _Atomic uint64_t *counter =
+        (_Atomic uint64_t *) (context->shim_data + SHIM_COUNTERS_OFF) +
+        SHIM_COUNTER_ATTN_BAIL;
+    el1_counter_increment(counter);
+}
+
+static inline bool el1_attention_pending(const el1_mmap_context_t *context)
+{
+    return atomic_load_explicit((_Atomic uint32_t *) (context->shim_data +
+                                                      SHIM_GLOBALS_OFF_ATTN),
+                                memory_order_acquire) != 0;
+}
+
+/* Consume a host-prepared, per-vCPU VA arena for the exact anonymous RW shape
+ * used by allocators:
+ *
+ *   mmap(NULL, len, PROT_READ|PROT_WRITE,
+ *        MAP_PRIVATE|MAP_ANONYMOUS[|MAP_NORESERVE], fd, off)
+ *
+ * Each vCPU is the sole producer of its publication ring and cursor.  The host
+ * acquire-drains all publications whenever it takes mmap_lock.
+ */
+static bool el1_mmap(el1_mmap_context_t *context, uint64_t *saved_gprs)
+{
+    shim_mmap_control_t *control = context->control;
+
+    if (el1_attention_pending(context)) {
+        el1_mmap_counter(context, SHIM_MMAP_COUNTER_ATTENTION);
+        el1_attention_counter(context);
+        return false;
+    }
+
+    uint64_t prot = saved_gprs[2];
+    uint64_t flags = saved_gprs[3];
+    uint64_t length;
+    if (saved_gprs[0] != 0 || prot != EL1_PROT_RW ||
+        (flags & ~EL1_MAP_NORESERVE) != (EL1_MAP_PRIVATE | EL1_MAP_ANONYMOUS) ||
+        !el1_page_round(saved_gprs[1], &length)) {
+        el1_mmap_counter(context, SHIM_MMAP_COUNTER_SHAPE_MISS);
+        return false;
+    }
+
+    uint32_t generation =
+        atomic_load_explicit(&control->generation, memory_order_acquire);
+    if (generation != atomic_load_explicit(&control->consumer_generation,
+                                           memory_order_relaxed)) {
+        /* A revocation deliberately leaves the consumer generation stale.  Ack
+         * only after the acquire load above, then make this syscall take HVC;
+         * the host will either leave the control disabled or publish a fresh
+         * arena.
+         */
+        atomic_store_explicit(&control->consumer_generation, generation,
+                              memory_order_relaxed);
+        el1_mmap_counter(context, SHIM_MMAP_COUNTER_GENERATION_STALE);
+        return false;
+    }
+
+    if (!(atomic_load_explicit(&control->flags, memory_order_relaxed) &
+          SHIM_MMAP_CTRL_ENABLED)) {
+        el1_mmap_counter(context, SHIM_MMAP_COUNTER_CAPACITY_MISS);
+        return false;
+    }
+
+    uint64_t address =
+        atomic_load_explicit(&control->cursor, memory_order_relaxed);
+    uint64_t cursor;
+    if (length >= EL1_ALIGN_THRESHOLD) {
+        if (el1_add_overflow(address, EL1_BLOCK_SIZE - 1, &address)) {
+            el1_mmap_counter(context, SHIM_MMAP_COUNTER_CAPACITY_MISS);
+            return false;
+        }
+        address &= ~(EL1_BLOCK_SIZE - 1);
+    }
+    if (el1_add_overflow(address, length, &cursor) ||
+        cursor >
+            atomic_load_explicit(&control->arena_limit, memory_order_relaxed)) {
+        el1_mmap_counter(context, SHIM_MMAP_COUNTER_CAPACITY_MISS);
+        return false;
+    }
+
+    uint32_t head = atomic_load_explicit(&control->head, memory_order_acquire);
+    uint32_t tail = atomic_load_explicit(&control->tail, memory_order_relaxed);
+    if ((uint32_t) (tail - head) >= SHIM_MMAP_RING_SIZE) {
+        el1_mmap_counter(context, SHIM_MMAP_COUNTER_RING_FULL);
+        return false;
+    }
+
+    shim_mmap_entry_t *entry = &control->ring[tail & (SHIM_MMAP_RING_SIZE - 1)];
+    entry->addr = address;
+    entry->len = length;
+    entry->prot = prot;
+    /* Publish the bump cursor before the entry. */
+    atomic_store_explicit(&control->cursor, cursor, memory_order_relaxed);
+    atomic_store_explicit(&control->tail, tail + 1, memory_order_release);
+
+    saved_gprs[0] = address;
+    el1_mmap_counter(context, SHIM_MMAP_COUNTER_HIT);
+    return true;
+}
+
+bool el1_mmap_fastpath(uint64_t saved_gprs[static EL1_SAVED_GPRS])
+{
+    if (saved_gprs[8] != EL1_SYS_MMAP)
+        return false;
+    el1_mmap_context_t context = el1_context(saved_gprs);
+    return el1_mmap(&context, saved_gprs);
+}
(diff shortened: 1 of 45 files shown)


jserv:
Can you explain benchmark results? In particular, dirty-every-page items.

maxliu0:
> Can you explain benchmark results? In particular, dirty-every-page items.

The 3–10x gap between PR95 and PR50 likely because improper measurement setup, where continuous munmap operations for each size were executed within the same process.

After isolating each size and attempt into a newly created process, the results show a much smaller difference between PR95 and PR50.
mmap-isolated-c-dirty

src/core/shim-mmap.c
```diff
@@ -0,0 +1,947 @@
...
+    uint32_t flags =
+        atomic_load_explicit(&candidate->flags, memory_order_acquire);
+    if (!(flags & SHIM_MMAP_CTRL_ENABLED))
+        return false;
+    uint32_t generation =
+        atomic_load_explicit(&candidate->generation, memory_order_acquire);
+    if (start <
+        atomic_load_explicit(&candidate->arena_base, memory_order_relaxed))
+        return false;
+    if (end > atomic_load_explicit(&candidate->cursor, memory_order_acquire))
```
jserv, line 392: The ownership test here is `[arena_base, cursor)`, but the host allocator can place a non-arena mapping inside that span. `munmap_retire_commit_locked` lowers the gap hint to the retired start (mem.c:297) and `mmap_fastpath_skip_reserved` reserves only `[cursor, limit)` (mem.c:1079), so a hole freed below the cursor is handed out to any mapping type.

Reproduced on this branch: three 64 KiB anon mappings, free the middle one, `mmap(fd)` into the hole, `munmap` it.

```
p1=0x200000000 p2=0x200010000 p3=0x200020000
freed p2
file q=0x200010000  reused-hole=1
FATAL src/syscall/mem.c:265: munmap retire: non-fast mapping in arena [0x200010000-0x200020000)
```

EL1 cleared the file mapping's PTEs and broadcast TLBI with the host uninvolved. `test-mmap-fastpath` (11/11) and `test-mmap-lazy` (16/16) pass, so nothing in the suite covers a non-arena mapping inside a live arena.

Simplest fix: reserve all of `[arena_base, arena_limit)` and drop the gap-hint lowering, so this bounds check stays a proof of ownership rather than a heuristic.

src/syscall/mem.c
```diff
@@ -37,6 +42,1049 @@
...
+        for (int slot = 0; slot < MAX_THREADS; slot++) {
+            shim_mmap_control_t *c = mmap_fastpath_control(g, slot);
+            if (!(atomic_load_explicit(&c->flags, memory_order_acquire) &
+                  SHIM_MMAP_CTRL_ENABLED))
+                continue;
+            uint64_t cursor =
+                atomic_load_explicit(&c->cursor, memory_order_acquire);
+            uint64_t limit =
+                atomic_load_explicit(&c->arena_limit, memory_order_relaxed);
+            if (cursor < limit && *start < limit && end > cursor) {
```
jserv, line 1079: This reserves only the unallocated tail. Holes below `cursor` stay visible to `find_free_gap_inner`, which is what lets a file-backed mapping land inside a live arena and be claimed later by `el1_munmap_match_arena`. Widening the condition to `[arena_base, arena_limit)` costs VA the arena already owns and restores the invariant EL1 depends on.

src/syscall/mem.c
```diff
@@ -37,6 +42,1049 @@
...
+                                    0, NULL, -1) < 0) {
+                /* EL1 already returned this address to the guest. Continuing
+                 * without semantic metadata would turn first touch into a false
+                 * SIGSEGV, so fail closed on the violated provisioning
+                 * invariant instead of silently corrupting process state.
+                 */
+                log_fatal(
+                    "mmap fast path: region metadata exhausted while "
+                    "draining vCPU slot %d",
+                    slot);
```
jserv, line 230: `log_fatal` does not terminate. `log_impl` returns after printing, and the convention in this tree is `log_fatal(...); abort();` (guest.c:311). All eight new `log_fatal` sites in this file lack the `abort()`, so every fail-closed guard added here is inert and execution continues into exactly the state the guard was written to prevent.

This site is the clearest case, since the comment above it names the outcome it is meant to rule out:

```
ERROR guest.c:2430: region table full (4096/4096), cannot track [0x202328000-0x202329000)
FATAL mem.c:230: mmap fast path: region metadata exhausted while draining vCPU slot 0
elfuse exit=139
```

The guest took SIGSEGV on first touch after fragmenting past `GUEST_MAX_REGIONS`. Beyond adding the `abort()`, this one is worth making unreachable: the slow path returns ENOMEM for the same condition, so gate the arena on free region slots instead of discovering exhaustion after EL1 has already handed the address to the guest.

src/core/shim-mmap.c
```diff
@@ -0,0 +1,947 @@
...
+        el1_mmap_counter(context, SHIM_MMAP_COUNTER_CAPACITY_MISS);
+        return false;
+    }
+
+    shim_mmap_entry_t *entry = &control->ring[tail & (SHIM_MMAP_RING_SIZE - 1)];
+    entry->addr = address;
+    entry->len = length;
+    entry->prot = prot;
+    /* publish the bump cursor before the entry */
+    atomic_store_explicit(&control->cursor, cursor, memory_order_relaxed);
```
jserv, line 351: `cursor` is stored relaxed before the release store to `tail`, so the acquire load of `cursor` in `el1_munmap_match_arena` (line 392) gets no synchronizes-with edge to that tail store.

A sibling vCPU can therefore observe the bumped cursor, accept the range, and publish a retirement, while the host acquire-loading this ring's `tail` still sees the old value. The drain then removes nothing, and the publication reinstates the region on a later pass: the inversion the PR description says the tail snapshot rules out. It can also surface as the `stale arena generation/range` guard further down.

Release-store `tail` first, then release-store `cursor`. The acquire load at line 392 then orders the publication ahead of the retirement, and the host inherits that ordering through the retire ring.

src/syscall/mem.c
jserv: Sharpening the reservation suggestion: counting currently free `g->regions[]` slots is not enough on its own. The reservation has to cover every entry already sitting unpublished in a ring (up to `SHIM_MMAP_RING_SIZE` per vCPU), since those addresses are in guest hands but not yet in the tracker. Reserve against that bound before enabling an arena, have EL1 bail to HVC once its reservation is spent, and give slots back as publications drain or retire.

jserv (changes requested):

src/core/shim-mmap.c
Max042004: Resolved in PR #372

src/syscall/mem.c
Max042004: Resolved in PR #372

src/syscall/mem.c
Max042004: Resolved in PR #372

src/core/shim-mmap.c
Max042004: Resolved in PR #372

Max042004:
Closed this PR since PR #372 sharing same content.

PR #318, opened 2026-08-21
Answer pselect6's poll fallback as select does

xalestar:
pselect6 falls back to poll() when a host descriptor is out of the fd_set range, causing the guest to receive poll()'s behavior in two ways: descriptors refused by macOS poll() (/dev/null, /dev/zero, /dev/urandom, directories, and kqueue fds) arrive as not ready, whereas Linux select() reports them as ready; additionally, the return value counts each descriptor once, whereas select() counts it once per set. The refused entries are routed through poll_eval_unpollable(), the helper sys_ppoll() already uses for the same descriptor set, and the count comes from the bits set by the write-back.

Reproduction is covered by the two high-descriptor cases in `tests/test-poll.c`, which fail on main and pass here. `make check` is clean, `make lint` is unchanged, and `tests/test-matrix.sh` passes all within baseline with zero failures (elfuse-aarch64 255, qemu-aarch64 234, elfuse-x86_64 78).

### DIFF
--- a/src/syscall/poll.c
+++ b/src/syscall/poll.c
@@ -104,6 +104,17 @@ static inline void host_fd_refs_close(host_fd_ref_t *refs, uint32_t n)
  */
 #define POLL_WRITE_EVENTS (POLLOUT | POLLWRBAND)
 
+/* One entry for poll_eval_unpollable(): the host descriptor the caller
+ * resolved, the events the guest asked about, and the answer written back. A
+ * negative fd marks an entry that is not part of the refused set, so the two
+ * callers can keep one array parallel to their own request list.
+ */
+typedef struct {
+    int fd;
+    short events;
+    short revents;
+} poll_unpollable_t;
+
 /* Evaluate the entries host poll() refused. macOS poll() answers POLLNVAL for
  * any descriptor it will not put on a kqueue: /dev/null, /dev/zero,
  * /dev/random, /dev/urandom, directories, and kqueue descriptors themselves.
@@ -118,11 +129,7 @@ static inline void host_fd_refs_close(host_fd_ref_t *refs, uint32_t n)
  *
  * Returns how many entries are ready.
  */
-static uint32_t poll_eval_unpollable(const struct pollfd *host_fds,
-                                     const host_fd_ref_t *refs,
-                                     const bool *unpollable,
-                                     short *revents,
-                                     uint32_t nfds)
+static uint32_t poll_eval_unpollable(poll_unpollable_t *entries, uint32_t n)
 {
     fd_set read_set, write_set, except_set;
     FD_ZERO(&read_set);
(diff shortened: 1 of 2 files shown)


src/syscall/poll.c
```diff
@@ -790,6 +864,16 @@ int64_t sys_pselect6(guest_t *g,
...
     }
 
+    /* Stamp the select()-evaluated answers over the zeroed revents the poll set
+     * left behind, and credit them to the return count.
+     */
+    if (ret >= 0 && fb.refused > 0) {
+        for (int i = 0; i < req_count; i++)
+            if (unp[i].fd >= 0)
+                reqs[i].revents = unp[i].revents;
+        ret += (int) fb.ready;
```
jserv: `fb.ready` counts refused entries, one per descriptor, but select's return value is the number of bits set across the three sets. The write-back below lights both the read and the write bit for a descriptor like `/dev/null`, so `ret` comes back one short. The direct pselect path returns the host's bit count, which is what the new "pselect always-ready devices" case asserts with `r == 2 * opened`; the same guest program now gets 2 for a low host fd and 1 for a high one. Count each refused entry once per bit it sets (read, write, except) instead of once per entry.

src/syscall/poll.c
```diff
@@ -720,42 +825,11 @@ int64_t sys_pselect6(guest_t *g,
...
-            ret = poll(poll_fds, (nfds_t) poll_count, timeout_ms);
-            if (ret >= 0) {
-                for (int i = 0; i < req_count; i++)
-                    reqs[i].revents = poll_fds[i].revents;
-                poll_wakeup_fired =
-                    added_wakeup && (poll_fds[req_count].revents & POLLIN);
-            }
-            free(poll_heap);
+            bool restart;
+            ret = pselect_fallback_pass(&fb, has_timeout ? &ts : &poll_ts,
```
jserv, line 831: When the pass fails its allocation it returns -1 with ENOMEM, but control now falls into the three interrupt predicates before `save_errno = errno`, and any of them can overwrite errno. The inlined code broke out of the loop at that point. Break out here too when the pass reports an allocation failure.

tests/test-poll.c
```diff
@@ -558,6 +558,86 @@ int main(void)
...
+
+    /* The same descriptors, reached through the poll fallback pselect6 takes
+     * when a host descriptor lands at or above FD_SETSIZE. Filling the fd table
+     * pushes the last one there. On a real kernel the fillers only raise the
+     * descriptor number; select() answers ready either way.
+     */
+    TEST("pselect always-ready device at a high fd");
+    {
+        static int fillers[FD_SETSIZE];
+        int n = 0, subject = -1;
```
jserv, line 609: The loop also breaks when `open()` returns -1, and `subject` is then a low descriptor whose host fd stays under FD_SETSIZE. The assertions still pass, so a run under a low RLIMIT_NOFILE reports green without ever entering the fallback this case exists to cover. Record which of the two breaks fired and fail loudly when the table never filled.

src/syscall/poll.c
xalestar: Measured with the same static binary. A /dev/null in both readfds and writefds answers select=2 on the qemu-system-aarch64 reference VM and 1 through the fallback. The count comes from the bits the write-back lights, which also covers the entries poll() accepts: a socket pair holding a byte, at fd 1022 with the table filled, answers 2 on the reference VM and answered 1 here. Both are pinned in tests/test-poll.c and both fail against main.

src/syscall/poll.c
xalestar: The pass leaves the wait loop on an allocation failure, ahead of the three predicates, so ENOMEM reaches save_errno.

tests/test-poll.c
xalestar: The case fails when the table stops short of FD_SETSIZE - 1 instead of passing. Both lanes reach 1023: the guest table is 1024 descriptors and the reference VM's RLIMIT_NOFILE is 1024, each filling 1018 before open() reports EMFILE.

src/syscall/poll.c
```diff
@@ -703,10 +804,14 @@ int64_t sys_pselect6(guest_t *g,
     if (added_wakeup && !RANGE_CHECK(wake_fd, 0, FD_SETSIZE))
         use_poll_fallback = true;
 
+    pselect_fallback_t fb = {.reqs = reqs,
+                             .unp = unp,
+                             .req_count = req_count,
+                             .wake_fd = added_wakeup ? wake_fd : -1};
```
jserv: Append comma (`,`) right after `-1`, so that clang-format can indent better.

xalestar: Done.

Issue #319, opened 2026-08-21
Implement general USB translation layer (Linux usbdevfs over IOKit)


jotpalch:
## Motivation

In #310, discussing the serial ioctl work, the maintainer asked for something wider ([comment](https://github.com/sysprog21/elfuse/issues/310#issuecomment-5345143498)):

> I wonder if there is a way to avoid the VM entirely by translating the Linux USB interface into its macOS counterpart. In other words, rather than limiting the approach to USB serial devices, could we provide a more general USB compatibility/translation layer that exposes arbitrary Linux USB device interfaces to macOS applications?

The named scenario is embedded development: let [probe-rs](https://github.com/probe-rs/probe-rs) flash and debug real hardware from inside elfuse, skipping the VM plus `VZUSBPassthroughDevice` route entirely. Demand outside that niche is old and unmet: [docker/for-mac#900](https://github.com/docker/for-mac/issues/900) (serial/USB device access from Linux containers on macOS) has been open since November 2016 and carries 239 upvotes, because a VMM needs hypervisor-level USB forwarding that macOS never offered. elfuse is positioned differently: guest syscalls already execute in a host process that can call IOKit directly, so there is no VMM device model and no guest driver stack to feed.

This issue proposes the layer, shows what a working prototype does on real hardware, and asks for a decision on scope and PR slicing before anything is submitted.

## What the hardware allows

Probed on an ESP32-S3 (VID 0x303a, PID 0x1001, a CDC + vendor-JTAG composite) from an unentitled, ad-hoc-signed, non-root process on macOS 15.6:

| Interface | Bound macOS driver | `USBInterfaceOpen` result |
|---|---|---|
| 0, CDC control (class 0x02) | AppleUSBACMControl | `kIOReturnExclusiveAccess`, always |
| 1, CDC data (class 0x0a) | AppleUSBACMData + IOSerialBSDClient | succeeds while nobody holds `/dev/cu.usbmodem*`; `ExclusiveAccess` once someone does |
| 2, vendor JTAG (class 0xff) | none | succeeds; both bulk pipes usable (0x02 OUT, 0x83 IN, mps 64) |

[...]

PR #325, opened 2026-08-23
Accept uevent netlink sockets as silent

jotpalch:
> [!NOTE]
> First piece of the general USB translation layer discussed in #319; it stands alone as a bug fix and does not depend on the rest of that series.

## Summary

`socket(AF_NETLINK, SOCK_RAW, NETLINK_KOBJECT_UEVENT)` now returns a working silent socket. libusb's non-udev build refuses to initialize at all without one (`linux_usbfs.c:416-430` aborts `libusb_init` when the hotplug monitor cannot start), so every libusb program in a guest died at init, whether or not it ever touched a device.

## What

- `NETLINK_KOBJECT_UEVENT` sockets are accepted next to `NETLINK_ROUTE`: never readable, so `poll` times out and a non-blocking `recv` reports `EAGAIN`, which is what a real Linux kernel with no hotplug activity looks like. Synthesizing uevents on device attach is deliberately out of scope here.
- `sendto` on a uevent socket reports `EPERM`, matching `uevent_net_rcv_skb()` for a sender without `CAP_SYS_ADMIN`.
- `SOL_SOCKET` options work on netlink fds; they previously failed `EBADF` because the sockopt path accepted only `FD_SOCKET`. The set follows `sk_setsockopt()`/`sk_getsockopt()` rather than accepting everything: `SO_RCVBUF`/`SO_SNDBUF` clamp to `rmem_max`/`wmem_max` as an unsigned value, double, then floor at `SOCK_MIN_RCVBUF` (2304) or `SOCK_MIN_SNDBUF` (4608), which differ; `SO_RCVTIMEO` is honored by the blocking receive path (deadline taken once on entry, `EAGAIN` on expiry) rather than accepted and dropped; `SO_LINGER` and the timeout options require their real payload length (8 and 16) instead of four bytes; `SO_DEBUG` with a nonzero value needs `CAP_NET_ADMIN` (`EACCES`); `SO_REUSEPORT` with a nonzero value is `EOPNOTSUPP` on a non-inet socket; `SO_SNDLOWAT` has no setter at all (`ENOPROTOOPT`); and everything settable reads back through `getsockopt` with the right payload length. Unrecognized options report `ENOPROTOOPT`.
- `SOCK_NONBLOCK`/`SOCK_CLOEXEC` are now published into the fd's linux_flags. That was dropped before, and it affected `NETLINK_ROUTE` sockets too.
- New guest test `tests/test-uevent-socket.c`, wired into `test-matrix.sh`. Its assertions also hold on a real Linux kernel: a payload below `NLMSG_HDRLEN` is swallowed there and `sendto` returns its length, so the test accepts either outcome and pins `EPERM` whenever the send fails.

## Evidence

A statically linked libusb test (current libusb git, `--disable-udev`, so the netlink monitor is the one in play), run under the same host and guest binary on both sides:

```
upstream main 628181b:   libusb_init = -99 (LIBUSB_ERROR_OTHER)
this branch:             libusb_init = 0 (LIBUSB_SUCCESS)
                         get_device_list = 0
```

`get_device_list` returns 0 because enumeration needs the synthetic `/sys/bus/usb` and `/dev/bus/usb` trees that come later in the series; init succeeding is what this change is for.

```
test-uevent-socket under elfuse: 40 passed, 0 failed
test-matrix elfuse-aarch64: 266 passed, 0 failed, 10 skipped
```

The test asserts relations rather than constants, because the floors derive from `sizeof(struct sk_buff)` and the caps are sysctls: the send floor is checked to be exactly twice the receive floor, `-1` and `INT_MAX` are checked to read back equal, and the honored `SO_RCVTIMEO` is timed with a jiffy of slack. Those assertions hold on a real Linux kernel, which matters because this test runs in the `qemu-aarch64` matrix lane as well.

`make check` exits 0 on a clean rebuild (host suite 27/27, guardrail PASS), and the `NETLINK_ROUTE` behavior is unchanged.

### DIFF
--- a/src/syscall/netlink.c
+++ b/src/syscall/netlink.c
@@ -12,19 +12,32 @@
  *
  * Supported operations:
  *   socket(AF_NETLINK, SOCK_RAW|SOCK_DGRAM, NETLINK_ROUTE) -> synthetic fd
- *   bind() -> always succeeds
+ *   bind() -> always succeeds, recording nl_pid and the nl_groups mask
  *   sendmsg(RTM_GETLINK) -> builds interface list from host getifaddrs
  *   sendmsg(RTM_GETADDR) -> builds address list from host getifaddrs
  *   recvmsg() / read() -> returns buffered response data
  *
- * Non-NETLINK_ROUTE protocols return -EAFNOSUPPORT at socket() time.
+ * NETLINK_KOBJECT_UEVENT sockets are also accepted, as a silent socket: no
+ * uevent is ever synthesized, so the fd never becomes readable, a non-blocking
+ * receive reports EAGAIN, and poll() times out. That is exactly the Linux
+ * kernel's behavior on a machine with no hotplug activity, and it is enough for
+ * libusb_init()'s netlink hotplug monitor (socket + bind(nl_groups=1) +
+ * setsockopt(SO_PASSCRED) + a poll loop) to come up. Sends on a uevent socket
+ * report -EPERM, matching uevent_net_rcv() for a sender without CAP_SYS_ADMIN
+ * (elfuse guests are not privileged over the host's device model). SOL_SOCKET
+ * and SOL_NETLINK options on any netlink fd are emulated per-fd in
+ * netlink_setsockopt/netlink_getsockopt below.
+ *
+ * Other protocols return -EAFNOSUPPORT at socket() time.
  */
 
+#include <limits.h>
 #include <pthread.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 #include <unistd.h>
 #include <net/if.h>
 #include <ifaddrs.h>
@@ -33,8 +46,10 @@
 
 #include "syscall/linux-wire.h"
 #include "syscall/internal.h"
-#include "syscall/io.h" /* io_wait_fd_or_interrupted */
+#include "syscall/io.h" /* io_wait_fd_timed_or_interrupted */
 #include "syscall/net.h"
+#include "syscall/proc.h"   /* syscall_restart_forbid, syscall_is_restarted */
+#include "runtime/thread.h" /* thread_stop_is_leader_work_only */
 #include "proved/netlink.h"
 #include "utils.h"
 #include <poll.h>
@@ -121,7 +136,23 @@ typedef struct {
 
 typedef struct {
     bool in_use;
-    int guest_fd;                  /* Guest fd number */
+    int guest_fd; /* Guest fd number */
+    /* Allocation order, unique and increasing. A guest fd number outlives the
+     * socket that had it: fd_cleanup_entry() runs netlink_close() after the
+     * number is already back in the fd table's free pool, so a socket() on
+     * another thread can be handed the same number while the previous slot is
+     * still in_use. Two slots can therefore carry one guest_fd, and the
+     * generation is what tells them apart -- nl_find() answers with the newest
+     * (the only one the guest can still reach) and netlink_close() retires the
+     * oldest.
+     *
+     * That pairing is exact rather than approximate. A number is only reissued
+     * after the previous holder's close was issued, so the pending closes for a
+     * number are always the oldest slots holding it, whatever order the threads
+     * running them arrive in; retiring oldest-first therefore never takes down
+     * the live socket, and every slot is retired exactly once.
+     */
+    uint64_t gen;
     uint8_t buf[NETLINK_BUF_SIZE]; /* Response buffer */
     size_t buf_len;                /* Bytes written into buf */
     size_t buf_pos;                /* Current read position */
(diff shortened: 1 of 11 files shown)


src/syscall/netlink.c
```diff
@@ -129,8 +141,27 @@ typedef struct {
...
 
+/* Linux net.core.rmem_default/wmem_default on a stock kernel: what
+ * getsockopt(SO_RCVBUF/SO_SNDBUF) reports before the guest ever sets one.
+ */
+#define NETLINK_DEFAULT_BUFSIZE 212992
+
+/* Floor for a doubled SO_RCVBUF/SO_SNDBUF, standing in for Linux's
+ * SOCK_MIN_RCVBUF/SOCK_MIN_SNDBUF (truesize-derived, ~2.2KB).
+ */
+#define NETLINK_MIN_BUFSIZE 2048
```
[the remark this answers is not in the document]
jotpalch: Confirmed and fixed in d395fd1. `nl_bufsize_store` now takes the direction's floor and follows `sk_setsockopt`'s order: clamp to the sysctl as an unsigned value, double, then `max` against `SOCK_MIN_RCVBUF` (2304) or `SOCK_MIN_SNDBUF` (4608). Measured on a real Linux kernel to pin the numbers: RCVBUF 0/128/512 all read back 2304 and 2048 reads back 4096, while SNDBUF stays at 4608 until the doubled value passes it. The unsigned clamp turned out to matter as much as the two floors, since `SO_RCVBUF = -1` is a large u32 and lands on `rmem_max` (425984 here) where the old code floored it to 2048. The test asserts the relation (send floor == 2x receive floor, and -1 == INT_MAX on readback) rather than the constants, because both derive from `sizeof(struct sk_buff)` and a sysctl.

src/syscall/netlink.c
```diff
@@ -504,6 +548,164 @@ int64_t netlink_bind(int guest_fd,
...
+        ns->opt_sndbuf = nl_bufsize_store(value);
+        break;
+    case LINUX_SO_DEBUG:
+    case LINUX_SO_REUSEADDR:
+    case LINUX_SO_DONTROUTE:
+    case LINUX_SO_BROADCAST:
+    case LINUX_SO_KEEPALIVE:
+    case LINUX_SO_OOBINLINE:
+    case LINUX_SO_LINGER:
+    case LINUX_SO_RCVTIMEO:
```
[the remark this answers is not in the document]
jotpalch: Confirmed and fixed in d395fd1, by honoring it rather than rejecting: Linux accepts `SO_RCVTIMEO` on a netlink socket, so refusing it would be the larger divergence. `nl_wait_readable_locked` now takes a `CLOCK_MONOTONIC` deadline once on entry, the way `skb_recv_datagram` reads `sock_rcvtimeo` once, and returns `-EAGAIN` on expiry with the buffer still empty. One correction to the premise while implementing it: the existing wait did not re-loop in slices at the call site, `io_wait_fd_or_interrupted` loops internally and returns only on readiness or EINTR, so a deadline checked in the outer loop would never have been reached. Added a timed variant that charges those slices against the caller's deadline; an interrupted finite wait forbids the SVC restart, matching `sock_intr_errno`. Measured against a real kernel first: a 200 ms timeout returns EAGAIN after 208 ms, and the new test pins that with a jiffy of slack and an alarm guard. `SO_SNDTIMEO` is stored and readable but inert, since a uevent send is EPERM before it can queue.

src/syscall/netlink.c
```diff
@@ -504,6 +548,164 @@ int64_t netlink_bind(int guest_fd,
...
+    case LINUX_SO_SNDBUF:
+        ns->opt_sndbuf = nl_bufsize_store(value);
+        break;
+    case LINUX_SO_DEBUG:
+    case LINUX_SO_REUSEADDR:
+    case LINUX_SO_DONTROUTE:
+    case LINUX_SO_BROADCAST:
+    case LINUX_SO_KEEPALIVE:
+    case LINUX_SO_OOBINLINE:
+    case LINUX_SO_LINGER:
```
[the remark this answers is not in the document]
jotpalch: Confirmed and fixed in d395fd1. The per-option minimum now applies after the generic four-byte gate: 8 for `SO_LINGER`, 16 for `SO_RCVTIMEO`/`SO_SNDTIMEO` (both the `_OLD` and `_NEW` numbers, the latter previously undefined here), with over-long lengths still accepted as Linux accepts them. Verified against a real kernel: `SO_LINGER` at optlen 4 gives EINVAL and at 8 succeeds, `SO_RCVTIMEO` at 4 gives EINVAL and at 16 succeeds. Reviewing the neighbouring options for the same class of problem turned up four more, all fixed in the same commit: `SO_REUSEPORT` with a nonzero value is EOPNOTSUPP on a non-inet socket, `SO_SNDLOWAT` has no setter in `sk_setsockopt` at all (ENOPROTOOPT), `SO_DEBUG` with a nonzero value needs CAP_NET_ADMIN (EACCES), and every option that can be set now reads back through `getsockopt` with its real payload length instead of ENOPROTOOPT.

tests/test-uevent-socket.c
```diff
@@ -0,0 +1,157 @@
...
+    len = sizeof(val);
+    CHECK(
+        getsockopt(fd, SOL_SOCKET, SO_TYPE, &val, &len) == 0 && val == SOCK_RAW,
+        "getsockopt(SO_TYPE) == SOCK_RAW");
+
+    /* No uevent is ever synthesized: an empty socket reports EAGAIN... */
+    drain(fd);
+    char buf[4096];
+    ssize_t r = recv(fd, buf, sizeof(buf), MSG_DONTWAIT);
+    CHECK(r == -1 && errno == EAGAIN, "recv(MSG_DONTWAIT) empty -> EAGAIN");
```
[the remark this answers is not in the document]
jotpalch: Confirmed and fixed in d395fd1. The recv assertion now uses the same bounded drain-and-retry as the poll one (four iterations, so a continuous uevent stream cannot spin it), which matters because this test runs in the `qemu-aarch64` lane against a real kernel with group 1 subscribed. The whole file was reworked in the same pass to stay honest on both sides: every new assertion was checked against a real Linux kernel first, and anything configuration-dependent is asserted as a relation rather than a constant.

src/syscall/netlink.c
```diff
@@ -504,6 +594,348 @@ int64_t netlink_bind(int guest_fd,
...
+    if (optlen < NETLINK_TIMEVAL_LEN)
+        return -LINUX_EINVAL;
+
+    nl_timeval_t tv;
+    if (guest_read_small(g, optval_gva, &tv, sizeof(tv)) < 0)
+        return -LINUX_EFAULT;
+
+    if (tv.tv_usec < 0 || tv.tv_usec >= 1000000)
+        return -LINUX_EDOM;
+    if (tv.tv_sec < 0 || tv.tv_sec >= NETLINK_TIMEO_MAX_SEC) {
```
jserv: This inverts Linux's two out-of-range cases. In `sock_set_timeout()`, `tv_sec < 0` does `WRITE_ONCE(*timeo_p, 0)`, and an `sk_rcvtimeo` of 0 means do not wait at all: `__skb_recv_datagram`'s `while (timeo && !__skb_wait_for_more_packets(...))` exits on the first pass with `EAGAIN`. Only `{0,0}` and the past-the-cutoff case leave `MAX_SCHEDULE_TIMEOUT`, which is forever. Here both land on `*out_us = 0`, and line 1297 reads 0 as forever.

So `setsockopt(SO_RCVTIMEO, {.tv_sec = -1})` then a blocking `recv` returns `EAGAIN` at once on Linux and blocks forever here. `getsockopt` cannot separate the two either, since `sock_get_timeout()` reports `{0,0}` for both.

This needs a third state: store `-1` for the negative `tv_sec` case, keep `0` for forever, and have `nl_wait_readable_locked` treat `opt_rcvtimeo_us < 0` exactly like `nonblock`.

src/syscall/netlink.c
```diff
@@ -837,12 +1288,39 @@ static int64_t nl_wait_readable_locked(netlink_state_t *ns,
...
                                        bool nonblock)
 {
+    /* The SO_RCVTIMEO deadline is taken once, here, and not refreshed by the
+     * waits below. That is skb_recv_datagram()'s shape: sock_rcvtimeo() is read
+     * at the top and every wait inside shares the one budget, so a wakeup that
+     * finds nothing does not buy the caller another full timeout.
+     */
+    int64_t deadline_us = 0;
+    if (!nonblock && ns->opt_rcvtimeo_us > 0)
+        deadline_us = nl_now_us() + ns->opt_rcvtimeo_us;
```
jserv: Signed overflow, and it turns the longest timeout into the shortest. `nl_timeout_store` accepts `tv_sec` up to `NETLINK_TIMEO_MAX_SEC - 1`, so the largest stored value is 9223372036852999999 us, leaving 1.78 seconds of headroom below `INT64_MAX`. Past that much `CLOCK_MONOTONIC` uptime the sum wraps negative, `left_us <= 0` on the first iteration, and a guest that asked for an effectively infinite receive timeout gets `EAGAIN` immediately.

Saturate the add, or lower `NETLINK_TIMEO_MAX_SEC` to something like `INT64_MAX / 2000000`. Separately, the current cutoff is three orders of magnitude below Linux's own (`MAX_SCHEDULE_TIMEOUT/HZ - 1`), so finite Linux timeouts above it silently become forever.

src/syscall/netlink.c
```diff
@@ -852,16 +1330,30 @@ static int64_t nl_wait_readable_locked(netlink_state_t *ns,
...
-        if (wait_rc < 0)
+        int64_t wait_rc =
+            io_wait_fd_timed_or_interrupted(rd_fd, POLLIN, timeout_ms);
+        if (wait_rc < 0) {
+            /* sock_intr_errno(): an interrupted receive reports ERESTARTSYS
+             * when the timeout is MAX_SCHEDULE_TIMEOUT and EINTR when it is
+             * finite, precisely so a restart cannot hand the caller its whole
+             * SO_RCVTIMEO a second time. Same split, same reason.
+             */
+            if (deadline_us)
```
jserv: `io_wait_fd_timed_or_interrupted` has two sources of `-EINTR` and only one is guest-visible. io.c:358 returns it for the leader-only execve handoff, which the comment at io.c:295 calls private to the dispatcher: the thread keeps running and the SVC is restarted. syscall.c:2838 arms that restart only when `!cpu_restart_req.forbidden`, so this unconditional forbid disables it.

Thread A blocks in `recvmsg` on a netlink socket with a finite `SO_RCVTIMEO`, thread B calls `execve`, and A's `recvmsg` returns `EINTR` with no signal ever delivered. No real kernel produces that, and the contract entry this PR adds claims the split is `sock_intr_errno()`'s, which never sees this case.

Gate on the guest-visible half: `if (deadline_us && wait_rc == -LINUX_EINTR && !thread_stop_is_leader_work_only())`. If the restart re-arming a fresh deadline is the worry, the deadline has to survive the restart rather than the restart being suppressed.

src/syscall/netlink.c
```diff
@@ -479,6 +557,18 @@ int64_t netlink_socket(int protocol, int type)
 
     ns->pipe_wr = pipefd[1];
     ns->pipe_rd = pipefd[0];
+    ns->proto = protocol;
```
jserv: `proto` is new state, and it is published outside the lock. `nl_alloc` at line 550 scans `nl_state[]` and sets `in_use` with no lock held, and lines 558-561 then write `pipe_wr`, `pipe_rd`, `proto` and `sock_type` the same way. Every other accessor takes `nl_lock`, including both sockopt entry points this PR adds.

The unlocked allocation predates the change, but `proto` is a behavior discriminator. Two guest threads calling `socket(AF_NETLINK, ...)` concurrently can pick the same free slot; the second overwrites `proto` and the pipe pair, so a uevent socket starts parsing rtnetlink requests or an rtnetlink socket answers `EPERM` on every send, and a pipe pair leaks.

`nl_lock` is a declared leaf, so taking it around `nl_alloc` and the field writes, and dropping it before `fd_publish_linux_flags`, is enough. `netlink_bind`'s unlocked `nl_find` and `ns->pid` write wants the same.

src/syscall/netlink.c
```diff
@@ -442,10 +519,11 @@ void netlink_init(void)
 
 int64_t netlink_socket(int protocol, int type)
```
jserv, line 647: `type` arrives unvalidated (`sys_socket` at net.c:338 passes it straight through) and this PR makes it guest-visible through `SO_TYPE`. `socket(AF_NETLINK, SOCK_STREAM, NETLINK_ROUTE)` succeeds and `getsockopt(SO_TYPE)` reports `SOCK_STREAM`, which no kernel can produce: `netlink_create()` refuses anything but `SOCK_RAW` and `SOCK_DGRAM` with `ESOCKTNOSUPPORT`, and an unrecognized flag bit is `EINVAL`. Two lines next to the protocol check.

src/syscall/netlink.c
```diff
@@ -504,6 +594,348 @@ int64_t netlink_bind(int guest_fd,
...
+                           uint64_t optval_gva,
+                           uint32_t optlen)
+{
+    /* SOL_NETLINK (NETLINK_ADD_MEMBERSHIP and friends) is out of scope: no
+     * emulated protocol delivers multicast events, so there is no membership to
+     * track. No consumer in scope sets them (libusb/nusb join groups via
+     * bind()); refuse the level the way sock_setsockopt refuses an unknown one
+     * rather than pretending a membership took effect.
+     */
+    if (level != LINUX_SOL_SOCKET)
```
jserv, line 1052: `netlink_bind` accepts and ignores any `nl_groups` mask, but `setsockopt(SOL_NETLINK, NETLINK_ADD_MEMBERSHIP)` reports `ENOPROTOOPT`. Those are two spellings of one operation and the emulation answers them oppositely; Linux returns 0 for both on `NETLINK_ROUTE` and `NETLINK_KOBJECT_UEVENT`.

Not a regression, since the level used to report `EBADF`, but it leaves libnl and libmnl monitors failing at subscription time, and group numbers above 31 have no other spelling. Accepting `NETLINK_ADD_MEMBERSHIP` and `NETLINK_DROP_MEMBERSHIP` under the same "recorded, nothing is ever delivered" contract `bind` already gives would close it.

tests/test-uevent-socket.c
```diff
@@ -0,0 +1,421 @@
...
+
+    /* The full-length spellings are accepted, and read back. SO_LINGER writes 8
+     * bytes and says so; the old code answered ENOPROTOOPT here.
+     */
+    struct linger lin = {.l_onoff = 1, .l_linger = 7};
+    CHECK(setsockopt(opt, SOL_SOCKET, SO_LINGER, &lin, sizeof(lin)) == 0,
+          "setsockopt(SO_LINGER, sizeof(struct linger))");
+    struct linger back = {.l_onoff = -1, .l_linger = -1};
+    socklen_t blen = sizeof(back);
+    CHECK(getsockopt(opt, SOL_SOCKET, SO_LINGER, &back, &blen) == 0 &&
```
jserv, line 511: Every `getsockopt` here passes a full-size buffer, so the length behavior the commit message calls load-bearing has no assertion: neither the `write_len = guest_optlen` truncation at netlink.c:889 nor the negative-optlen `EINVAL` at netlink.c:884 is exercised. A regression that always writes `lv` bytes, or reports `lv` in `*optlen` regardless, passes the suite while writing past the caller's buffer. A `SO_LINGER` read with `len = 4` into a poisoned buffer, plus a `SO_RCVBUF` read with `len = -1`, both hold identically on a real kernel.

Related: the three two-world assertions (:203 sendto, :363 SO_DEBUG, :366 SO_REUSEPORT) accept either outcome, so none can fail in the elfuse lane. Line 225 already computes a lane discriminator; hoisting it into a `bool on_elfuse` and asserting the strict half turns three tautologies into real tests.

jserv (commented):
The silent-socket shape is right and the `SO_RCVBUF`/`SO_SNDBUF` clamp-double-floor matches `sk_setsockopt` exactly, floors included. Seven comments inline; the first three are correctness, the next two are ABI fidelity, and the last two are scope and coverage.

Checked and clean: optlen validation order and the `lv` values for both timeout spellings, the getsockopt copy-out bounds, the new `SO_*` numbers against `asm-generic/socket.h`, the `io_wait_fd_or_interrupted` refactor for `timeout_ms < 0`, and the `linux_flags` publication (`FD_NETLINK` is synthetic so `fd_host_flag_mask` is 0 and `F_GETFL` answers from the shadow; no in-tree `NETLINK_ROUTE` caller passes `SOCK_NONBLOCK`, so rtnetlink is unaffected).

One thing deliberately not filed as a comment because it predates this PR and is untouched by it: netlink state is keyed by guest fd number, so a `dup`ed netlink fd gets `EBADF` from every netlink path, and closing the original tears down the pipe under the survivor.

src/syscall/netlink.c
jotpalch: Right, and inverted in the worst direction. Timeouts are now stored the way `sk_rcvtimeo` is, as a tri-state: 0 forever, -1 do-not-wait, positive finite. `nl_wait_readable_locked` treats a negative value exactly like nonblock, and the readback still reports `{0,0}` for both, as `sock_get_timeout` does. Measured on a real kernel before and after: `{-1,0}` then a blocking recv returns EAGAIN at 0 ms there and now here, where the old code parked forever.

src/syscall/netlink.c
jotpalch: Valid, and it was not merely theoretical: a UBSan build of the old code reports `netlink.c:1298: signed integer overflow: 2025428727468 + 9223372036852000000`. It happened to be benign under this build's codegen, which is the worst kind of benign. The add now saturates, and I took the second half of your point rather than lowering the cutoff: it is now `MAX_SCHEDULE_TIMEOUT/HZ - 1`, probed rather than assumed, so a finite Linux timeout no longer becomes forever on the way in. Boundary measured on both lanes: `{...772,0}` and `{...773,999999}` read back intact, `{...774,0}` reads back `{0,0}`.

src/syscall/netlink.c
jotpalch: Correct on both halves. The forbid is now gated on `!thread_stop_is_leader_work_only()`, so the dispatcher-private handoff no longer surfaces as a guest EINTR. On the restart question I did not take the documented-deviation route: the wait stashes its absolute deadline in a thread-local that is consumed only when `syscall_is_restarted()` and the fd matches, so the re-executed receive continues the original budget rather than arming a fresh one, which is what `restart_block` does. Verified with a sibling that fails `execve` every 20 ms across a 3 s `SO_RCVTIMEO`: the old build returned EINTR at 61 ms, the fixed build returns EAGAIN at 3002 ms after 53 handoffs, and a real Linux kernel returns EAGAIN at 3030 ms. A fresh-deadline restart would have come back 53 timeouts later.

src/syscall/netlink.c
jotpalch: Confirmed, and the lock alone was not enough. `nl_alloc` now takes `nl_lock` and publishes `proto`, `sock_type` and the pipe pair with `in_use` written last, and `netlink_bind` reads the guest address before locking, then looks up and writes under it. The concurrent-open test I added still failed deterministically after that: `fd_cleanup_entry` runs the netlink teardown after the fd number is already back in the free pool, so two slots can exist for one number. Fixed by tagging each allocation with a generation, with lookups taking the newest slot and teardown retiring the oldest; the pairing is exact because a number is only reissued after the previous close was issued. `nl_lock` stays a leaf, and `netlink_socket` drops it before `fd_publish_linux_flags`.

src/syscall/netlink.c
jotpalch: Fixed, screened before the protocol check: unknown flag bits give EINVAL, and a base type outside {SOCK_RAW, SOCK_DGRAM} gives ESOCKTNOSUPPORT. The ordering is measured rather than assumed, since `socket(AF_NETLINK, SOCK_STREAM, 99)` returns 94 on a real kernel, not a protocol error. While testing this I noticed a related deviation that predates the PR and that I have deliberately left alone for now: an unknown protocol returns EAFNOSUPPORT here where `netlink_create` gives EPROTONOSUPPORT. The new test currently uses that difference as its lane discriminator, so fixing it needs a different discriminator first; say the word and I will do both in one go.

src/syscall/netlink.c
jotpalch: Agreed, and I probed all twelve `SOL_NETLINK` optnames on both families before choosing. ADD/DROP_MEMBERSHIP now return 0 under the same recorded-nothing-delivered contract `bind` gives, with a 1-based range check; LIST_MEMBERSHIPS reads the bitmap back; PKTINFO, BROADCAST_ERROR, NO_ENOBUFS, CAP_ACK, EXT_ACK and GET_STRICT_CHK set and read as flags; LISTEN_ALL_NSID is EPERM; RX/TX_RING and anything unknown stay ENOPROTOOPT. Where it was a judgement call I favoured accept-and-record, since a libnl or libmnl monitor sets several of these on the way up and treats a refusal as a broken socket. One measured short edge: the group count is 32 for both families here, which is exact for uevent (group 32 joins, 33 is EINVAL, LIST_MEMBERSHIPS reports 4) but short for rtnetlink, where a real kernel accepts group 33 and reports 8. Nothing is delivered on any group either way, so over-accepting looked like the friendlier error; happy to widen `NETLINK_ROUTE` if you prefer.

tests/test-uevent-socket.c
jotpalch: Right on both counts. Added the two length cases, a `SO_LINGER` read with `len = 4` into a poisoned buffer (rc 0, `*optlen` 4, bytes 4 to 7 untouched) and a `SO_RCVBUF` read with `len = -1` (EINVAL), plus an over-long optlen. The discriminator at :225 is hoisted into a `bool on_elfuse` and all three two-world assertions now pin the strict half per lane. Being straight about the length pair: it passes against the old code too, since the copy-out already did `min(len, lv)` and refused a negative length, so those two are regression pins rather than something that caught a bug. What does fail against the previous head is 24 assertions, deterministically: the type screening, the whole `SOL_NETLINK` surface, the 1e13-second readback, the concurrent open, the sibling execve, and the blocking receive, which hangs the old build until the watchdog fires. The file is 88 assertions now and passes identically under elfuse and on a real Linux kernel.

jotpalch:
Also picked up your unfiled note about fd-number keying. Confirmed, and I would keep it out of this PR: the generation tag above removes the stale-slot half, but a `dup` still yields a second guest fd with no state, and closing the original still tears the pipe down under the survivor. The real fix hangs the state off the open file description rather than the fd number, and the fd layer already publishes that identity as `ofd_id` (epoll uses it for exactly this). That needs a small accessor plus alias-aware teardown, with tests over dup/dup2/F_DUPFD and CLOEXEC across execve, so it wants its own change; when it lands, netlink's private generation counter should go away with it.

jserv:
Thank @jotpalch for contributing!

PR #332, opened 2026-08-25
Decide epoll_ctl EPERM from the target alone

xalestar:
## Summary

`sys_epoll_ctl` forwards macOS `kevent()` failures to the guest, and that disagrees with Linux in three directions at once.

Linux reads `EPERM` off the target file and nothing else, and it does so before it looks at the operation. A target with no poll method is refused whatever the event mask asks for and whatever the operation is; a target that never becomes writable still accepts `EPOLLOUT`.

kqueue refuses a knote on a directory or a character device with `EINVAL` rather than `EPERM`. It also refuses `EVFILT_WRITE` on a kqueue, and a guest timerfd and a guest epoll fd are both kqueues here, so `epoll_ctl(ADD, timerfd, EPOLLIN|EPOLLOUT)` failed where a real kernel returns 0. And in the other direction, kqueue accepts a knote on a plain file, which Linux refuses.

## What it changes for a guest

An event loop that registers a timerfd for both readiness directions used to get `EINVAL` and, on the common spelling, no timer at all. `epoll_ctl` on `/dev/null`, `/dev/zero`, a directory or a plain file now answers `EPERM` the way `epoll_ctl(2)` documents, for `ADD`, `MOD` and `DEL` alike, instead of `EINVAL` for `ADD` and `ENOENT` for the other two.

A mask that names no readiness filter at all -- `events = 0`, or `EPOLLET` on its own -- registers nothing and so used to ask kqueue nothing, which let a directory and `/dev/null` through. It consults the target directly now.

The batched changelist also leaked. `kevent()` stops at the first failed change, so a timerfd registered for `EPOLLIN|EPOLLOUT` left the read filter armed behind a call the guest was told had failed. Each filter is registered on its own now, and a refused write filter is dropped rather than failing the whole `ADD`, which matches a target Linux accepts but never reports writable.

## Why fstat cannot decide the plain-file half

The fd type cannot answer "is this a plain file": `FD_REGULAR` also covers a fifo, a socket or a tty opened by path, which `src/syscall/asyncio.c` already says in the `fd_keeps_fasync` comment. So the check asks the host object.

`fstat` alone is not enough either, and this is the part that took measuring. Several intercepted trees are served from ordinary host files: `/proc` from an anonymous temp file, `/sys/devices/system/cpu` and `/etc/mtab` from a scratch directory. Linux polls all three, so a rule keyed on `S_ISREG` refuses targets a real kernel accepts. An earlier revision of this branch did exactly that and regressed `/sys/devices/system/cpu/online` and `/etc/mtab` from 0 to `EPERM`.

The revision after it recorded one bit -- "this open went through an intercept" -- and exempted it from the plain-file rule. That bit is too coarse, because it conflates two groups Linux separates: `/proc/<pid>/mountinfo` is pollable and `/proc/<pid>/stat` is not, and both are ordinary host files here. So the fd table now records what Linux would actually answer, decided from the guest path by `path_intercept_poll_capable` and carried as `path_poll_capable`.

The split is not one a reader would guess from what the files contain. A per-process procfs file is opened through `proc_single_file_operations` and has no `->poll`, while every `proc_create` entry gets `proc_reg_poll` whether or not it has anything to report, which is why `/proc/meminfo` is accepted and `/proc/self/stat` is not. `/proc/<pid>/mounts`, `mountinfo` and the `/proc/<pid>/net` subtree are the exceptions inside the process directory; `mountstats` sitting next to them is not one. Every arm was measured against Linux 6.12 rather than read off the kernel source.

The classification lives beside `path_might_use_open_intercept` in `src/syscall/path.c` because the two enumerate the same surface: an intercept added to one without the other is a target answered from the wrong object.

The answer belongs to the open file description rather than to the name. It travels with the fd table across fork, an alias inherits it so that a dup and its source agree, and an opened magic link keeps what its source carried rather than being classified from `/proc/self/fd/N`, which is a per-process procfs path whatever it points at. The new fd table entry is on the fork IPC wire, so `FORK_IPC_PROTOCOL_MAGIC` moves `ELFP` to `ELFQ` and `tests/test-fork-ipc-protocol-host.c` gains the matching assertion. An old and a new host process can no longer fork to each other, which is what the magic exists to enforce.

## Why the probe knote is disabled

`epoll_target_pollable` exists because an `EPOLLOUT`-only `ADD` never asks kqueue about the read filter, so a refused write filter cannot otherwise be told from a target that takes no knote at all.

Its probe knote is added with `EV_DISABLE`, which is load bearing rather than tidiness. `sys_epoll_pwait` blocks in `kevent()` on this same kqueue without holding `inst->lock`, and an enabled probe on an already-ready target wakes that wait and hands it an event whose NULL `udata` reads back as guest fd 0, which `sys_epoll_pwait` then drops and returns 0 early with no retry. Measured on the host: with the probe enabled, a `kevent()` blocked for 2s on another thread returns after 300ms carrying `udata 0x0`; with `EV_DISABLE` it rides out the full 2s, and the probe still tells `/dev/null` and a directory (`EINVAL`) from a plain file and a kqueue (accepted).

## The DEL path leaked the same way (second commit)

The first revision of this branch listed the batched `EPOLL_CTL_DEL` under "what this does not fix", on the reading that the order of its two changes made it safe. That reading was wrong, and this branch is what makes it reachable.

`EPOLL_CTL_DEL` builds its changelist from `reg->events`, which names the filters the registration asked for rather than the filters kqueue still holds. The two diverge in two ways: `EPOLLONESHOT` removes only the filter that fired, and the `ADD` path above now drops a write filter the target refuses. So the first `EV_DELETE` is often for a filter that is already gone, `kevent()` stops there with `ENOENT`, and the deletes behind it never run.

The knote left behind wakes the next `epoll_pwait` on that instance with an event for an fd that is no longer registered, which the wait drops before returning 0 ahead of its timeout: the same spurious wakeup the disabled probe above avoids, reached through a different door.

Measured end to end on a socketpair whose send buffer is full, so the target is readable but not writable. With `EPOLLIN|EPOLLOUT|EPOLLONESHOT` registered, the read filter fires and `EV_ONESHOT` removes it, `DEL` then fails on the read delete and skips the write delete, and the `epoll_wait(300)` that follows returns after 0ms. Each delete goes in its own call now, the way `MOD` already did, and that wait rides out its full 300ms.

## What this does not fix

elfuse does not serve `/proc/<pid>/net` today, so that arm of the classification is unreachable here. It is present, and the comment above it says so, because the day the tree does serve it the rule would otherwise be silently wrong.

## Checks

Rebased on `d410047`. `2ab2139` and `3f488c5` moved the event copy to the top of `sys_epoll_ctl` and reordered the errors around it, so the plain-file test was rebased onto that shape and moved ahead of the `EEXIST` and `ENOENT` checks, which is where `do_epoll_ctl` reads the target's poll support. `tests/test-epoll.c`, 26 cases including the new error-order ones, passes under elfuse and under the reference kernel with this change on top.

`make check`, rc=0: 89 in the manifest suite, 7/7, 82/84 (the two skips need a raw socket and an interactive terminal), 27/27, 4/4, guardrail PASS.

`make check-format`, rc=0. `make indent`, a no-op. `make lint`, rc=0, with no new clang-tidy warning on a touched line.

`make test-matrix-elfuse-aarch64`, 270 passed, 0 failed. `make test-matrix-qemu-aarch64`, 249 passed, 0 failed.

`tests/test-epoll-del-leak.c` is three cases pinning the wait after a `DEL`. It is registered in both `tests/manifest.txt` and `tests/test-matrix.sh`, since the assertion is a timeout the reference kernel adjudicates as readily as elfuse does. On the build without the `DEL` change its third case fails, the wait returning after 0ms instead of 300ms.

`tests/test-epoll-unsupported.c` is 57 cases and produces byte-identical output under elfuse and under `qemu-system-aarch64` running the fixtures' Linux 6.12 kernel. It is registered in `tests/test-matrix.sh` rather than `tests/manifest.txt`, since a real kernel can adjudicate it. Reverting only the `src/` changes turns 20 of its assertions red, so the cases pin the change rather than restating it.

Beyond the suite, a probe over timerfd, eventfd, a nested epoll fd, a pipe, a socket, a plain file, a fifo opened by path, a directory, character devices, opened magic links, and forty-four paths under `/proc`, `/sys` and `/etc`, against `EPOLLIN`, `EPOLLOUT`, both, an empty mask, and all three operations, answers identically under elfuse and the reference kernel.

Fifteen of those paths disagreed before this change: `/etc/passwd`, `/etc/group`, `/proc/<pid>/stat` in its numeric spelling, `/proc/self/fdinfo/0`, and `stat`, `status`, `maps`, `smaps`, `cmdline`, `environ`, `auxv`, `io`, `limits`, `oom_score` and `oom_score_adj` under `/proc/self`. None of them is refused wrongly now, and no target the reference kernel accepts became refused.

---
## Summary by cubic
Decides EPERM in epoll_ctl from the target alone to match Linux. Previously we forwarded kqueue errors and rejected EPOLLOUT on timerfd/nested epoll; now unsupported targets return EPERM for ADD/MOD/DEL, EPOLLOUT on timerfd/nested epoll is accepted, and partial registrations and deletes no longer leak.

- Plain files and other non-pollable fds return EPERM; intercepted opens record a path_poll_capable flag and carry it across dup and fork.
- Unregistered MOD/DEL return ENOENT only for pollable targets; non-pollable targets return EPERM.
- An empty or non-readiness mask consults the target and returns EPERM on unsupported fds.
- Read/write filters register via separate kevent calls; a refused write filter is dropped; EINVAL on the read filter (or when no knote is accepted) maps to EPERM.
- EPOLL_CTL_DEL deletes each filter in its own kevent call and ignores delete errors, so leftover knotes no longer wake epoll_pwait.
- Adds a disabled-read probe to detect unsupported targets without waking sys_epoll_pwait.
- Adds tests/tests-epoll-unsupported.c and tests/test-epoll-del-leak.c; wires them into tests/test-matrix.sh and manifest.

**Migration**
- Bumps fork IPC protocol magic to ELFQ; mixed old/new processes are incompatible. Upgrade all components together.

<sup>Written for commit 2548742d930360a3cf93ed1bd942b79c1e579b84. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/332?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/runtime/fork-state.h
+++ b/src/runtime/fork-state.h
@@ -19,7 +19,7 @@
 /* Fork IPC protocol identity. Bump this whenever the header layout or ordered
  * fork payload changes incompatibly.
  */
-#define FORK_IPC_PROTOCOL_MAGIC 0x454C4650U /* "ELFP" */
+#define FORK_IPC_PROTOCOL_MAGIC 0x454C4651U /* "ELFQ" */
 
 #define IPC_MAGIC_HEADER FORK_IPC_PROTOCOL_MAGIC
 #define IPC_MAGIC_SENTINEL 0x454C4F4BU /* "ELOK" */
@@ -116,8 +116,12 @@ typedef struct {
      * again: whether the description came from outside elfuse (the launcher's
      * stdio, or an alias of it) and whether elfuse owns its O_NONBLOCK. See
      * fd_alias_carried.
+     *
+     * path_poll_capable rides along for the same reason and cannot be rebuilt
+     * on the far side: the child holds the descriptor, not the guest path the
+     * parent classified it from.
      */
-    int32_t foreign_description, nonblock_owned;
+    int32_t foreign_description, nonblock_owned, path_poll_capable;
     char proc_path[FD_VIRTUAL_PATH_MAX];
 } ipc_fd_entry_t;
 
--- a/src/syscall/fdtable.c
+++ b/src/syscall/fdtable.c
@@ -307,6 +307,7 @@ static inline void fd_init_entry(int fd,
                 (src->linux_flags & FD_DESCRIPTION_FLAGS);
             fd_alias_spec.foreign_description = src->foreign_description;
             fd_alias_spec.nonblock_owned = src->nonblock_owned;
+            fd_alias_spec.path_poll_capable = src->path_poll_capable;
         }
     }
 
@@ -339,6 +340,14 @@ static inline void fd_init_entry(int fd,
         fd_table[fd].linux_flags = accmode < 0 ? 0 : accmode;
     fd_table[fd].dir = NULL;
     fd_table[fd].proc_path[0] = '\0';
+
+    /* An alias shares the description, so it shares the answer: a dup of an
+     * intercepted open is the same file by another name, and epoll_ctl on the
+     * two names has to agree. Only an open that resolved a path decides this
+     * from scratch, which is why fd_alloc_opened_host leaves an alias alone.
+     */
+    fd_table[fd].path_poll_capable =
+        fd_alias_pending && fd_alias_spec.path_poll_capable;
     fd_table[fd].seals = 0;
 
     /* Whether a host read/write can block, so the fast-path and slow-path
@@ -851,6 +860,7 @@ fd_lifetime_t *fd_mark_closed_unlocked(int fd)
     fd_table[fd].ofd_id = 0;
     fd_table[fd].dir = NULL;
     fd_table[fd].proc_path[0] = '\0';
+    fd_table[fd].path_poll_capable = false;
     fd_table[fd].linux_flags = 0;
     fd_table[fd].seals = 0;
     fd_table[fd].fasync_owner_type = FASYNC_OWNER_NONE;
--- a/src/syscall/fs.c
+++ b/src/syscall/fs.c
@@ -492,6 +492,14 @@ static int fd_alloc_opened_host(int host_fd,
         if (have_proc_path)
             memcpy(fd_table[guest_fd].proc_path, proc_path_buf,
                    sizeof(proc_path_buf));
+
+        /* An alias already carries the description's answer, and the path would
+         * describe the magic link rather than the file behind it:
+         * /proc/self/fd/N is a per-process procfs name whatever it points at.
+         */
+        if (!spec)
+            fd_table[guest_fd].path_poll_capable =
+                path_intercept_poll_capable(virtual_path);
         bool readable_urandom =
             type == FD_URANDOM &&
             (linux_flags & LINUX_O_ACCMODE) != LINUX_O_WRONLY;
--- a/src/syscall/path.c
+++ b/src/syscall/path.c
@@ -76,6 +76,11 @@ bool path_might_use_open_intercept(const char *path)
  * when the path names no process directory at all: "/proc/self/stat" and
  * "/proc/41/stat" both yield "stat", "/proc/self" yields "", and
  * "/proc/meminfo" yields NULL.
+ *
+ * A "task/<tid>/" segment is stripped with the process directory it sits in, so
+ * "/proc/41/task/41/mounts" yields "mounts" too. Linux answers the thread
+ * spelling of these names the way it answers the process one, measured on 6.12
+ * over mounts, mountinfo, net/dev and stat.
  */
 static const char *proc_pid_dir_suffix(const char *path)
 {
@@ -96,7 +101,18 @@ static const char *proc_pid_dir_suffix(const char *path)
             return NULL;
         p = d;
     }
-    return *p == '/' ? p + 1 : p;
+    if (*p == '/')
+        p++;
+
+    if (!strncmp(p, "task/", 5)) {
+        const char *d = p + 5;
+        const char *tid = d;
+        while (*d >= '0' && *d <= '9')
+            d++;
+        if (d != tid && (*d == '\0' || *d == '/'))
+            p = (*d == '/') ? d + 1 : d;
+    }
+    return p;
 }
 
 /* Whether Linux gives the file behind this intercepted guest path a poll
--- a/src/syscall/path.c
+++ b/src/syscall/path.c
@@ -72,6 +72,75 @@ bool path_might_use_open_intercept(const char *path)
     return false;
 }
 
+/* The part of a guest /proc path that follows the process directory, or NULL
+ * when the path names no process directory at all: "/proc/self/stat" and
+ * "/proc/41/stat" both yield "stat", "/proc/self" yields "", and
+ * "/proc/meminfo" yields NULL.
+ */
+static const char *proc_pid_dir_suffix(const char *path)
+{
+    if (strncmp(path, "/proc/", 6) != 0)
+        return NULL;
+
+    const char *p = path + 6;
+    if (!strncmp(p, "self", 4) && (p[4] == '\0' || p[4] == '/')) {
+        p += 4;
+    } else if (!strncmp(p, "thread-self", 11) &&
+               (p[11] == '\0' || p[11] == '/')) {
+        p += 11;
+    } else {
+        const char *d = p;
+        while (*d >= '0' && *d <= '9')
+            d++;
+        if (d == p || (*d != '\0' && *d != '/'))
+            return NULL;
+        p = d;
+    }
+    return *p == '/' ? p + 1 : p;
+}
+
+/* Whether Linux gives the file behind this intercepted guest path a poll
+ * method, which is the only thing epoll_ctl reads EPERM off. fstat cannot
+ * answer it: elfuse serves several intercepted trees from ordinary host files,
+ * so the host object describes elfuse's staging rather than the file the guest
+ * named, and the path is what still knows.
+ *
+ * The answer per family was measured against Linux 6.12 rather than reasoned
+ * from the file's contents, because the two do not track each other: a
+ * per-process procfs file is opened through proc_single_file_operations, which
+ * carries no poll, while every proc_create entry gets proc_reg_poll whether or
+ * not it has anything to report. /proc/<pid>/mounts and mountinfo are the
+ * exception that makes the split visible -- they go through mounts_operations
+ * so a guest can wait for the mount table to change, while mountstats next to
+ * them does not and is refused. The /proc/<pid>/net subtree is the other
+ * exception, measured on Linux rather than reached here: elfuse serves
+ * /proc/net but not yet the per-process spelling of it, so that arm is
+ * unreachable today and is present so it does not become wrong the day it is.
+ */
+bool path_intercept_poll_capable(const char *path)
+{
+    if (!path || path[0] != '/')
+        return false;
+
+    const char *pid_rel = proc_pid_dir_suffix(path);
+    if (pid_rel)
+        return !strcmp(pid_rel, "mounts") || !strcmp(pid_rel, "mountinfo") ||
+               !strncmp(pid_rel, "net/", 4);
+
+    /* sysfs attributes are pollable through kernfs, and /etc/mtab is a symlink
+     * onto the mount table. What is left of the intercept surface --
+     * /etc/passwd and /etc/group, the utmp files, /dev and the FUSE mounts --
+     * is a plain file, a character device or a fifo, all of which the host
+     * object describes correctly, so the caller's fstat and kqueue probe answer
+     * for them.
+     */
+    if (!strncmp(path, "/proc/", 6))
+        return true;
+    if (path_prefix_match(path, SYSFS_CPU_PREFIX, sizeof(SYSFS_CPU_PREFIX) - 1))
+        return true;
+    return !strcmp(path, "/etc/mtab");
+}
+
 bool path_might_use_stat_intercept(const char *path)
 {
     if (!path || path[0] != '/')
--- a/src/syscall/path.c
+++ b/src/syscall/path.c
@@ -145,14 +145,22 @@ bool path_intercept_poll_capable(const char *path)
 
     /* sysfs attributes are pollable through kernfs, and /etc/mtab is a symlink
      * onto the mount table. What is left of the intercept surface --
-     * /etc/passwd and /etc/group, the utmp files, /dev and the FUSE mounts --
-     * is a plain file, a character device or a fifo, all of which the host
-     * object describes correctly, so the caller's fstat and kqueue probe answer
-     * for them.
+     * /etc/passwd and /etc/group, the utmp files, most of /dev and the FUSE
+     * mounts -- is a plain file, a character device or a fifo, all of which the
+     * host object describes correctly, so the caller's fstat and kqueue probe
+     * answer for them.
+     *
+     * /dev/random is the exception, and it does not extend to the /dev/urandom
+     * beside it. random_fops carries .poll and urandom_fops does not, since a
+     * read from urandom never waits, so Linux 6.12 answers 0 for the first and
+     * EPERM for the second. macOS refuses a knote on both, which leaves the
+     * host object unable to tell them apart.
      */
     if (!strncmp(path, "/proc/", 6))
         return true;
     if (path_prefix_match(path, SYSFS_CPU_PREFIX, sizeof(SYSFS_CPU_PREFIX) - 1))
+        return true;
+    if (!strcmp(path, "/dev/random"))
         return true;
     return !strcmp(path, "/etc/mtab");
 }
--- a/src/syscall/poll.c
+++ b/src/syscall/poll.c
@@ -18,6 +18,7 @@
 #include <errno.h>
 #include <limits.h>
 #include <sys/event.h>
+#include <sys/stat.h>
 #include <poll.h>
 
 #include "utils.h"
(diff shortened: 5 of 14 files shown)


src/runtime/fork-state.h
```diff
@@ -117,7 +117,7 @@ typedef struct {
      * stdio, or an alias of it) and whether elfuse owns its O_NONBLOCK. See
      * fd_alias_carried.
      */
-    int32_t foreign_description, nonblock_owned;
+    int32_t foreign_description, nonblock_owned, synthesized;
```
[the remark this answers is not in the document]
xalestar: Valid, fixed in `b05c1a9`. tests/test-fork-ipc-protocol-host.c pins the magic with a _Static_assert, so the bump to ELFQ comes with the matching reject-ELFP assertion.

src/syscall/poll.c
```diff
@@ -1533,11 +1588,32 @@ int64_t sys_epoll_ctl(guest_t *g, int epfd, int op, int fd, uint64_t event_gva)
...
+    /* Linux decides EPERM from the target alone, never from the requested
+     * events: epoll_ctl(2) accepts EPOLLOUT on a timerfd or on a nested epoll
+     * fd, neither of which ever reports itself writable. Both are kqueues here,
+     * and kqueue refuses EVFILT_WRITE on a kqueue with EINVAL, so register one
+     * filter per call and drop a refused write filter instead of failing the
+     * whole ADD. A batched call would also stop at the first failed change and
+     * leave the survivor registered. EINVAL on the read filter, or on a write
+     * filter whose target takes no knote at all, is the EPERM case.
+     */
+    bool read_registered = false;
```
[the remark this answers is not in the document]
xalestar: Valid, fixed in `b05c1a9`. A mask naming no readiness filter left nchanges at zero, so the loop asked kqueue nothing. It probes the target directly in that case now.

src/syscall/fs.c
```diff
@@ -491,6 +491,7 @@ static int fd_alloc_opened_host(int host_fd,
         if (have_proc_path)
             memcpy(fd_table[guest_fd].proc_path, proc_path_buf,
                    sizeof(proc_path_buf));
+        fd_table[guest_fd].synthesized = virtual_path != NULL;
```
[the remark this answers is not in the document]
xalestar: Valid, and fixed differently than suggested, in `b05c1a9`. Preserving the marker was not enough: it recorded "an intercept ran", which conflates /proc/<pid>/mountinfo (pollable on Linux) with /proc/<pid>/stat (not), and both are ordinary host files here. path_intercept_poll_capable classifies by guest path instead, and the field is renamed path_poll_capable to say what it answers. Measured by diffing one binary's transcript under elfuse and under Linux 6.12 over 44 paths: 15 disagreed before, 0 now, and nothing Linux accepts became refused.

The magic-link half goes the other way round: /proc/self/fd/N is a per-process procfs path whatever it points at, so an alias keeps its source's answer rather than being classified from the link. All four link-to-{mountinfo, /proc/self/stat, regular file, pipe} cases match the reference kernel.

src/syscall/fdtable.c
```diff
@@ -339,6 +339,7 @@ static inline void fd_init_entry(int fd,
         fd_table[fd].linux_flags = accmode < 0 ? 0 : accmode;
     fd_table[fd].dir = NULL;
     fd_table[fd].proc_path[0] = '\0';
+    fd_table[fd].synthesized = false;
```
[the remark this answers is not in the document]
xalestar: Valid, fixed in `b05c1a9`. install_fd_alias_metadata_atomic does not exist in this tree; the reset was in fd_init_entry. The flag now travels through fd_alias_spec_t beside foreign_description and nonblock_owned, so fd_alias_of and fd_alias_host_shared both carry it. Covered both ways: a dup of /proc/self/mountinfo is accepted, a dup of /proc/self/stat is refused.

tests/test-epoll-unsupported.c
```diff
@@ -0,0 +1,255 @@
...
+#include <sys/wait.h>
+
+#include "test-harness.h"
+
+int passes = 0, fails = 0;
+
+/* ADD fd to a fresh epoll instance and report the raw result. */
+static int add_to_fresh_epoll(int fd, uint32_t events)
+{
+    int epfd = epoll_create1(0);
```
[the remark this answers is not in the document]
xalestar: Valid, fixed in `b05c1a9`.

tests/test-epoll-unsupported.c
```diff
@@ -0,0 +1,255 @@
...
+    expect_op_errno("regular file MOD", "/etc/hosts", EPOLL_CTL_MOD, EPERM);
+    expect_op_errno("regular file DEL", "/etc/hosts", EPOLL_CTL_DEL, EPERM);
+    expect_op_errno("/dev/null MOD", "/dev/null", EPOLL_CTL_MOD, EPERM);
+    expect_op_errno("/dev/null DEL", "/dev/null", EPOLL_CTL_DEL, EPERM);
+    expect_op_errno("unregistered MOD", "/proc/self/mountinfo", EPOLL_CTL_MOD,
+                    ENOENT);
+    expect_op_errno("unregistered DEL", "/proc/self/mountinfo", EPOLL_CTL_DEL,
+                    ENOENT);
+
+    unlink("/tmp/epoll-unsupported-fifo");
```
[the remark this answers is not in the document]
xalestar: Valid, fixed in `b05c1a9`. The path is pid-suffixed and now follows the /tmp/elfuse-* naming the rest of tests/ uses.

tests/test-epoll-unsupported.c
```diff
@@ -0,0 +1,402 @@
...
+{
+    printf("epoll_ctl target-support tests\n");
+
+    /* No poll support: EPERM whichever events are asked for. The EPOLLOUT-only
+     * case is the one that cannot read the answer off the write filter alone.
+     */
+    expect_eperm("/dev/null EPOLLIN", open("/dev/null", O_RDWR), EPOLLIN);
+    expect_eperm("/dev/null EPOLLOUT", open("/dev/null", O_RDWR), EPOLLOUT);
+    expect_eperm("/dev/zero EPOLLIN", open("/dev/zero", O_RDONLY), EPOLLIN);
+    expect_eperm("/dev/urandom EPOLLIN", open("/dev/urandom", O_RDONLY),
```
jserv: This asserts the opposite of what the reference kernel answers. drivers/char/random.c sets `.poll = random_poll` on both `random_fops` and `urandom_fops`, so `file_can_poll()` is true and `epoll_ctl(ADD, /dev/urandom, EPOLLIN)` returns 0 on Linux, not EPERM.

That makes it a src-side divergence too, not just a wrong assertion. macOS kqueue refuses a knote on /dev/urandom, which this tree already records in the `poll_eval_unpollable` comment at src/syscall/poll.c:121, so `sys_epoll_ctl` now answers EPERM for a target the reference kernel accepts. /dev/random and /dev/urandom need the same override `path_poll_capable` gives /proc, since the host object cannot answer for them either.

Worth rechecking how the qemu run agreed here. A fixture where /dev/urandom is a plain file rather than a device node would make this case pass for the wrong reason and hide the src half.

src/syscall/poll.c
```diff
@@ -1565,11 +1627,42 @@ int64_t sys_epoll_ctl(guest_t *g, int epfd, int op, int fd, uint64_t event_gva)
...
+    for (int i = 0; i < nchanges; i++) {
+        if (kevent(epoll_ref.fd, &changes[i], 1, NULL, 0, NULL) == 0) {
+            if (changes[i].filter == EVFILT_READ)
+                read_registered = true;
+            continue;
+        }
+        if (errno != EINVAL) {
```
jserv, line 1750: This exit leaks the same way the batched call did. If the read filter registered on the previous iteration and the write filter fails with anything other than EINVAL, the knote is already armed and carries `udata` naming a guest fd whose `reg->active` never becomes true, because the store below is skipped.

`sys_epoll_pwait` does reap it (the `!inst->regs[gfd].active` arm around line 1945 issues the EV_DELETE), but only after the wait has been woken and the event counted, so `nout` stays 0 and the call returns 0 ahead of its timeout. That is the spurious wakeup the second commit removes from the DEL path, reached through a third door.

Deleting the filters registered so far before the goto closes it, and covers MOD as well, where the deletes above have already run against the old `reg->events` and the re-registration then fails halfway.

src/syscall/poll.c
```diff
@@ -1458,36 +1495,60 @@ int64_t sys_epoll_ctl(guest_t *g, int epfd, int op, int fd, uint64_t event_gva)
...
         }
         ret = 0;
         goto out_locked;
     }
 
+    /* The plain-file half of epoll_target_supported, inline so an ADD does not
+     * pay for the kqueue probe: the registration below already asks kqueue the
+     * rest of the question. It sits ahead of the EEXIST and ENOENT checks
+     * because do_epoll_ctl tests the target's poll support before it looks up
+     * the registration, so EPERM outranks both.
```
jserv: The comment states the ordering slightly short. `do_epoll_ctl` puts `file_can_poll()` ahead of `is_file_epoll(f.file)` and ahead of the op switch, not only ahead of the registration lookup, so EPERM outranks two EINVALs this function still answers first:

- `epoll_ctl(epfd, 99, plain_file_fd, &ev)` is EPERM on Linux, EINVAL here (the op check earlier in this function).
- `epoll_ctl(regular_fd, ADD, plain_file_fd, &ev)` is EPERM on Linux, EINVAL here (`epoll_instance_acquire` fails before the target is looked at).

tests/test-epoll-unsupported.c only ever passes a real epfd and a valid op, so neither case is pinned.

src/syscall/path.c
```diff
@@ -72,6 +72,75 @@ bool path_might_use_open_intercept(const char *path)
...
+ */
+bool path_intercept_poll_capable(const char *path)
+{
+    if (!path || path[0] != '/')
+        return false;
+
+    const char *pid_rel = proc_pid_dir_suffix(path);
+    if (pid_rel)
+        return !strcmp(pid_rel, "mounts") || !strcmp(pid_rel, "mountinfo") ||
+               !strncmp(pid_rel, "net/", 4);
```
jserv, line 144: The /proc/<pid>/net arm is kept for the day the tree serves it, which is the right instinct. The /proc/<pid>/task/<tid>/ spelling of these same names reaches `proc_pid_dir_suffix` as `task/<tid>/net/...` and falls through to false.

If Linux answers the tid directory the way it answers the tgid one, that arm wants the same forward-looking treatment. If it does not, the comment above is the place to say so, since a reader who has just been told the net arm is present for a tree elfuse does not serve yet will ask why the tid spelling is not.

jserv (commented):
The design holds up: the probe cannot clobber a sibling knote (every guest fd owns its own host fd, since sys_dup, epoll_dup_fd and the magic-link open all go through dup), the blanket /proc/ rule is right (proc_get_inode installs proc_reg_file_ops, which carries .poll unconditionally, for every proc_create entry), and the fork wire, the EV_DISABLE argument, the lock order and the errno handling around the ignored kevent calls all check out. The comments below are the parts that did not.

jserv (changes requested):
Rebase latest `main` branch which resolves unintended CI breakage.

tests/test-epoll-unsupported.c
xalestar: Half of this holds and the other half measures the other way, so the assertion stays and the src fix covers /dev/random only.

Measured on the fixture kernel, Linux 6.12.103, after confirming both nodes are device nodes rather than plain files (crw-rw-rw- 1, 8 and 1, 9), so the qemu agreement was not the artifact you suspected:

```
epoll_ctl(ADD, /dev/random, EPOLLIN)    0, and epoll_wait reports EPOLLIN at once
epoll_ctl(ADD, /dev/random, EPOLLOUT)   0, and epoll_wait reports nothing
epoll_ctl(ADD, /dev/urandom, EPOLLIN)   EPERM
```

random_fops carries .poll and urandom_fops does not, since a read from urandom never waits and there is nothing for a wait to report. The existing /dev/urandom assertion was right and is unchanged.

/dev/random is fixed in 73077e1. path_intercept_poll_capable answers for it the way it already answers for /proc, since the host object cannot tell the two nodes apart.

The errno was only half of it. With no knote to carry the readiness, accepting the ADD would have replaced EPERM with a wait that blocks to its deadline for an event that cannot arrive. The registration records that its read readiness is not coming through kqueue and the wait answers from the registration, reusing the shape the pty hangup scan established, including the instance counter that keeps other instances off the scan. EPOLLOUT stays unreported, matching random_poll, which offers it only before the pool is seeded.

The 44 path transcript now matches the reference kernel on every line except the pid two labels embed.

src/syscall/poll.c
xalestar: Valid, fixed in ad6ee4a. epoll_undo_changes deletes what the pass armed before the bail, which covers the MOD case you name: it arrives with the old filters already deleted, so a re-registration failing halfway would otherwise leave the instance holding one filter for an entry describing two.

The EPERM arm needs no undo, and the comment says why: it is reachable only while nothing is armed, since a refused read filter is the first change submitted, and a refused write filter that finds no registered read filter means the mask named no read filter to submit.

No test drives it. Provoking it needs a kevent failing with something other than EINVAL on the second filter of a target that accepted the first, and nothing in the fd table reaches that state on demand.

src/syscall/poll.c
xalestar: Valid on both halves, fixed in c3496a9. Measured on Linux 6.12 with a pipe as the control that stays EINVAL:

```
epoll_ctl(ep, 99, plain_file, &ev)            EPERM, was EINVAL
epoll_ctl(plain_file, ADD, plain_file, &ev)   EPERM, was EINVAL
```

The second never reached the target at all, since epoll_instance_acquire returned before the snapshot was taken. The snapshot moves ahead of the instance lookup, matching the kernel reaching both fdgets first.

Answering there needs the target's poll support with no instance to probe on, so epoll_target_supported_standalone opens its own kqueue, and only on a call that is already failing. The plain-file half answers most of them without one.

The comment is rewritten to state the full ordering rather than only the registration lookup. tests/test-epoll-unsupported.c pins both cases, each paired with the pollable control, since a lone EPERM would also pass if it came from the wrong reason.

The pairing check after it is left alone: a guest fd equal to the epfd is an epoll descriptor, which has a poll method, so EPERM never outranks that EINVAL.

src/syscall/path.c
xalestar: Linux does answer the tid directory the way it answers the tgid one, so that arm gets the same treatment, in a8d1364. Measured on 6.12 over mounts, mountinfo, net/dev and stat: each pair agrees, epoll_ctl accepting the first three through both spellings and refusing stat through both.

proc_pid_dir_suffix now strips a task/<tid> segment with the process directory it sits in, so /proc/41/task/41/mountinfo reduces to mountinfo. Like the net arm it is unreachable here, since the tree serves no path under task/<tid> that the classification would accept, and the comment says what was measured so the strip does not read as a guess.

PR #336, opened 2026-08-26
Support OCI image pull, run, unpack, inspect, and cleanup

### LINKED ISSUE #31: OCI image support for Linux userspace execution on macOS

henrybear327:
We leverage the OCI image to build the sysroot, but, we don't need to support the file OCI image lifecycle. 

Thus, in this PR, we only implement pull, inspect, unpack, and run, leveraging existing open source OCI image libraries. For clean up, we will just remove the image layers directly.

We have also added several workloads, such as Redis, to make sure that elfuse can properly leverage OCI image to build sysroot, and the execution can be run properly.

The total go code that is responsible for the logic is about 2k lines, the rest is test, or supporting scripts or configuration code.

Closes https://github.com/sysprog21/elfuse/issues/31

---
## Summary by cubic
Adds `elfuse-oci`, a separate Go CLI that pulls OCI images into a local cache, materializes them as sysroots, and runs them under elfuse, closing #31. OCI handling stays out of elfuse itself.

- Supports `pull`, `inspect`, `unpack`, `run`, and `clean`, with Docker config credentials and Linux `arm64`/`amd64` platform selection.
- `run` auto-pulls missing images, resolves the OCI launch configuration, injects host networking files, and propagates the guest exit status.
- macOS uses case-sensitive APFS sparsebundles with per-run COW clones by default; other platforms use plain rootfs directories.
- Unpacking handles whiteouts, devices and fifos, Darwin mode-bit restrictions, absolute symlinks, and atomic digest-keyed cache publication.
- `clean --cache` removes unpacked caches while retaining blobs and pins; full cleanup removes the store and orphaned bundle volumes. Corrupt caches require cleanup and a fresh pull.
- Build targets, usage documentation, and OCI design notes are included; Go targets are skipped when the required toolchain is unavailable.
- CI adds Linux unit tests, macOS filesystem checks, end-to-end HVF smoke tests, and real guest workloads for C, Go, JVM, Node, Python, and Redis.

<sup>Written for commit 67ca4740fa94f4a6a6646b290e2446cb943b155f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/336?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


henrybear327:
We need to rely on sparsebundle due to the potential issues with the case-insensitive MacOS filesystem causing problem during unpacking.

jserv:
Merge docs/oci-design.md and docs/oci-testing.md into docs/oci-images.md

README.md
```diff
@@ -37,6 +37,19 @@ linker resolved against an external sysroot via `--sysroot`.
 - Self-contained test matrix that cross-checks elfuse against QEMU
   and exercises a separate Rosetta acceptance suite
 
+## OCI Images
```
jserv, line 41: Refer to this subsection in prior statements about `sys root`.

jserv (changes requested):
Indent all shell scripts with `shfmt`.

henrybear327:
> Indent all shell scripts with `shfmt`.

Sorry for missing this. Will add checks on CI in a follow-up PR.

README.md
henrybear327: Referenced.

.github/actions/hvf-elfuse-setup/action.yml
```diff
@@ -0,0 +1,74 @@
...
+          -H "Accept: application/vnd.github+json" \
+          "https://api.github.com/repos/$REPO/pulls/$PR_NUMBER" \
+          | python3 -c 'import json,sys; print(json.load(sys.stdin)["head"]["sha"])') \
+          || latest=""
+        echo "Run targets : $RUN_SHA"
+        echo "PR HEAD now : ${latest:-<unknown>}"
+        if [ -n "$latest" ] && [ "$latest" != "$RUN_SHA" ]; then
+          echo "::error::This run targets $RUN_SHA, but PR #$PR_NUMBER HEAD is now $latest, so the commit is no longer the latest. Failing instead of re-testing stale code on the self-hosted runner; re-run CI on the current commit."
+          exit 1
+        fi
```
jserv: Move these commands to dedicated scripts in top-level `.ci/` directory.

henrybear327: Attempted.

jserv (commented):
Reviewed the new `cmd/elfuse-oci` package and the CI scripts. The package builds, `go vet` and `go test ./cmd/...` are clean, `shellcheck -S warning` over `scripts/ci` is clean, and `docs/oci-images.md` already covers most of what would otherwise be worth raising (the shared-cache `/etc` injection, the publish-by-rename rule, the `st_dev` mount probe caveat).

Five things left, inline.

cmd/elfuse-oci/csrootfs_darwin.go
```diff
@@ -0,0 +1,372 @@
...
+	if err != nil {
+		return err
+	}
+	if err := prepareRootfsForRun(sysroot, spec); err != nil {
+		return err
+	}
+	code, err := spawnForRun(rc.bin, sysroot, spec)
+	if err != nil {
+		return err
+	}
```
jserv, line 95: `provisionCaseSensitive` runs outside the store lock, so two cold `run`s of the same digest can interleave: B observes `isMountPoint(mnt) == false`, A's attach then completes, and B's `clearDir` walks the volume A just mounted and `RemoveAll`s it, taking out the base tree and any live run clone with it.

Take `s.withLock` around provisioning, or at minimum re-check `isMountPoint` after `clearDir` and before attaching.

cmd/elfuse-oci/unpack.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	return unpackImageFresh(ctx, s, m, dest)
+}
+
+// unpackImageFresh stages in a sibling temp directory and renames, never
+// merges: dest's existence means "fully unpacked" to every reader.
+func unpackImageFresh(ctx context.Context, s *store, m ocispec.Manifest, dest string) error {
+	parent := filepath.Dir(dest)
+	if err := os.MkdirAll(parent, 0o755); err != nil {
+		return err
```
jserv, line 136: A failed rename is read as "the concurrent unpacker won", but the only evidence is that something now exists at `dest`. A regular file or a symlink appearing there produces the same answer, and `run` then boots from it.

Gate this on `storeRootfsPublished(dest)` returning true with a nil error, and return the rename error otherwise.

cmd/elfuse-oci/csrootfs_darwin.go
```diff
@@ -0,0 +1,372 @@
...
+	return exitStatus(code)
+}
+
+// provisionCaseSensitive creates the sparsebundle image if absent,
+// attaches it at the slot's mount path unless already mounted, and
+// returns the actual mount path (hdiutil's plist answer on a fresh attach).
+func provisionCaseSensitive(bundleDir, size string) (string, error) {
+	mountPath := csMountPath(bundleDir)
+	if err := os.MkdirAll(bundleDir, 0o755); err != nil {
```
jserv, line 105: The detach error is dropped here and again at line 109, so a sparsebundle that stays attached is invisible to the caller; joining it into the returned error would at least report the leak.

Related: warm reuse at line 93 probes `mountPath`, but the function returns `actual` from the plist, so an attach that landed anywhere else is never found again and the next run tries to attach the same bundle a second time.

cmd/elfuse-oci/csrootfs_darwin.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+	out, err := hdiutil("attach", "-mountpoint", mountPath, "-plist", csBundleImage(bundleDir))
+	if err != nil {
+		return "", err
+	}
+	// attach has already mounted the volume, so a failure past this point
+	// detaches, or the slot leaks a live mount.
+	actual, err := parseMountpoint(out)
+	if err != nil {
+		_ = detachForce(mountPath)
```
jserv: If `hdiutil create` fails after creating the bundle directory, the partial `.sparsebundle` stays on disk, every later run takes the `Lstat` early return at line 119, skips creation, and fails at attach instead. Nothing short of `clean --cache` recovers.

`os.RemoveAll(image)` on this error path makes the next run retry cleanly.

cmd/elfuse-oci/tarfilter.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// Special-mode bits in a tar header (setuid, setgid, sticky).
+const tarSpecialBits = 0o4000 | 0o2000 | 0o1000
+
+// clearSpecialBits names the host capability, not the OS: an
+// unprivileged darwin chmod EPERMs on a foreign-group setgid file.
+const clearSpecialBits = runtime.GOOS == "darwin"
+
+// layerFilter adapts one layer's entries for archive.Apply, which reads
```
jserv: `dropped` is keyed on `path.Clean(hdr.Name)`, which keeps a leading slash, while `hdr.Linkname` for the same entry may be spelled without one, so the lookup at line 32 misses and the hardlink reaches `archive.Apply`, which fails the whole layer linking to a file that was never created.

The same miss happens across layers, since `layerFilter()` is constructed per layer in `unpackInto`, so a hardlink to a device dropped by an earlier layer never matches. Trim the leading slash before keying, and build the filter once per image.

henrybear327:
This branch served as PoC for the feature set. 

I will close this PR, in favor of a split PR of this one for easier review.

PR #341, opened 2026-08-29
Add elfuse-oci pull into a local image store

henrybear327:
## Summary by cubic
Adds `elfuse-oci`, a Go CLI that pulls OCI images into a local store, keeping elfuse an OCI-agnostic runtime.

**New Features**

- `pull` fetches an image via `go-containerregistry` into an OCI layout, pins it in `index.json` by canonical ref and platform, and refuses a legacy `refs.json` store.
- A single flock serializes layout init and pin-table updates; the pin table is rewritten via a synced temp file and rename, so a crash can't publish a truncated table.
- `--timeout` bounds the whole pull including the lock wait; a lock freed after the deadline is released, not used. `--platform` only accepts `linux/arm64` and `linux/amd64`.
- A bare manifest or index child that states no platform is pinned only when its config's platform matches; a child that is itself an index is refused.
- Credentials come from Docker config inline auths or helpers, with the pull anonymous when no entry matches.
- `mk/oci.mk` adds `elfuse-oci`, `oci-test`, and `oci-lint` targets; CI runs race tests and lint on Linux.
- Documents the store, pull flow, and validation in `docs/oci-images.md` and the user guide.

<sup>Written for commit 2f7710bdfa6aad0e0c8c6809d1efb1c603693996. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/341?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


go.mod
```diff
@@ -0,0 +1,37 @@
+module github.com/sysprog21/elfuse
+
+go 1.26.3
```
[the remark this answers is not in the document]
henrybear327: containerd v2.3.4 declares go 1.26.3 in its own go.mod, so that is the real go version which will be used.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,185 @@
...
+	return d, nil
+}
+
+func (s *store) manifestFor(ctx context.Context, dgst string) (ocispec.Manifest, error) {
+	var m ocispec.Manifest
+	d, err := digest.Parse(dgst)
+	if err != nil {
+		return m, fmt.Errorf("store: manifest %s: %w", dgst, err)
+	}
+	b, err := content.ReadBlob(ctx, s.content, ocispec.Descriptor{Digest: d})
```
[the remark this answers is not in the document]
henrybear327: Not so: content.ReadBlob sizes the read from ReaderAt.Size(), which the local store takes from the blob file's stat, not from desc.Size; manifestOf and TestStore* read manifests through exactly this path and pass.

cmd/elfuse-oci/pull.go
```diff
@@ -0,0 +1,169 @@
...
+	resolver := newResolver(ctx)
+	name, desc, err := resolver.Resolve(ctx, name)
+	if err != nil {
+		return fmt.Errorf("pull %s: %w", ref, err)
+	}
+	fetcher, err := resolver.Fetcher(ctx, name)
+	if err != nil {
+		return fmt.Errorf("pull %s: %w", ref, err)
+	}
+	matcher := platforms.OnlyStrict(cf.platform)
```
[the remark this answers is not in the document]
henrybear327: OnlyStrict normalizes both sides, and a probe with platforms@ the pinned version prints strict matches v8: true (and arm/v7: false, which is the point of choosing it over Only).

cmd/elfuse-oci/pull_test.go
```diff
@@ -0,0 +1,196 @@
...
+		Manifests: []ocispec.Descriptor{
+			child(amdDigest, ocispec.Platform{OS: "linux", Architecture: "amd64"}),
+			child(armDigest, defaultPlatform),
+		}}
+	desc := pushBlob(t, s, ocispec.MediaTypeImageIndex, mustJSON(t, idx))
+	got, err := selectManifest(context.Background(), s.content, desc, platforms.OnlyStrict(defaultPlatform))
+	if err != nil || got.Digest.String() != armDigest {
+		t.Fatalf("arm64 child = %v, %v; want %s", got.Digest, err, armDigest)
+	}
+	_, err = selectManifest(context.Background(), s.content, desc, platforms.OnlyStrict(ocispec.Platform{OS: "linux", Architecture: "s390x"}))
```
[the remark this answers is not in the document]
henrybear327: LimitManifests returns ErrNotFound itself when the filtered slice is empty (handlers.go: "no match for platform in manifest"), so children[0] is never reached; TestSelectManifestPicksIndexChild passes as written.

cmd/elfuse-oci/creds.go
```diff
@@ -0,0 +1,206 @@
...
+	var cfg dockerConfig
+	dir := os.Getenv("DOCKER_CONFIG")
+	if dir == "" {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return cfg, nil
+		}
+		dir = filepath.Join(home, ".docker")
+	}
+	b, err := os.ReadFile(filepath.Join(dir, "config.json"))
```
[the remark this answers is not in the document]
henrybear327: Real, but it is a docker-cli 28+ addition layered as a fallback store on top of config.json; this PR covers what docker login writes, and the env form is a follow-up rather than part of this change.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return m, fmt.Errorf("store: read manifest %s: %w", dgst, err)
+	}
+	if err := json.Unmarshal(b, &m); err != nil {
```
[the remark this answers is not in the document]
henrybear327: Pins are only written by pull after selectManifest has checked the media type and config platform, so a mismatched pin means a hand-edited refs.json, which this PR does not try to defend against; a reader hits the empty config digest and errors anyway.

mk/oci.mk
```diff
@@ -0,0 +1,58 @@
...
+endif
+
+elfuse-oci: $(OCI_BIN)
+
+# rm -f first: go build -o refuses a path that already exists and is not an
+# object file, so without it a stale non-binary at $@ would fail the build
+# instead of being replaced.
+# The directory itself covers a deleted or renamed source, which OCI_SRCS
+# no longer lists.
+$(OCI_BIN): go.mod go.sum cmd/elfuse-oci $(OCI_SRCS) $(VERSION_DEPS) | $(BUILD_DIR)
```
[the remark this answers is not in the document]
henrybear327: Same contract as $(BUILD_DIR)/version.h in mk/shim.mk, which tracks .git/HEAD, .git/index and mk/config.mk and not an ad-hoc override; keeping the Go binary consistent with the C build rather than special-casing it.

henrybear327:
Final switch of the open source library after trying out different projects and attempting to make it maintainable in the long run

.github/workflows/build.yml
```diff
@@ -126,6 +126,27 @@ jobs:
...
+  oci-linux:
+    name: elfuse-oci (Linux)
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
```
jserv: Bump to actions/setup-go@v7

jserv:
Use shorter and still meaningful paths. e.g., `cmd/oci/`.

docs/oci-images.md
```diff
@@ -0,0 +1,54 @@
+# OCI Images
+
+`elfuse-oci` is a separate Go command that pulls images into a local cache. It
+does not unpack or run them. Command syntax is in
+[usage.md](usage.md#oci-images).
```
jserv, line 13: Address more background about OCI.

mk/oci.mk
```diff
@@ -0,0 +1,54 @@
...
+endif
+export HAVE_GO
+
+OCI_BIN  := $(BUILD_DIR)/elfuse-oci
+OCI_SRCS := $(shell find cmd/elfuse-oci -type f -name '*.go' ! -name '*_test.go' 2>/dev/null)
+OCI_VERSION_FILE := $(BUILD_DIR)/.elfuse-oci-version
+
+.PHONY: elfuse-oci oci-test oci-vet oci-fmt-check oci-lint oci-version-force
+
+ifeq ($(HAVE_GO),yes)
```
jserv: The `HAVE_GO` probe proves only that a toolchain exists, not that the module graph is available. `GOTOOLCHAIN=local go list -m` reads go.mod and nothing else, so it succeeds offline against a cold module cache (checked with `GOPROXY=off`), `HAVE_GO` becomes `yes`, and then a plain `make` dies at `go build ./cmd/elfuse-oci` with `module lookup disabled`. There is no `vendor/` directory in the tree.

That makes the C build depend on network reachability for a component it does not need. Either keep `make elfuse-oci` opt-in and drop `all`/`check` here, or vendor the modules and build with `-mod=vendor`.

cmd/elfuse-oci/pull.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	img, err := desc.Image()
+	if err != nil {
+		return nil, err
+	}
+	actual, err := imagePlatform(img)
+	if err != nil {
+		return nil, err
+	}
+	if !actual.Satisfies(platform) {
```
jserv: `Satisfies` treats an empty requested variant as a wildcard: in go-containerregistry, `satisfies(want, have)` returns true whenever `want == ""`. But `parsePlatform` in common.go rejects every variant the user could type.

So `linux/amd64` accepts an image whose config declares `linux/amd64/v3`, and store.go:392 then rewrites the descriptor platform to the requested one, so `index.json` pins it as baseline `linux/amd64`. A later lookup for `linux/amd64` hands back an image that needs AVX2. Same shape at line 83 for the index children.

Compare the fields directly, or require `actual.Variant == ""` after the arm64/v8 collapse that `parsePlatform` already performs.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,483 @@
...
+	marker, err := os.ReadFile(filepath.Join(s.root, markerName))
+	if os.IsNotExist(err) {
+		if _, legacyErr := os.Lstat(filepath.Join(s.root, "refs.json")); legacyErr == nil {
+			return fmt.Errorf("store: legacy refs.json layout; remove the store and pull again")
+		}
+		entries, readErr := os.ReadDir(s.root)
+		if readErr != nil {
+			return readErr
+		}
+		if len(entries) != 0 {
```
jserv: A crash between the temp-file create and the rename inside `replaceFile` leaves a `.index.json.tmp.<pid>.<nanos>` (or the marker equivalent) in the root. This check then sees a non-empty directory with no marker and refuses the store permanently with `is not an elfuse OCI store`, with nothing telling the user that a manual `rm` is the remedy.

Skip entries carrying the `.<name>.tmp.` prefix here, and remove them while the lock is held.

cmd/oci/store.go
```diff
@@ -0,0 +1,483 @@
...
+func openStore(root string) (*store, error) {
+	root = filepath.Clean(root)
+	if err := os.MkdirAll(root, 0o755); err != nil {
+		return nil, err
+	}
+	fi, err := os.Lstat(root)
+	if err != nil {
+		return nil, err
+	}
+	if !fi.IsDir() {
```
jserv, line 45: `MkdirAll` follows symlinks, `Lstat` does not, so a store reached through a symlink is rejected here with a message that is also untrue. Confirmed on a symlink pointing at a real directory: `MkdirAll` returns nil, then `Lstat(root).IsDir()` returns false.

The default store lives under `~/.local/share`, which people do relocate to another volume with a symlink. `os.Stat` is the right call for a "must be a directory" check.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,483 @@
...
+	if err := tmp.Close(); err != nil {
+		return err
+	}
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	if err := os.Rename(tmpPath, path); err != nil {
+		return err
+	}
+	d, err := os.Open(dir)
```
jserv: Only `blobs/<algorithm>` is fsynced after the rename. On a fresh store the `MkdirAll` at line 210 also creates `blobs/` and `blobs/<algorithm>` themselves, and those directory entries are never made durable.

After power loss `index.json` can be durable while the directory naming its blobs is gone, which is exactly the failure the temp-file, fsync, rename sequence exists to prevent. Fsync each directory `MkdirAll` had to create, up to the store root, before `pinLocked` writes the pin table.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,483 @@
...
+	defer r.Close()
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	dir := filepath.Join(s.root, "blobs", desc.Digest.Algorithm)
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		return err
+	}
+	path := filepath.Join(dir, desc.Digest.Hex)
+	if b, err := os.ReadFile(path); err == nil {
```
jserv: The already-present path reads the whole blob into memory just to rehash it. Re-pulling an image with a 2 GB layer allocates 2 GB per layer on a pull that has no work to do, on a machine that streamed the same layer fine the first time.

Open the file, `io.Copy` it through the hasher, and take the size from `Stat` instead of `len(b)`.

cmd/elfuse-oci/pull.go
```diff
@@ -0,0 +1,161 @@
...
+	return img, nil
+}
+
+func pullImage(ctx context.Context, s *store, ref string, platform ocispec.Platform) error {
+	parsed, err := normalizeRef(ref)
+	if err != nil {
+		return fmt.Errorf("pull %s: %w", ref, err)
+	}
+	fmt.Fprintf(os.Stderr, "Pulling %s...\n", parsed.Name())
+	return s.withLock(ctx, func() error {
```
jserv: The exclusive store lock covers the whole registry transfer, not just the index update. A multi-minute `debian:stable-slim` pull blocks a concurrent `alpine:3` pull for its entire duration, or fails it on its deadline, for two images that share nothing.

Blob writes are content-addressed and already land via temp file plus rename, so they do not need the lock. Taking it around `ensureLayoutLocked` and again around `pinLocked` is enough to keep the pin table serialized.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,483 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if !fi.IsDir() {
+		return nil, fmt.Errorf("store: %s is not a directory", root)
+	}
+	return &store{root: root}, nil
+}
+
+func (s *store) lockPath() string { return s.root + ".lock" }
```
jserv: The lock is a sibling of the store rather than a member of it, so creating it needs write permission on the parent, not on the store. With `/srv` root-owned at 0755 and `/srv/oci` chowned to the user, a pull fails with `permission denied` on `/srv/oci.lock`, a path the user never named, even though the store itself is writable.

`filepath.Join(root, ".lock")` fixes it; exclude that name from the emptiness check at line 77.

cmd/oci/store.go
```diff
@@ -0,0 +1,483 @@
...
+		}
+		size, err := layer.Size()
+		if err != nil {
+			return v1.Descriptor{}, err
+		}
+		media, err := layer.MediaType()
+		if err != nil {
+			return v1.Descriptor{}, err
+		}
+		r, err := layer.Compressed()
```
jserv, line 363: A pull that fails partway leaves every layer already written in the store with nothing referencing it, and no subcommand removes it. Retry against a moved tag and the previous layers stay behind too, so the store grows without bound.

Either record what `publishImage` wrote and unwind on error, or land a prune subcommand and say so in docs/oci-images.md.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,483 @@
...
+	kept := nested.Manifests[:0]
+	for _, desc := range nested.Manifests {
+		if !samePlatform(desc.Platform, platform) {
+			kept = append(kept, desc)
+		}
+	}
+	manifest.Platform = toV1Platform(platform)
+	nested.Manifests = append(kept, manifest)
+	sort.Slice(nested.Manifests, func(i, j int) bool {
+		return nested.Manifests[i].Platform.String() < nested.Manifests[j].Platform.String()
```
jserv: `v1.Descriptor.Platform` is a `*v1.Platform` and `Platform.String()` has a value receiver, so a child with no platform panics here rather than producing an error. `samePlatform` returns false for nil, so such a child survives the filter at line 388 and reaches this comparator.

Reachable for any nested index blob not written by this code. Sort on a nil-safe key that yields `""` when `desc.Platform == nil`.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,483 @@
...
+
+var errNotPulled = errors.New("not pulled")
+
+type store struct {
+	root string
+}
+
+func openStore(root string) (*store, error) {
+	root = filepath.Clean(root)
+	if err := os.MkdirAll(root, 0o755); err != nil {
```
jserv: 0755 on the store and blob directories, 0644 on the metadata, 0444 on the blobs. On a shared host every local user can read the layers of a private image and the ref names in `index.json`.

0700 for the directories and 0600 for the metadata costs nothing here, and matches what other local image stores do.

jserv (commented):
Checked and found sound, recording them so they do not get re-raised: `--timeout` really does bound the lock wait plus the whole transfer, because the deadline context is built before `withLock` and `remote.Get` threads it into the descriptor; `authn.DefaultKeychain` propagates config-load and helper errors instead of silently going anonymous, so the anonymous path only triggers on a genuinely absent entry; and `v1.Hash.parse` rejects any non-hex body and any algorithm other than sha256, so no reference or digest can walk out of `filepath.Join`.

The store design itself reads well. The pin table as a nested index annotated with `org.opencontainers.image.ref.name` is the right data structure, and it keeps elfuse itself free of OCI knowledge.

cmd/elfuse-oci/store.go
```diff
@@ -0,0 +1,483 @@
...
+	})
+	b, err := json.MarshalIndent(index, "", "  ")
+	if err != nil {
+		return err
+	}
+	b = append(b, '\n')
+	return replaceFile(ctx, s.root, "index.json", b, 0o644)
+}
+
+func (s *store) digestFor(ref string, platform ocispec.Platform) (string, error) {
```
jserv: `digestFor`, `manifestFor`, `notPulledError` and `errNotPulled` have no caller outside `_test.go`, since `pull` is the only command. Roughly 60 lines of production surface that only the tests exercise, so nothing checks the shape against what the consumer will actually want.

Related: `writeShellStub`, `prependPath` and `manifestOf` in helpers_test.go have no caller at all (one occurrence each, their own definition). They look like leftovers from a deleted credential-helper test, which would also explain why nothing covers the Docker config and helper resolution that docs/usage.md documents.

Either land the consumer here, or drop the read half and reintroduce it with the command that needs it, and add the credential test those helpers were written for.

.github/workflows/build.yml
henrybear327: Done.

henrybear327:
> Use shorter and still meaningful paths. e.g., `cmd/oci/`.

Renamed.

docs/oci-images.md
henrybear327: Attempted.

cmd/oci/store.go
henrybear327: As we have agreed in the previous iteration that we would not be handling OCI image lifecycle, this is intentionally left to the `clean` command to deal with (it's in the subsequent PR). WDYT?

jserv:

PR #342, opened 2026-08-29
Spend one host descriptor on a directory fd

xalestar:
## Summary

The host descriptor budget in `src/elfuse-limits.h` is `FD_TABLE_SIZE` guest fds plus `HOST_FD_RESERVE` for elfuse's own plumbing, and that arithmetic assumes one host descriptor per guest fd. A directory took two. `fd_alloc_opened_host` dup'd the descriptor so `fdopendir` could adopt one copy while the fd table kept the other, and the duplicate lived as long as the guest fd rather than as long as a call.

Measured with the host limit at the minimum elfuse itself raises it to, which is what any launch from a lower soft limit gets:

| | before | after | Linux 6.12, `RLIMIT_NOFILE` 1024 |
|---|---|---|---|
| `open(dir, O_DIRECTORY)` until `EMFILE` | 636 | 1021 | 1021 |
| `open` a plain file until `EMFILE` | 1021 | 1021 | 1021 |
| 400 open + `dup` pairs on a directory | 318 pairs | 400 pairs | 400 pairs |

Overrunning the table is the visible half. The other half is that the overflow comes out of the reserve, so the first thing to break is not the open that caused it: with 636 directories held, `fork` failed inside `src/runtime/forkipc.c` with `socketpair` returning `EMFILE`, and the guest saw `ENOMEM` from a fork a real kernel completes. `HOST_NOFILE_MIN` is unchanged — this spends fewer descriptors rather than asking for more.

## Why the duplicate was there, and why it could not just go

`fdopendir` adopts the descriptor it is handed, and only `closedir` gives it back. A slot that also closed its own `host_fd` would close it twice, so two owners with independent lifetimes needed two descriptors. That is a real constraint, not an oversight: an in-flight `getdents64` must keep reading after a sibling closes the guest fd.

`fdclosedir(3)` returns the descriptor without closing it and would remove the need for a second owner outright. It is macOS 26.4, and this tree supports macOS 13, so it is not available.

So the ownership goes the other way instead. The stream owns the descriptor, and everything that needs it to stay valid holds a reference on the stream rather than a descriptor of its own. There are three such holders: the fd table slot, an in-flight `getdents64` (which already took one), and an `fd_lifetime` pin, which now delegates its close to the stream instead of calling `close` itself. The descriptor returns to the host at the stream's last release, so `dir_stream_release` is the single point where a directory fd is given back.

## Ordering, and why the stream is part of the slot's identity

Two steps in `fd_cleanup_entry` are keyed by the descriptor's number rather than by the slot: the pty side tables and the readiness watch. The slot's reference on the stream is therefore released after both, so the number cannot be recycled underneath them.

A slot published as `FD_DIR` before its stream exists is a slot whose descriptor looks plain to a concurrent close, which would close it and leave the opening thread's stream sitting on a number the next open can claim. Both creation paths — `openat` and `dup` — now build the stream first and publish it in the `fd_lock` window that creates the slot, which is what the `fd_alloc_dir` entry points already do for `FD_EPOLL`. The alias allocators hand back the allocation generation so the metadata install that follows can still prove the slot is its own, and `install_fd_alias_metadata_atomic` no longer installs the stream at all.

`fd_alloc_opened_host` now takes ownership of `host_fd` on every path including its failures, because for a directory that transfer is mandatory rather than tidy: once `fdopendir` has succeeded there is no way to hand the descriptor back. Its four callers no longer close on failure.

One host behaviour stops being invisible once the descriptor is shared: Darwin's `fdopendir` sets `FD_CLOEXEC` on the descriptor it adopts. That was harmless on a private duplicate. On the guest's own descriptor it flips a flag elfuse never asked for, so `dir_stream_open` clears it back — the guest's `O_CLOEXEC` lives in `fd_entry_t.linux_flags`, never on the host descriptor.

## Test

`tests/test-dir-fd-budget.c` pins the budget from both directions, how many directories fit and whether dup'ing them fits too, and runs at `host_nofile=elfuse-minimum` where the doubling has nowhere to hide. Against the parent commit it reports 636 of a promised 1021 and 318 of 400 pairs. It is registered in the matrix rather than exempted, because the reference kernel agrees with every number it asserts.

`tests/test-getdents-refcount.c` already covers the window this change makes load-bearing — `close`/`dup2` racing an in-flight `getdents64` — and passes unchanged.

## What this does not fix

A dup'd directory fd reads as an empty stream: `getdents64` on it returns 0 entries, not even `.` and `..`, where Linux lists the directory. This predates the change and is unrelated to descriptor accounting — verified by running the same probe against the parent commit, which behaves identically. The test above therefore checks the surviving dup by using it as an `openat` anchor rather than by reading it, so it pins ownership without depending on that defect.

`/dev/ptmx` is the other guest fd that costs two host descriptors, since elfuse holds a keepalive slave alongside the master. That one is bounded by the host's `kern.tty.ptmx_max` rather than by the descriptor budget, so it is a different problem and is left alone here.

## Checks

Based on `300bd1d`.

`make check`, rc=0: 90 in the manifest suite, 7/7, 82/84 (the two skips need a raw socket and an interactive terminal), 27/27, 4/4, 6/6, guardrail PASS.

`make test-matrix`, rc=0: elfuse-aarch64 271 passed, 0 failed; qemu-aarch64 250 passed, 0 failed. `test-dir-fd-budget` passes in both lanes.

`make lint`, rc=0, with no new clang-tidy warning on a touched line. `make check-format`, rc=0.

---
## Summary by cubic
Spends one host descriptor per directory fd instead of two, so directory-heavy workloads reach the full guest fd table and stop draining the host descriptor reserve that fork and other elfuse plumbing depend on.

- The directory stream now owns the slot's descriptor and closes it at its last release; an in-flight `getdents64` or an `fd_lifetime` pin holds a stream reference instead of a duplicate descriptor, and fork restore adopts the received descriptor the same way.
- The stream is published in the slot's creation window so a concurrent close never sees a directory entry whose descriptor looks plain.
- Darwin's `fdopendir` sets `FD_CLOEXEC` on the descriptor it adopts, so `dir_stream_open` clears it back; the guest's `O_CLOEXEC` still lives only in `fd_entry_t.linux_flags`.
- Adds `tests/test-dir-fd-budget.c`, which pins the one-descriptor budget at the minimum host limit and reports 636 of a promised 1021 directories before the fix.

<sup>Written for commit 87321da93dc15859cc157d29dc1e3f97c28beaf7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/342?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/runtime/fork-state.c
+++ b/src/runtime/fork-state.c
@@ -518,25 +518,19 @@ int fork_ipc_recv_fd_table(int ipc_fd, guest_t *g)
                  * yields EBADF, an honest signal that the fd did not survive
                  * the fork.
                  */
-                int dir_fd = dup(host_fds[i]);
-                if (dir_fd < 0) {
-                    log_error("fork-child: dup failed for DIR gfd %d: %s", gfd,
-                              strerror(errno));
-                    fd_retire_published(gfd, host_fds[i]);
-                    continue;
-                }
-                DIR *dir = fdopendir(dir_fd);
-                if (!dir) {
-                    close(dir_fd);
-                    log_error("fork-child: fdopendir failed for gfd %d", gfd);
-                    fd_retire_published(gfd, host_fds[i]);
-                    continue;
-                }
-                void *ds = dir_stream_create(dir);
+
+                /* The stream adopts the descriptor that arrived over
+                 * SCM_RIGHTS; it does not take one of its own. On failure the
+                 * descriptor is still this side's, which is what lets
+                 * fd_retire_published close it -- and it must, because a slot
+                 * left published as FD_DIR with a NULL dir owns its descriptor
+                 * the ordinary way (see fd_cleanup_entry).
+                 */
+                void *ds = dir_stream_open(host_fds[i]);
                 if (!ds) {
-                    closedir(dir);
-                    log_error("fork-child: dir_stream_create failed for gfd %d",
-                              gfd);
+                    log_error(
+                        "fork-child: dir_stream_open failed for gfd %d: %s",
+                        gfd, strerror(errno));
                     fd_retire_published(gfd, host_fds[i]);
                     continue;
                 }
--- a/src/syscall/fdtable.c
+++ b/src/syscall/fdtable.c
@@ -47,6 +47,14 @@ struct fd_lifetime {
     _Atomic unsigned int refs;
     int host_fd;
     int type;
+
+    /* FD_DIR only: the directory stream that owns host_fd. fdopendir() adopted
+     * the descriptor and only closedir() releases it, so a pin on a directory
+     * cannot close host_fd itself -- it holds a reference on the stream and
+     * lets the stream's last release do it. NULL for every other type, which
+     * closes host_fd directly.
+     */
+    void *dir;
 };
 
 /* RLIMIT_NOFILE tracking. Guest-side soft limit for RLIMIT_NOFILE. fd_alloc
@@ -575,7 +583,8 @@ int fd_alloc_dir_from(int minfd,
                       int host_fd,
                       void (*cleanup)(int),
                       void *dir,
-                      int linux_flags)
+                      int linux_flags,
+                      uint64_t *out_gen)
 {
     fd_host_probe_t probe = fd_probe_host(type, host_fd);
     pthread_mutex_lock(&fd_lock);
@@ -584,6 +593,8 @@ int fd_alloc_dir_from(int minfd,
         fd_table[fd].dir = dir;
         fd_table[fd].linux_flags =
             fd_flags_with_accmode(fd_table[fd].type, linux_flags);
+        if (out_gen)
+            *out_gen = fd_table[fd].generation;
     }
     pthread_mutex_unlock(&fd_lock);
     return fd;
@@ -598,7 +609,8 @@ int fd_alloc_dir_at(int fd,
                     int host_fd,
                     void (*cleanup)(int),
                     void *dir,
-                    int linux_flags)
+                    int linux_flags,
+                    uint64_t *out_gen)
 {
     if (!RANGE_CHECK(fd, 0, FD_TABLE_SIZE))
         return -1;
@@ -616,6 +628,8 @@ int fd_alloc_dir_at(int fd,
     fd_table[fd].dir = dir;
     fd_table[fd].linux_flags =
         fd_flags_with_accmode(fd_table[fd].type, linux_flags);
+    if (out_gen)
+        *out_gen = fd_table[fd].generation;
     pthread_mutex_unlock(&fd_lock);
 
     if (old.type != FD_CLOSED)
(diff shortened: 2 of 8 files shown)


xalestar:
The flags concern on `fd_alloc_dir_*` does not have the window it describes. Both writes happen inside one `fd_lock` hold: `fd_alloc_locked` never touches the mutex, so `fd_init_entry`'s write and the allocator's own write are in the same critical section, and every alias sweep that could race them -- `fd_set_shadow_flags`, `fd_apply_guest_nonblock` -- takes that same lock. The two writes also compute the same value, since the flags handed to the allocator are the alias spec's own. This double write is not new here either: `fd_alloc_alias_dir` has been used this way by the epoll dup path.

The other four findings across these rounds were real and are fixed: the clobbered `errno`, the misattributed `FD_CLOEXEC` failure, the unbounded dirent walk and its missing header bounds check, and the failure message that read an unreadable stream as a missing entry. The `FD_CLOEXEC` one was a genuine regression -- an unconditional clear wiped the flag from a directory the guest opened with `O_CLOEXEC`, measured before and after.

xalestar:
Three findings from these rounds were not acted on, with reasons.

**The flags race on `fd_alloc_dir_*`** has no window: both writes are inside one `fd_lock` hold -- `fd_alloc_locked` never touches the mutex -- and every alias sweep that could race them takes that same lock. The two writes also compute the same value.

**The fork-restore race** has no window either: `fork_ipc_recv_fd_table` runs at `forkipc.c:307`, and the child's first vCPU is created at `:394`, registered at `:464` and started at `:553`, so no guest thread exists to issue a `close`. The whole restore loop depends on this -- `proc_path`, `path_poll_capable`, `seals`, `ofd_id` and the fasync fields are all written after publication too.

**`getdents64` is 61 on aarch64**, which is the target here; 217 is its x86_64 number. Measured under elfuse: `getdents64=61 lseek=62 openat=56 close=57 dup=23 fcntl=25`, which is the same asm-generic table the other numbers in that line come from.

The hardcoded thresholds were worth fixing but not by including `src/elfuse-limits.h`: this binary also runs in the reference lane, where those host-side constants describe nothing, and no guest test reaches into `src/`. They are derived from the guest's own `RLIMIT_NOFILE` instead, which under elfuse is seeded from `FD_TABLE_SIZE` and so tracks it. The dup-pair count went from a written-down 400 to a derived 488, and still separates the two: 318 of 488 against the parent commit.

Everything else raised was real and is fixed: the clobbered `errno`, the misattributed `FD_CLOEXEC` failure, the unbounded dirent walk and its missing header bounds check, the failure message that read an unreadable stream as a missing entry, and the fixture directory left behind on one error path.

jserv:

Issue #12, opened 2026-05-06
`make test-busybox` fails at Macbook m5 air


devarajabc:
On Apple M5 / macOS 26.4.1 (Darwin 25.4.0), every guest crashes on the first                                                                                                                                EL1 shim instruction. 

```
━━━ Busybox 1.37.0 applet smoke tests ━━━

── Core utilities ──
echo           [ OK ]
printf         [ OK ]
cat            [ FAIL ] (pattern 'hello world' not found, rc=128)
  19:33:54 ERROR src/syscall/proc.c:1699: elfuse: guest exception vec=0x100 ESR=0x2000000 FAR=0x0 ELR=0x100000 SPSR=0x3c5

head           [ FAIL ] (pattern 'line1' not found, rc=128)
  19:33:54 ERROR src/syscall/proc.c:1699: elfuse: guest exception vec=0x100 ESR=0x2000000 FAR=0x0 ELR=0x100000 SPSR=0x3c5

tail           [ FAIL ] (pattern 'line5' not found, rc=128)
  19:33:54 ERROR src/syscall/proc.c:1699: elfuse: guest exception vec=0x100 ESR=0x2000000 FAR=0x0 ELR=0x100000 SPSR=0x3c5

wc             [ OK ]
sort           [ FAIL ] (pattern 'apple' not found, rc=128)
  19:33:54 ERROR src/syscall/proc.c:1699: elfuse: guest exception vec=0x100 ESR=0x2000000 FAR=0x0 ELR=0x100000 SPSR=0x3c5

uniq           [ OK ]
cut            [ OK ]
tr             [ FAIL ] (pattern 'HELLO' not found, rc=127)
  /Users/devaraja/Work/elfuse/tests/lib/test-runner.sh: line 169: timeout: command not found
sed            [ FAIL ] (pattern 'HELLO' not found, rc=127)
  /Users/devaraja/Work/elfuse/tests/lib/test-runner.sh: line 169: timeout: command not found
awk            [ OK ]
grep           [ FAIL ] (pattern 'hello' not found, rc=127)
  /Users/devaraja/Work/elfuse/tests/lib/test-runner.sh: line 169: timeout: command not found
true           [ FAIL ] (got 128, expected 0)
  19:33:54 ERROR src/syscall/proc.c:1699: elfuse: guest exception vec=0x100 ESR=0x2000000 FAR=0x0 ELR=0x100000 SPSR=0x3c5

false          [ FAIL ] (got 128, expected 1)
[...]

PR #344, opened 2026-08-31
Clear ARM64 instruction cache on mprotect exec

doanbaotrung:
Flush data cache and invalidate instruction cache using __builtin___clear_cache when sys_mprotect marks memory with LINUX_PROT_EXEC.

This prevents Apple Silicon CPUs from executing stale cached instructions when dynamic JIT compilers like Android ART generate and execute new ARM64 machine code.

---
## Summary by cubic
Fixes `sys_mprotect` clearing the ARM64 instruction cache when marking memory executable, so JIT-generated code (e.g. Android ART) runs correctly on Apple Silicon instead of executing stale cached instructions.

- Flushes cache via `__builtin___clear_cache` on the slow path and the same-protection fast path, so repeated RX `mprotect` after in-place rewrites also invalidates stale instructions.
- Resolves the host range per region: `host_ptr_for_gpa` for high-VA regions, `g->host_base + mprot_off` for primary memory.
- Uses `host_start + length` as the exclusive upper bound so the range doesn't end exactly on a guest-size boundary where `host_ptr_for_gpa` returns `NULL`.
- Adds `test-jit-icache`, which rewrites guest code in place and verifies RX `mprotect` executes the rewritten version on both ARM64 and x86_64, including the fast path.
- Casts `seals` to `unsigned int` in `fd_block_state_of` to silence a warning.

<sup>Written for commit 0c88d3b0f26a8d8d255f6192d92262e64f37916e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/344?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/internal.h
+++ b/src/syscall/internal.h
@@ -952,7 +952,7 @@ static inline fd_block_state_t fd_block_state_of(const fd_entry_t *e)
     return (fd_block_state_t) {
         .type = e->type,
         .generation = e->generation,
-        .seals = e->seals,
+        .seals = (unsigned int) e->seals,
         .can_block = e->can_block,
         .nonblock_owned = e->nonblock_owned,
         .guest_nonblock = (e->linux_flags & LINUX_O_NONBLOCK) != 0,
--- a/src/syscall/mem.c
+++ b/src/syscall/mem.c
@@ -4523,8 +4523,16 @@ int64_t sys_mprotect(guest_t *g, uint64_t addr, uint64_t length, int prot)
             if (mprotect_same_prot_fast_path_safe(prot) &&
                 !g->regions_tracker_stale &&
                 guest_region_range_prot_uniform(g, addr, mprot_end, prot) &&
-                !guest_region_range_has_noreserve(g, addr, mprot_end))
+                !guest_region_range_has_noreserve(g, addr, mprot_end)) {
+                if (prot & LINUX_PROT_EXEC) {
+                    char *host_start = (char *) host_ptr_for_gpa(g, addr);
+                    if (host_start) {
+                        __builtin___clear_cache(host_start,
+                                                host_start + length);
+                    }
+                }
                 return 0;
+            }
 
             /* Do PTE work BEFORE updating the tracker. If page-table
              * maintenance fails, regions[] still reflects the old prot, so a
@@ -4540,6 +4548,12 @@ int64_t sys_mprotect(guest_t *g, uint64_t addr, uint64_t length, int prot)
                     return -LINUX_ENOMEM;
             }
             guest_region_set_prot(g, addr, mprot_end, prot);
+            if (prot & LINUX_PROT_EXEC) {
+                char *host_start = (char *) host_ptr_for_gpa(g, addr);
+                if (host_start) {
+                    __builtin___clear_cache(host_start, host_start + length);
+                }
+            }
             return 0;
         }
         uint64_t mprot_off = addr - g->ipa_base;
@@ -4563,8 +4577,13 @@ int64_t sys_mprotect(guest_t *g, uint64_t addr, uint64_t length, int prot)
                 !g->regions_tracker_stale &&
                 guest_region_range_prot_uniform(g, mprot_off, mprot_end,
                                                 prot) &&
-                !guest_region_range_has_noreserve(g, mprot_off, mprot_end))
+                !guest_region_range_has_noreserve(g, mprot_off, mprot_end)) {
+                if (prot & LINUX_PROT_EXEC) {
+                    char *host_start = (char *) g->host_base + mprot_off;
+                    __builtin___clear_cache(host_start, host_start + length);
+                }
                 return 0;
+            }
 
             if (prot != LINUX_PROT_NONE) {
                 int page_perms = prot_to_perms(prot);
@@ -4578,6 +4597,10 @@ int64_t sys_mprotect(guest_t *g, uint64_t addr, uint64_t length, int prot)
                     return -LINUX_ENOMEM;
             }
             guest_region_set_prot(g, mprot_off, mprot_end, prot);
+            if (prot & LINUX_PROT_EXEC) {
+                char *host_start = (char *) g->host_base + mprot_off;
+                __builtin___clear_cache(host_start, host_start + length);
+            }
         }
     }
     return 0;
--- a/tests/test-jit-icache.c
+++ b/tests/test-jit-icache.c
@@ -1,0 +1,105 @@
+/*
+ * Self-modifying JIT instruction cache invalidation test
+ *
+ * Copyright 2026 elfuse contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Tests: mprotect(PROT_READ | PROT_EXEC) invalidates instruction cache
+ *        when guest JIT code is updated in place.
+ *
+ * Syscalls exercised: mmap(222), mprotect(226)
+ */
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+#include "test-harness.h"
+
+int passes = 0, fails = 0;
+
+typedef int (*jit_fn_t)(void);
+
+static void test_jit_icache_flush(void)
+{
+    TEST("JIT mprotect instruction cache flush");
+
+    uint32_t *buf = mmap(NULL, 4096, PROT_READ | PROT_WRITE,
+                         MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+    if (buf == MAP_FAILED) {
+        FAIL("mmap failed");
+        return;
+    }
+
+    /* Write 1st version: return 1 0x52800020 : mov w0, #1 0xd65f03c0 : ret */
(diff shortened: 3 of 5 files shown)


jserv (changes requested):
Provide an MRE ([Minimal reproducible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example)) demonstrating an mprotect execution issue caused by self-modifying code.

doanbaotrung:
**MRE Added (`tests/test-jit-icache.c`)**: Added a self-modifying JIT unit test that writes initial ARM64 machine code (`mov w0, #1; ret`), executes it under `mprotect(PROT_READ|PROT_EXEC)`, rewrites the buffer in-place to return `42` (`mov w0, #42; ret`), and asserts that the second invocation returns `42` after `mprotect(PROT_READ|PROT_EXEC)` invalidates the instruction cache.

src/syscall/mem.c
```diff
@@ -4523,8 +4523,16 @@ int64_t sys_mprotect(guest_t *g, uint64_t addr, uint64_t length, int prot)
             if (mprotect_same_prot_fast_path_safe(prot) &&
                 !g->regions_tracker_stale &&
                 guest_region_range_prot_uniform(g, addr, mprot_end, prot) &&
-                !guest_region_range_has_noreserve(g, addr, mprot_end))
+                !guest_region_range_has_noreserve(g, addr, mprot_end)) {
+                if (prot & LINUX_PROT_EXEC) {
+                    char *host_start = (char *) host_ptr_for_gpa(g, addr);
```
jserv, line 4528: `host_ptr_for_gpa` takes a GPA, but this branch is reached only when `addr >= g->guest_size`, so `addr` is a high guest VA. High-VA mappings are backed at a separately allocated `backing_gpa_start` (`mem.c:1048-1060`); the VA is not the GPA.

For a normal high-VA mapping the lookup misses and returns NULL, so this flush silently does nothing on the one path the PR is meant to fix. If a VA does collide numerically with some region's GPA range, it flushes unrelated host memory instead. Translate VA to GPA through the region's `gpa_base` first.

src/syscall/mem.c
```diff
@@ -4523,8 +4523,16 @@ int64_t sys_mprotect(guest_t *g, uint64_t addr, uint64_t length, int prot)
...
+                    if (host_start) {
+                        __builtin___clear_cache(host_start,
+                                                host_start + length);
```
jserv, line 4531: Resolving only the start and adding `length` assumes contiguous host backing for the whole range. `guest_is_valid_range` (`src/core/guest.c:985-995`) exists to reject exactly that: "host pointers cannot safely span discontiguous backing."

A range crossing two high-VA mappings or an overflow-segment boundary sweeps past the end of the first host allocation. Walk the intersecting regions and cap each call at its backing end.

src/syscall/mem.c
```diff
@@ -4540,6 +4548,12 @@ int64_t sys_mprotect(guest_t *g, uint64_t addr, uint64_t length, int prot)
                     return -LINUX_ENOMEM;
             }
             guest_region_set_prot(g, addr, mprot_end, prot);
+            if (prot & LINUX_PROT_EXEC) {
+                char *host_start = (char *) host_ptr_for_gpa(g, addr);
+                if (host_start) {
+                    __builtin___clear_cache(host_start, host_start + length);
```
jserv, line 4554: This is a guest-reachable, unguarded host touch of guest memory. `src/debug/gdbstub.c:359-366` documents the one existing `sys_icache_invalidate` outside `HOST_SIGBUS_GUARD` and says explicitly not to treat it as precedent, because a truncated `MAP_SHARED` overlay faults it the same way a `memcpy` would.

`mprotect` previously never touched the mapped bytes. Route through the guarded path, or drop the flush. Same applies to the three sibling call sites.

src/syscall/mem.c
```diff
@@ -4563,8 +4577,13 @@ int64_t sys_mprotect(guest_t *g, uint64_t addr, uint64_t length, int prot)
                 !g->regions_tracker_stale &&
                 guest_region_range_prot_uniform(g, mprot_off, mprot_end,
                                                 prot) &&
-                !guest_region_range_has_noreserve(g, mprot_off, mprot_end))
+                !guest_region_range_has_noreserve(g, mprot_off, mprot_end)) {
+                if (prot & LINUX_PROT_EXEC) {
+                    char *host_start = (char *) g->host_base + mprot_off;
```
jserv, line 4582: The same-protection fast path exists to return without doing work, and `mprotect_same_prot_fast_path_safe` admits `PROT_EXEC`, so this is the common repeat-`mprotect` case.

A 64 MiB RX region now costs roughly a million cache lines of DC plus IC work per call, for a request where neither permissions nor bytes changed. Flush only on an actual transition, if at all.

src/syscall/internal.h
```diff
@@ -952,7 +952,7 @@ static inline fd_block_state_t fd_block_state_of(const fd_entry_t *e)
     return (fd_block_state_t) {
         .type = e->type,
         .generation = e->generation,
-        .seals = e->seals,
+        .seals = (unsigned int) e->seals,
```
jserv, line 955: This only silences a signedness warning. `fd_entry_t.seals` is `int` (`linux-wire.h:591`) while this field is `unsigned` (`internal.h:481`). Seal bits are non-negative flags, so make the two declarations agree instead of casting at the copy site.

Also worth splitting out: it has nothing to do with I-cache maintenance.

tests/test-jit-icache.c
```diff
@@ -0,0 +1,105 @@
...
+    /* Re-issue mprotect RX while region is ALREADY tracked as RX to exercise
+     * same-protection fast path icache invalidation.
+     */
+    if (mprotect(buf, 4096, PROT_READ | PROT_EXEC) != 0) {
+        FAIL("same-protection mprotect RX failed");
+        munmap(buf, 4096);
+        return;
+    }
+
+    int res3 = fn();
```
jserv, line 89: Nothing is written between `res2` and `res3`, so `res3 == 42` holds whether or not the fast path flushes. This check passes with all four `__builtin___clear_cache` calls deleted.

The same is largely true of the `res2` check: a line that was never retained, or was already evicted, passes by accident. This is not a deterministic stale-I-cache repro.

jserv (commented):
The premise needs to be settled before the mechanics. `src/core/shim.S:1018-1045` already traps guest `DC CVAU` / `IC IVAU` under `HCR_EL2.TPU` and runs `IC IALLU`, and `SCTLR_EL1.UCI` is set at `src/core/bootstrap.c:765`, so a guest JIT that calls `__builtin___clear_cache` should already get a coherent I-cache. Real Linux does no I-cache maintenance in `mprotect`, and the new `QEMU_SKIP` note in this PR concedes exactly that.

If a guest is genuinely reading stale instructions, the evidence to attach is a minimal guest that calls `__builtin___clear_cache` plus a trace showing the trap does not reach HVC #12, or that `IC IALLU` does not make the next fetch coherent. That would point the fix at the trap path, not at `mprotect`.

If instead the intent is that elfuse flushes on the guest's behalf, the contract is incomplete: `bootstrap.c:493` and `exec.c:2047` flush executable segments, but `mmap(PROT_EXEC)`, executable `mremap`, and `process_vm_writev` (`src/syscall/io.c:2466`) do not. Covering only `mprotect` leaves a guest able to hit the same staleness by other routes.

The inline notes below are independent of how that question lands: the address math is wrong regardless.

tests/test-matrix.sh
```diff
@@ -341,6 +342,10 @@ QEMU_SKIP="
 #   elfuse's own vCPU/TLBI shootout timing; the qemu reference lane's nested
 #   virtualization has different scheduling latency and trips the same
 #   thresholds for real, unrelated races.
+# test-jit-icache: real Linux mprotect does not perform user-space instruction
```
jserv, line 345: The skip note is accurate, and it is also the argument against the change: real Linux requires the JIT to do its own maintenance.

Separately, the PR description says the test verifies behavior "on both ARM64 and x86_64", but the `elfuse-x86_64` lane returns after `run_rosetta_x86_64_suites` (`test-matrix.sh:1275`) and never reaches `run_unit_tests`. The test hardcodes AArch64 opcodes with no arch guard; it is AArch64-only.

doanbaotrung:
Thank you @jserv for the comprehensive architectural review and clarification.

You are completely right — since `src/core/shim.S:1018-1045` traps guest `DC CVAU` / `IC IVAU` under `HCR_EL2.TPU` and `SCTLR_EL1.UCI` is set at `bootstrap.c:765`, any guest JIT compiler that calls `__builtin___clear_cache()` in user space already receives host instruction cache coherence through the EL2 trap handler.

Because real Linux `sys_mprotect` does not perform instruction cache maintenance for user space, attempting to flush inside `sys_mprotect` deviates from Linux semantics and introduces overhead.

I am closing this PR based on these findings. Thank you for taking the time to explain the trap design!

PR #348, opened 2026-09-01
Serve the non-blocking futex shapes from EL1

### LINKED ISSUE #315: Serve the non-blocking futex shapes from EL1
## Summary

Issue #292 established that `futex` is 56.4% of the guest syscalls in the
sampled corpus. What that ranking cannot say is how many of those calls
actually block. A `FUTEX_WAIT` that returns `EAGAIN` because the word already
moved, and a `FUTEX_WAKE` that finds no waiter, both cost a full HVC round trip
today and neither needs anything the host knows.

I instrumented `sys_futex` to find out, and would like direction before writing
any EL1 assembly.

## What the instrumentation shows

A host-side matrix counting every futex call by (command, outcome), gated on an
env var and dumped at exit. Across five runs of `tests/bench-futex-pingpong`
(70k - 76k futex calls per run):

| shape | share of all futex calls |
| --- | --- |
| untimed `FUTEX_WAIT` returning `EAGAIN` | 40.5% - 45.9% |
| `FUTEX_WAKE` returning 0 | 10.4% - 24.8% |
| combined | ~60% |

The two trade off against one another across runs, which is what a ping-pong
should do. **This is a microbenchmark, not a workload** - I am not claiming the
ratio holds for a JVM. If there is a reproducible recipe for the corpus behind
#292, I am happy to run it and report real numbers before anything else is
decided.

This is not `ELFUSE_STARTUP_TRACE=syscalls` in a different shape.
`src/debug/syscall-hist.c` freezes at the first `execve` and is keyed by
syscall number alone, so it answers which syscalls dominate linker bring-up.
This asks what shape the futex traffic of a steady-state run has.

## Candidate A: answer a value-compare miss at EL1

`futex_wait()` (`src/runtime/futex.c:668`) takes the bucket lock, loads the
guest word, and returns `-EAGAIN` when it does not match.

`futex_os_sync_wait()` (`:501`) - the path plain `FUTEX_WAIT` actually takes on
macOS 14.4 and later - already makes that same check with no bucket lock at
all: a plain `futex_word_load` and an early `-EAGAIN`.

So the tree already treats a mismatch as safe to answer lock-free, and it is.
The bucket lock orders the enqueue against a concurrent wake, but the futex
word is guest memory that another guest thread stores to without holding it. A
mismatch means a waker already moved the word, so no wakeup can be lost and the
caller's userspace loop re-checks. Moving that check to EL1 is not a new
invariant, only an earlier place to make it.

The mechanics already exist in `src/core/shim.S`:

- the SVC fast-path dispatch at `:441-452`
- the `at s1e0w` / `PAR_EL1` translation probe at `:558` (a futex read needs
  only `s1e0r`)
- `handle_el1_data_abort_recover` at `:970`, which turns a fault inside a named
  code range into `-EFAULT` instead of halting the VM

so the shape is `cmp x10, #98` plus a body modeled on `urandom_read_fast`.

Bails to HVC on: any command outside `FUTEX_WAIT` / `FUTEX_WAIT_BITSET`, a zero
bitset, unaligned `uaddr`, a set attention bit, and a failed AT probe. The
`EINVAL` cases (zero bitset, unaligned address) and the `EFAULT` case (probe
failure) stay on the host so their ordering against the rest is untouched; the
attention bit is the gate every existing fast path already honors.

**The one real problem.** Both wait paths build the deadline *before* checking
the word, so an invalid `timespec` combined with a mismatched word currently
returns `EINVAL`/`EFAULT`, not `EAGAIN`. Linux orders it the same way, since
`get_timespec64` runs at syscall entry. Preserving that from EL1 means either:

1. bail whenever the timeout argument is non-zero - simple and obviously
   correct, but it gives up the timed waits, which is where `parkNanos`,
   `sem_timedwait` and `pthread_cond_timedwait` live; or
2. validate the guest `timespec` at EL1 too, so the ordering is preserved and
   the timed waits stay eligible - more EL1 code, and a second guest structure
   to probe.

**Q1: which of those two do you want?** My instrumentation splits timed from
untimed rows precisely so this can be decided on numbers rather than taste.

## Candidate B: answer a wake with no waiter at EL1

Larger, and it needs host bookkeeping that does not exist yet.

Waiters live in two places. `futex_wake()` (`:879`) walks the hash bucket, then
`futex_wake_topup_osync()` (`:656`) drains the Darwin queue. The second has no
count at all - `futex_os_sync_wake_n()` (`:464`) simply calls
`os_sync_wake_by_address_any` and watches for a negative return. So EL1 has
nothing to consult today.

A host-published occupancy bitmap in `shim_data`, indexed by a hash of `uaddr`,
would give it one. `SHIM_URANDOM_OFF_BITMAP` (`src/core/shim-globals.h:137`) is
the existing precedent for exactly this shape of host-maintained summary.

Two things make it harder than the urandom bitmap:

- **Ordering is load-bearing.** A false positive costs one wasted HVC. A false
  negative is a lost wakeup and a hang. The waiter must publish its bit before
  re-reading the word, and the waker must `dmb ish` after storing the word and
  before reading the bitmap - the usual store-buffer pattern.
- **Both queues must be covered.** `futex_os_sync_wait()` bypasses the bucket
  entirely, so it needs to maintain the bitmap too, on a path that is currently
  free of that bookkeeping.

`futex_hash()` (`:174`) spreads over 64 buckets, which is probably too coarse
here - a bitmap over it would read as occupied nearly always under a thread
pool. The right width depends on how many distinct futex addresses are live at
once, which the instrumentation does not record today. I can add that if the
direction is worth pursuing.

**Q2: is a host-published waiter bitmap in `shim_data` an acceptable direction
at all, or does the added invariant on the wait paths outweigh the round trips
it saves?**

## What is not proposed

Blocking at EL1. The shim has no scheduler, so a wait that must sleep has to
leave. Only the calls that were never going to block are in scope. `LOCK_PI`,
`REQUEUE`, `CMP_REQUEUE` and `WAKE_OP` stay on the host path.

## Sequencing, and #307

Candidate A touches the `shim.S` dispatch and Candidate B touches the
`shim_data` layout - both of which #307 is currently rewriting. I would rather
not create that conflict.

Proposed order:

1. the measurement patch alone: host-side only, no `shim.S`, no `shim_data`,
   zero conflict surface with #307. It is also independently useful - "which
   futex shapes does this run have" is a question worth being able to ask
   whether or not a fast path ever lands.
2. Candidate A, after #307 settles.
3. Candidate B, only if the answer to Q2 is yes.

**Q3: is step 1 welcome as its own PR?** It is written and passing locally
(`make check` minus the `test-hello` lane, which needs a toolchain I cannot
install on this machine). It is +866/-32 across 11 files, but 766 of those
lines are three new files - `src/runtime/futex-stats.c`, its header, and a host
unit test; the change to existing code is small. I can open it whenever.

jserv:
Serves the two futex shapes that cannot block from the EL1 shim and separates an unrelated vCPU watchdog rewrite that
had been sitting in the same change.

## Commits

1. **Arm one repeating vCPU watchdog, not two a call.** The run loop armed
   and disarmed `alarm()` around every `hv_vcpu_run`, and on macOS that is
   `setitimer(ITIMER_REAL)`, so it cost two real syscalls per guest syscall:
   693 ns against a 1226 ns bare HVF round trip. One repeating timer plus a
   seqlock progress word replaces it. Unrelated to futex, hence its own
   commit.
2. **Serve the non-blocking futex shapes from EL1.** A `FUTEX_WAIT` whose
   word already moved answers EAGAIN at EL1 (51 ns against a 2000 ns HVC
   round trip), and a `FUTEX_WAKE` with nobody parked answers 0 from a
   per-bucket census the wait paths publish into `shim_data` (1927 ns to
   77 ns). Also widens the bucket table to 1024 with a new hash, and fixes
   undefined behavior in the `FUTEX_WAKE_OP` operand decode, which was
   undefined for half its input range and included the shift operand that
   was CVE-2018-6927.
3. **Refuse to requeue a PI waiter, as Linux does.**
4. **Give the analyzer Darwin's shapes, not Linux's.** Five declarations
   where Frama-C's modeled libc uses the Linux layout and macOS does not,
   which is what kept the socket and abstract-socket paths outside the
   analyzer.
5. **Stop rebuilding guest state at EL1.**
6. **Spin at EL1, and serve a timed futex wait too.** Measured against the
   branch point: wake-wait-handoff 1104 ns to 103, wait-timed 1503 ns to 52.
7. **Prove the futex deadline arithmetic.** `verify-futexdeadline` discharges
   128 of 128 obligations with `-wp-rte`.

Close #315 

---
## Summary by cubic
Serves non-blocking, timed, and wake futex shapes from the EL1 shim instead of round-tripping to the host, replaces the per-call `alarm()` watchdog with a repeating timer, fixes signal and ptrace handling inside the shim, and lets the Frama-C analyzer parse 5 more sources.

**Futex**

- A `FUTEX_WAIT` with a moved word answers `EAGAIN` at EL1 (51 ns vs 2000 ns); the shim also spins on a matching word and serves timed waits, cutting a contended wake-wait-handoff from 1104 ns to 103 ns.
- The EL1 wake path (77 ns vs 1927 ns for a wake with nobody parked) serves both wake shapes; it was briefly routed to the host when a contract assert flagged the bucket census, but the assert read a transient population and never survived a re-read, so the census holds and `test-shim-futex-stats` asserts the counters move.
- Fixes undefined behavior in the `FUTEX_WAKE_OP` operand decode and masks an out-of-range shift operand to its low five bits, matching Linux; the shift fix is Linux commit 30d6e0a4190d, not CVE-2018-6927, which names a negative-count `futex_requeue` overflow still unmeasured in TODO.md.
- Refuses to requeue a PI waiter, matching Linux.
- Open defect: a signal during SVC return can re-execute the SVC as `ENOSYS` under load.

**Watchdog, signals, and tooling**

- The run loop arms one repeating watchdog and publishes progress via a seqlock; this removes the two real syscalls per guest syscall that `alarm()` cost.
- Signal delivery and ptrace stops taken inside the shim now establish EL0 before building a frame; `PTRACE_INTERRUPT` uses a new attention bit.
- Adds Darwin-shaped stubs for `sockaddr_un` (one-byte family), `sockaddr_storage`, `ipc_perm`, `msqid_ds`, `stat`, and `pthread_setname_np`, plus `check-stub-shadow` to keep the rename trick sound; the gate honors the `FRAMAC` override, including an empty `FRAMAC=` as a clean skip.
- Proves the futex deadline arithmetic (`verify-futexdeadline` discharges 128 of 128 obligations); `futex_make_deadline` stays unproved since the analyzer has no body for `guest_read_small`.
- Five checks could not report what they guard: `test-ptrace-interrupt`'s stop-PC ceiling never separated guest from shim PCs (now 16 GiB), `test-matrix.sh` counted that test without running it, `bench-futex-contend` ignored spawn results, two tests read errno after `pthread_create`, which never sets it, `test-sleep-long` accepted an unparsable argument, and `test-shim-futex-stats` aborted before its counter checks.

<sup>Written for commit 96be098de7f05145b23c59814cd4a8da830fcc5e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/348?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jserv:
The wake fast path is back on, and the reason it was disabled was my mistake.

## The lost wakeup was not real

I disabled both wake shapes because `futex_wake`'s contract assert aborted on
every run of `test-signal-in-shim`, reporting a published bucket census of 0
against a live waiter. I reported that as a lost wakeup, this PR recorded it as
one, and the P0 in the review inherited it from me. It was the assert.

Its three readings are not one instant. The chain is stable, since the assert
holds the bucket lock, but `os_sync_waiters` is not: the Darwin address-wait
path takes no lock, which is the whole reason it is counted separately. Either
read order is wrong in one direction. Read the census last and a waiter that
finishes in between drops it under a population already counted. Read it first
and a waiter that arrives in between raises the population above a census
already read. Neither is an understatement, and no ordering fixes it, because
the two populations cannot be sampled together without stopping the world.

A real understatement is a charge that is missing rather than one in flight, so
it does not clear when looked at again. The assert re-reads now and believes a
violation only when it survives. Measured before the change: `pub=0` against
`parked=1` on every run, surviving 0 of 5 re-reads every time, across
`test-signal-in-shim`, `test-shim-futex-fast`, `test-futex-wake-nowaiter`,
`test-futex-requeue-account` and `bench-futex-contend`.

What settled it was tracing, not reading. Tagging every publish site by line and
dumping them for the offending bucket showed only the os_sync wrapper's own
pair, whose nesting is already correct, which is what left the assert as the
remaining suspect. I had read that code twice and concluded the opposite.

## What that changes

Both wake shapes are served from EL1 again, so #315 is complete rather than
half done: a `FUTEX_WAIT` whose word already moved answers EAGAIN at EL1, and a
`FUTEX_WAKE` with nobody parked answers 0 there.
`test-shim-futex-stats` asserts those counters move again rather than that they
stay at zero, and `make check-contracts` passes end to end for the first time,
which it never did before this branch.

## Validation

`make verify` 21 targets, `make verify-mutants` 114 of 114, `make check` green,
`make check-contracts` green, guardrail PASS, and the qemu lane at 256 passed 0
failed, which is the reference kernel agreeing.

## Still open, unchanged by this

The X8=2 SVC re-execution race under load is untouched and remains recorded in
TODO.md. So are the ptrace attention-bit races, the `lock_pi` publish on the
uncontended path, `EL0_PC_CEILING` on 36-bit IPA, and the two test-diagnostic
P3s from the earlier review.

Issue #225, opened 2026-07-18
Integrate scoped gVisor syscall conformance


henrybear327:
Similar to https://github.com/sysprog21/elfuse/issues/171

Integrate a scoped [gVisor](https://github.com/google/gvisor) syscall conformance suite into elfuse.

henrybear327:
Initial prototyping discovered several issues blocking the integration. I will submit PRs fixing the blocker, and the initial PoC implementation in a bit.

henrybear327:
Ground work https://github.com/sysprog21/elfuse/pull/339

Issue #224, opened 2026-07-18
elfuse real-workload syscall frequency ranking (Top 50)


Max042004:
Environment: origin/main (435a546) + PR #191 (OCI tooling, 9 commits) + instrumentation
commit 3312482 + unpack setuid-EPERM fallback fix (not separately committed).

## Methodology

- Added an `ELFUSE_SYSCALL_STATS_DIR` whole-lifetime stats mode (built on the
  existing `src/debug/syscall-hist.c`): recording does not freeze at the first
  execve, fork children (fresh posix_spawn images) record independently, and
  each guest process writes `<host-pid>.stats` (`name count total_ns`) at exit
  (atexit / elfuse_launch). Summing the per-process files yields exact
  process-tree totals.
- The watchdog was raised to `--timeout 3600` via an `ELFUSE_BIN` wrapper so
  long compile bursts are not killed (the 10 s default per-iteration alarm;
  note `--timeout 0` is not "no timeout" — 0 selects worker-thread semantics).

## Workloads (five categories, 116 guest processes)

| workload | image | content | syscalls | processes |
|---|---|---|---:|---:|
| jvm | eclipse-temurin:21 | javac compile of a multi-class program + java run (collections, file I/O, SHA-256, 8 threads, subprocess) | 3,028,653 | 8 |
| python | python:3.12-slim | JSON/regex churn, sqlite3 (20k inserts + queries), 400 small-file I/O, os.walk over the stdlib, 10 subprocesses, 4 threads | 1,364,400 | 11 |
| go | golang:1.23-alpine | `go version` + `gofmt` over stdlib trees (net/runtime/crypto) + `gofmt -d fmt` | 46,037 | 6 |
| node | node:22-alpine | module loading, fs/crypto/zlib/JSON, then an HTTP server taking 302 host-side curl requests before /quit | 18,740 | 1 |
| c-compile | gcc:14 | 30-file C project via `make -j1` + sqlite amalgamation (8 MB TU) `gcc -O1` compile+link + a sqlite3 query | 82,606 | 90 |

Total: **4,540,436 syscalls, 89 distinct**.

## Top 50 ranking (five workloads combined, descending)

| # | syscall | count | share% | cum% | jvm | python | go | node | c-compile |
|--:|---|--:|--:|--:|--:|--:|--:|--:|--:|
| 1 | futex | 2,559,392 | 56.369 | 56.37 | 1,517,216 | 1,040,892 | 166 | 1,118 | 0 |
[...]

Issue #171, opened 2026-07-08
Integrate scoped Linux Test Project syscall conformance


jserv:
## Goal

Integrate a scoped [Linux Test Project](https://github.com/linux-test-project/ltp) (LTP) syscall conformance suite into elfuse.

Use a pinned shallow LTP checkout, apply local patches if needed, and run a curated syscall subset through the existing `make check` flow.

Rules:
- Do not use git submodules.
- Do not copy LTP files into elfuse.
- Retrieve LTP with `git clone --depth=1`.
- Keep LTP under `build/ltp-src`.
- Treat LTP as a vendored test dependency and syscall semantics oracle.
- Keep elfuse-specific changes as local patches.
- Keep the default `make check` slice fast and deterministic.

### 1. Fetch pinned LTP with shallow clone

Add a reproducible fetch step for LTP.

Implementation:
- Add `tests/fetch-ltp.sh`.
- Fetch LTP into:
  ```text
  build/ltp-src
  ```
- Use shallow clone:
  ```sh
  git clone --depth=1 --branch <tag-or-branch> \
      https://github.com/linux-test-project/ltp \
      build/ltp-src
  ```
- Resolve and record the exact commit:
  ```sh
  git -C build/ltp-src rev-parse HEAD > build/ltp-src/.ltp-ref
  ```
- If `build/ltp-src/.ltp-ref` already matches the requested SHA, reuse the checkout.
- If it does not match, delete `build/ltp-src` and fetch again.
- Expose override variables:
  ```sh
  LTP_REF=<tag-or-branch-or-sha>
  LTP_URL=https://github.com/linux-test-project/ltp
  LTP_DIR=build/ltp-src
  ```

Considerations:
- No git submodule.
- No generated LTP build output should be committed.
- A shallow clone by SHA may need a fetch fallback because `git clone --depth=1 --branch <sha>` is not portable for raw SHAs. Accept tag/branch by default, then verify resolved SHA.
- Network access should happen only in explicit fetch/build targets, not every `make check`.
- The script should be idempotent.

Validation:
```sh
rm -rf build/ltp-src
LTP_REF=<tag-or-branch> tests/fetch-ltp.sh
test -s build/ltp-src/.ltp-ref
git -C build/ltp-src rev-parse --is-shallow-repository
```

Expected:
- `.ltp-ref` exists.
- checkout is shallow.
- rerunning the script reuses the checkout when the ref matches.

---

### 2. Apply elfuse-specific LTP patches

[...]

PR #350, opened 2026-09-01
Prove the futex words behind Linux's refusals

alanhc:
Follows #348, which added the published waiter census the shim's wake fast
path reads, and #315.

## The bug

Linux answers `EINVAL` when a requeue without `FUTEX_CMP_REQUEUE_PI` meets a
waiter holding an `rt_waiter` or a `pi_state`. `futex_requeue()` in
`kernel/futex/requeue.c` runs that test from inside its walk, **before** it
decides whether the waiter in hand is woken or moved, and without consulting
either address.

elfuse narrowed it three ways:

| | Linux | elfuse |
| --- | --- | --- |
| address condition | none | required `uaddr != uaddr2` |
| PI waiter inside `nr_wake` | refused | skipped, then woken |
| `nr_requeue == 0` | still walks and checks | did not look at all |

Each is reachable from unprivileged guest code with a thread parked in
`FUTEX_LOCK_PI(X)`: `futex(X, FUTEX_REQUEUE, 0, 1, X)`,
`futex(X, FUTEX_REQUEUE, 1, 1, Y)`, and `futex(X, FUTEX_REQUEUE, 1, 0, Y)`.

## Why the first one is more than an errno

That rejection is also what keeps the census charge move away from a PI
waiter, whose bucket charge `futex_lock_pi`'s wrapper owns for the whole call.
Letting one through runs the move, which credits the destination but debits
`w->pub_bucket` only under `pub_follows`. The bucket is then charged for a
waiter that has gone, `futex_wake_fast` reads it as occupied and declines from
then on, and repeating the call wraps the `uint32`, at which point a zero
reading is a wake answered 0 at EL1 while a waiter is parked.

## The fix

Walk every waiter the call could act on, bounded by
`wake_count + requeue_count` the way Linux's own
`task_count - nr_wake >= nr_requeue` break bounds it. The sum is taken in 64
bits because both halves are guest-supplied and wake-all passes `INT_MAX`.

Both halves of the charge move go under `pub_follows` as well. It is
unreachable for a PI waiter now, but a credit sitting outside the branch that
carries its debit is what let the two come apart, and pairing them keeps a
later narrowing of the guard from reintroducing it.

## The test

`tests/test-futex-requeue-pi.c` parks a PI waiter in `FUTEX_LOCK_PI` and puts
one case to each narrowing, retrying the first while the call reports zero
because that is its answer before the waiter parks. Every assertion is an
errno, so the matrix can put the same source to a reference kernel; registered
in `run_unit_tests` and only there, since the manifest's scope note keeps a
cross-checkable test out of that file.

The census gets no assertion. An over-count is invisible from the guest (the
host answers every wake correctly either way), and the restored rejection
makes it unreachable.

## Scope

Only the first narrowing was in the report this PR started from. The other two
were found while checking that one against `futex_requeue()`, and they are the
same `if`, three lines apart, answered by the same paragraph of kernel source.
Fixing one and leaving two would have made the claim in this PR's title untrue.

## Verification

`tests/driver.sh` passes 97 of 98, and `make check-format`, `check-ascii`,
`check-syscall-coverage`, `check-eintr-contract`, `check-lock-order`,
`check-atomics` and `check-skill-refs` are clean. The new test fails before the
fix and passes after it. Every existing futex lane still passes, including
`test-futex-pi`, `test-futex-requeue-account`, `test-futex-waitv`,
`test-shim-futex-fast` and `test-shim-futex-toctou`.

`test-hello` and the `test-matrix` lanes want `aarch64-none-elf-as`, whose
Homebrew cask needs a sudo installer this machine cannot run, so those are
unbuilt rather than passing. For the same reason the `EXPECTED_BASELINES`
floors are untouched: a raised floor would be a number nobody observed, and the
comment above them already says a floor too low costs nothing.

Across two full driver runs `test-clone3` failed once, on `stack reuse mmap
returned -1`, and passed on the rerun and on three isolated runs. It is in
`QEMU_SKIP` already, and its failure mode is address-space placement, disjoint
from anything here.


jserv (commented):
The three narrowings are real and the 64-bit bound is right, including the wake=0/requeue=0 case, which matches Linux breaking at the top of the loop before it checks anything. Two things on the source change and three on the test, inline.

Checked and not raised: pub_follows as the is-PI proxy is already answered by the note at futex.c:1900; the unlock order on the early return is pre-existing and identical to the do_cmp path; there is no scan-cost regression, since musl's broadcast passes requeue_count = INT_MAX and the old gated scan already walked the whole chain; and the manifest placement matches the scope note in tests/manifest.txt. check-matrix-lists, check-commentflow, check-newline and check-security pass here.

src/runtime/futex.c
```diff
@@ -1440,33 +1440,48 @@ static int64_t futex_requeue(guest_t *g,
...
-
-        for (futex_waiter_t *w = b_src->head; w; w = w->next) {
-            if (w->uaddr != uaddr)
-                continue;
-            if (skips != 0) {
-                skips--;
-                continue;
-            }
+    uint64_t checked = (uint64_t) wake_count + requeue_count;
+    for (futex_waiter_t *w = b_src->head; w && checked != 0; w = w->next) {
```
jserv, line 1507: The pre-scan makes the EINVAL all-or-nothing, which requeue.c does not. Linux marks each plain waiter with `this->wake(&wake_q, this)` inside the same loop that later hits the PI waiter and breaks, and `wake_up_q(&wake_q)` runs at `out_unlock` regardless of `ret`. So `futex(X, FUTEX_REQUEUE, 1, 1, Y)` with a plain waiter queued ahead of the PI waiter wakes that waiter and still returns EINVAL there, and wakes nobody here.

The comment above says the scope is what Linux checks, but not that the effect differs. Either fold the check into the existing walk, or state the atomicity as a deliberate departure. The new test cannot tell the two apart: it queues exactly one waiter.

src/runtime/futex.c
```diff
@@ -1440,33 +1440,48 @@ static int64_t futex_requeue(guest_t *g,
...
-        uint32_t remaining = requeue_count;
-
-        for (futex_waiter_t *w = b_src->head; w; w = w->next) {
-            if (w->uaddr != uaddr)
-                continue;
-            if (skips != 0) {
-                skips--;
-                continue;
-            }
+    uint64_t checked = (uint64_t) wake_count + requeue_count;
```
jserv: The sibling guard from the same function is still missing: `futex_requeue()` opens with `if (nr_wake < 0 || nr_requeue < 0) return -EINVAL;`.

`sc_futex` passes val as `(uint32_t) x2` and val2 as `(uint32_t) timeout_gva`, so `futex(X, FUTEX_REQUEUE, -1, -1, Y)` reads both as 0xffffffff and succeeds here where Linux refuses. Same function, and the same guest-supplied pair this new 64-bit sum is written to survive, so it looks like it belongs in this fix rather than a follow-up.

tests/test-futex-requeue-pi.c
```diff
@@ -0,0 +1,205 @@
...
+ */
+static int pi_lock;      /* the PI futex the waiter parks on */
+static int holder_ready; /* set once the holder owns pi_lock */
+static int release;      /* set to tell the holder to unlock */
+static int waiter_done;  /* set once the waiter has taken and released it */
+
+static int holder_stack[16384] __attribute__((aligned(16)));
+static int waiter_stack[16384] __attribute__((aligned(16)));
+
+static long futex_lock_pi(int *addr)
```
jserv, line 59: These helpers drop `FUTEX_PRIVATE_FLAG`, unlike the two neighbours: `test-futex-pi.c` sets it on `LOCK_PI` and `UNLOCK_PI`, and `test-futex-requeue-account.c` sets it on every op including `CMP_REQUEUE`.

elfuse masks the bit off in `FUTEX_CMD_MASK`, so it makes no difference here. The header says the file is meant to go to a reference kernel, and there private and shared hash to different keys.

tests/test-futex-requeue-pi.c
```diff
@@ -0,0 +1,205 @@
...
+    set_and_wake(&release);
+    wait_until_set(&waiter_done);
+
+    /* Both threads are done with pi_lock, so nobody is parked on it and a wake
+     * has nothing to report. A rejected requeue must not have left the waiter
+     * somewhere a later wake still finds it.
+     */
+    TEST("wake the drained address");
+    for (int i = 0; i < 8; i++) {
+        if (raw_futex_wake(&pi_lock, 1) != 0) {
```
jserv, line 252: Direct consequence of the flag above: `raw_futex_wake` ORs `FUTEX_PRIVATE_FLAG`, while `pi_lock` was parked on and requeued without it. On a real kernel this wakes a key nothing ever waited on, so the check passes for free. It only asserts anything under elfuse.

tests/test-futex-requeue-pi.c
```diff
@@ -0,0 +1,205 @@
...
+    while (__atomic_load_n(addr, __ATOMIC_ACQUIRE) == 0)
+        raw_futex_wait(addr, 0);
+}
+
+/* Takes pi_lock, reports it, and holds it until told to let go. The waiter
+ * cannot park until someone else owns the word.
+ */
+static void holder_fn(void)
+{
+    if (futex_lock_pi(&pi_lock) == 0) {
```
jserv, line 120: If `futex_lock_pi` fails, the holder exits without setting `holder_ready` and main blocks at line 158. The matrix's 30s wrapper turns that into a reported failure rather than a hang, but the line reads `timeout` instead of naming the lock that did not come up. Setting a failure word before `raw_exit` and checking it in main would keep the diagnosis in the test.

tests/test-matrix.sh
```diff
@@ -892,6 +892,8 @@ run_unit_tests()
         "$bindir/test-futex-wake-nowaiter"
     test_rc "$runner" "test-futex-requeue-account" 0 \
         "$bindir/test-futex-requeue-account"
+    test_rc "$runner" "test-futex-requeue-pi" 0 \
```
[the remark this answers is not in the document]
alanhc: Not acting on this: both halves of the premise are wrong, and the change it asks for is the one the manifest header forbids.

**Binaries are not built from the manifest.** They come from a glob: `mk/config.mk:54` sets `TEST_C_SRCS := $(wildcard tests/*.c)`, `:55` maps it to `TEST_C_BINS`, `:56` folds that into `TEST_DEPS`, and `mk/tests.mk:1470` makes the matrix depend on it (`test-matrix: $(ELFUSE_BIN) $(TEST_DEPS)`). Dropping a `.c` into `tests/` is what builds it. `driver.sh` reads `manifest.txt` to decide what to *run* on the `make check` lane, not what to build.

**A cross-checkable test belongs only in the matrix.** From the scope note at the top of `manifest.txt`: "Every other aarch64 test -- anything that is meaningful to cross-check against a real Linux kernel -- lives exclusively in tests/test-matrix.sh's run_unit_tests... Don't add a portable/cross-checkable test here; add it to test-matrix.sh instead." Every assertion in this test is an errno a reference kernel can adjudicate, so it is squarely in that category. The sibling `test-fchmodat-empty-path` and `test-fchownat-empty-path` are registered in the matrix and absent from the manifest for the same reason.

**The invariant runs the other way.** `.ci/check-matrix-lists.sh` fails on "a tests/manifest.txt binary that no test_* call registers at all": manifest is a subset of matrix, not the reverse. It passes on this branch.

Worth noting that an earlier run of this same review took the opposite position: on commit `83475a4` it flagged the manifest entry as violating that scope note, the entry was removed, and this comment now asks for it back.

tests/test-futex-requeue-pi.c
```diff
@@ -0,0 +1,185 @@
...
+
+#include "test-harness.h"
+#include "raw-syscall.h"
+
+int passes = 0, fails = 0;
+
+/* One page, so the three futex words below cannot share a cache line with the
+ * child stacks and slow the handshakes down.
+ */
+static int pi_lock;      /* the PI futex the waiter parks on */
```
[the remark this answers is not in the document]
alanhc: Miscount on the review's side, not the file's. The line reads "The lock under test **and** the three handshake words": `pi_lock` is the lock under test; `holder_ready`, `release` and `waiter_done` are the three handshake words. Four declarations, and the sentence accounts for all four. The comment quoted here as "the three futex words below" is not what the file says.

The line has since changed anyway: `dfbd075` adds a fourth handshake word, `holder_failed`, so it now reads "the four handshake words".

tests/test-futex-requeue-pi.c
alanhc: Fixed in `dfbd075`, the way you describe: `holder_failed` is set before `raw_exit`, `holder_ready` is set after it so main is not left parked, and main checks the word and fails with "FUTEX_LOCK_PI refused the holder, nothing to requeue against".

Your read on the wrapper is the correction to mine. I had found the same gap and described it as wedging the suite, which is only true off the matrix. Measured against `FUTEX_LOCK_PI` stubbed to ENOSYS, run directly as `make check` runs it:

```
before: 60s cap reached, exit 124, no output at all
after:  exit 1 in under a second
  clone holder and waiter        OK
  holder takes pi_lock           FAIL: FUTEX_LOCK_PI refused the holder, nothing to requeue against
```

The empty output before is its own small point: stdout is block-buffered into a pipe and `timeout` SIGKILLs, so even the results already printed were lost. On the matrix the 30s wrapper does convert it to a reported failure, but as you say the line says `timeout` and names nothing.

Healthy tree is 6 passed, 0 failed, unchanged otherwise.

src/runtime/futex.c
alanhc: Added in `d40ade5`, at the top of `futex_requeue` where Linux has it, ahead of the alignment check, since `futex_requeue()` refuses the counts before `get_futex_key` runs.

Confirmed the path you describe: `syscall.c:460` forwards as `sys_futex(g, x0, (int) x1, (uint32_t) x2, x3, x4, (uint32_t) x5)`, and `futex.c:2196` passes `val` plus `(uint32_t) timeout_gva`, so both counts reach `futex_requeue` as `uint32_t` with the sign surviving only as the top bit. Reading them back as `int32_t` is what recovers it:

```c
    if ((int32_t) wake_count < 0 || (int32_t) requeue_count < 0)
        return -LINUX_EINVAL;
```

Two cases in the test hold it, and they need no waiter, which is the point: on an empty address the wake and the requeue are both no-ops, so an unguarded build answers 0 and the assertion separates that from EINVAL. Verified by disabling the guard: both fail with `errno=0` and pass with it in place.

```
  a negative wake count is EINVAL FAIL: a negative wake count must be refused (errno=0)
  a negative requeue count is EINVAL FAIL: a negative requeue count must be refused (errno=0)
```

Agreed it belongs here rather than in a follow-up: it is the same function and the same guest-supplied pair the 64-bit sum already exists to survive.

tests/test-futex-requeue-pi.c
alanhc: Fixed in `d40ade5`: `FUTEX_LOCK_PI`, `FUTEX_UNLOCK_PI` and `FUTEX_REQUEUE` all carry `FUTEX_PRIVATE_FLAG` now, matching `test-futex-pi.c` and `test-futex-requeue-account.c`.

The comment above the helpers records why, since the bit is invisible under elfuse and a later reader would have no reason to keep it: `FUTEX_CMD_MASK` strips it here, but the file is meant to go to a reference kernel, where private and shared hash to different keys.

Healthy tree is 8 passed, 0 failed (the two negative-count cases are new), and the futex family around it is unchanged: `test-futex-pi`, `test-futex-requeue-account`, `test-futex-waitv`, `test-futex-timed`, `test-futex-wake-op`, `test-futex-wake-nowaiter`, `test-robust-futex`, `test-shim-futex-fast` and `test-pthread` all pass, plus `tests/driver.sh` at 97 passed.

tests/test-futex-requeue-pi.c
alanhc: Resolved by the flag change above. `pi_lock` is now parked on, requeued and woken through the same private key, so this check asserts on a reference kernel rather than waking a key nothing ever waited on. Nothing else about the case changed.

src/runtime/futex.c
alanhc: Confirmed, and left open deliberately: it changes what the fix does rather than how it is spelled.

The asymmetry is exactly as you describe. The pre-scan returns EINVAL before `int woken = 0` is ever reached, so nothing is woken; `requeue.c` marks each plain waiter with `this->wake(&wake_q, this)` inside the walk and flushes at `out_unlock` regardless of `ret`, so the same call there wakes the waiters ahead of the PI one and still fails. And you are right that the test cannot tell them apart: it queues exactly one waiter, so both behaviours look identical to it.

Of your two options I lean toward folding the check into the existing walk. The comment I wrote argues fidelity to `requeue.c` at some length, and a "deliberate departure" note sitting directly under it would read badly: if the case for the scope is that Linux checks every waiter the call could touch, the case for the effect is the same case. Either way it needs a second waiter in the test to pin the difference, since nothing currently distinguishes them.

Your call on that before it is written, then. Items 2 through 5 are done in `d40ade5`; this one is untouched.

src/runtime/futex.c
```diff
@@ -1440,33 +1451,48 @@ static int64_t futex_requeue(guest_t *g,
...
+     *     the wake decision: a PI waiter inside the wake budget is rejected,
+     *     not woken;
+     *   - no requeue_count != 0 gate, since Linux walks and checks for a
+     *     wake-only requeue too.
+     *
+     * The budget mirrors Linux's "task_count - nr_wake >= nr_requeue" break, so
+     * the walk stops after as many matches as the call could act on. The sum is
+     * taken in 64 bits because both halves are guest-supplied and wake-all
+     * passes INT_MAX.
      */
```
jserv: Shorten comments. They are intended to be read by humans.

alanhc: Fair, and taken across the whole branch in `d3585f8`. The diff was 45 comment lines to 14 of code in `futex.c`. The ratio alone says the reviewer was reading an essay, not a change.

`futex.c` is now 16 comment lines to 14, and the test file lost about half of its prose too. What went: the three-way account of how the old guard was narrowed, the recitation of `requeue.c` around the check, and the paragraph justifying why the credit and the debit sit together. All of that is history and argument, which belongs in the commit message where it already is: the code needs only what a reader cannot get from the code.

The negative-count guard reads as:

```c
    /* Linux refuses these before taking either key. Both arrive as uint32_t,
     * so the guest's sign survives only as the top bit.
     */
```

down from seven lines, and the pre-scan from twenty-eight to six.

Nothing executable changed: 8 passed, 0 failed, the futex family and `tests/driver.sh` at 97 unchanged.

Still holding on your call for the other thread, whether to fold the PI check into the walk so the waiters ahead of it are still woken, or to record the atomicity as a departure. That one changes what the fix does, so I have left it alone.

src/runtime/futex.c
alanhc: I withdraw the lean I stated above. I wrote the fold, and writing the test that was supposed to justify it is what changed my mind.

**The fold itself is straightforward** and I have it working locally (not pushed): budget check, then the PI check, then the wake/requeue decision, `rc = -EINVAL; break` on a hit, and the unlock path returns `rc` before the os_sync top-up. Net +26/-27, existing cases 8 passed 0 failed, `tests/driver.sh` at 97 and the futex family unchanged.

**The test cannot be written, because the premise does not hold here.** I staged a plain `FUTEX_WAIT` waiter on `pi_lock`, cloned only after the retry loop proved the PI waiter was parked, so the head-insert would put it first. It failed against the folded build too. Instrumenting `futex_requeue` to count the chain it walks:

```
[probe] bucket matches=1 pi=1 wake=1 req=1     (identical on every one of 10 calls)
```

One waiter in the chain, and it is the PI one. The plain waiter was demonstrably parked at that moment (its `FUTEX_WAIT` had not returned, a sentinel in its return slot was still untouched), and `futex_requeue` returned `-EINVAL` without ever seeing it.

The mechanism is `futex_os_sync_wait`: on macOS 14.4+ a plain wait parks on the Darwin address-wait queue and is counted in `b->os_sync_waiters`, not linked into `b->head`. The comment above that call site says it outright: "futex_wake reads this bucket's census as the chain plus os_sync_waiters". A control probe with no PI waiter confirms the other half: `futex(W, FUTEX_REQUEUE, 1, 1, W)` does wake such a waiter, but through `futex_wake_topup_osync` after the walk finds nothing, which is exactly the step an EINVAL return skips.

**So folding the check into the walk does not buy Linux's behaviour here.** "The waiters ahead of the PI one" is a statement about one ordered list; elfuse has two unordered-with-respect-to-each-other populations, and the one the walk traverses holds almost only PI waiters. The fold would grant the Linux semantics to a set that is nearly always empty, while the waiters a guest actually has stay unreachable on that path.

That makes your second option the honest one, and for a better reason than I had: not "we chose not to", but "the ordering the kernel semantics are defined over does not exist in this implementation". Folding it anyway, with up to `wake_count` os_sync wakes on the EINVAL path as an approximation, would be guessing at what "ahead of it" means across the two populations. The departure is written down instead.

The measurement goes in the commit message, beside the argument for the departure.

alanhc: I put the premise itself to a kernel, and it does not survive. The current all-or-nothing behaviour matches Linux in every state I can reach; the fold buys nothing.

Linux 6.18.44 aarch64, via the reference VM this repo already ships (`tests/qemu-runner.sh` over the `fetch-fixtures.sh` tree). A holder takes `pi_lock` with `FUTEX_LOCK_PI`, then two threads queue on it in each order, then `futex(pi_lock, FUTEX_REQUEUE, 1, 1, pi_lock)`.

**Plain waiter queues first, PI waiter second**, and the PI side is refused:

```
requeue = 1  plain_woken=1  pi_rc=-22 (EINVAL)  word=0x1a9
```

`pi_rc` is the `FUTEX_LOCK_PI` return. Linux will not put a PI waiter behind a non-PI one on the same futex: `word` never gains FUTEX_WAITERS, so the mixed state the fold exists to serve is never created. The requeue then sees one plain waiter, wakes it, returns 1.

**PI waiter queues first, plain waiter second**, the state exists, and:

```
requeue = -22 (EINVAL)  plain_woken=0  plain_rc=<untouched>  word=0x800001a9
```

FUTEX_WAITERS is set and the plain waiter's `FUTEX_WAIT` never returned, so both were genuinely queued. Linux answered EINVAL and woke nobody, which is what this branch already does.

So the two reachable orderings give "refuse the PI waiter outright" and "EINVAL, no wakes". `wake_q` is flushed at `out_unlock` as you said, but with equal priorities the plist is FIFO, so a plain waiter can only be reached before the PI one in an ordering Linux refuses to create.

One case stays unmeasured: a plist is priority-ordered, so an RT-priority plain waiter queueing after a PI waiter would sit ahead of it. Pinning it needs RT scheduling in the lane, for a shape no guest workload here produces.

On that evidence I would leave the pre-scan as it is and drop the idea of a departure note, since there is no departure to record. The fold is written and sitting on a local branch if you disagree; say the word and I will push it.

Separately, now that the VM is running: `test-futex-requeue-pi` reports 8 passed, 0 failed against 6.18.44, so the negative-count guard and all three PI shapes are confirmed there, and the `FUTEX_PRIVATE_FLAG` change you asked for is what lets that run mean anything. `make test-matrix-qemu-aarch64` is 257 passed, 0 failed.

jserv: You should consolidate contracts for futex system call. See [Model Checking Futexes](https://www.doc.ic.ac.uk/~afd/papers/2023/SPIN.pdf) and utilize [frama-c-mcp](https://github.com/sysprog21/frama-c-mcp) to proceed.

alanhc: Started on it, scoped to what this PR already touches, and set the tooling up locally so the rest is reachable.

**In `5d0d0c1`:** the negative-count guard moves to `src/proved/futexreq.h`, next to the `futexhash` and `futexop` fragments this syscall already had, and `make verify-futexreq` discharges **14 of 14** obligations on alt-ergo/z3. Two functions: `futex_requeue_counts_valid` states the rule Linux applies before it takes either key, and `futex_requeue_budget` states what the walk is then bounded by, with the first as its precondition. That precondition is the part worth having: a validated pair leaves the sum inside 32 bits, where an unvalidated one bounds the walk at 2^33 - 2, which is the overflow CVE-2018-6927 reached through this same argument pair. `futexop.h` already names that CVE next door, so the contract closes a gap its own comment had pointed at.

It also fixed something I had got wrong: the guard read `(int32_t) wake_count < 0`, and converting a value above INT32_MAX to a signed type is implementation-defined before C23, so it was reading a sign the standard does not promise. It is a comparison against 2^31 now, which also keeps the provers in the value theory, the reason `futexop.h` and `futexhash.h` spell their masks as remainders.

Whole suite still closes: **1093 of 1093** across 23 targets, 63s on this machine, with Frama-C 33.0 / Alt-Ergo 2.6.3 / Z3 4.16.0, the pins `.github/actions/framac` carries, in a switch named `frama-c-elfuse` so a local run and CI are the same run.

**frama-c-mcp is installed and registered**, plugin built into that switch. Pointing its `check` at the new header is already informative: it returns `incomplete` with `VALID_UNDER_HYP: 5` and `EVA_NOT_RUN: 1`. Neither is a defect (the five are the goals that hold under `futex_requeue_budget`'s precondition, which is what makes it a conditional contract), but "is the budget contract conditional or total" is exactly the first question a consolidation pass has to answer, and the tool asks it without being prompted.

**What I have not done**, and left as its own change rather than growing this one: consolidating `futexhash`, `futexop`, `futexdeadline` and `futexreq` into one contract surface, against the model in the SPIN paper. That is a different piece of work from rejecting a PI waiter, it wants EVA in the loop rather than WP alone, and this PR has five rounds of review and a green matrix behind it. It goes to its own issue once this one lands.

jserv: > Happy to open it as its own PR or an issue and carry it there — say which you prefer.

Don't let AI agents make decisions for you. Always judge by yourself.

alanhc: Filed as #356 so this PR can stop here.

It records the four targets and their 160 obligations, and what they have in common: all four `_UNPROVED` lines defer the same thing in different words, the queue and bucket state the walks maintain. `futex.c` has 14 `futex_*` functions and 27 sites answering EINVAL or EAGAIN; the contracts reach a handful.

One thing I want to flag from reading the paper, since it changes what "consolidate" can mean. Evrard and Donaldson model futex-based **mutex and condition variable implementations** in Promela and check them with Spin, the userspace protocols built on top of the API, not a kernel-side implementation of it. So it does not hand us ACSL for `futex_requeue`. What it hands us is the register the fragments are missing: the properties those protocols need the API to hold. A per-function WP proof cannot state "a waiter this call enqueues is either woken or still queued when it returns".

That points at two tools rather than one, which the issue says plainly: WP/ACSL keeps the sequential arithmetic it already proves well, and the concurrent claims (no lost wakeup, no waiter charged to no bucket, the published census never understating the chain) want a model. That last one is not hypothetical: `futex_wake` already carries an `ELFUSE_CONTRACT_ASSERT` that re-reads before believing a violation, because it cannot see the chain and `os_sync_waiters` in one instant. That comment is a specification, written in prose, inside a runtime assert.

The issue opens on a survey rather than a rewrite: for each of the 9 proved functions, is the contract total or conditional, and was that decided or inherited? `futex_bucket_index` and `futex_op_shift_arg_mask` are total, `futex_remaining_ns` and now `futex_requeue_budget` are not, and nothing says which of those is deliberate, which is the same question `frama-c-mcp check` raises on its own with `VALID_UNDER_HYP`.

The survey goes to #356 as its own PR once this one lands.

jserv: > Filed as #356 so this PR can stop here.

No, formalize futex operations with Frama-C within this PR.

jserv:
To proceed before implementing #171 and #225, we need to ensure that verifiable contracts for critical system calls like `futex` are available. This ensures that all Frama-C based verifications are consolidated within this pull request.

alanhc:
Writing the contracts turned up a behavior difference rather than only
proving the existing behavior.

Both selectors in `val3` have encodings Linux does not implement: `op` is
three bits with 0..4 assigned, `cmp` is four bits with 0..5. Linux answers
`ENOSYS` for the rest, and refuses each at a different point, which the
modify makes visible. Measured on 6.18.44 through the qemu lane, with one
waiter parked on `uaddr`:

| op | cmp | ret | `*uaddr2` | waiter |
| --- | --- | --- | --- | --- |
| SET | EQ | 0 or 1 | modified | woken |
| 5..7 | EQ | `-ENOSYS` | untouched | parked |
| SET | 6, 15 | `-ENOSYS` | modified | parked |

elfuse treated both as no-ops and went on to wake up to `nr_wake` waiters
and report the count. It now answers the same as the reference in every
column, and `tests/test-futex-wake-op-enosys.c` fails 7 of 11 before the
change and passes on the reference kernel unchanged.

Separately, the waiter cases retry on `EINTR` rather than counting it as a
wake: elfuse returns `EINTR` from a `FUTEX_WAIT` no signal was delivered
to, which a first draft of the test read as the refused call answering it.
Whether that is itself a divergence is not investigated here and nothing
in this branch changes it.

alanhc:
Rebased onto `main` (f70ee7a). All runtime legs pass, including the five mutation legs that cover the new `futexreq` target.

The one red check, scan-build, is unrelated to this branch: it fails on `src/syscall/proc-identity.c:32`, which this pull request does not touch. `guest_has_ctty` is written in three places and read in none, so `-Wunused-but-set-global` under `-Werror` fails the analyze leg. Filed as #359 with the details.

Locally on macOS Apple Silicon, the rebased branch passes `make check-sanitizer` with exit status 0 and no `[ FAIL ]` lines, and `test-futex-requeue-pi` reports 8 passed, 0 failed.

alanhc:
The PI word transitions are proved now. Unlike the `WAKE_OP` selectors they
needed no behavior change: measured on 6.18.44, a robust owner's death
leaves `0x40000000` on an uncontended word and `0xC0000000` on one that
already had waiters, and elfuse answers the same both before and after.

`test-robust-futex` asserted only that `OWNER_DIED` was set, so the TID
clear and the `WAITERS` survival had no coverage. It now puts a case to
each, including the contended word.

The death transition is spelled the way `handle_futex_death()` spells it,
`(uval & FUTEX_WAITERS) | FUTEX_OWNER_DIED`, rather than as a TID clear.
The two agree on every input, since the OR sets bit 30 whatever the mask
left there, and the kernel's form is the one that can be checked against
the reference.

The waiters bit is read as a magnitude and edited as a remainder rather
than with the bit operators. Under the bitwise spelling the "leaves the
rest alone" clause on `set_waiters` times out at 30s on both provers while
every other clause in the file discharges in seconds, which is the wall
`futexop.h` documents. The TID and `OWNER_DIED` clauses stay bitwise,
because those discharge.

What the proofs do not reach: the queue walks, the CAS loops, the bucket
lock ordering and the blocking paths. Those are stateful and concurrent,
and `mk/verify.mk` records that ignoring `_Atomic` is sound for
per-function obligations one thread at a time and would not be for a
concurrency analysis. They stay test-covered.

tests/test-robust-futex.c
```diff
@@ -76,48 +80,75 @@ static int child_fn(void *arg)
...
+                     0, (long) &child_tid_addr);
+    if (ret == 0) {
+        child_fn(NULL);
+        test_unreachable();
+    }
     if (ret < 0) {
+        *ok = 0;
+        return 0;
+    }
+    usleep(100000); /* grace period for the exit-time walk */
```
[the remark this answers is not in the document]
alanhc: Valid, fixed in 4808451.

Confirmed the premise first: `robust_list_walk` runs immediately before the `CLONE_CHILD_CLEARTID` store in the worker teardown (`src/runtime/forkipc.c`), so a cleared ctid word really is proof the walk finished. `join_child` now waits on it with plain `FUTEX_WAIT` (matching the non-private wake CLEARTID issues), bounded so a stuck teardown is a reported failure rather than a driver timeout.

One thing the report missed, which made the naive version a no-op: the clone flags asked for `CLONE_CHILD_CLEARTID` but not `CLONE_PARENT_SETTID`, so `child_tid_addr` was zero from the start and a futex wait on it returned immediately. The address is now passed as ptid too, as `test-thread.c` and `test-futex-pi.c` already do.

The join also closes a hazard not mentioned here: `child_stack` is static and reused by the second run, and only a join proves the first child has left it.

For the record, I could not reproduce the flakiness locally -- the old version passed 48 concurrent runs plus a CPU-loaded sweep. The change stands on the guarantee, not on a measured failure.

tests/test-robust-futex.c
```diff
@@ -76,48 +80,75 @@ static int child_fn(void *arg)
...
+    } else {
+        EXPECT_TRUE(plain & FUTEX_OWNER_DIED, "FUTEX_OWNER_DIED not set");
+    }
+
+    /* The walk owes the word two more things than the flag. A TID left behind
+     * is an owner no LOCK_PI can displace, and it is what separates the robust
+     * path from an ordinary abandoned lock.
+     */
+    TEST("robust-futex: owner-died clears the TID");
+    EXPECT_TRUE(ok && (plain & FUTEX_TID_MASK) == 0,
```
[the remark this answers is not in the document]
alanhc: Valid, fixed in 4808451.

The two dependent checks, and their `TEST()` labels, are now inside `if (!err)`, so a setup failure yields one failure instead of three.

Went slightly further than the suggestion: `int *ok` became `const char **err`, because the join added a second way for setup to fail and reporting a join timeout as "clone failed" would have been its own misleading message.

tests/test-robust-futex.c
```diff
@@ -76,48 +80,104 @@ static int child_fn(void *arg)
...
+     * turns a stuck teardown into a reported failure.
+     */
+    struct timespec ts = {.tv_sec = 1, .tv_nsec = 0};
+
+    for (int i = 0; i < 10; i++) {
+        int seen = __atomic_load_n(ctid, __ATOMIC_SEQ_CST);
+        if (seen == 0)
+            return 0;
+        raw_syscall6(98, (long) ctid, FUTEX_WAIT, seen, (long) &ts, 0, 0);
+    }
```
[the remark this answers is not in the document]
alanhc: This one does not hold: `FUTEX_WAIT` is the single futex op whose timeout is *relative*. From futex(2):

> Note: for `FUTEX_WAIT`, timeout is interpreted as a relative value. This differs from other futex operations, where timeout is interpreted as an absolute value.

Measured it rather than trusting the man page. A probe that `FUTEX_WAIT`s on a word nobody wakes, using the same `{.tv_sec = 1}` timespec:

```
real Linux 6.18.44 (tests/qemu-runner.sh):  ret=-110 (ETIMEDOUT) after 1008.2ms
elfuse:                                    ret=-110 (ETIMEDOUT) after 1000.2ms
```

The wait blocks for a second on both lanes, so `join_child` does not degenerate into back-to-back polls, and elfuse is not diverging from the kernel here -- `futex_make_deadline` is called with `is_absolute = 0` for `FUTEX_WAIT` (`src/runtime/futex.c`) precisely because Linux treats it as relative. `test-robust-futex` itself also passes 6/6 under the qemu reference lane.

An absolute CLOCK_MONOTONIC deadline would be required for `FUTEX_WAIT_BITSET`, `FUTEX_LOCK_PI` and `FUTEX_WAIT_REQUEUE_PI`, none of which this code uses. Leaving `join_child` as it is.

tests/test-futex-wake-op-enosys.c
```diff
@@ -0,0 +1,226 @@
...
+
+    unsupported_op_case("op 5 is ENOSYS", 5);
+    unsupported_op_case("op 6 is ENOSYS", 6);
+    unsupported_op_case("op 7 is ENOSYS", 7);
+
+    unsupported_cmp_case("comparison 6 is ENOSYS", 6);
+    unsupported_cmp_case("comparison 15 is ENOSYS", 15);
+
+    /* Both unassigned: the op is read first, so the word stays untouched. */
+    unsupported_op_case("an unsupported pair stops at the op", 5);
```
jserv: `unsupported_op_case` hardcodes cmp to 0 in its `encode` call, so this case runs a supported comparison and is a byte-for-byte repeat of `op 5 is ENOSYS` above. The comment's claim (both selectors unassigned, the op is read first) is never exercised.

Take cmp as a parameter, or spell this one out: `encode(5, 6, 0x111, 0x111)`.

tests/test-futex-wake-op-enosys.c
```diff
@@ -0,0 +1,226 @@
...
+    }
+    PASS();
+
+    /* The waiter is still there, so a supported pair reaches it. This is also
+     * what keeps the two cases above from passing on a waiter that never
+     * parked: one that had gone leaves nothing to wake here.
+     */
+    TEST("a supported op wakes the waiter left behind");
+    __atomic_store_n(&park_word, 1, __ATOMIC_RELEASE);
+    if (futex_wake_op(1, 1, valid_val3()) != 1) {
```
jserv, line 212: `target_word` is `0x111` here, left by the refused-comparison case above, so `valid_val3`'s EQ does fire and the second wake runs. The call returns 1 only because nothing is parked on `target_word`.

Reset `target_word` to 0 before this assertion, or drop the "the second wake does not fire" claim from the comment on `valid_val3`.

tests/test-futex-wake-op-enosys.c
```diff
@@ -0,0 +1,226 @@
...
+    raw_futex_wake(addr, 1);
+}
+
+/* Parks once. EINTR is a retry rather than a wake, so a build that interrupts
+ * the wait does not read as one that answered it.
+ */
+static void waiter_fn(void)
+{
+    long r;
+    set_and_wake(&waiter_parking);
```
jserv, line 73: `waiter_parking` is published before `FUTEX_WAIT` is entered, so the two "leaves the waiter parked" cases rest on `settle()` rather than on an ordering guarantee. The final supported call does catch a waiter that never parked, so this is lane flakiness rather than a false pass, but 10000 `sched_yield` calls are buying a guarantee they cannot make.

tests/test-futex-requeue-pi.c
```diff
@@ -0,0 +1,227 @@
...
+ * reaching the case.
+ *
+ * Syscalls exercised: futex(98), clone(220), gettid(178), exit(93), sched_yield
+ */
+
+#include <limits.h>
+#include <stdint.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/mman.h>
```
jserv: `string.h`, `unistd.h`, `sys/mman.h` and `stdint.h` are all unused in this file. The header at line 17 also names `gettid(178)`, which the test never issues.

jserv:
Always avoid agent-specific words like
> If you would rather the source read uniformly one way or the other, say
so and I will move the remaining three — it costs a longer prover budget
on one goal, not correctness.

Let's concentrate on real discussions!

alanhc:
The EINTR noted above is #361, on main and independent of this branch.

The premise was wrong rather than the timing: clone(2) sends no signal
when a CLONE_THREAD thread terminates, and forkipc.c raised the futex
interrupt for that signal anyway. It reached ppoll and epoll_pwait as
well, since they read the same flag.

jserv (changes requested):
Rebase onto the latest `main` branch, respond to review comments, and rework the commits so that each is functionally minimal, free of known issues, and similar changes are folded together.

jserv:
After integrating this PR, it is time to update #224 .

src/proved/futexpi.h
```diff
@@ -0,0 +1,107 @@
...
+ * Layout, which Linux fixes and a guest can write any bit pattern into:
+ *
+ *   bits 0-29  owner TID (FUTEX_TID_MASK)
+ *   bit 30     FUTEX_OWNER_DIED
+ *   bit 31     FUTEX_WAITERS
+ *
+ * The death transition is the one worth a contract rather than a mask: it has
+ * to clear the TID, set OWNER_DIED and leave WAITERS alone, and an owner whose
+ * waiters bit it dropped is a lock nobody is ever woken from.
+ */
```
jserv, line 15: Append upstream reference.

alanhc: Added: the three fields come from `include/uapi/linux/futex.h` (`FUTEX_WAITERS` 0x80000000, `FUTEX_OWNER_DIED` 0x40000000, `FUTEX_TID_MASK` 0x3fffffff), now named in the layout block. The death transition further down already cites `handle_futex_death()` in `kernel/futex/core.c`.

Squashed into `Prove the PI lock word transitions`; `make verify-futexpi` still discharges 38 of 38.

src/runtime/futex.c
```diff
@@ -1512,12 +1507,13 @@ static int64_t futex_requeue(guest_t *g,
...
-            /* Move the publication with the waiter. The destination is charged
-             * before the source is debited, so the shim never sees this parked
-             * waiter charged to no bucket.
+            /* Credit the destination before debiting the source, so the shim
+             * never sees this waiter charged to no bucket. Both halves sit
+             * under pub_follows: a waiter carrying no charge of its own must
+             * not gain one here, which is how a charge outlived its waiter.
              */
-            shim_globals_futex_waiters_add(g, idx_dst, +1);
             if (w->pub_follows) {
```
jserv: When `idx_src == idx_dst`, `b_dst` is `b_src`, and the move below (`*pp = w->next`, then `w->next = b_dst->head; b_dst->head = w`) puts the waiter straight back into `*pp` whenever `pp` is still `&b_src->head`. With `uaddr == uaddr2` its `uaddr` still matches on the re-read, so `futex(X, FUTEX_REQUEUE, 0, INT_MAX, X)` against a single waiter parked on `X` re-links that one waiter `INT_MAX` times while holding `b_src->lock`, never reaching the second entry, then reports `INT_MAX` requeued. A standalone model of this loop runs 8 s at `-O2`; here every other futex op on the bucket blocks for the duration.

Linux's `requeue_futex()` skips the list move entirely when `&hb1->chain == &hb2->chain` and only rewrites `q->key`, so its walk advances. Setting `w->uaddr = uaddr2; pp = &w->next;` and skipping the unlink/relink when `idx_src == idx_dst` gets both the termination and the count.

This predates the change, but `uaddr == uaddr2` is the case this PR teaches the function to reason about, so it belongs here or in the follow-up it names.

src/runtime/futex.c
```diff
@@ -1673,6 +1656,16 @@ static int64_t futex_wake_op(guest_t *g,
         return -LINUX_EFAULT;
     }
 
+    /* A comparison Linux does not implement stops here: the modify above has
+     * already landed, and neither wake runs.
+     */
+    if (!futex_wake_cmp_supported(wake_cmp)) {
```
jserv, line 1706: The kernel runs the same test this PR restores for requeue inside `futex_wake()` and `futex_wake_op()` too: `if (this->pi_state || this->rt_waiter) { ret = -EINVAL; break; }`, from inside their own walks. The two wake loops below still treat a `pub_follows == false` entry as an ordinary waiter, so `futex(X, FUTEX_WAKE, 1)` against a thread parked in `FUTEX_LOCK_PI(X)` answers 1 where Linux answers `EINVAL`.

The note above `futex_lock_pi` argues a PI waiter has to be visible to the census, which is a different claim from wakeable. Follow-up: apply the same `pub_follows` test in these two loops and in `futex_wake`.

tests/test-futex-requeue-pi.c
```diff
@@ -0,0 +1,239 @@
...
+#include "raw-syscall.h"
+
+int passes = 0, fails = 0;
+
+/* The lock under test, a second address distinct enough to land in a different
+ * bucket, and the four handshake words that sequence the two threads around
+ * them.
+ */
+static int pi_lock;       /* the PI futex the waiter parks on */
+static int pi_lock_dst;   /* requeue destination for the cross-bucket case */
```
jserv, line 46: Nothing checks the claim that `pi_lock_dst` lands in a different bucket. The two words are adjacent in BSS and `futex_hash` collides for roughly one address pair in 1024, and on a collision the cross-bucket case silently repeats the same-bucket one and still passes. `tests/test-futex-waitv-buckets.c` already mirrors the hash as `bucket_of`; the same helper here would let this assert the split rather than assume it.

tests/test-futex-waitv-buckets.c
```diff
@@ -0,0 +1,171 @@
...
+    int64_t tv_sec;
+    int64_t tv_nsec;
+};
+
+#define PARK_MS 200
+#define EARLY_MS (PARK_MS / 2)
+
+#define ARENA_WORDS (1u << 20)
+static uint32_t arena[ARENA_WORDS];
+static int32_t first_at[FUTEX_BUCKETS];
```
jserv: `first_at` is filled with -1 at the top of `main` and never read again. Drop the array and its loop.

tests/test-futex-wake-op-enosys.c
alanhc: Fixed in `2fa99a3`. The supported-op assertion now resets the word first, so the EQ it rests on is a comparison against 0 rather than against the `0x111` the refused-comparison case left behind:

```c
     TEST("a supported op wakes the waiter left behind");
     __atomic_store_n(&park_word, 1, __ATOMIC_RELEASE);
+    __atomic_store_n(&target_word, 0, __ATOMIC_RELEASE);
     if (futex_wake_op(1, 1, valid_val3()) != 1) {
```

The comment on `valid_val3` keeps its claim, which is now the one the code makes.

tests/test-futex-wake-op-enosys.c
```diff
@@ -0,0 +1,226 @@
...
+ * Copyright 2026 elfuse contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Both selectors in val3 are guest-supplied and both have unassigned encodings.
+ * Linux refuses each at a different point, which the modify makes visible: an
+ * unimplemented op stops before it, an unimplemented comparison after it.
+ * Neither wakes anybody.
+ *
+ * Every assertion is an errno or a word the guest can read, so the same source
+ * runs against a reference kernel.
```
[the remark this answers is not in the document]
alanhc: This does not hold. The test runs unmodified on the reference kernel the tree cross-checks against, Linux 6.18.44 aarch64 via `tests/qemu-runner.sh`:

```
=== futex wake_op selector tests ===
  a supported pair modifies the word OK
  op 5 is ENOSYS                 OK
  op 6 is ENOSYS                 OK
  op 7 is ENOSYS                 OK
  comparison 6 is ENOSYS         OK
  comparison 15 is ENOSYS        OK
  ...
test-futex-wake-op-enosys: 11 passed, 0 failed - PASS
```

That covers all four assertions named above (lines 103, 119, 181, 193). Linux answers `-ENOSYS` for an unassigned op or comparison selector, not `-EINVAL`; the measured table is in the commit message of `2fa99a3`, one row per selector class with the resulting errno, the state of `*uaddr2` and whether a parked waiter was woken. `EINVAL` is what an out-of-range operand shift gets, which is a different refusal and not what these cases encode.

The reference-kernel claim in the header is the one being tested here, and it passes. No change.

src/runtime/futex.c
alanhc: Carried in this PR rather than in #356. Four new headers under `src/proved/`, on top of the `futexhash.h` and `futexop.h` already on main, 180 proof obligations:

| header | commit | what it pins | WP |
| --- | --- | --- | --- |
| `futexreq.h` | `d88f158` | plain-requeue counts. A negative pair reaches elfuse as unsigned and makes the walk budget `nr_wake + nr_requeue` reach 2^33-2; refusing the pair first is what holds the sum inside 32 bits. Same argument pair CVE-2018-6927 went through, on the same syscall. | 14 of 14 |
| `futexwakeop.h` | `2fa99a3` | both `val3` selectors and the two switches they guard, one value per op rather than a range | 92 of 92 |
| `futexpi.h` | `ced3cba` | the PI lock word: TID, OWNER_DIED and WAITERS reads and edits, with the death transition pinned bit for bit rather than by mask | 38 of 38 |
| `futexwaitv.h` | `6866109` | the bucket set a `futex_waitv` locks comes out sorted and repeat-free. A repeat locks one non-recursive mutex twice; an unsorted set takes the bucket locks out of order against every other futex path. | 36 of 36 |

"PROVED 180 of 180" on its own says nothing about whether the clauses are load bearing, so `scripts/check-mutants.py` carries 16 mutations across the four, one per property that would otherwise be free: accept any count pair, accept every op, accept every comparison selector, drop the complement from ANDN, widen LT to LE, read the flag bits as part of the TID, drop the waiters bit from the death transition, stop the bucket scan one short. All 16 are caught.

One qualification on that number: all 16 are caught by exhausting the prover rather than by refutation, which is a weaker verdict. What makes it evidence is the baseline proving every goal while each mutant exhausts on the goal its own function owns. `--escalate SECONDS` is what would separate unprovable from slow, and I have not run it.

What stays out of the proofs is the same thing in all four `_UNPROVED` lines: the walks and CAS loops around these functions, which stay test-covered. That is the queue and bucket state #356 records, and it is not reachable without the concurrency model that issue is about.

jserv: > What stays out of the proofs is the same thing in all four _UNPROVED lines: the walks and CAS loops around these functions, which stay test-covered. That is the queue and bucket state https://github.com/sysprog21/elfuse/issues/356 records, and it is not reachable without the concurrency model that issue is about.

Employ E-ACSL to catch concurrency specification violations dynamically during execution.

jserv: See also:
* https://frama-c.com/fc-plugins/mthread.html
* https://www.frama-c.com/fc-plugins/conc2seq.html

alanhc: Ran all three against the census, which is the smallest piece of the
concurrent surface: `shim_globals_futex_waiters_add` publishes a count with
`memory_order_seq_cst`, the shim's wake fast path reads it without a lock, and
each side stores one location and loads the other. The property is the one
`src/core/shim-globals.h` states in prose -- if the shim reads zero, any waiter
that exists loads the word after that read, sees the store, and returns EAGAIN
rather than parking. A stale-low count is a lost wakeup.

## E-ACSL

It translates, and it drops the atomics on the way.

Frama-C's own `share/libc/stdatomic.h:38`:

```c
// _Atomic is currently ignored by Frama-C
#define _Atomic
```

A probe with three `_Atomic` globals and the assertion above comes out of
`-e-acsl -then-last -print` as:

```c
static uint32_t census;
static uint32_t word;
static int parked;
```

The assertion itself survives and is emitted at the right point. What it is
checked against is a program with no atomicity, so the store-buffer shape the
claim rests on is not there to check. The memory order does reach the monitor
-- `__gen_e_acsl___fc_atomic_store_explicit_marker(..., memory_order_seq_cst,
...)` -- as an argument recorded beside a plain variable.

This is not a macOS artifact. The `#define` is in Frama-C's model, so Linux
gives the same translation.

Two macOS-only obstacles sit in front of it, both avoidable and worth
recording: `e-acsl-gcc` fails with `unexpected output of system getopt` under
BSD getopt (`brew install gnu-getopt` fixes it), and Frama-C cannot parse
Apple's `pthread.h` because of `_Nonnull`/`_Nullable` (Frama-C's own libc
parses instead).

The manual also states two limits that bear on this directly. `--concurrency`
exists and monitors pthreads, but Known Limitations 3.1.2 says "The runtime
library is also not thread-safe", and 2.2.5 says the E-ACSL language "does not
support concurrent access indications", with the documented workaround being to
demote a function contract to a statement contract inside a lock. The census
has no lock to move it inside; that is the point of it.

## Mthread

From its own page: "The detection of race condition supposes that shared zones
are protected by mutexes. Lock-free algorithms are not detected as such."

The bucket chain is under a mutex and is not where the open questions are. The
census, the shim's read, and `os_sync_waiters` are lock-free, and those are the
three the `_UNPROVED` lines defer.

## Conc2Seq

Its page marks it "Archived: no maintenance or support, incompatible with
recent Frama-C versions". It also assumes the input is sequentially consistent,
"which must be verified independently (e.g. by using the Mthread plugin)".

## Where that leaves it

`mk/verify.mk:115` already recorded the shape of this before I measured it:
ignoring `_Atomic` is sound for per-function runtime-error obligations one
thread at a time, and would not be sound for a concurrency analysis. The
measurement adds that the concession is not ours to withdraw -- it is the
frontend's, and every plugin downstream of it inherits it.

What E-ACSL can still do here is narrower than catching concurrency violations,
and worth doing on its own terms: `futex_wake` carries an
`ELFUSE_CONTRACT_ASSERT` that re-reads before believing a violation, and that
comment is a specification in prose inside a runtime assert. Written as ACSL and
monitored, it becomes a checked one, for the properties that do not rest on
atomicity. That is a smaller claim than the four `_UNPROVED` lines make, and it
does not close them.

The probe and its translation are reproducible with Frama-C 33.0:

```
frama-c -machdep gcc_x86_64 census-probe.c -e-acsl -then-last -print -ocode probe.frama.c
```

src/runtime/futex.c
alanhc: Fixed in `be22079`, the way you describe: when `idx_src == idx_dst` the walk rewrites `w->uaddr` and advances `pp = &w->next` without touching the list, and the unlink/relink stays for the cross-bucket case. That is what `requeue_futex()` does for `hb1 == hb2`.

`tests/test-futex-requeue-samebucket.c` parks one waiter and issues `FUTEX_REQUEUE(X, X)` with the largest requeue budget a guest can pass. Reverting the fix and running it under elfuse times out; with the fix it returns in well under a second reporting `rc=1` rather than `INT_MAX`.

It needs no fixture and is not in `QEMU_SKIP`, so both `EXPECTED_BASELINES` floors in `test-matrix.sh` moved up by one: elfuse-aarch64 251 and qemu-aarch64 226, both observed here.

src/runtime/futex.c
alanhc: Applied in `a295121`, and the premise measured first. On Linux 6.18.44 aarch64 through the qemu lane, with a holder owning the lock and one thread parked in `FUTEX_LOCK_PI` on it:

```
call                                    Linux     elfuse before
FUTEX_WAKE(pi_lock, 1)                  -EINVAL   1 woken
FUTEX_WAKE_OP, PI waiter at uaddr       -EINVAL   1 woken
FUTEX_WAKE_OP, PI waiter at uaddr2      -EINVAL   1 woken
```

`futex_pi_waiter_within()` reads the same `pub_follows` bit `futex_requeue` reads. Both wake paths consult it before anything is woken rather than partway through the walk, so a refused call leaves the chain as it found it -- the all-or-nothing shape this PR already settled on for requeue. The walk is bounded by the wake budget, because a waiter Linux never reaches cannot refuse the call: its loops stop once `nr_wake` entries are woken. `wake_op` tests both walks up front, which needs the comparison evaluated before either runs; the modify on `uaddr2` still lands, since Linux applies it before its walks reach the waiter that answers EINVAL.

One consequence worth naming: the robust-list walk wakes through `futex_wake()` too, so a PI lock whose owner died now takes EINVAL there and wakes nobody, as it does in `handle_futex_death()`. The waiter is not stranded -- the untimed `FUTEX_LOCK_PI` loop re-reads the word every 100 ms and recovers on `OWNER_DIED` by itself. `test-robust-futex` exercises that path and still passes.

`tests/test-futex-wake-pi.c` puts a case to each of the three calls above. It fails on the previous build in all three and passes unmodified on the reference kernel, so the errno it asserts is measured rather than assumed. Both matrix lanes are clean with it: elfuse-aarch64 286 passed / 0 failed, qemu-aarch64 264 passed / 0 failed, and both `EXPECTED_BASELINES` floors move up by one.

jserv (changes requested):
Rework the commits to be functionally minimal and free of known issues.

jserv:
Thank @alanhc for contributing!

PR #352, opened 2026-09-01
Bound the pty hangup reads with an alarm

### LINKED ISSUE #275: Pty hangup checks in test-pty block instead of failing
The hangup checks in tests/test-pty.c read from a master that is
expected to be hung up. Without the support in place those reads block
forever rather than returning EIO, so a run against a tree that lacks it
hangs instead of failing -- a control run took ten minutes and had to be
killed.

A test that guards a fix should fail when the fix is absent, not wedge
the suite. tests/test-devpts.c already takes this position for its pty
round trip: signal(SIGALRM, handler) plus alarm(10) around the blocking
reads, cleared with alarm(0) once they complete.

Applying the same guard to the hangup block would keep a regression
visible in CI rather than turning it into a timeout.

alanhc:
Closes #275.

The hangup checks in `tests/test-pty.c` read from a master whose slaves have
all closed. Linux answers EIO there, but on a tree without the support nothing
fails the read at all -- elfuse's keepalive slave holds the host pty open, so
the host has no reason to. The read parks, and a run against such a tree stalls
instead of reporting.

`tests/test-devpts.c` already takes the position that a test guarding a fix has
to fail when the fix is absent, and the `readv` case at the bottom of this file
does too: `signal(SIGALRM, ...)` plus `alarm(10)` around the blocking call,
cleared with `alarm(0)` once it returns. The first hangup block never got the
same treatment. This applies it there.

Each read gets its own window. The second is the wedge; the drain above it
reads the same hung-up master with no deadline of its own. Each alarm is
cleared as soon as its read returns, so nothing prints while one is armed.
Hoisting the second read out of `EXPECT_TRUE` is what lets the alarm be
cleared before the verdict is printed, as the `readv` case does.

Two smaller things came out of measuring it. The handler was silent, and its
`_exit(2)` discarded whatever stdout still held, so a wedged run through a pipe
produced no output at all -- neither the results collected so far nor a word
about why it stopped. It now names the timeout on stdout, and stdout is
line-buffered, so the log ends with every check that did complete followed by
the reason the run went no further.

## Evidence

Measured against a tree with `proc_pty_master_hung_up()` stubbed to `return
false`, which is what "without the support" means here.

| | before | after |
|---|---|---|
| outcome | still running at a 60s cap, killed | **exit 2 at 10s** |
| output | nothing | 37 lines of results, `master reports POLLHUP once the slave closes FAIL`, then the timeout line |

After, on the stubbed tree:

```
  packet-mode read carries a status byte OK
  master reports POLLHUP once the slave closes FAIL: no POLLHUP after the last slave closed (errno=2)
  queued output survives the hangup OK

test-pty: TIMEOUT waiting on a hung-up master read
```

On an unmodified tree the file still reports `70 passed, 0 failed - PASS`,
exit 0, in 3 seconds.

`make check-format` passes (clang-format 22.1.8, commentflow), `make indent`
leaves the file unchanged, and the cross build is clean under `-Wall -Wextra`.
A full `make check` did not run locally for want of `aarch64-none-elf-as`, so
everything outside `test-pty` is on CI.

---
## Summary by cubic
Fixes #275 by bounding the hung-up master reads in `tests/test-pty.c` with an alarm: on trees without hangup support those reads previously blocked forever, now the run exits 2 at 10 seconds with a timeout message.

**Bug Fixes**
- Applies the existing `SIGALRM` + `alarm(10)` pattern from the `readv` case to both hangup reads, clearing the alarm before printing the verdict.
- Writes the timeout reason to stdout, not stderr (`test-matrix.sh` discards stderr), and line-buffers stdout so the log retains all completed checks when the alarm fires.

<sup>Written for commit 43bbd8c9961e1911ea529843c8dfeac119998c2c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/352?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tests/test-pty.c
```diff
@@ -88,11 +88,17 @@
...
+ * that into a visible failure instead of a wedged run. The message says why the
+ * run stopped: the log otherwise just ends after the last completed check, with
+ * nothing to tell a timeout apart from a crash.
  */
 static void hup_on_alarm(int sig)
 {
     (void) sig;
+    static const char msg[] =
+        "\ntest-pty: TIMEOUT waiting on a hung-up master read\n";
+    ssize_t ignored = write(2, msg, sizeof(msg) - 1);
```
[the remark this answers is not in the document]
alanhc: Valid, and fixed in 2848eb0. Thanks.

I checked before acting: `run_elfuse` in `tests/test-matrix.sh` ends both of its branches with `2> /dev/null` (lines 170 and 173), and `run_elfuse_sysroot` does the same, so nothing this handler wrote to stderr would have reached the matrix log. That is the lane most likely to meet a timeout with nobody watching, which is exactly where the message earns its keep.

The diagnostic now goes to `STDOUT_FILENO`. Verified against a tree with `proc_pty_master_hung_up()` stubbed to `return false`, run as the matrix runs it:

```
$ ./build/elfuse ./build/test-pty 2>/dev/null > log; echo $?
2                                   # after 11s, not a hang
$ tail -3 log
  master reports POLLHUP once the slave closes FAIL: no POLLHUP after the last slave closed (errno=2)
  queued output survives the hangup OK

test-pty: TIMEOUT waiting on a hung-up master read
```

37 lines of results survive with it, because the same commit makes stdout line-buffered. That also settles the interleaving question a raw `write(2)` onto a stdio stream would otherwise raise: both guarded reads sit at a line boundary, so the write cannot land inside a partly-built line. The comment above the handler now records both points.

Worth noting for whoever looks next: `tests/test-devpts.c` has the same shape: its `on_alarm` writes `test-devpts: TIMEOUT waiting on pty read` to fd 2, and `test-devpts` is registered in the matrix at line 739, so that message is discarded there too. I have left it alone as unrelated to this change, but it is the same blind spot.

tests/test-pty.c
```diff
@@ -804,6 +824,17 @@ int main(void)
...
+                /* Guard both reads, as the readv case below does. The second
+                 * one is the wedge: on a tree without the support nothing ever
+                 * fails it -- elfuse's keepalive slave holds the host pty open,
+                 * so the host has no reason to -- and the run stalls where it
+                 * should report. The drain above reads the same hung-up master
+                 * with no deadline of its own, so it belongs in the same window
+                 * rather than in one of its own.
+                 */
+                signal(SIGALRM, hup_on_alarm);
+                alarm(10);
```
jserv, line 836: The alarm stays armed while the drain verdict prints. If the drain read burns most of the window, SIGALRM can fire after that read already returned, and it can land between `TEST()`'s label (printed without a newline, so still buffered) and the newline `EXPECT_TRUE` adds. The handler comment claims both guarded reads sit at a line boundary; that holds for the reads, not for the gap between them, and the TIMEOUT line would then name a read that did not block. Clearing at line 850 and re-arming just before the EIO read costs two lines.

jserv (commented):
One note outside the diff: the description says "It now names the timeout on stderr", but the handler writes `STDOUT_FILENO` and the new comment argues at length for stdout. One word to fix in the prose.

jserv (changes requested):
Rebase and refine per review.

tests/test-pty.c
alanhc: Split into two windows in 43bbd8c: the drain read and the EIO read are each bracketed by `alarm(10)`/`alarm(0)`, so nothing prints while the alarm is armed. The handler comment now states that invariant, and the "stderr" in the description is corrected. Rechecked on the stubbed tree: exit 2 at 11s, log ends with the drain OK and the TIMEOUT line. Rebased onto e03b72e.

Issue #292, opened 2026-08-13
elfuse real-workload syscall frequency ranking (Top 50)


maxliu0:
- Counts are guest syscalls actually handled by the host dispatcher.

## Workloads (five categories, 116 guest processes)

| workload | image | content | syscalls | processes |
|---|---|---|---:|---:|
| jvm | eclipse-temurin:21 | javac compile of a multi-class program + java run (collections, file I/O, SHA-256, 8 threads, subprocess) | 3,028,653 | 8 |
| python | python:3.12-slim | JSON/regex churn, sqlite3 (20k inserts + queries), 400 small-file I/O, os.walk over the stdlib, 10 subprocesses, 4 threads | 1,364,400 | 11 |
| go | golang:1.23-alpine | `go version` + `gofmt` over stdlib trees (net/runtime/crypto) + `gofmt -d fmt` | 46,037 | 6 |
| node | node:22-alpine | module loading, fs/crypto/zlib/JSON, then an HTTP server taking 302 host-side curl requests before /quit | 18,740 | 1 |
| c-compile | gcc:14 | 30-file C project via `make -j1` + sqlite amalgamation (8 MB TU) `gcc -O1` compile+link + a sqlite3 query | 82,606 | 90 |

Total: **4,540,436 syscalls, 89 distinct**.

## Top 50 ranking (five workloads combined, descending)

| # | syscall | count | share% | cum% | jvm | python | go | node | c-compile |
|--:|---|--:|--:|--:|--:|--:|--:|--:|--:|
| 1 | futex | 2,559,392 | 56.369 | 56.37 | 1,517,216 | 1,040,892 | 166 | 1,118 | 0 |
| 2 | clock_gettime | 1,849,178 | 40.727 | 97.10 | 1,507,075 | 308,525 | 30,326 | 3,251 | 1 |
| 3 | readlinkat | 48,107 | 1.060 | 98.16 | 109 | 3 | 1 | 15 | 47,979 |
| 4 | openat | 12,457 | 0.274 | 98.43 | 240 | 1,213 | 1,572 | 633 | 8,799 |
| 5 | read | 9,449 | 0.208 | 98.64 | 409 | 972 | 2,903 | 1,230 | 3,935 |
| 6 | close | 6,978 | 0.154 | 98.79 | 209 | 1,255 | 1,573 | 936 | 3,005 |
[...]

PR #353, opened 2026-09-01
Answer AT_EMPTY_PATH before the path resolver

### LINKED ISSUE #322: fstat() of a pipe or socket fails with ENOTDIR under a sysroot
# fstat() of a pipe or socket fails with ENOTDIR under a sysroot

## Summary

`fstatat(fd, "", &st, AT_EMPTY_PATH)` returns `ENOTDIR` for any descriptor that
has no host path behind it — a pipe, a socket, an eventfd. Since glibc 2.33 that
is how `fstat(fd)` is spelled, so on a sysroot with such a glibc, plain
`fstat()` on a pipe fails.

The empty path names the descriptor itself and resolves nothing, but
`stat_at_path()` runs `path_translate_at()` before it looks at `AT_EMPTY_PATH`.
That resolver measures the (relative, empty) name against `dirfd`, and
`dirfd_guest_base_path()` correctly reports `ENOTDIR` for a descriptor that is
not a directory — correct for a name, wrong for the empty path.

The most visible casualty is coreutils `cat`, which `fstat`s its stdout as its
first action, so `anything | cat` dies with
`cat: standard output: Not a directory`. That takes down `dpkg`/`apt` flows that
pipe a tar stream through `cat`.

## Environment

- elfuse `b1aac9a` (reproduced on a clean checkout of that commit)
- macOS 26 (Darwin 25.6.0), Apple silicon
- Affected sysroot: Debian 12 (bookworm) aarch64, glibc 2.36
- Unaffected sysroot: Ubuntu 26.04 aarch64, glibc 2.43

## Reproduction

```sh
elfuse --sysroot DEBIAN12_ROOT /bin/sh -c 'cat /etc/hostname | cat'
```

Actual:

```
cat: standard output: Not a directory
cat: standard output: Not a directory
rc=1
```

Expected: the hostname, as on the host.

The same command against an Ubuntu 26.04 (glibc 2.43) sysroot with the same
elfuse binary prints the hostname — that glibc issues `__NR_fstat` (syscall 80),
which lands in `sys_fstat()` and is not affected.

Another one-liner on the Debian sysroot, same cause:

```sh
elfuse --sysroot DEBIAN12_ROOT /bin/sh -c 'stat -c %F /proc/self/fd/1 | cat'
```

## Trace

`--verbose` on the failing run:

```
syscall 79@0x2000ed07c(0x1, 0x2001690a0, 0x7ffef08, 0x1000, 0x0, 0x0)
  -> -20 (0xffffffffffffffec)
```

syscall 79 = `newfstatat`, dirfd = 1 (the pipe), flags = `0x1000` =
`AT_EMPTY_PATH`, result = `-ENOTDIR`.

The failing `cat` call site is:

```c
/* coreutils src/cat.c */
if (fstat (STDOUT_FILENO, &stat_buf) < 0)
  error (EXIT_FAILURE, errno, _("standard output"));
```

which is where the message text comes from.

## Root cause

`stat_at_path()` (`src/syscall/fs-stat.c`) reads the path and then immediately
calls `path_translate_at()`; its `AT_EMPTY_PATH` branch is further down, after
translation:

```c
    if (guest_read_path(...) < 0)
        return -LINUX_EFAULT;

    if (pathp[0] == '/' && fuse_path_matches_mount(pathp)) { ... }

    path_translation_t tx;
    if (path_translate_at(dirfd, pathp, ..., &tx) < 0)
        return linux_errno();          /* <-- ENOTDIR comes from here */
    ...
    if ((flags & LINUX_AT_EMPTY_PATH) && pathp[0] == '\0') { ... }
```

Inside `path_translate_at()` the empty path is a relative path
(`guest_path[0] != '/'`), so with a sysroot active it reaches
`path_check_relative_sysroot_containment()` →
`dirfd_reconstruct_abs_path()` → `dirfd_guest_base_path()`
(`src/syscall/path.c`), which ends in:

```c
    /* fd_snapshot already proved dirfd is open, so a valid-but-wrong-type fd
     * (pipe, socket, epoll, ...) belongs here, not in the "bad fd" case: Linux
     * resolves a relative path against such a dirfd with ENOTDIR, reserving
     * EBADF for a closed or out-of-range descriptor.
     */
    if (snap.type != FD_DIR) {
        errno = ENOTDIR;
        return -1;
    }
```

That reasoning is right for a *name*, but the empty path is not a lookup, so the
whole translation is a false negative here — its result is not even used by the
`AT_EMPTY_PATH` branch, which only ever `fstat`s the descriptor.

Note the failing branch is gated on `proc_get_sysroot()`, so this appears to
need `--sysroot` (which is the normal way to run a distro rootfs).

`sys_fchmodat()` and `sys_fchownat()` already handle this correctly — both
short-circuit `AT_EMPTY_PATH` + empty path *before* translation, with a comment
explaining why. `stat_at_path()` is the odd one out. Since `sys_statx()` shares
`stat_at_path()`, `statx(fd, "", AT_EMPTY_PATH, ...)` has the same bug.

## Impact

Anything that `fstat`s a non-directory descriptor with no host path, on a sysroot
whose glibc is in the range that spells `fstat()` as `fstatat(..., AT_EMPTY_PATH)`
(2.33 through at least 2.36; 2.43 issues `__NR_fstat` again). In practice:
`cat` into a pipe fails, which breaks `dpkg-deb --fsys-tarfile | dpkg` style
package unpacking, and anything else that stats stdout/stdin when it is a pipe.

## Suggested fix

Answer `AT_EMPTY_PATH` + empty path from the descriptor before any path
translation, the way `sys_fchmodat`/`sys_fchownat` do. FUSE descriptors go
through the fd-based shim (as `sys_fstat` does), O_PATH descriptors through the
`/proc` stat intercept, everything else through `fstat()`. `AT_FDCWD` keeps the
existing behavior — the cwd always has a base path, so it was never affected.

Patch below is against `b1aac9a`; it is `clang-format` clean. I verified it by
hand on a Debian 12 aarch64 sysroot (the reproduction above passes, `stat -c %F
/proc/self/fd/1` reports `fifo`, and a full `dpkg -i` / `apt-get install` run
completes), but I could not run the guest test lanes — they need the aarch64
cross toolchain at `/opt/toolchain`, which I do not have set up. Happy to open it
as a PR if you would like it in that form.

```diff
--- a/src/syscall/fs-stat.c
+++ b/src/syscall/fs-stat.c
@@ -162,6 +162,52 @@ static void translate_statfs(const struct statfs *mac, linux_statfs_t *lin)
     lin->f_frsize = mac->f_bsize;
 }
 
+/* AT_EMPTY_PATH with an empty path names the descriptor itself, so nothing is
+ * resolved against it. That has to be answered before path_translate_at runs:
+ * the resolver measures a relative name against dirfd and owes ENOTDIR for a
+ * descriptor that is not a directory, which is right for a name but wrong for
+ * the empty path -- it fails fstat() on a pipe, a socket, or any other fd with
+ * no host path behind it, since glibc since 2.33 spells fstat() as
+ * fstatat(fd, "", AT_EMPTY_PATH). sys_fchmodat and sys_fchownat short-circuit
+ * the same way and for the same reason.
+ *
+ * FUSE descriptors go through the fd-based shim rather than a path, matching
+ * sys_fstat: the emulation layer, not the host file, is what answers for them.
+ *
+ * Returns 0 on success or a negative Linux errno.
+ */
+static int64_t stat_empty_path_fd(int dirfd, struct stat *mac_st)
+{
+    int frc = fuse_fstat_fd(dirfd, mac_st);
+    if (frc == 0)
+        return 0;
+    if (frc != -LINUX_EBADF)
+        return frc;
+
+    fd_entry_t snap;
+    host_fd_ref_t ref = {.fd = fd_snapshot_and_dup(dirfd, &snap),
+                         .owned = true};
+    if (ref.fd < 0)
+        return -LINUX_EBADF;
+
+    int64_t rc = 0;
+    if (snap.type == FD_PATH && snap.proc_path[0] != '\0') {
+        int intercepted = proc_intercept_stat(snap.proc_path, mac_st);
+        if (intercepted == 0)
+            goto done;
+        if (intercepted == -1) {
+            rc = linux_errno();
+            goto done;
+        }
+    }
+    if (fstat(ref.fd, mac_st) < 0)
+        rc = linux_errno();
+
+done:
+    host_fd_ref_close(&ref);
+    return rc;
+}
+
 /* Resolve the directory + path arguments of a *at-style stat operation and fill
  * *mac_st via the appropriate host call (proc intercept where applicable).
  * Shared by sys_newfstatat and sys_statx; the caller copies the result into the
@@ -192,6 +238,10 @@ static int64_t stat_at_path(guest_t *g,
                         sizeof(path), &pathp) < 0)
         return -LINUX_EFAULT;
 
+    if ((flags & LINUX_AT_EMPTY_PATH) && pathp[0] == '\0' &&
+        dirfd != LINUX_AT_FDCWD)
+        return stat_empty_path_fd(dirfd, mac_st);
+
     if (pathp[0] == '/' && fuse_path_matches_mount(pathp)) {
         int frc = fuse_stat_path(pathp, mac_st, flags);
         if (frc < 0)
@@ -224,36 +274,14 @@ static int64_t stat_at_path(guest_t *g,
     host_fd_ref_t dir_ref = {.fd = -1, .owned = false};
     if ((flags & LINUX_AT_EMPTY_PATH) && pathp[0] == '\0') {
         /* Linux: AT_EMPTY_PATH with dirfd == AT_FDCWD operates on the current
-         * working directory.
+         * working directory. Every other descriptor was already answered by
+         * stat_empty_path_fd() above.
          */
-        if (dirfd == LINUX_AT_FDCWD) {
-            dir_ref.fd = AT_FDCWD;
-            int mac_flags = translate_at_flags(flags);
-            if (fstatat(AT_FDCWD, ".", mac_st, mac_flags) < 0) {
-                rc = linux_errno();
-                goto done;
-            }
-        } else {
-            fd_entry_t snap;
-            dir_ref.fd = fd_snapshot_and_dup(dirfd, &snap);
-            dir_ref.owned = true;
-            if (dir_ref.fd < 0) {
-                rc = -LINUX_EBADF;
-                goto done;
-            }
-            if (snap.type == FD_PATH && snap.proc_path[0] != '\0') {
-                int intercepted = proc_intercept_stat(snap.proc_path, mac_st);
-                if (intercepted == 0)
-                    goto done;
-                if (intercepted == -1) {
-                    rc = linux_errno();
-                    goto done;
-                }
-            }
-            if (fstat(dir_ref.fd, mac_st) < 0) {
-                rc = linux_errno();
-                goto done;
-            }
+        dir_ref.fd = AT_FDCWD;
+        int mac_flags = translate_at_flags(flags);
+        if (fstatat(AT_FDCWD, ".", mac_st, mac_flags) < 0) {
+            rc = linux_errno();
+            goto done;
         }
     } else {
         if (host_dirfd_ref_open(dirfd, &dir_ref) < 0)
```

## Suggested regression test

A guest test in the shape of `tests/test-fchmodat-empty-path.c` would cover it:
for a pipe read end, a pipe write end, a socketpair fd and an eventfd, assert
that `fstatat(fd, "", &st, AT_EMPTY_PATH)` succeeds and reports the right
`S_IFMT`, and that plain `fstat(fd)` agrees. Also assert `statx(fd, "",
AT_EMPTY_PATH, ...)` succeeds on the same descriptors, since it shares the code
path. Worth running under `--sysroot`, since the failing branch is gated on a
sysroot being active.

alanhc:
Closes #322.

**Credit where it is due:** the diagnosis, the root cause and the shape of this
fix are @doanbaotrung's, from the report in #322, including the observation
that `sys_fchmodat`/`sys_fchownat` already get this right and `stat_at_path` is
the odd one out. The patch in that issue is written against `b1aac9a` and
predates #324's move from dup to pin, so it no longer applies; this is the same
fix rebuilt on `fd_host_ref_acquire`, plus the regression test the report
proposed and the verification it says it could not run (no cross toolchain).
@doanbaotrung reported this and may have a patch in flight; this one is
redundant if so.

## The bug

glibc has spelled `fstat(fd)` as `fstatat(fd, "", AT_EMPTY_PATH)` since 2.33, so
an ordinary `fstat()` of a pipe reaches `stat_at_path` with an empty path. The
empty path is relative, and `stat_at_path` resolved it before it looked at the
flag: `path_translate_at` measures a relative name against `dirfd`, and a
descriptor that is not a directory owes ENOTDIR for a name. Correct for a name,
wrong for the empty path, which names the descriptor and resolves nothing.

`sys_statx` shares `stat_at_path`, so it failed identically.

## Reproduction

It needs no distro rootfs and no glibc of a particular vintage: an empty
directory passed to `--sysroot` is enough, because the gate is only that a
sysroot is active. A guest program that opens a pipe, a socketpair and an
eventfd and calls `fstatat(fd, "", &st, AT_EMPTY_PATH)` on each:

```
### before, --sysroot <empty dir> ###
  pipe-read    fstatat(fd,"",AT_EMPTY_PATH) rc=-1 errno=20(Not a directory)
  pipe-read    fstat(fd)                    rc=0  mode=10000
  socketpair   fstatat(fd,"",AT_EMPTY_PATH) rc=-1 errno=20(Not a directory)
  eventfd      fstatat(fd,"",AT_EMPTY_PATH) rc=-1 errno=20(Not a directory)
  dirfd        fstatat(fd,"",AT_EMPTY_PATH) rc=0  mode=40000
```

Without `--sysroot` every line already returned 0, which matches the report.

## The fix

The empty path is answered from the descriptor before anything tries to resolve
it. The new `stat_empty_path_fd` is the branch that already existed for this
case, moved in front of the resolver rather than left behind it, so the FUSE
shim, the `/proc` intercept for O_PATH and the plain `fstat` keep their order,
the order `sys_fstat` uses, since it answers the same question. AT_FDCWD stays
where it was: the cwd always has a base path, so it was never affected, and the
branch that served it is now the only thing left there.

## Test

`tests/test-fstatat-empty-path.c`, in the shape of `test-fchmodat-empty-path.c`
as the report suggested. Every case asserts the `fstatat` answer against what
plain `fstat()` returns for the same descriptor, since the two are the same
question and the divergence is the bug: pipe read end, pipe write end,
socketpair, eventfd, plus a regular fd, an O_PATH fd and a directory fd for the
paths this change moves; `statx` over the same set; AT_FDCWD; and EBADF for a
closed fd.

| | pass | fail |
|---|---|---|
| unfixed, `--sysroot` | 8 | **7**, every one ENOTDIR |
| fixed, `--sysroot` | 15 | 0 |
| fixed, no sysroot | 15 | 0 |

Registered in two places, deliberately: `tests/test-matrix.sh` so a real kernel
adjudicates the semantics, and a `make check` lane in `mk/tests.mk` that runs it
under `--sysroot`, since without one the empty path never reaches the code that
used to fail it and the test would pass without proving anything.

## What I ran

`make test-fstatat-empty-path`, `bash tests/driver.sh -e build/elfuse -d build`
(97 passed, 1 failed, `test-hello`, missing binary: this machine has no
`aarch64-none-elf-as`, and it fails the same way on an unmodified tree),
`test-pty`, `test-file-ops`, `test-proc`, `test-fd-family`, `test-chown-overlay`,
and the `test-sysroot-nofollow`, `test-sysroot-path-matrix`,
`test-sysroot-case-exact`, `test-sysroot-dotdot` and `test-usb-sysfs-sysroot`
lanes, all green. `make check-format`, `make check-ascii`,
`make check-skill-refs` and `.ci/check-matrix-lists.sh` pass. Full `make check`
stops at `build/test-hello` for the toolchain reason above, so everything past
that is on CI.

## Two things left deliberately undone

- **`EXPECTED_BASELINES` not raised.** This adds one registered test to both
  aarch64 lanes, but I cannot run the matrix here (no fixtures), and the comment
  above that table is explicit that a floor too low costs nothing while an
  unobserved one asserts a run that did not happen.
- **An adjacent deviation, not fixed here.** Under `--sysroot`,
  `fstatat(pipe_fd, "", &st, 0)` (no AT_EMPTY_PATH) returns ENOTDIR, where
  Linux returns ENOENT: `getname_flags()` rejects an empty name before
  `path_init()` ever reaches the `d_can_lookup()` check, and that check is
  gated on `*s` being non-empty anyway. Same resolver, different question, and
  the closed-fd case next to it (EBADF here, ENOENT on Linux by the same
  reasoning) would need deciding with it. I could not test either against a real
  kernel from this machine, so the test asserts only that the call fails without
  the flag rather than asserting an errno I have not measured. Filed separately
  as #354, with the measurements and the kernel reading behind them.

### DIFF
--- a/src/syscall/fs-stat.c
+++ b/src/syscall/fs-stat.c
@@ -162,6 +162,90 @@ static void translate_statfs(const struct statfs *mac, linux_statfs_t *lin)
     lin->f_frsize = mac->f_bsize;
 }
 
+/* Stat the descriptor an *at-style call names with AT_EMPTY_PATH and an empty
+ * path.
+ *
+ * The empty path names the descriptor rather than anything beneath it, so it
+ * resolves nothing and must be answered before a resolver ever sees it. A
+ * resolver measures a relative name against dirfd, and the empty path is
+ * relative, so it owes ENOTDIR for a dirfd that is not a directory -- correct
+ * for a name, wrong for the descriptor itself, and glibc has spelled fstat(fd)
+ * as fstatat(fd, "", AT_EMPTY_PATH) since 2.33. sys_fchmodat and sys_fchownat
+ * short-circuit ahead of translation for the same reason.
+ *
+ * The order mirrors sys_fstat, since this answers the same question: the FUSE
+ * shim first, because a FUSE descriptor is answered by the emulation layer
+ * rather than by a host file; then the /proc intercept, which an O_PATH
+ * descriptor needs because its host fd cannot be stat'ed for the emulated
+ * object; then the host.
+ *
+ * sys_fstat calls this rather than keeping a body of its own, since glibc's
+ * spelling makes this the form most fstat() calls arrive in and there is no
+ * second policy worth having. The classification and the descriptor come from
+ * one fd_lock window here, so a sibling thread cannot close and reuse the slot
+ * between the two: answering the /proc intercept for one object and then
+ * stat'ing another is what separate reads allow.
+ *
+ * AT_FDCWD is not handled here -- it names the current directory, which always
+ * has a base path to resolve against, so the caller keeps answering it the way
+ * Linux specifies it.
+ *
+ * Returns 0 on success or a negative Linux errno.
+ */
+static int64_t stat_empty_path_fd(int dirfd, struct stat *mac_st)
+{
+    int frc = fuse_fstat_fd(dirfd, mac_st);
+    if (frc == 0)
+        return 0;
+    if (frc != -LINUX_EBADF)
+        return frc;
+
+    fd_entry_t snap;
+    host_fd_ref_t ref = HOST_FD_REF_INIT;
+    if (thread_is_single_active()) {
+        /* No sibling can mutate the slot, so the pin has nothing to defend
+         * against and the snapshot alone is already consistent -- the rule
+         * fd_block_state states for the same reason. A closed slot, or one
+         * carrying no host descriptor, is the EBADF that fd_to_host would have
+         * reported on this path before.
+         */
+        if (!fd_snapshot(dirfd, &snap) || snap.host_fd < 0)
+            return -LINUX_EBADF;
+        ref.fd = snap.host_fd;
+    } else {
+        /* Pin, not dup: the descriptor has to stay valid across the stat, not
+         * become private, and retiring a dup would drop the guest's record
+         * locks on the file (fcntl(2)).
+         */
+        if (fd_host_ref_acquire(dirfd, &snap, &ref.lifetime) < 0)
+            return linux_errno(); /* EBADF or ENOMEM; the helper picks. */
+        ref.fd = snap.host_fd;
+    }
+
+    int64_t rc = 0;
+    if (snap.type == FD_PATH && snap.proc_path[0] != '\0') {
+        /* The descriptor already names one object: an O_PATH open that followed
+         * refers to the target, one made with O_NOFOLLOW to the link itself, so
+         * the open's flag is what selects here.
+         */
+        bool follow = !(snap.linux_flags & LINUX_O_NOFOLLOW);
+        int intercepted =
+            proc_intercept_stat_at(snap.proc_path, mac_st, follow);
+        if (intercepted == 0)
+            goto done;
+        if (intercepted == -1) {
+            rc = linux_errno();
+            goto done;
+        }
+    }
+    if (fstat(ref.fd, mac_st) < 0)
+        rc = linux_errno();
+
+done:
+    host_fd_ref_close(&ref);
+    return rc;
+}
+
 /* Resolve the directory + path arguments of a *at-style stat operation and fill
  * *mac_st via the appropriate host call (proc intercept where applicable).
  * Shared by sys_newfstatat and sys_statx; the caller copies the result into the
@@ -191,6 +275,10 @@ static int64_t stat_at_path(guest_t *g,
     if (guest_read_path(g, path_gva, short_path, sizeof(short_path), path,
                         sizeof(path), &pathp) < 0)
         return -LINUX_EFAULT;
+
+    if ((flags & LINUX_AT_EMPTY_PATH) && pathp[0] == '\0' &&
+        dirfd != LINUX_AT_FDCWD)
+        return stat_empty_path_fd(dirfd, mac_st);
 
     if (pathp[0] == '/' && fuse_path_matches_mount(pathp)) {
         int frc = fuse_stat_path(pathp, mac_st, flags);
@@ -224,46 +312,15 @@ static int64_t stat_at_path(guest_t *g,
     host_fd_ref_t dir_ref = HOST_FD_REF_INIT;
     if ((flags & LINUX_AT_EMPTY_PATH) && pathp[0] == '\0') {
         /* Linux: AT_EMPTY_PATH with dirfd == AT_FDCWD operates on the current
-         * working directory.
+         * working directory. Every other descriptor was answered by
+         * stat_empty_path_fd above, before anything tried to resolve the empty
+         * path against it.
          */
-        if (dirfd == LINUX_AT_FDCWD) {
-            dir_ref.fd = AT_FDCWD;
-            int mac_flags = translate_at_flags(flags);
-            if (fstatat(AT_FDCWD, ".", mac_st, mac_flags) < 0) {
-                rc = linux_errno();
-                goto done;
-            }
-        } else {
-            /* Pin, not dup: fstatat needs the descriptor to stay valid, not to
-             * be private, and retiring a dup would drop the guest's record
-             * locks on the file (fcntl(2)).
-             */
-            fd_entry_t snap;
-            if (fd_host_ref_acquire(dirfd, &snap, &dir_ref.lifetime) < 0) {
-                /* EBADF or ENOMEM; the helper distinguishes them. */
-                rc = linux_errno();
-                goto done;
-            }
-            dir_ref.fd = snap.host_fd;
-            if (snap.type == FD_PATH && snap.proc_path[0] != '\0') {
-                /* The descriptor already names one object: an O_PATH open that
-                 * followed refers to the target, one made with O_NOFOLLOW to
-                 * the link itself, so the open's flag is what selects here.
-                 */
-                bool follow = !(snap.linux_flags & LINUX_O_NOFOLLOW);
-                int intercepted =
-                    proc_intercept_stat_at(snap.proc_path, mac_st, follow);
-                if (intercepted == 0)
-                    goto done;
-                if (intercepted == -1) {
-                    rc = linux_errno();
-                    goto done;
-                }
-            }
-            if (fstat(dir_ref.fd, mac_st) < 0) {
-                rc = linux_errno();
-                goto done;
-            }
+        dir_ref.fd = AT_FDCWD;
+        int mac_flags = translate_at_flags(flags);
+        if (fstatat(AT_FDCWD, ".", mac_st, mac_flags) < 0) {
+            rc = linux_errno();
+            goto done;
         }
     } else {
         int64_t ref_err = host_dirfd_ref_open(dirfd, &dir_ref);
@@ -305,52 +362,20 @@ int64_t sys_fstat(guest_t *g, int fd, uint64_t stat_gva)
      * from it.
      */
     struct stat mac_st = {0};
-    int frc = fuse_fstat_fd(fd, &mac_st);
-    if (frc == 0) {
-        if (write_linux_stat(g, stat_gva, &mac_st) < 0)
-            return -LINUX_EFAULT;
-        return 0;
-    }
-    if (frc != -LINUX_EBADF)
-        return frc;
 
-    fd_entry_t snap;
-    if (fd_snapshot(fd, &snap) && snap.type == FD_PATH &&
-        snap.proc_path[0] != '\0') {
-        /* As in stat_at_path's AT_EMPTY_PATH branch: the open already chose
-         * between the link and its target, so replay that choice here.
-         */
-        bool follow = !(snap.linux_flags & LINUX_O_NOFOLLOW);
-        int intercepted =
-            proc_intercept_stat_at(snap.proc_path, &mac_st, follow);
-        if (intercepted == 0) {
-            if (write_linux_stat(g, stat_gva, &mac_st) < 0)
-                return -LINUX_EFAULT;
-            return 0;
-        }
-        if (intercepted == -1)
-            return linux_errno();
+    /* fstat(fd) and fstatat(fd, "", AT_EMPTY_PATH) are the same question --
+     * glibc has spelled the first as the second since 2.33 -- so they are
+     * answered by one body rather than two that can drift.
+     */
+    int64_t rc = stat_empty_path_fd(fd, &mac_st);
+    if (rc < 0) {
+        log_debug("fstat(%d): failed rc=%lld", fd, (long long) rc);
+        return rc;
     }
 
-    host_fd_ref_t host_ref;
-    int64_t ref_err = host_fd_ref_open(fd, &host_ref);
-    if (ref_err < 0) {
-        log_debug("fstat(%d): invalid guest fd", fd);
-        return ref_err;
-    }
-    if (fstat(host_ref.fd, &mac_st) < 0) {
-        log_debug("fstat(%d->%d): host fstat failed errno=%d", fd, host_ref.fd,
-                  errno);
-        host_fd_ref_close(&host_ref);
-        return linux_errno();
-    }
-
-    if (write_linux_stat(g, stat_gva, &mac_st) < 0) {
-        host_fd_ref_close(&host_ref);
+    if (write_linux_stat(g, stat_gva, &mac_st) < 0)
         return -LINUX_EFAULT;
-    }
 
-    host_fd_ref_close(&host_ref);
     return 0;
 }
 
--- a/tests/test-fstatat-empty-path.c
+++ b/tests/test-fstatat-empty-path.c
@@ -1,0 +1,295 @@
+/*
+ * fstatat/statx AT_EMPTY_PATH tests
+ *
+ * Copyright 2026 elfuse contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * fstatat(fd, "", &st, AT_EMPTY_PATH) stats fd itself rather than a name
+ * beneath it, so it has to work for descriptors that name no path at all: a
+ * pipe, a socket, an eventfd. glibc has spelled fstat(fd) exactly that way
+ * since 2.33, which is what makes this ordinary rather than exotic -- a program
+ * that never mentions AT_EMPTY_PATH still arrives here through plain fstat().
+ *
+ * The interesting failure is a resolver reached before the empty path is
+ * recognized: the empty path is relative, so measuring it against dirfd owes
+ * ENOTDIR for a dirfd that is not a directory, which is right for a name and
+ * wrong for the descriptor itself. Every case below therefore asserts the
+ * fstatat answer against the plain fstat answer for the same descriptor, since
+ * the two are the same question and a divergence is the bug.
+ *
+ * statx(2) shares the same resolution path and is checked on the same
+ * descriptors. AT_FDCWD is checked too: it names the current directory, which
+ * always has a path, so it is the case that keeps working when the rest breaks.
+ */
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/eventfd.h>
+#include <sys/socket.h>
+#include <sys/stat.h>
+#include <sys/syscall.h>
+#include <sys/sysmacros.h>
+#include <unistd.h>
+
+#include "test-harness.h"
+
+#ifndef O_PATH
+#define O_PATH 010000000
+#endif
+#ifndef AT_EMPTY_PATH
+#define AT_EMPTY_PATH 0x1000
+#endif
+#ifndef SYS_statx
+#define SYS_statx 291
+#endif
+
+/* STATX_BASIC_STATS. Spelled out because the mask is what the guest sends and
+ * the header's name for it is not present on every libc this cross-builds
+ * against.
+ */
+#define STATX_MASK_BASIC 0x7ff
+
+int passes = 0, fails = 0;
+
+static char tmp_file[] = "/tmp/elfuse-fstatat-empty-XXXXXX";
+
+/* Both spellings of the same question must return the same answer. st_dev and
+ * st_ino identify the object, st_mode carries the type the caller switches on,
+ * and nothing else is stable enough across descriptor kinds to compare.
+ */
+static void expect_agrees_with_fstat(const char *what, int fd, mode_t want_fmt)
+{
+    TEST(what);
+
+    struct stat direct;
+    memset(&direct, 0, sizeof(direct));
+    if (fstat(fd, &direct) != 0) {
+        FAIL("fstat failed, nothing to compare against");
+        return;
+    }
+
+    struct stat viaat;
+    memset(&viaat, 0, sizeof(viaat));
+    errno = 0;
+    if (fstatat(fd, "", &viaat, AT_EMPTY_PATH) != 0) {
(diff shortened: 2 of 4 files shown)


src/syscall/fs-stat.c
```diff
@@ -162,6 +162,71 @@ static void translate_statfs(const struct statfs *mac, linux_statfs_t *lin)
...
+ * descriptor needs because its host fd cannot be stat'ed for the emulated
+ * object; then the host.
+ *
+ * AT_FDCWD is not handled here -- it names the current directory, which always
+ * has a base path to resolve against, so the caller keeps answering it the way
+ * Linux specifies it.
+ *
+ * Returns 0 on success or a negative Linux errno.
+ */
+static int64_t stat_empty_path_fd(int dirfd, struct stat *mac_st)
```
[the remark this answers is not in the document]
alanhc: Half of this is right and is fixed in 28f34d6. The other half stays as it is.

**The comment drift was real, and my fault.** `sys_fstat` said "As in `stat_at_path`'s AT_EMPTY_PATH branch", but this change is exactly what moved that logic out of `stat_at_path` and into `stat_empty_path_fd`, so the pointer went stale in the same commit that created the thing it should point at. It now names the helper, and `stat_empty_path_fd` says in so many words that the rule is written out there and pointed at from here, so there is one copy of the reasoning rather than two drifting ones. (I checked the other two cross-references while I was there: `fs.c:3100` and `fs.c:3258` both still describe `stat_at_path` accurately after the move, since the AT_FDCWD case they refer to stayed put.)

**Sharing the body is a different matter.** The two do not differ only in "pin-vs-open acquisition and write-timing". The acquisition *is* the reason they are separate:

- `sys_fstat` goes through `host_fd_ref_open`, which has a `thread_is_single_active()` fast path (`internal.h:869`) that returns a plain `fd_to_host` result and never touches `fd_lock`.
- `stat_empty_path_fd` needs `snap.proc_path` and `snap.linux_flags` for the O_PATH branch from the *same* window that pins the descriptor, so it has to call `fd_host_ref_acquire` directly. That one takes `pthread_mutex_lock(&fd_lock)` unconditionally (`fdtable.c:1059`) plus a lifetime pin and release.

Routing `sys_fstat` through the shared helper would therefore put a table-lock acquire and a refcount pair on every `fstat()` in a single-threaded guest, where today it is an array read. `fstat` is not a rare call: the syscall ranking in #292 puts it tenth across five real workloads, 5,330 calls, 3,012 of them in the C-compile one. I am not willing to pay that for fifteen lines.

The alternative (one body parameterized by an acquisition strategy) needs a callback or a mode flag to express a difference that two call sites state plainly today, and it would hide the very thing a reader of either function needs to know. `host_fd_ref_open_state` would have been the natural shared vehicle, but it hands back `fd_block_state_t`, which carries neither `proc_path` nor the open flags, so it cannot serve the /proc branch.

I have written that rationale into the helper's comment so the next person reading it does not have to re-derive it.

tests/test-fstatat-empty-path.c
```diff
@@ -0,0 +1,287 @@
...
+#include "test-harness.h"
+
+#ifndef O_PATH
+#define O_PATH 010000000
+#endif
+#ifndef AT_EMPTY_PATH
+#define AT_EMPTY_PATH 0x1000
+#endif
+#ifndef SYS_statx
+#define SYS_statx 291
```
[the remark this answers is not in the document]
alanhc: Not fixing this one, because the situation it describes cannot arise in this tree today, and the guard would be the kind of dead config that reads as deliberate policy while protecting nothing.

Two things have to be true for the wrong constant to be reached: a guest test built for x86_64, and a libc whose headers do not define `SYS_statx`. Neither holds.

**There is no x86_64 guest test build.** Every `tests/*.c` binary comes from one pattern rule, `$(BUILD_DIR)/%: tests/%.c`, which invokes `$(CROSS_COMPILE)gcc`, and `CROSS_COMPILE` resolves to `aarch64-linux-gnu-` (`mk/toolchain.mk:34`). `grep -n x86_64 Makefile` returns nothing. The `elfuse-x86_64:*` rows in `EXPECTED_BASELINES` are the Rosetta sub-suites, which run prebuilt x86_64 fixtures, not these cross-built tests.

**The fallback is already unreachable on the arch that is built.** `#ifndef SYS_statx` only fires for a libc that omits the define, and the toolchain in use defines it, which is why the statx cases in this test pass rather than returning ENOSYS.

**And the value is the established spelling here.** Three existing files carry the identical two lines: `tests/test-pty.c:84`, `tests/test-case-collision.c:47`, `tests/test-file-ops.c:26`. Guarding the arch in the one new file would leave four files disagreeing about the same constant, which is a worse state than the one it fixes.

If x86_64 guest tests ever do land, all four files need this at once, and the right shape then is a shared `tests/` header holding the syscall numbers once, not an `#if` ladder repeated four times. That is a follow-up; this PR does not encode an architecture the build cannot produce.

src/syscall/fs-stat.c
```diff
@@ -162,6 +162,79 @@ static void translate_statfs(const struct statfs *mac, linux_statfs_t *lin)
...
+ * as fstatat(fd, "", AT_EMPTY_PATH) since 2.33. sys_fchmodat and sys_fchownat
+ * short-circuit ahead of translation for the same reason.
+ *
+ * The order mirrors sys_fstat, since this answers the same question: the FUSE
+ * shim first, because a FUSE descriptor is answered by the emulation layer
+ * rather than by a host file; then the /proc intercept, which an O_PATH
+ * descriptor needs because its host fd cannot be stat'ed for the emulated
+ * object; then the host.
+ *
+ * The two are deliberately not one body. sys_fstat reaches its descriptor
```
jserv: This paragraph argues acquisition cost, but the two bodies also differ in correctness, and the patch puts the safe one on the hot path.

`sys_fstat` reads `fd_snapshot` unpinned (line 364) and then pins separately with `host_fd_ref_open`. In a multithreaded guest the fd can be closed and reused between those two reads, so the /proc intercept can answer for the old `proc_path` while the `fstat` lands on the new object. `stat_empty_path_fd` takes the snapshot and the pin in one `fd_lock` window and has no such gap.

Since glibc 2.33 spells `fstat(fd)` as the AT_EMPTY_PATH form, the atomic body is now the common one and the racy body is the leftover. Having `sys_fstat` call `stat_empty_path_fd(fd, &mac_st)` leaves one body and one policy. If the `fd_lock` hit matters, recover it inside that one helper with the `thread_is_single_active` fast path `host_fd_ref_open` already uses, rather than by keeping two versions apart.

src/syscall/fs-stat.c
```diff
@@ -192,6 +265,10 @@ static int64_t stat_at_path(guest_t *g,
                         sizeof(path), &pathp) < 0)
         return -LINUX_EFAULT;
 
+    if ((flags & LINUX_AT_EMPTY_PATH) && pathp[0] == '\0' &&
```
jserv, line 279: One more gap in the same family as the one you filed as #354. `statx(2)` accepts a NULL pathname when AT_EMPTY_PATH is set, since 6.11, and `sys.h` advertises 6.18.0. `guest_read_path` four lines up returns EFAULT for `path_gva == 0`, so this branch never sees that spelling. `fstatat` still owes EFAULT for NULL, so the check has to know which caller it is serving and does not belong inline here. Worth folding into #354 rather than growing this PR.

tests/test-fstatat-empty-path.c
```diff
@@ -0,0 +1,287 @@
...
+
+    struct statx sx;
+    memset(&sx, 0, sizeof(sx));
+    errno = 0;
+    long rc = syscall(SYS_statx, fd, "", AT_EMPTY_PATH, STATX_MASK_BASIC, &sx);
+    if (rc != 0) {
+        FAIL("statx with AT_EMPTY_PATH failed");
+        return;
+    }
+    EXPECT_TRUE((sx.stx_mode & S_IFMT) == (direct.st_mode & S_IFMT),
```
jserv: `expect_agrees_with_fstat` compares mode, dev and ino; this one compares only `S_IFMT`. statx is the call that splits dev into major and minor, so an answer carrying the right type from the wrong object passes here. Tightening it costs two terms: `sx.stx_ino == direct.st_ino`, and `stx_dev_major`/`stx_dev_minor` against `major()`/`minor()` of `direct.st_dev`.

jserv (commented):
The diagnosis holds and the fix is in the right place. Things I checked and found sound, so they need no reply: reference lifetime in `stat_empty_path_fd` releases exactly once on every exit including both proc-intercept error paths; the leftover `AT_FDCWD` branch is still reachable and is now the only thing that reaches it; FUSE descriptors are an improvement rather than a regression, since the old path reached `fd_host_ref_acquire` with `host_fd == -1` and owed EBADF; dropping `AT_SYMLINK_NOFOLLOW` in the helper is right, because Linux ignores it once the path is empty and the `O_PATH` branch replays the open's own flag instead; `$(BUILD_DIR)/%: tests/%.c` builds the new binary, and the `make check` lane really does put a resolver in front of the empty path.

On `EXPECTED_BASELINES`: leaving it alone is the right call, and it is the call the comment above that table already makes. No need to bump it here.

Three comments inline.

src/syscall/fs-stat.c
alanhc: You are right, and my rationale was measured against the wrong baseline. Done in 27f0930: `sys_fstat` is now `stat_empty_path_fd(fd, &mac_st)` plus the guest write.

The correctness point is the one I missed entirely. `sys_fstat` read `fd_snapshot` at line 364 and pinned separately at line 383, so in a multithreaded guest a close-and-reuse between them let `proc_intercept_stat_at` answer for the old `proc_path` while the `fstat` landed on the new object. That is the gap #324 and #327 were closing elsewhere, still open on this path.

On cost, my objection does not survive checking. I had compared the helper against `host_fd_ref_open` alone and forgotten that `sys_fstat` already paid an unconditional `fd_snapshot` (which takes `fd_lock`) before ever reaching it. Counting acquisitions per call:

| | before | after |
|---|---|---|
| single active thread | `fd_snapshot` (1 lock) + `host_fd_ref_open` fast path (0) = **1** | `fd_snapshot` (1 lock) = **1** |
| siblings alive | `fd_snapshot` (1) + `fd_host_ref_acquire` (1 + pin) = **2 + pin** | `fd_host_ref_acquire` (1 + pin) = **1 + pin** |

So it is unchanged on the hot path and strictly cheaper under contention. I took the `thread_is_single_active` fast path you suggested anyway, since the pin defends against a mutator that cannot exist there; it reads the entry through `fd_snapshot` rather than `fd_to_host`, because the /proc branch needs `proc_path` and the open flags and not just the descriptor. A closed slot, or one with no host fd, is the same EBADF `fd_to_host` gave.

Verified: `tests/driver.sh` 97 passed (the 1 failure is `test-hello`, no `aarch64-none-elf-as` on this machine, same on an unmodified tree), plus `test-pty`, `test-file-ops`, `test-proc`, `test-fd-family`, `test-chown-overlay` and the fd-race set (`test-fd-race`, `test-fd-pin-lock`, `test-dup-setfl-race`, `test-mt-fork`, `test-thread-churn`), all green, and `make check-lock-order` and `make check-atomics` pass.

src/syscall/fs-stat.c
alanhc: Agreed, and folded into #354 rather than into this PR: nothing here changed for it.

Confirmed both halves before writing it up: `src/syscall/sys.h:24` advertises `GUEST_KERNEL_RELEASE "6.18.0"`, and `guest_read_path` reads the guest string at `path_gva` with no NULL case, so a `path_gva` of 0 is refused as EFAULT before the AT_EMPTY_PATH branch is reached. Same shape as the other two: a step that assumes there is a name running ahead of the rule that says there is not one.

Your point about it needing to know its caller is what makes it a poor fit for an inline check, and I recorded that as the reason: `statx` with AT_EMPTY_PATH takes NULL as empty since 6.11 via `vfs_empty_path`, while `fstatat` still owes EFAULT, and `sys_statx` and `sys_newfstatat` reach `stat_at_path` today with nothing to tell them apart. #354 now carries all three cases so they can be decided together.

tests/test-fstatat-empty-path.c
alanhc: Fixed in 27f0930, with both terms you named.

`expect_statx_agrees` now checks `sx.stx_ino == direct.st_ino` and `stx_dev_major`/`stx_dev_minor` against `major()`/`minor()` of `direct.st_dev`, alongside the S_IFMT comparison it already had (`<sys/sysmacros.h>` added for the accessors). The comment above it now says why identity and not just type: an answer carrying the right type from the wrong object is exactly what a resolver reached too early produces, which a type-only check waves through.

Still 15 passed, 0 failed with and without `--sysroot`, and still 7 of 15 failing on an unfixed tree.

jserv:
Thank @alanhc for contributing!

PR #360, opened 2026-09-03
Add a security skill for the guest boundary

jserv:
## Add a security skill for the guest boundary

The tree has skills for adding a syscall, the guest ABI, verification,
debugging, refactoring and conventions, but none saying what makes the guest
hostile or what counts as a security finding against it. Reviews kept
rediscovering the same out-of-scope categories, and generic checklists got
applied to a boundary they were not written for: there is no server, no
session and no browser here.

The skill states the threat model as a decision procedure. Five impacts
count: host state the guest was never handed, sysroot escape, host memory
corruption through a trusted length, host resource exhaustion, and host
privilege the launching user did not lend. What the guest does to itself is a
correctness bug at most, which is what stops a wedged vCPU being filed as a
vulnerability.

Each rule names what breaking it costs rather than restating mechanism a
sibling skill already owns. The gate inventory is split by the target that
runs it, because "make check", "make verify" and "make verify-mutants" do not
run the same set, and telling a reviewer to skip ground a gate covers is
unsafe when that gate did not run. Two limits are stated where they change
what review still owes: a contract proves the arithmetic and not its callers,
and check-atomics.py leaves plain-operator access to an _Atomic object to
review.

Two reference files carry what only one branch of the work reaches: the
finding schema, and the detectors a whole-tree sweep runs together with what
each one gets wrong in this tree.

## Match the lock-order rule to what the gate checks

The syscall skill said to add a new lock to the ordering block before using it
in a second module. check-lock-order.py fails any file-scope mutex or rwlock
the block does not name, whether one module uses it or ten.

## Give the getsockopt truncation one spelling

Four sites in sys_getsockopt clamped the reply to the caller buffer in three
different forms, one with the comparison inverted. ip_copyval_clamp had
already taken the IP-specific half out, so the extraction was half finished.

## Seed AT_RANDOM from a call that cannot fail

build_linux_stack drew the 16 AT_RANDOM bytes from getentropy and, on failure,
filled them with a constant, so every process in that run would have shared
one publicly known canary. arc4random_buf returns void, so the fallback branch
goes with it. Not a boundary finding: it weakens the guest against its own
memory-safety bugs and hands it nothing of the host.

### DIFF
--- a/.claude/skills/elfuse-security/SKILL.md
+++ b/.claude/skills/elfuse-security/SKILL.md
@@ -1,0 +1,310 @@
+---
+name: elfuse-security
+description: The guest as an attacker - where the trust boundary runs, the rules a handler on it obeys, what the gates already catch, and what is out of scope. Use when a change parses a guest-chosen length, translates a guest address, resolves a guest path, allocates on the guest's behalf, or blocks holding shared state, and when auditing a diff or writing a finding up.
+---
+
+# Security at the guest boundary
+
+The guest binary is untrusted, and it is not a remote attacker. It picks every
+register, every length, and the order the calls arrive in, at machine speed,
+with a debugger attached and as many retries as it likes. A window that needs
+a race to hit is a window it hits.
+
+That is the whole threat model, and it decides what counts as a finding here:
+the guest reading or writing host state it was never handed, escaping the
+sysroot, corrupting host memory through a length the host trusted, exhausting
+a host resource the guest does not own, or reaching a host privilege the
+launching user did not intend to lend it. What the guest does to itself,
+including wedging its own vCPU and exhausting its own address space, is a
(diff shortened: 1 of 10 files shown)


jserv:
Fixed in "Require a terminal fd for TIOCGSID".

The finding was correct and the guard was checking only process state. TIOCGSID
now also requires the descriptor to be a terminal, decided with tcgetattr, which
is how every termios case in the same switch establishes it. A pipe or a regular
file gets ENOTTY instead of the caller session id.

One case the fix does not cover, recorded in a comment at the site rather than
left silent: an fd that is a terminal but not this caller controlling one.
elfuse records whether a controlling terminal is held, never which device it is,
so there is nothing to compare the descriptor against. Closing that needs
proc_set_ctty to record the terminal identity, which is a larger change than
this pull request.

PR #362, opened 2026-09-03
Add an AGENTS.md that routes rather than restates

alanhc:
elfuse carries no file at the path an agent's harness opens first.
`CONTRIBUTING.md` is found by name, and the skills under `.claude/skills/`
are loaded by description rather than read in order, so an agent starting
cold chooses its own entry point.

`AGENTS.md` is that entry point, and it routes rather than restates: the
scope in a paragraph, `CONTRIBUTING.md` before any C, one line per skill for
what each covers, the gate commands, the two prose rules that bind a first
sentence, and the rule about other people's untracked working docs. A rule
copied into it is a rule that can drift, so it carries none of its own.
`CLAUDE.md` is a symlink to it, the way `.agents/skills` already symlinks
`.claude/skills`.

`scripts/check-skill-refs.py` already took a routing document as an argument,
so `make check-skill-refs` and the lint workflow name `AGENTS.md` now. Every
path, make target and skill name in it resolves under the gate a skill file
gets.

Verified on macOS 26.5.2 (25F84), SDK 26.5, Apple M1 Pro:
`make check-skill-refs` passes its 29 self-test cases and resolves 13 files,
and `scripts/check-commit-log.sh` passes over the commit. `make check` does
not run on this machine, which has no `aarch64-none-elf-as`; the change
touches no C, only one markdown file, one symlink, one make recipe, one
workflow line, and a comment.

The prose-register change that shared this branch is #363 now, since it is a
different subject from a routing file.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -177,7 +177,7 @@ jobs:
         # current. This found two dead references to a workflow and a makefile
         # that the CI split had already removed.
         if: ${{ !cancelled() }}
-        run: python3 scripts/check-skill-refs.py
+        run: python3 scripts/check-skill-refs.py AGENTS.md
 
       # What follows needs LINT_PKGS, and the three checks that use it read
       # only C sources, the style file, and the .ci scripts that implement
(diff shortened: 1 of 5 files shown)


jserv:
Don't do that. Both `AGENTS.md` and `CLAUDE.md` should be personal and never checked in repository. Consolidate existing skills instead.

PR #364, opened 2026-09-03
Remove duplicated contracts

jserv:
Behavior-preserving cleanup: each contract that was stated in two or more
places now has one owner. No guest-visible change intended.

Two test guards also stop reporting false failures. One awk exited at the
first match and killed its own pipeline under pipefail, which surfaced as
an intermittent 141 under load. The wget probe tested TCP reachability
rather than content, so a captive portal answering port 80 read as a
busybox failure.

Validation on macOS 26.6.2, Apple M1, 8 cores, SDK 26.5:

    make check                       rc=0, 283 pass
    make verify                      21/21 targets
    make verify-mutants              117/117 caught
    make test-matrix-elfuse-aarch64  279 pass, 0 fail
    make check-format                clean

Also ran all 202 guest test binaries under this branch and under
origin/main, comparing output and exit status. Twelve differ, and all
twelve are clock values, ephemeral ports, random bytes, race counters, or
first-run state.

Not addressed here: sys_sendmmsg's vlen == 1 fast path lacks the
FD_NETLINK check sys_sendmsg has, so sendmmsg on a netlink fd sends on the
descriptor instead of dispatching the request. It predates this branch and
recvmmsg has the same shape.

---
## Summary by cubic
Gives each duplicated contract a single owner, fixes the `sys_wait4` already-reaped path to return EFAULT when the guest rusage pointer can't be written, and stops two test guards from reporting false failures. Rewrites the `elfuse-refactor` skill so its lint-gate claims match `.clang-tidy`, and adds `references/measuring.md` with the scans behind reported numbers.

**Cleanup**
- Extracts shared helpers for the wait4 paths, netlink dump framing, file-overlay unmapping, single-iovec send loop, futex signal polling, transport opens, dirfd-pair opens, and USB intercept refusal, and drops the unused `emit_forget` parameter.
- Moves socket `MSG_*` flags, execve argument caps, the SysV message size ceiling, the inotify overflow event, and the capget version into named constants, and updates `check-eintr-contract.py` to track the moved wait site.

**Test fixes**
- The nameserver and help-synopsis awk pipelines now read to EOF instead of exiting early, which closed the pipe and failed under `pipefail`; the synopsis awk latches after the first blank line so a later `usage:` line cannot re-arm the block.
- The wget guard now fetches example.com and requires the same `Example Domain` marker the guest check uses, so captive portals and transparent proxies skip rather than fail.

<sup>Written for commit 3bcd93004037acae9737b2bf7905ad6b5bc5b099. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/364?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.claude/skills/elfuse-refactor/SKILL.md
+++ b/.claude/skills/elfuse-refactor/SKILL.md
@@ -1,251 +1,152 @@
 ---
 name: elfuse-refactor
-description: Behavior-preserving cleanup of elfuse code, calibrated for C-specific smells and false alarms. Use when simplifying, deduplicating, or restructuring working code; reviewing maintainability; or facing an over-long function, repeated cleanup, a 0/-1 helper used as a boolean, or prose that narrates code or asserts a number. Not for syscall implementation (elfuse-syscall), naming or commit style (elfuse-conventions), or live failures (elfuse-debug).
+description: Behavior-preserving cleanup of elfuse code. Use when simplifying, deduplicating, or restructuring working code; reviewing maintainability; or reducing prose that narrates code or makes unchecked claims. Not for syscall changes (elfuse-syscall), naming or commit style (elfuse-conventions), or a live failure (elfuse-debug).
 ---
 
 # Refactoring elfuse code
 
-Refactoring is behavior-preserving change, and in this tree the behavior that
-must be preserved is what the guest observes, not what the host code returns.
-That makes the two halves of this file: what is worth changing, and what has
-to stay green before the change counts.
+Refactoring preserves what the guest observes, not merely a host return value.
+Do not call a behavior change a cleanup to make it easier to review.
 
-## What the gates actually measure
+## Route the change first
 
-Worth knowing before treating a green build as a verdict. `.clang-tidy` enables
-`bugprone-*`, `cert-*`, `clang-analyzer-*`, and `performance-*`, which answer
-"is this a bug". The one structural check is `readability-function-size`, and
-it measures line count only: nesting, statement, branch, and parameter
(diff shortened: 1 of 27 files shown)


jserv:
All six addressed, in three commits.

Fixed as reported: the measuring reference defaulted to 60 lines while
its text described the 400-line ceiling, its duplication scan never
formed the last window, and its nesting grep matched exactly 20 leading
spaces rather than 20 or more. That last one was worse than described:
764 lines against 4196 for the corrected pattern on this tree.

The synopsis awk now latches, verified against a synthetic two-block
input where the unfixed version printed both. The wget guard now
requires "Example Domain".

On the proc.c finding: correct, and pre-existing rather than introduced
here. The same call was unchecked inline before the extraction, so it
was kept out of that commit and landed on its own as a behavior change.

Validation after the fixes: make check 283 pass, make verify 21/21,
make test-matrix-elfuse-aarch64 279 pass 0 fail, make check-format
clean.

One unexplained result, recorded rather than dismissed: test-thread-churn
timed out once at 60s during a make check run and did not recur in 60
further runs, 40 idle and 20 under load, each completing in under 0.4s.
It calls no wait4, rusage, waitpid or fork, so the proc.c change is not
on its path. Not understood, not attributed to this branch.

PR #368, opened 2026-09-04
Saturate timerfd timeout conversions

Suzu1Dev:
Large valid timerfd_settime requests can expire immediately or return
negative timerfd_gettime intervals because clamping seconds leaves the
nanosecond addition able to overflow. Reuse the existing saturating timespec
helper for the initial value, interval and absolute deadline, and derive the
host timeout from the same bounded value.

The regression includes tv_sec=INT64_MAX/1000000000, tv_nsec=999999999,
plus microsecond-boundary, maximum-seconds, interval and absolute-clock
cases. All six fail before the fix and pass afterward. The baseline UBSAN
build aborts on signed overflow.

- make elfuse and make check-format pass.
- make -j6 check passes, including the throughput guardrail and sharun.
- make test-matrix-elfuse-aarch64: 280 pass, 0 fail, 10 skip, compared
  with the baseline's 279 pass, 0 fail, 10 skip.
- Existing timerfd tests: 24/24; new tests: 6/6 in release and UBSAN builds.
- QEMU Linux reference: all six new cases pass.
- make verify-timespec: 45/45 proof obligations discharged with Frama-C
  33.0, Why3 1.8.2, Alt-Ergo 2.6.3 and Z3 4.16.0.
- make verify-mutants MUTANT_TARGET=timespec MUTANT_JOBS=3: all nine
  mutations caught. The existing helper and ACSL contracts are unchanged.

---
## Summary by cubic
Fixes timerfd timeout conversions so large valid `timerfd_settime` requests no longer expire immediately or return negative intervals, and applies the same saturating conversion to host clock reads.

- Clamping only seconds could overflow the nanosecond addition; the fix reuses the existing saturating `timespec` helper for the initial value, interval, absolute deadline, and host clock conversions.
- Adds seven regression tests covering nanosecond, microsecond, maximum-seconds, interval, absolute-clock, and rearmed-interval boundaries, validating remaining values against clock samples bracketing settime and gettime.

<sup>Written for commit 28388eda9355125f7b68c145316a42d36eb08a1c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/syscall/fd.c
+++ b/src/syscall/fd.c
@@ -28,6 +28,7 @@
 #include "debug/log.h"
 #include <sys/event.h>
 
+#include "proved/timespec.h"
 #include "syscall/linux-wire.h"
 #include "syscall/fd.h"
 #include "syscall/internal.h"
@@ -47,7 +48,10 @@ static void eventfd_close(int guest_fd);
 static void signalfd_close(int guest_fd);
 
 #define NS_PER_SEC 1000000000LL
-#define US_PER_SEC 1000000LL
+
+_Static_assert(
+    NS_PER_SEC == TIMESPEC_NSEC_PER_SEC,
+    "timerfd and proved timespec conversions must use the same scale");
 
 /* All special-FD state arrays store guest_fd as their first field. Keep the
  * common slot walk in one place so timerfd/eventfd/signalfd stay consistent.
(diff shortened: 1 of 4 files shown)


src/syscall/fd.c
```diff
@@ -293,26 +297,22 @@ int64_t sys_timerfd_settime(guest_t *g,
...
+     * bounded value reported by timerfd_gettime.
      */
-    int64_t val_sec = its.it_value_sec, int_sec = its.it_interval_sec;
-    if (val_sec > INT64_MAX / US_PER_SEC)
-        val_sec = INT64_MAX / US_PER_SEC;
-    if (int_sec > INT64_MAX / NS_PER_SEC)
-        int_sec = INT64_MAX / NS_PER_SEC;
-    int64_t value_us = val_sec * US_PER_SEC + its.it_value_nsec / 1000;
-    int64_t interval_ns = int_sec * NS_PER_SEC + its.it_interval_nsec;
+    int64_t value_ns = timespec_to_ns_sat(its.it_value_sec, its.it_value_nsec);
```
jserv, line 311: `timespec_to_ns_sat` saturates at `sec > TIMESPEC_SEC_MAX`; `ktime_set` saturates at `sec >= KTIME_SEC_MAX`, same 9223372036. For `{9223372036, 0}` Linux stores `KTIME_MAX` and `timerfd_gettime` reports `9223372036.854775807`, this reports `9223372036.000000000`. 0.85 s on a 292 year timer, so the behavior may not be worth changing, but the comment above claims the conversion matches `timespec64_to_ktime` and the thresholds differ. Either clamp `sec >= TIMESPEC_SEC_MAX` here or soften the claim.

src/syscall/fd.c
```diff
@@ -293,26 +297,22 @@ int64_t sys_timerfd_settime(guest_t *g,
                              ? CLOCK_REALTIME
                              : CLOCK_MONOTONIC;
         clock_gettime(host_clock, &now);
-        int64_t target_sec = its.it_value_sec;
-        if (target_sec > INT64_MAX / NS_PER_SEC)
-            target_sec = INT64_MAX / NS_PER_SEC;
-        int64_t target_ns = target_sec * NS_PER_SEC + its.it_value_nsec;
+        int64_t target_ns =
+            timespec_to_ns_sat(its.it_value_sec, its.it_value_nsec);
         int64_t now_ns = now.tv_sec * NS_PER_SEC + now.tv_nsec;
```
jserv: The guest value now saturates, but `now_ns` one line down still does the raw `tv_sec * NS_PER_SEC` product this patch exists to remove, as do lines 278, 361 and 385. `CLOCK_REALTIME` crosses `INT64_MAX / 1e9` in 2262, so no urgency, but `timespec_to_ns_sat` is already included in this file and each site is a one line change.

tests/test-timerfd-overflow.c
```diff
@@ -0,0 +1,89 @@
...
+        }
+        uint64_t count;
+        errno = 0;
+        ssize_t got = read(fd, &count, sizeof(count));
+        int read_errno = errno;
+        bool interval_ok =
+            !cases[i].interval ||
+            (current.it_interval.tv_sec == INT64_MAX / 1000000000 &&
+             current.it_interval.tv_nsec == INT64_MAX % 1000000000);
+        if (current.it_value.tv_sec > 30 && interval_ok && got == -1 &&
```
jserv: `> 30` only catches the value collapsing to zero. A truncation turning 9223372036 s into an hour still passes. The table already knows each case, so give it a per-case floor: 1e9 seconds for the five saturating cases (the ABSTIME ones land near 7.4e9 after subtracting realtime now), 30 for the interval case whose `it_value` is 60 s.

jserv:

PR #369, opened 2026-09-04
Drain buffered GDB packets before polling

Suzu1Dev:
The RSP transport can read several requests at once, but the session polls
the socket before consuming each request. A debugger waiting for replies
without sending more data leaves the buffered requests stalled. Consume
buffered input before polling again.

Reproduce by negotiating QStartNoAckMode, then sending eight framed
qAttached requests in one write. The regression requires all eight exact,
checksum-valid replies without another socket write. The subprocess exit
status also participates in the raw RSP test verdict.

- Focused baseline reproduction: 3/3 timeouts; fixed: 10/10 passes.
- make elfuse and make test-gdbstub: all 18 tests pass.
- Split-packet/shutdown probe: baseline and fixed each pass 3/3.
- make check-format passes.

---
## Summary by cubic
Fixes the GDB stub stalling when a single socket read buffers multiple RSP packets. Previously the session polled for new input before consuming each buffered request, so a client waiting for replies without sending more data could stall indefinitely. The session now drains buffered packets before polling again, with the buffer-exhaustion check shared between the session loop and transport decoder. A deterministic host regression forces two requests into one transport read, requires both exact, checksum-valid replies before any further client input, and runs in the shared and GDB check lanes.

<sup>Written for commit 6ad8748664ef2aa3de244d7c7ad30318e03e62b2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/369?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/debug/gdbstub.c
+++ b/src/debug/gdbstub.c
@@ -1146,7 +1146,7 @@ static void gdb_client_session(void)
         /* A socket read can buffer several packets. Consume them before waiting
          * for more input that the client may never send.
          */
-        if (rsp_ctx.read_pos >= rsp_ctx.read_len) {
+        if (!gdb_rsp_pending(&rsp_ctx)) {
             struct pollfd pfd = {
                 .fd = gdb.client_fd,
                 .events = POLLIN,
--- a/src/debug/gdbstub.c
+++ b/src/debug/gdbstub.c
@@ -1143,23 +1143,25 @@ static void gdb_client_session(void)
     gdb.rsp_ctx = &rsp_ctx;
 
     while (gdb.client_fd >= 0) {
-        /* Wait for either a packet from GDB or a stop event from a vCPU. Use
-         * poll() so the GDB stub can wake up when a thread stops.
+        /* A socket read can buffer several packets. Consume them before waiting
+         * for more input that the client may never send.
          */
-        struct pollfd pfd = {
-            .fd = gdb.client_fd,
-            .events = POLLIN,
-        };
+        if (rsp_ctx.read_pos >= rsp_ctx.read_len) {
+            struct pollfd pfd = {
+                .fd = gdb.client_fd,
+                .events = POLLIN,
+            };
 
-        int pr = poll(&pfd, 1, -1);
-        if (pr <= 0) {
-            if (pr < 0 && errno == EINTR)
-                continue;
-            break;
+            int pr = poll(&pfd, 1, -1);
+            if (pr <= 0) {
+                if (pr < 0 && errno == EINTR)
+                    continue;
+                break;
+            }
+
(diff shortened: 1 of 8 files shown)


src/debug/gdbstub.c
```diff
@@ -1143,23 +1143,25 @@ static void gdb_client_session(void)
...
-            .events = POLLIN,
-        };
-
-        int pr = poll(&pfd, 1, -1);
-        if (pr <= 0) {
-            if (pr < 0 && errno == EINTR)
-                continue;
-            break;
-        }
+        if (rsp_ctx.read_pos >= rsp_ctx.read_len) {
```
jserv: This restates the buffer test `gdb_rsp_recv` already does at `gdbstub-rsp.c:302`, and reaches into `gdb_rsp_ctx_t` from outside the transport. If the transport ever changes how it buffers, the stall comes back here with nothing to catch it. A `bool gdb_rsp_pending(const gdb_rsp_ctx_t *)` next to `gdb_rsp_reset` keeps the invariant in one file.

tests/test-gdbstub.sh
```diff
@@ -187,6 +191,8 @@ def recv_packet():
             break
         body.extend(ch)
     cksum = recv_exact(2).decode("ascii")
+    if int(cksum, 16) != sum(body) & 0xFF:
```
jserv: `int(cksum, 16)` accepts `+1` and ` 1`, so a malformed two byte trailer still passes this check. Compare the text instead:

```python
if cksum != f"{sum(body) & 0xFF:02x}":
```

tests/test-gdbstub.sh
```diff
@@ -207,9 +213,18 @@ print(f"stop_ack={ack3 or 'none'}")
 print(f"stop_body={body3}")
 print(f"stop_cksum={cksum3}")
 
+if sys.argv[2] == "batch":
+    # No further socket writes may wake the stub while replies are pending.
+    sock.sendall(frame_packet("qAttached") * 8)
```
jserv, line 219: Nothing forces the stub to coalesce these eight frames into one `read()`. If they arrive split, the unfixed loop polls, finds data, and answers all eight, so the guard passes against the bug it exists to catch. 112 bytes in one loopback write makes that unlikely, but the test does not say it is relying on that.

jserv:

PR #370, opened 2026-09-04
Avoid SIGPIPE in output pattern checks

Suzu1Dev:
A successful command with a match near the start of a large output can fail
run_check or run_pipe: grep exits early, printf receives SIGPIPE, and pipefail
reports a failed pattern check. Feed captured output through a here-string.

The regression emits a matching first line followed by 30,000 lines. Both
early-match cases fail before the fix; all eight cases pass afterward. Late
matches, missing patterns and nonzero command exits are also covered.

- make test-runner: 8 checks pass under Bash 3.2 and Bash 5.2.
- BusyBox integration using the changed runner: 80 pass, 0 fail, 4 skip.
- make check-format: passes, including 57 shell scripts.
- Baseline make check stopped at an UNMEASURED throughput guardrail; its
  separate rerun after parallel builds stopped passes. The baseline
  AArch64 matrix reports 279 pass, 0 fail, 10 skip.

---
## Summary by cubic
Fixes flaky failures in `run_check`, `run_pipe`, and the matrix and static-bins pattern checks when a successful command matched early in large output: `grep` exited before the pipe finished writing, SIGPIPE surfaced under `pipefail`, and the check was reported as failed. All captured-output pattern checks now read from a here-string instead of piping through `echo`/`printf`.

- Replaces piped `grep -qE` checks with `grep -qE "$pattern" <<< "$output"` in `tests/lib/test-runner.sh`, `tests/test-matrix.sh`, and `tests/test-static-bins.sh`, preserving `diff`'s expected exit status of 1.
- Adds `tests/test-runner.sh` covering early matches, late matches, missing patterns, and nonzero command exits, and wires it into `make check` via `mk/tests.mk`; the generated output exceeds pipe capacity so the early-exit path is actually exercised.

<sup>Written for commit 6a8518633e8fa52424469d03ecbdeb8b3ca59208. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/370?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/lib/test-runner.sh
+++ b/tests/lib/test-runner.sh
@@ -264,7 +264,7 @@ run_check()
         test_report fail "$tool" " (exit rc=$rc)"
         test_excerpt "$output"
         fail=$((fail + 1))
-    elif printf "%s\n" "$output" | grep -qE "$pattern"; then
+    elif grep -qE "$pattern" <<< "$output"; then
         test_report ok "$tool"
         pass=$((pass + 1))
     else
@@ -324,7 +324,7 @@ run_pipe()
         test_report fail "$tool" " (exit rc=$rc)"
         test_excerpt "$output"
         fail=$((fail + 1))
-    elif printf "%s\n" "$output" | grep -qE "$pattern"; then
+    elif grep -qE "$pattern" <<< "$output"; then
         test_report ok "$tool"
         pass=$((pass + 1))
     else
--- a/tests/test-runner.sh
+++ b/tests/test-runner.sh
@@ -8,6 +8,7 @@ set -euo pipefail
 if [ "${1:-}" = "--emit" ]; then
     awk 'BEGIN {
         print "match-begin"
+        # Keep output larger than pipe capacity to expose early-match SIGPIPE.
         for (i = 0; i < 30000; i++)
             print "abcdefghijklmnopqrstuvwxyz"
         print "match-end"
@@ -31,6 +32,7 @@ expect_result()
     local runner="$1" label="$2" pattern="$3" rc="$4" want="$5"
     local output
     checks=$((checks + 1))
+    # Each subshell inherits zero pass and fail counts and discards updates.
     if output=$(
         if [ "$runner" = run_pipe ]; then
             run_pipe test-runner.sh "$pattern" input --emit "$rc"
--- a/tests/test-runner.sh
+++ b/tests/test-runner.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+
+# Copyright 2026 elfuse contributors
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+if [ "${1:-}" = "--emit" ]; then
+    awk 'BEGIN {
+        print "match-begin"
+        for (i = 0; i < 30000; i++)
+            print "abcdefghijklmnopqrstuvwxyz"
+        print "match-end"
+    }'
+    exit "$2"
+fi
+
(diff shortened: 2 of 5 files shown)


tests/lib/test-runner.sh
```diff
@@ -264,7 +264,7 @@ run_check()
         test_report fail "$tool" " (exit rc=$rc)"
         test_excerpt "$output"
         fail=$((fail + 1))
-    elif printf "%s\n" "$output" | grep -qE "$pattern"; then
+    elif grep -qE "$pattern" <<< "$output"; then
```
jserv, line 267: Six identical copies of this pipeline survive outside this file, all under `set -euo pipefail`: tests/test-matrix.sh:565,633 and tests/test-static-bins.sh:93,127,161,358. They lose a matching pattern on large output exactly the way these two did, and the new lane does not reach them. Apply the same here-string to all six.

tests/test-runner.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# Copyright 2026 elfuse contributors
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+if [ "${1:-}" = "--emit" ]; then
+    awk 'BEGIN {
+        print "match-begin"
+        for (i = 0; i < 30000; i++)
```
jserv, line 12: 30000 is load-bearing: it is what pushes the emitted output past the pipe buffer so the pre-fix writer takes SIGPIPE. Nothing here says so, so a later trim to something that fits in 64 KB leaves this lane green against the bug it exists to catch. State the constraint in one line above the loop.

tests/test-runner.sh
```diff
@@ -0,0 +1,57 @@
...
+    local runner="$1" label="$2" pattern="$3" rc="$4" want="$5"
+    local output
+    checks=$((checks + 1))
+    if output=$(
+        if [ "$runner" = run_pipe ]; then
+            run_pipe test-runner.sh "$pattern" input --emit "$rc"
+        else
+            run_check test-runner.sh "$pattern" --emit "$rc"
+        fi
+        [ "$pass" -eq "$want" ] && [ "$fail" -eq "$((1 - want))" ]
```
jserv, line 42: This reads the counters from the sourced lib and only sees a zero baseline because the command substitution gives each call its own subshell. Drop that subshell later and every assertion after the first compares accumulated totals instead. Name the dependency in one line so the next reader does not refactor it away.

jserv (changes requested):
Rebase latest `main` branch, resolve conflicts, and response to review messages.

tests/lib/test-runner.sh
Suzu1Dev: Applied the here-string change to all six checks in 6a85186.

tests/test-runner.sh
Suzu1Dev: Added the pipe-capacity constraint above the loop in d1275e0.

tests/test-runner.sh
Suzu1Dev: Documented the subshell isolation of the pass/fail counters in d1275e0.

Suzu1Dev:
Rebased onto b54747a and resolved the conflict in mk/tests.mk.

jserv:

PR #371, opened 2026-09-05
Make the proof and mutation gates say what they prove

jserv:
Three commits, each standing on its own.

### Emit verify profiles from the recipe's variables

The frama-c MCP server can run the proofs `make verify` runs, but only if it loads the same program. Every field that decides that had to be restated by hand, and a hand-written function set is the drift a proof gate cannot afford: it produces a confident answer about a program the build system does not prove.

`print-verify-profiles` emits that JSON from the variables the `verify-<name>` recipe consumes. Include directories and force-includes move into shared variables so neither list is spelled twice; they were duplicated at first, under a comment claiming they could not drift.

`FRAMAC_ISYSTEM_DIRS` is used unquoted rather than through `patsubst`: its value carries a shell substitution containing a space, so a `patsubst` over it splits mid-command. That mangling still proves green on any target whose sources need no libc, which is most of them.

### Contract the two proved functions that had none

`elf_place_segment` and `futex_uaddr_is_aligned` sat in their targets' `-wp-fct` lists with no ACSL contract, so WP proved only that their bodies commit no runtime error. Nothing there can reject a broken source.

`elf_place_segment` now states the property its caller depends on: an accepted placement never overlaps the runtime infra reserve, which holds the page-table pool, the shim code and the EL1-only shim data.

Both raise the obligation count, so both floors move with them. `MIN_GOALS` is a tripwire against an emptied body or a dropped contract, which prove 0 of 0, and it only means that while it sits at the target's baseline. Left alone it would have tolerated `verify-elf` losing 15 obligations and `verify-futexdeadline` losing 2, silently.

### Make the mutation gate say what it actually proves

A function proved by two targets could only be mutated in the one whose `VERIFY_<T>_SRC` named its file. `hex_nibble` lives in `src/utils.h` and is proved by both `verify-elf` and `verify-rsp` precisely because assuming it once was a past mistake, and neither could mutate it. The runner now shadows a mutant header through `MUTANT_INCDIR`.

The staged path mirrors the original's path under `src/` and `MUTANT_INCDIR` is the staging root, not the copy's parent. Deriving it as the parent got `src/utils.h` right by luck and every nested header wrong: `"proved/netlink.h"` fell through `-Isrc` to the real header, so the run proved unmutated code while reporting a mutation nobody caught. A `#error` probe now confirms each shadow is reached; the unmutated baseline structurally cannot, since it stages a byte-identical copy.

Mutation runs pass `-wp-cache none`, because WP stores a timeout as a verdict and replays it, which would be a catch with no prover run behind it.

### One thing this makes visible

The gate reported "121 caught", which read as 121 proofs demonstrably rejecting their broken sources. It does not mean that. **All 121 are caught by exhausting the prover, not by refutation** - `[Unknown]`/`[Failed]` never appear in this tree, and raising the budget eightfold to 240s on a quiet host changes nothing.

They still count, and the reasoning is recorded rather than assumed: what makes exhaustion evidence is the baseline proving every goal while each mutant exhausts on the one its own function owns. A merely-hard goal would exhaust in the baseline too, and a failing baseline is fatal rather than scored.

---
## Summary by cubic
Makes the Frama-C proof gate and the mutation gate prove what they claim: verify profiles are emitted from the same variables the recipes consume, proved functions gain the contracts their callers rely on, included headers can finally be mutated, and mutation verdicts count only goals the mutated function owns.

**Proof gate**
- `make print-verify-profiles` emits the frama-c-mcp JSON from the same variables `verify-<name>` consumes, so the server can no longer prove a different program; include dirs and force-includes moved into shared `FRAMAC_INCLUDE_DIRS` / `FRAMAC_FORCE_INCLUDES` instead of being spelled twice.
- `elf_place_segment` now states an accepted placement never overlaps the runtime infra reserve, and `futex_uaddr_is_aligned` states its result in the bit domain; `elf_place_segment`'s `assigns` clause was dropped after proving only against `log_impl`'s WP-generated frame — `check-wp-result.py` now reports that frame, matching both wordings Frama-C uses for a missing spec — so `VERIFY_ELF_MIN_GOALS` went 156→172 with the new contract and back to 159 once the false frame fell, and `VERIFY_FUTEXDEADLINE_MIN_GOALS` went 128→130.
- `check-proof-targets` now runs on `make check`, and `BUILD_FLAVOR_GOALS` skips `print-%` goals so that scanner's sub-make cannot wipe a sanitizer build's objects.

**Mutation gate**
- Header mutants are staged under `MUTANT_INCDIR` (the staging root, not the copy's parent) so they shadow the real files, and a `#error` probe confirms each shadow is reached — run serially, since a concurrent sibling could spoil its exit-status answer — because the byte-identical baseline cannot detect a fall-through.
- Mutation runs pass `-wp-cache none` because WP otherwise replays a stored timeout as a catch; baselines that fail only while the pool is busy are retried serially, and their targets' mutations then run serially too.
- Verdicts report refutation and resource exhaustion separately — every catch here is `[Timeout]` — and `MUTANT_ESCALATE=<n>`, off by default, re-runs resource verdicts at a larger budget, reporting MISSED for a mutation that merely turned an easy goal into a slow one. Every open goal must belong to the mutated function, decided by the longest matching name in the target's own proved set and checked by a self-test, with caller obligations naming the callee qualifying for both; coverage is counted by function and by (target, function) pair.

<sup>Written for commit b6ff88cd13c3acf1c956f02a6a20d57a51f52130. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/371?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.claude/skills/elfuse-verify/SKILL.md
+++ b/.claude/skills/elfuse-verify/SKILL.md
@@ -1,12 +1,42 @@
 ---
 name: elfuse-verify
-description: How elfuse validates a change - choosing the lanes for the area you touched, the test matrix, make check, and the Frama-C proof targets declared in mk/verify.mk, including how to drive the frama-c MCP server on a stuck proof. Use when adding bounds math to src/proved/, writing or repairing ACSL contracts, running or debugging make verify / verify-mutants, touching frama-c-stubs/, adding a test lane, or deciding what to run before calling work done.
+description: How elfuse validates a change - choosing the lanes for the area you touched, the test matrix, make check, and the Frama-C proof targets declared in mk/verify.mk, including how to drive the frama-c MCP server on a stuck proof and how to read its proof_coverage report. Use when adding bounds math to src/proved/, writing or repairing ACSL contracts, running or debugging make verify / verify-mutants, asking how much of a target is actually proved, touching frama-c-stubs/, adding a test lane, or deciding what to run before calling work done.
 ---
 
 # Validating an elfuse change
 
 Two independent gates: the runtime tests and the proofs. A change to
 attacker-facing bounds math needs both.
+
+Independent in what they prove, not in what they consume. Run them one after
+the other, never concurrently. `make verify` re-invokes itself parallel and
+`verify-mutants` fans out too, while the runtime lanes are wall-clock
+sensitive, so overlapping them makes the machine fail tests that a serial run
+passes. Measured here: `make check` alongside the proof gate drove an 8-core
+host to a load average of 43.8 against a busy threshold of 4.8, and
+`test-thread-churn` timed out twice at 60 s and reported FAIL. The same binary
+on the same tree at load 4.0 finishes in 0.34 s, three runs out of three.
+Nothing was wrong with it. Neither mitigation saves you at that load, since
+`test_host_is_busy` only skips the throughput guardrail and the runner only
+re-runs once. A timing FAIL is worth nothing as evidence either way: it is not
+a regression you can act on, and a serial re-run is the only thing that tells
+you whether it was real.
+
+The pure source scanners are the exception, and they are the cheap early
+signal while something long is in flight: `check-lock-order`,
+`check-eintr-contract`, `check-atomics`, `check-proof-targets`,
+`check-stub-shadow` and `check-syscall-coverage` read the tree, cost seconds,
+and fail long before a full lane would. Five of the six are on `make check`;
+`check-stub-shadow` is a prerequisite of every `verify-*` target instead, so it
+is not reached by `make check` alone. Running one directly with
+`python3 scripts/<name>.py` costs nothing and needs no arguments.
+
+Five of the six write nothing. `check-proof-targets` is the one that does:
+it shells out to `make print-verify-targets` rather than reading
+`mk/verify.mk`, and a sub-make evaluates the build-flavor guard while it reads
+the makefiles. `print-%` goals are skipped by that guard for exactly this
+reason (`mk/common.mk`), so the scanner is safe to run beside a build; if you
+add a scanner that invokes make on some other goal, it is not.
 
 ## Choosing what to run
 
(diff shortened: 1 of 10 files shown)


jserv:
Thanks - fixed three of the four in 03de1be, and I want to give the measurements behind the fourth rather than just disagree.

### P2 (elf.c) - valid, fixed

Correct, and it was a real gap. `elf_place_segment` is the only contracted function in `elf.c` that both declares `assigns` and calls `log_error`, and `src/debug/log.h` declared `log_impl` with no spec - so Frama-C generated one and that frame was discharged against an assumption nothing stated.

`log_impl` now declares `assigns \nothing;` with a comment saying what it does and does not claim: what it writes is a file descriptor, outside the memory the analyzer models, so this is a statement about modeled objects rather than a claim that nothing is written anywhere.

No proof moved (`verify-elf` 172/172, `verify-futexdeadline` 130/130), which is the point. It also surfaced the assumption in the gate's own output, where it had been invisible:

```
before: assumed (spec generated, body not analyzed): memcpy pread
after:  assumed (spec generated, body not analyzed): log_impl memcpy pread
```

### P3 x2 - valid, fixed

`Update` -> `update` (WP's mode names are lowercase). `BUSY_LOAD_PER_CPU` was genuinely dead once the resource summary was rewritten to print the load beside each verdict instead of classifying the host once at the end; the constant and its comment described a classification no code performed.

### P1 - measurements, and what I did instead

The concern is right in principle. Two things make the stated remedies not apply here, both measured rather than argued.

**Refutation is unreachable in this configuration.** Across every mutation log this tree produces the tag is `[Timeout]`; `[Unknown]` and `[Failed]` never appear. Alt-Ergo and Z3 do not refute these goals, they grind. Raising the budget eightfold to 240s on a host at 0.3-0.6 runnable threads per CPU left all four `futexdeadline` mutations exhausting exactly as they did at 30s. So making exhaustion fatal would not make the gate stricter - it would make `caught` unreachable and the gate permanently red.

**Budget escalation as a blanket policy is unaffordable.** Measured on `futexhash`, serial:

| budget | 2 mutations |
|---|---|
| 30s | 116s |
| 120s | 454s |

3.9x, which extrapolates to roughly 1.9h -> 7.6h for all 121.

What does separate a broken contract from a hard one here is the baseline, not the verdict tag: the unmutated source proves every goal, and the mutant exhausts on the one its own function owns. A goal that were merely hard would exhaust in the baseline too, and a failing baseline is fatal rather than scored.

That said, the review pointed at a prior question nothing was checking: **did the mutation break the function it claims to break?** 120 of the 121 entries narrow the run to the mutated function via `FCT_ARG`, where this holds by construction - but the narrowing is dropped for a mutation that edits a contract, and there the target proves everything, so some unrelated goal getting slower would have scored as a catch. That case now fails as `ELSEWHERE` unless an opened goal belongs to the mutated function. It guards 1 of 121 today and makes the guarantee stated rather than incidental for the rest.

Resource verdicts are still reported apart from refutations, with the load each was measured under, so the summary never reads as "these proofs refute their mutants".

Happy to make `RESOURCE` fatal if you would rather have the red gate - it is a one-line change. The consequence is that no mutation can pass until the prover setup can actually refute, which needs counterexample generation rather than a longer budget.

PR #375, opened 2026-09-08
Run usbdevfs URBs asynchronously over IOKit

jotpalch:
NOTE: piece D of the USB layer described in #319, on top of #366 (piece C, the usbdevfs descriptor and its synchronous ioctls) and through it on #334 and #351. #366 answered `ENOTTY` for `SUBMITURB` / `REAPURB` / `DISCARDURB` and reported `GET_CAPABILITIES` as 0; this piece implements them and raises the two bits it can honour. The tty aliasing piece follows and is not here. Two commits: the engine, then the loopback fixture that tests it.

## Summary

`SUBMITURB`, `REAPURB`, `REAPURBNDELAY`, `DISCARDURB`, `poll`/`select`/`epoll` on a usbdevfs fd, and `CAP_REAP_AFTER_DISCONNECT`'s disconnect drain, over `IOUSBDeviceInterface650` / `IOUSBInterfaceInterface800` async completions delivered on one lazily-started CFRunLoop host thread -- libusb's own darwin model. `GET_CAPABILITIES` goes from 0 to `0x11`.

The engine is verified three ways: a native host binary over the arithmetic (781 assertions), a new IOKit loopback fixture that lets the whole engine run without a board (44 assertions, plus the entire #366 fd-contract lane re-run against a device that can actually complete a transfer), and the attached ESP32-S3 for everything the fixture cannot show. All numbers below are from runs made while writing this description.

## The two commits

1. **Run usbdevfs URBs asynchronously over IOKit** -- the engine. `SUBMITURB`, `REAPURB`, `REAPURBNDELAY`, `DISCARDURB`, the per-endpoint queue, the disconnect drain, `poll`/`select`/`epoll`, and the native host binary over the arithmetic. 16 files, 2968 insertions, 182 deletions. No fixture file exists in the tree at this commit, and it builds, passes every gate, and passes `make check` there.
2. **Model a loopback USB device behind the IOKit seam** -- the device model, the stub that stands in for it, the build variable that picks between them, the two lanes that run it, and the docs. 14 files, 2482 insertions, 34 deletions.

The boundary is the shipped binary. The model does not go into one (below), so the second commit is a translation unit no default build links plus the wiring that keeps it that way, and the engine can be read without it.

## What

**One transfer per endpoint in flight, and what it costs.** Linux `DISCARDURB` cancels exactly one URB. IOKit's `AbortPipe` (`USBDeviceAbortPipeZero` for ep0) cancels every outstanding transfer on the pipe. There is no IOKit call with Linux's granularity, so the Linux semantics are kept by never giving IOKit more than one transfer per endpoint to abort: later submissions queue inside elfuse and are started from the completion callback.

The cost is real and worth stating plainly. A guest gets queue depth but no overlap: an endpoint is idle for a full host round trip between one transfer completing and the next starting, so a bulk-streaming guest that submits a ring to keep the pipe busy gets its URBs served one at a time. The ring itself is honoured -- 300 URBs queued on one fd on the board all queued (first error 0) and all 300 reaped back -- and the loopback lane asserts both halves of the price: *never more than one in flight per endpoint*, and *the follower started only after the leader completed*. Throughput on an endpoint is therefore bounded by round-trip latency, not by IOKit's pipelining. Removing the queue would restore overlap and lose `DISCARDURB`'s one-URB granularity; that trade is the one design decision this piece is built around.

One URB on the wire is not by itself sufficient. The abort is issued with `async_lock` dropped, and in that window the target can complete normally and the completion callback can start the queued follower for the abort to hit instead. So the endpoint's FIFO stays shut while an abort for it runs, and `DISCARDURB` waits (bounded at 2s) for the record to leave the pending list -- which is also what makes an immediately following `REAPURBNDELAY` find the URB the way `usb_kill_urb` leaves it. On the board, across 63 runs, 12600 rounds of discarding one of two queued ep0 URBs produced 0 bystander `ECONNRESET`, and 1260 bulk-IN discards all reaped `ENOENT`, with the `REAPURBNDELAY` straight afterwards finding the URB every time.

**Argument validation follows `proc_do_submiturb`, not `do_proc_bulk`.** Flags mask, then `USBFS_XFER_MAX`, then a NULL buffer with a positive length, then the endpoint (`devio.c:1631-1658`). So an absent endpoint outranks an unknown transfer type, the ISO refusal and the control arm's eight-byte minimum, and a 2 GB URB on the default control pipe is `EINVAL` rather than an allocation. That is deliberately *not* the order #366 pins for the synchronous `BULK` ioctl, which resolves the endpoint first; the two kernel paths genuinely differ.

**Resource limits are Linux's, and two of elfuse's own are gone.** The 16 MB budget is the one the synchronous transfers already charge, because `usbfs_memory_usage` is a single kernel-wide static (`devio.c:134, 143-178`) and a second counter would let this path queue a budget behind whatever a `BULK` ioctl holds. Records are charged alongside their buffers so zero-length URBs are bounded, and there is no URB-count cap: the old per-fd 256-record backstop refused a 257th eight-byte URB, which is exactly the ring depth libusb's async API builds for bulk streaming.

**The disconnect path.** `usbdev_remove` runs `destroy_all_async` before it wakes the reapers. IOKit completes nothing of its own when a device terminates, so the first reap that finds the completion list empty on a disconnected fd issues the kill itself. The disconnect-watch refcon packs the slot index in a field sized from `USBDEV_MAX_FDS` behind a static assertion rather than a hand-written four bits: 32 slots through four bits decoded slot 16+k as slot k, which the host binary now pins -- reverting it fails 144 assertions, exactly 128 round trips plus 16 collisions.

**poll/select/epoll** (644 lines added to `poll.c`). The fd's completion pipe raises host `POLLIN`, remapped to guest `POLLOUT|POLLWRNORM` (`devio.c:2830-2843`); epoll registers `EVFILT_READ` and reports `EPOLLOUT` through a per-registration flag. Host-side interest is armed regardless of what the guest asked for, because a disconnect must wake even a read-only wait with the unmaskable `POLLERR|POLLHUP` (`devio.c:2839-2842`). A wake that maps to nothing guest-visible re-blocks: ppoll/pselect withdraw the woken entry's host interest (unreaped completions keep the pipe readable, so leaving it armed busy-spins at 100% CPU) and resume in bounded 200ms slices; epoll mutes the fired knote and re-enters the kevent slice loop rather than returning 0 before the timeout, which Linux `ep_poll` never does. `read()` on a disconnected fd is `ENODEV` (`devio.c:320-322`), and `REAPURB` returns `EINTR` with `syscall_restart_forbid()` because `reap_as` does not restart (`devio.c:2113-2114`).

The writability half of the remap reads the same derived capability the ioctl and write gates read rather than testing the access mode against `O_RDONLY`, matching `usbdev_poll`'s `FMODE_WRITE` gate (`devio.c:2837`) -- access mode 3 is not `O_RDONLY` and carries no `FMODE_WRITE` either. This is offered as a consistency fix with no test, and deliberately not as a bug: no guest can observe the difference, because reaching `POLLOUT` needs a completion and `SUBMITURB` on such an fd is already `EPERM`, so an assertion here would pass against the literal test too.

## Capability bits, and the evidence each is honoured

An earlier piece in this series was caught raising a bit it did not honour, so each bit is listed with what proves it.

- **`CAP_ZERO_PACKET` (0x01)** -- honoured. The trailing zero-length write is emitted from the completion callback with a bounded timeout and `async_lock` dropped, and a failure folds into `urb->status`. The loopback fixture logs what crossed the seam, so the lane asserts *the trailing zero-length packet reached the wire*, *a short OUT gets no terminating packet*, and *a failed terminating packet lands in urb->status* -- observed on a wire, not inferred from the predicate. Wiring the predicate to false fails 2 assertions; the untimed-under-lock version reported a failed terminating packet as success.
- **`CAP_REAP_AFTER_DISCONNECT` (0x10)** -- honoured. Loopback asserts *CAP_REAP_AFTER_DISCONNECT hands back every in-flight URB*, *each one carries the errno usb_kill_urb leaves*, and *REAPURB is ENODEV once the drain is done*; a reap that skips the drain fails 2. On the board, `RESET` returned all 3 queued URBs (rc=0, 3 of 3).
- **`CAP_BULK_CONTINUATION` (0x02)** -- **stays clear**, on purpose. The flag is accepted but its error-cascade unlink is not implemented, and a guest that read the bit would rely on the cascade.
- `NO_PACKET_SIZE_LIM` and `BULK_SCATTER_GATHER` stay clear; neither is implemented.

`GET_CAPABILITIES` reads `0x11` on the board in every run, and the loopback lane asserts it too.

## The loopback fixture: what it is, and what it cannot show

The engine needed a device that can complete a URB, and IOKit publishes no loopback one, so `ELFUSE_USB_FIXTURE=loopback` is one.

**The seam is exactly two IOKit COM vtables and nothing above them.** Every wire call in `usbdev.c` goes through `IOUSBDeviceInterface650 **` or `IOUSBInterfaceInterface800 **` as `(*h)->Method(h, ...)`, so `src/syscall/usbdev-fixture.c` hands back an object whose first member is a vtable of the same shape. Everything above stays the code that runs against a board: the URB records, the per-endpoint FIFO, `usbdev_async_cb`, the `SHORT_NOT_OK` and `ZERO_PACKET` predicates, the readiness and disconnect maps, `REAPURB`, the drain, and all 644 poll.c lines. Completions arrive from a one-shot `CFRunLoopTimer` on the event thread, where `IODispatchCalloutFromCFMessage` would have delivered them, and aborts are asynchronous the way IOKit's are.

What the fixture does is data rather than a flag: `ELFUSE_USB_LOOPBACK` scripts a per-endpoint sequence of outcomes, each naming the IOReturn it stands for -- `short(n)` `kIOReturnUnderrun` 0xe00002e7, `stall` `kIOUSBPipeStalled` 0xe000404f, `timeout` `kIOUSBTransactionTimeout` 0xe0004051, `nodev` `kIOReturnNoDevice` 0xe00002c0, plus `ok`, `ok(n)`, `err`, `refuse`, `delay(ms)`, `never`, `terminate` and `zlpfail` -- and a guest rewrites the script, reads back a log of what crossed the seam, and terminates the device through vendor control requests.

**What it cannot show, stated so the coverage is not overread.** Three of the engine's own answers are board-only, and breaking any of the three leaves the lane at 44 passed, 0 failed:

1. the `ZERO_PACKET` write's dropped `async_lock` -- the fixture answers a zero-length write immediately, so nothing NAKs and nothing stalls behind the lock;
2. its bounded timeout -- the fixture ignores both timeout arguments;
3. the bystander window `ep_aborting` shuts -- the fixture retargets an aborted transfer's timer under its own lock, so it can never start a follower into an abort that is still running.

The 200-round board driver, run 63 times for 12600 rounds with 0 bystander `ECONNRESET`, remains the only evidence for (3). More broadly, `docs/testing.md` records what stays on hardware: real timing, NAKs, maxpacket segmentation, DMA alignment and throughput; that IOKit really delivers completions on the runloop, and the `IODispatchCalloutFromCFMessage` opacity behind the URB record's atomic owner, which a timer callout fully visible to ThreadSanitizer does not reproduce; exclusive-access arbitration and kernel-driver binding; a real `SET_CONFIGURATION`, altsetting renumbering and port `RESET`; that a device actually receives the zero-length packet; and a physical unplug mid-transfer.

**None of it is in a default build.** `src/syscall/usbdev-fixture.c` is a translation unit under `src/` that exists only so a test can run, and this tree has no precedent for one. The env-var fault hooks that `usb-sysfs.c`, `usbdev.c`, `fs-stat.c` and `fs.c` carry are precedented, but each is a handful of lines inside a file that is there for production reasons; and the synthetic USB tree in `usb-sysfs.c` earns its place the same way -- it lets `lsusb` work on a machine with no devices, so a non-testing user gets something from it. A device that echoes back what was written to it does nothing for anyone outside an assertion.

So the seam has two implementations and the build picks one. `src/syscall/usbdev-fixture-stub.c` -- 72 lines, `usbdev_fixture_loopback()` answering false and the rest inert -- is in `SRCS` in every build; `usbdev-fixture.c` takes its place only when `USB_LOOPBACK_FIXTURE=1` asks for it (`mk/config.mk`, and the `SRCS` block in the top-level `Makefile`). Listing both would be a duplicate-symbol link error, which is the property that keeps a default build from quietly acquiring the model. Not one call site in `usbdev.c` is conditionally compiled, so it is one program in both builds and the fixture cannot drift into code a default build never sees.

To get the fixture, ask for it:

```
make USB_LOOPBACK_FIXTURE=1                        # the model, in build/elfuse
ELFUSE_USB_FIXTURE=loopback build/elfuse <guest>   # and one loopback device
```

`make check` needs no flag. `mk/tests.mk` re-enters make with `USB_LOOPBACK_FIXTURE=1 ELFUSE_BIN=build/elfuse-loopback` and the two loopback lanes run that binary, so what they run is the build a reader gets from the variable and cannot drift from it, while a plain `make` still leaves `build/elfuse` without the model. Overriding `ELFUSE_BIN` rather than `BUILD_DIR` shares every object but the fixture's, so a loopback build compiles exactly one extra file, and a command-line override reaches the sub-make through `MAKEFLAGS`, so the sanitizer lanes get a loopback binary of their own flavor. `make lint` reads a new `ALL_SRCS` rather than `SRCS`, so gating the model out of the build does not gate it out of clang-tidy, and cppcheck walks the tracked sources, so it sees both either way.

What a default build carries of it is measured rather than asserted. On a clean rebuild of the tip:

```
build/syscall/usbdev-fixture.o        never compiled (no such file)
nm build/elfuse, fixture symbols      the 7 seam entry points, all the stub's
size -m usbdev-fixture-stub.o __text          44 bytes
size -m usbdev-fixture.o      __text       10104 bytes
defined symbols, default vs USB_LOOPBACK_FIXTURE=1     1771 vs 1811
  the 40 extra are all fx_* or fixture; none is present only in the default
the loopback binary is about 18 KB larger, the model's text being the difference
```

After a full `make check` -- which links `elfuse-loopback` for the lanes -- `nm build/elfuse` still finds zero `fx_*` symbols.

**Off, the seam is five `if (u->fake)` branches, one has-device probe and one bind call**, behind a mode resolved once per process and set for one modeled location only. Every other device takes the path it took before -- the existing fixture modes' descriptor blobs come out byte-identical seam-in versus seam-out. The fixture device carries no synthetic `io_service_t` (`u->service` stays `IO_OBJECT_NULL` and the flag says so), which keeps every `IOObjectRelease` and `usbdev_arm_disconnect_watch`'s NULL-service guard correct by construction. A vtable slot `usbdev.c` calls that the fixture left NULL would be a null call rather than a compile error, so the list is checked once when the fixture stands up.

To show the seam changes nothing on real hardware, the four board drivers were run against the board twice: once on the engine commit, where no fixture file exists at all, and once on the tip's default build, where the seam exists and the stub answers it. The two logs are byte-identical once the wall-clock figures are normalized. Those figures jitter run to run -- the three timeouts read 5019, 5046 and 6176 ms in the first against 5082, 5033 and 6146 ms in the second -- and timing figures in this description are quoted with the run they came from and compared as jitter, never as values to reproduce.

## Deviations

Eight XFAIL rows, printed by `test-usbdev-ioctl` with both values so they live in the lane's own output rather than only here.

```
XFAIL open-limit: Linux unbounded, elfuse 32 simultaneous
XFAIL reset: Linux re-enumerates the port, elfuse kills the device's URBs and
  clears claimed pipes' stalls, logging rather than reporting a clear that
  fails, and returns 0
XFAIL clear-halt-collateral: Linux warns and leaves a queued URB on the
  endpoint alone (check_reset_of_active_ep, devio.c:1379-1391), elfuse has only
  ClearPipeStallBothEnds, which aborts the pipe, so CLEAR_HALT and RESETEP make
  an in-flight URB reap -ECONNRESET
XFAIL urb-signal: Linux raises the URB's signr at completion
  (kill_pid_usb_asyncio, devio.c:654) and DISCSIGNAL's at disconnect, elfuse
  accepts both, returns 0 and delivers neither
XFAIL iso: Linux serves isochronous URBs, elfuse answers EINVAL once the
  endpoint has resolved
XFAIL discard-latency: Linux's usb_kill_urb returns with the URB already
  completed, elfuse waits for IOKit's abort callback and gives up after 2s
  rather than parking the vCPU thread
XFAIL driver-name: Linux GETDRIVER reports the driver's name (cdc_acm), elfuse
  reports the IOKit class (AppleUSBACMControl), and DISCONNECT_CLAIM's name
  filters compare against it
XFAIL short-bulk-out: Linux reports the byte count actually sent, elfuse has no
  length from WritePipeTO and reports EIO
```

The last three are inherited unchanged from #366. Separately, `docs/usage.md` now states what a stock distribution `lsusb` actually does here: usbutils reaches libusb through libudev, whose monitor wants `SO_ATTACH_FILTER`, elfuse's netlink layer answers `ENOPROTOOPT`, and `libusb_init` gives up with -99. That gap belongs to the netlink layer, not to this piece, but the worked example had to stop claiming otherwise.

## Evidence

Both commits were checked out on their own, each from `make clean`, and each ran the whole set. The engine commit first, with no fixture anywhere in the tree:

```
make check BAREMETAL_CROSS=aarch64-elf-        exit 0,  All 99 tests passed
test-usbdev-urb-host                           781 passed, 0 failed
test-usbdev-ioctl        (FIXTURE=1)           103 passed, 0 failed
test-usb-sysfs / -sysroot / -overflow          138 / 13 / 5 passed, 0 failed
nm build/elfuse, fixture symbols                0
check-asan / check-ubsan / check-tsan          exit 0, 0 sanitizer reports
board: adj, full, board, conc                  4 of 4 exit 0,
                                               adj-board 7 passed, 0 failed
```

Then the tip, from a fully clean tree, with `build/elfuse-loopback` linked inside the lane that needs it:

```
make check BAREMETAL_CROSS=aarch64-elf-        exit 0,  All 99 tests passed
test-usbdev-urb-host                           781 passed, 0 failed
test-usbdev-urb-loopback (FIXTURE=loopback)     44 passed, 0 failed
test-usbdev-ioctl        (FIXTURE=loopback)    108 passed, 0 failed
test-usbdev-ioctl        (FIXTURE=1)           103 passed, 0 failed
test-usb-sysfs / -sysroot / -overflow          138 / 13 / 5 passed, 0 failed
check-asan / check-ubsan / check-tsan          exit 0, 0 sanitizer reports;
                                               both loopback lanes ran under
                                               each, 44/0 and 108/0 every time
make lint (clang-tidy, 63 files)               exit 0
.ci/check-cppcheck.sh (63 files)               clean
```

The gate set, run on each commit separately: commit-log, `commentflow --check`,
check-format (clang-format 22.1.8), `.ci/check-cppcheck.sh`, check-newline,
check-security, check-matrix-lists, check-eintr-contract, check-lock-order,
check-atomics, check-ascii, check-svc-tails, check-skill-refs,
check-proof-targets, check-syscall-coverage and `make lint` -- all exit 0 on
both.

The loopback run answers 108 where the `FIXTURE=1` run answers 103. The five extra are exactly the in-flight transfer allowance -- with no claimable device the claim answers `ENODEV` before any length is looked at, so those five skip there and print why. Every assertion both runs make answers the same in both.

Negative controls, because a lane that only ever passes proves nothing:

- Reverting each of the five things the host binary pins to the shape it had fails it: 144, 3, 3, 1 and 1 assertions (the refcon one is a compile error while the static assertion stands).
- Of the eleven new `SUBMITURB` argument-order assertions in `test-usbdev-ioctl`, seven fail against the order the engine had first -- endpoint resolved ahead of the gate, no `USBFS_XFER_MAX` bound at all.
- Ten deliberate breaks of the engine were each applied, rebuilt and run against the loopback lane. Completion callback leaving `actual_length` at zero: **7**. `urb_complete_locked` always writing status 0: **10**. `usbdev_ep_may_start` ignoring the in-flight URB: **6**. `DISCARDURB` returning 0 without doing anything: **5**, then the lane runs into its per-stage alarm. A reap that skips the disconnect drain: **2**. A readiness map set unconditionally: **1** -- the disconnect that must report `POLLERR|POLLHUP` alone reports 0x11c, because the wake byte keeps the pipe readable for good. `ZERO_PACKET` predicate wired to false: **2**. `SHORT_NOT_OK` ignored: **1**. Transferred-count clamp removed: **1** (`actual_length` 999 for a 32-byte buffer). A refused start routed through the syscall map: **1** -- that one has to be a queued follower, because a leader's refusal is a syscall return value and `kIOReturnNotOpen` is `ENODEV` in both maps, so only a follower reaches the one row where they differ (`kIOReturnAborted`, `ECONNRESET` in `urb->status` against the `EINTR` the syscall map answers and the kernel never writes into a URB).
- Two of those ten were re-applied to the tree in its present shape and rebuilt through `make USB_LOOPBACK_FIXTURE=1`, to confirm the lane still kills a broken engine now that the fixture is behind a variable: `urb_complete_locked` always writing status 0 gives 34 passed, 10 failed; the readiness map set unconditionally gives 43 passed, 1 failed, on `revents=0x11c` where `POLLERR|POLLHUP` alone is wanted. Both exit 1.

On the attached ESP32-S3 at `303a:1001`, re-measured for this description against a clean rebuild of the tip: 63 consecutive runs of the review driver, and one run each of the other six.

```
GET_CAPABILITIES                                 0x11, in all 63 runs
control URB through poll+reap    revents 0x4 (POLLOUT), status 0, actual 18
short IN                                         actual 18 (not 255),
                                                 tail still 0xaa,
                                                 SHORT_NOT_OK status -121
bulk IN discard reaps ENOENT                     20 of 20 in each of 63 runs
                                                 (1260 discards, no exception)
REAPURBNDELAY straight afterwards finds it       20 of 20 in each of 63 runs
300 URBs queued on one fd                        300 queued (first error 0),
                                                 300 reaped, in each of 63 runs
200 rounds discarding 1 of 2 queued ep0 URBs     0 bystander ECONNRESET in each
                                                 of 63 runs (12600 rounds),
                                                 A settled 200/200
RESET with 3 URBs queued                         rc=0, 3 of 3, in each of 63
mixed 400-URB poll/epoll/select driver           submit 400, reap 400,
                                                 pollout 58, pollerr 0
ThreadSanitizer / AddressSanitizer+UBSan         0 reports, 7 drivers x 2
                                                 sanitizer builds, including
                                                 the 400-URB poll/epoll/select
                                                 driver whose first run once
                                                 reported the record-owner race
```

That last row is why the URB record's owning slot is `_Atomic`: it is the one field the completion callback reads before it can take any lock, and ThreadSanitizer cannot see through `IODispatchCalloutFromCFMessage`, so it reported the access on a cold-start run.

---
## Summary by cubic
Implements the usbdevfs async URB path over IOKit, so `SUBMITURB`, `REAPURB*`, and `DISCARDURB` now transfer instead of returning `ENOTTY`; `GET_CAPABILITIES` reports `0x11`. Completions and disconnect watches ride one lazy CFRunLoop thread, and each endpoint keeps at most one IOKit transfer in flight to preserve Linux's single-URB `DISCARDURB`, so queued transfers don't pipeline.

- Applies Linux's URB validation order, shared 16 MB usbfs accounting, `SHORT_NOT_OK`, `ZERO_PACKET`, and disconnect-drain semantics; the old 256-URB-per-fd limit is gone.
- Propagates disconnects to every open usbfs fd; a generated table pins each ioctl's departed-device answer and fixed wrong ones on claim/release, discard, the default arm, interface enumeration ops, and `RESET`.
- `REAPURBNDELAY` issues the post-disconnect aborts and answers `-EAGAIN` until they land instead of blocking up to 2s.
- Sync `CONTROL` now charges the shared byte allowance; per-endpoint FIFOs make submit/complete/restart O(1) under load; drain-timeout and late-callback paths hand back orphaned URBs and pin handles safely.
- Adds an opt-in IOKit loopback fixture with scripted transfer outcomes; the default build links only a stub, and loopback tests use a separate binary so the shipped one can't become a stale fixture build.
- Adds native URB tests, loopback completion/disconnect tests, departed-device ioctl coverage, and regression tests for epoll masks (`EPOLLWRNORM`, `EPOLLRDNORM`, `EPOLLRDHUP`).
- Documents hardware-only gaps: isochronous URBs, async signals, clear-halt collateral, short bulk OUT status, and bounded `DISCARDURB` latency.

<sup>Written for commit 5e7bad470f6fdd365b321f6c3b5d29df38e4c1cd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/375?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src/syscall/usbdev-urb.h
```diff
@@ -0,0 +1,154 @@
...
+ * zero-length packet (darwin_usb.c:3193-3204). ep0 is excluded because the
+ * control transfer carries its own status stage.
+ */
+static inline bool usbdev_urb_needs_zlp(int32_t status,
+                                        bool zero_packet,
+                                        uint8_t pipe,
+                                        uint32_t data_len,
+                                        uint16_t mps)
+{
+    return status == 0 && zero_packet && pipe > 0 && data_len > 0 && mps != 0 &&
```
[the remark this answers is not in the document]
jotpalch: Checked against Linux v6.18, and this does not hold for interrupt OUT: devio.c sets allow_zero for bulk OUT (1710-1711) and for interrupt OUT (1734-1735), and urb.c:516-519 puts URB_ZERO_PACKET in the allowed set for XFER_BULK and XFER_INT alike when is_out. Restricting the predicate to bulk OUT would make us withhold a ZLP the kernel emits. The control half is unreachable rather than wrong, twice over: pipe > 0 already excludes ep0, and for a non-ep0 control URB mps stays 0, because usbdev.c:3235 initializes it and only the bulk/interrupt arm at 3348 assigns it, while the predicate requires mps != 0. One limit worth stating: the fixture and the board both expose interrupt IN only, so the interrupt-OUT half rests on the kernel source rather than on a measurement here. No change made.

src/syscall/usbdev.c
```diff
@@ -1958,6 +3017,13 @@ static int64_t usbdev_do_setconfiguration(usbdev_t *u, guest_t *g, uint64_t arg)
     usbdev_lazy_device_open(u);
     if (!u->dev_open)
         return -LINUX_EBUSY; /* exclusive holder elsewhere */
+
+    /* usb_set_configuration -> usb_disable_device kills every URB on the
+     * device, ep0 included. The -EBUSY check above only looks at interfaces,
+     * and the default control pipe is exactly the queue no interface claim
+     * covers, so a control URB could otherwise ride across the change.
+     */
+    (void) usbdev_kill_urbs_locked(u, NULL);
```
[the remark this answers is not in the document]
jotpalch: Right, and fixed. Both sites now go through a single usbdev_drain_for_change (usbdev.c:1614) that answers 0 or -EBUSY, so the condition is decided once instead of separately at each caller. Reaching it needed a transfer that can actually miss the 2 s deadline, which no existing script step could produce, so the fixture grew a wedge(ms) step: with the fix reverted the new assertion reports rc=0 after 2011 ms, and with it in place -EBUSY. RESET still proceeds on a failed drain, deliberately, since a wedged transfer is the state a guest issues RESET to escape and a stall clear retires no pipeRef; that is recorded as a known deviation in internals.md.

src/syscall/usbdev.c
```diff
@@ -689,11 +965,639 @@ static int64_t usbdev_build_pipe_map(usbdev_iface_t *fi)
...
+                break;
+            }
+            next = r;
+            break;
+        }
+        if (!usbdev_ep_may_start(u->draining, u->ep_aborting[key], inflight))
+            return;
+        if (!next)
+            return;
+        IOReturn ir = usbdev_urb_start(next);
```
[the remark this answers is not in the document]
jotpalch: Right, and fixed. The stamp is now issued at the queued follower's start, and all three sites that can learn the device is gone share one usbdev_ioret_device_gone predicate rather than repeating the pair of code comparisons. With the fix reverted the two new assertions report revents=0x0 and GET_CAPABILITIES rc=0; with it in place, 0x18 and -ENODEV.

src/syscall/poll.c
```diff
@@ -1920,6 +2254,105 @@ static int epoll_collect_hung_up(epoll_instance_t *inst,
...
+        pthread_mutex_lock(&inst->lock);
+        epoll_reg_t *reg = &inst->regs[gfd];
+        bool consumed = reg->active && reg->generation == muted->gens[i] &&
+                        reg->oneshot_armed;
+        pthread_mutex_unlock(&inst->lock);
+        struct kevent kev;
+        EV_SET(&kev, muted->idents[i], EVFILT_READ,
+               consumed ? EV_DELETE : EV_ENABLE, 0, 0,
+               (void *) (uintptr_t) gfd);
+        kevent(kq, &kev, 1, NULL, 0, NULL);
```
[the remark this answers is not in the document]
jotpalch: Right, and it fires. An earlier pass on this branch judged the race not constructible, and that was wrong. With a delay inserted between the unlock and the kevent, a second thread's EPOLL_CTL_MOD landed inside the window at t=538 ms; epoll_unmute then issued EV_DELETE against the knote MOD had just installed, and the following epoll_wait returned 0 after its full 1500 ms timeout, where the same probe without the delay returns EPOLLERR|EPOLLHUP in about 5 ms. Fixed by holding inst->lock across the kevent, which is what the hangup and mute stamping loops beside it already do; the kevent is changelist-only, so nothing blocks under the lock. The same probe now serializes the MOD to t=959 ms and reports 0x18.

docs/usage.md
```diff
@@ -250,6 +250,39 @@ Practical notes:
...
+
+Linux USB tools see the Mac's attached devices. `/dev/bus/usb` and
+`/sys/bus/usb/devices` are synthesized from the IOKit registry, and opening
+a device node yields a usbfs-compatible fd driven over IOKit (see
+[internals.md](internals.md), section "USB Device Passthrough").
+
+What works unprivileged is what macOS itself leaves unclaimed: vendor-class
+and bulk interfaces, which is the debug-probe and DFU population. `lsusb`,
+libusb programs, and nusb-based tools such as the probe-rs family
+enumerate, claim, and transfer against those directly, including async
```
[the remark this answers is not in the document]
jotpalch: Right, and there was a second instance you did not name: README.md carried the same stale sentence as docs/usage.md. Both are deleted. I swept every doc and comment this branch touches in both directions afterwards: nothing left denies async URBs, and the only surviving "not implemented" statements are the ones about isochronous URBs and about async guest-signal injection, which are both still true.

src/syscall/usbdev.h
```diff
@@ -72,3 +75,28 @@ int64_t usbdev_fstat(int fd, struct stat *st);
...
+ */
+bool usbdev_poll_host_events(int guest_fd,
+                             short guest_events,
+                             short *host_events);
+short usbdev_poll_guest_revents(int guest_fd,
+                                short guest_events,
+                                short host_revents);
+
+/* Lock-free "device gone" test for the epoll merge path. */
+bool usbdev_fd_disconnected(int guest_fd);
```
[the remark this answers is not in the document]
jotpalch: Measured rather than argued, and it does not reproduce. A probe drives a usbfs fd to a disconnected state (poll reports 0x18), closes it, and then takes the same fd number again: an eventfd landing on that number polls 0x0, and a fresh usbfs fd on that number also polls 0x0. Two clearing sites bracket the reuse window. Close clears the maps only while no live entry still answers to that fd number (usbdev.c:2119-2131), and open clears them before the number can be polled through (usbdev.c:2458-2465), which is what keeps a sibling's bit from being erased and a stale bit from being read. No change made.

tests/test-usbdev-urb-loopback.c
```diff
@@ -0,0 +1,736 @@
...
+
+    /* refuse: IOKit turns the start itself down, and the URB has to come back
+     * through the URB-status map rather than the syscall map, whose Aborted row
+     * would write -EINTR into urb->status.
+     */
+    fx_reset();
+    fx_script("ep81:refuse(0xe00002cd)");
+    mk_bulk(&u, EP_IN, in, sizeof(in));
+    TEST("a start IOKit refuses is a SUBMITURB error");
+    CHECK(submit(&u) == -ENODEV, "submit rc=%ld", submit(&u));
```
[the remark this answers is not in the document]
jotpalch: Right, and fixed by capturing the return once. The failure mode is exactly the one described: injecting refuse(0xe0004061),refuse(0xe00002cd) so the two calls differ, the old shape printed submit rc=-19, which is the value the assertion tests against, for a call that had actually returned -2; the new shape prints -2.

src/syscall/usbdev-fixture.c
```diff
@@ -0,0 +1,1254 @@
...
+                         : (uint32_t) sizeof(buf) - 1;
+        if (n && req->pData)
+            memcpy(buf, req->pData, n);
+        else
+            n = 0;
+        buf[n] = '\0';
+        pthread_mutex_lock(&usbdev_fixture_lock);
+        fx_script_load(d, buf);
+        pthread_mutex_unlock(&usbdev_fixture_lock);
+        req->wLenDone = req->wLength;
```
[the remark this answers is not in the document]
jotpalch: Right, and fixed by refusing an oversize script whole instead of truncating it, with wLenDone set to n. With the fix reverted the new assertion reports rc=595 len=595, and the assertion after it shows why the truncation mattered rather than merely being untidy: the installed prefix had quietly become a different scenario, so a later stall answered status=0 where the intended script gives -EPIPE.

src/syscall/usbdev-fixture.c
```diff
@@ -0,0 +1,1254 @@
...
+    for (int i = 0; i < FX_MAX_WATCH; i++) {
+        if (d->watch[i].used && d->watch[i].refcon == refcon) {
+            free_slot = i;
+            break;
+        }
+        if (free_slot < 0 && !d->watch[i].used)
+            free_slot = i;
+    }
+    if (free_slot >= 0) {
+        d->watch[free_slot].used = true;
```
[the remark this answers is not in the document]
jotpalch: Right, and fixed. One point of detail on the shape of the fix: the callback is not invoked directly from usbdev_fixture_watch, because that runs under usbdev.c's per-entry lock while usbdev_interest_cb takes usbdev_table_lock, and internal.h says those two are never held together. A zero-delay timer re-runs the same delivery on the event thread instead, where fx_deliver_terminate already delivers and holds neither lock. With the fix reverted the new assertion reports revents=0x0 after a 500 ms poll; with it in place, 0x18.

src/syscall/usbdev-fixture.c
```diff
@@ -0,0 +1,1254 @@
...
+                         IOReturn result,
+                         const void *payload)
+{
+    if (idx < 0 || idx >= (int) d->nlog)
+        return;
+    fx_rec_t *r = &d->log[idx];
+    r->actual = actual;
+    r->ioreturn = (uint32_t) result;
+    if (payload && (r->flags & 1u)) {
+        uint32_t n = actual < 8 ? actual : 8;
```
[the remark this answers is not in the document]
jotpalch: Right, and fixed by clamping to r->requested as well as to the eight-byte field. AddressSanitizer on the reverted shape reports a heap-buffer-overflow READ of size 8 in fx_timer_cb, and is clean once the clamp is back. One caveat on the assertion that went with it: it states the contract but is not a deterministic pre-fix failure, because the bytes past the four-byte allocation happened to read as zero. ASan is what actually detects this one.

tests/test-usbdev-urb-loopback.c
```diff
@@ -0,0 +1,736 @@
...
+    mk_bulk(&u, EP_IN, in, sizeof(in));
+    submit(&u);
+
+    /* Asking for both bits, because do_pollfd masks the answer by the events
+     * the caller demanded: a poll for POLLOUT alone is answered POLLOUT alone,
+     * on Linux as here.
+     */
+    short want = POLLOUT | POLLWRNORM;
+    TEST("poll reports nothing while the URB is in flight");
+    CHECK(poll_once(want, 40) == 0, "revents=0x%x", poll_once(want, 0));
```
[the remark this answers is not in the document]
jotpalch: Right, and fixed in two places rather than one: the same double-call shape sat on line 539 as well as the 533 you named, and both now poll once and reuse the result. On what is actually lost, to be exact about it: the two calls cannot disagree in this lane, because the completion pipe is level-triggered and the message poll only runs on the failure path, so there is no constructible pass/fail divergence. What was wrong is that the printed value came from a second syscall issued with a different timeout than the one the assertion tested.

src/syscall/poll.c
```diff
@@ -610,6 +705,184 @@ static void pselect_bits_init(pselect_bits_t *b,
...
+{
+    int word = req->word;
+    uint64_t bit = BIT64(req->bit_index);
+    int gfd = (int) req->word * 64 + req->bit_index;
+    bool ready = use_poll_fallback
+                     ? (req->revents & (POLLIN | POLLHUP | POLLERR)) != 0
+                     : (RANGE_CHECK(req->host_fd, 0, FD_SETSIZE) &&
+                        FD_ISSET(req->host_fd, read_set));
+    bool disc = usbdev_fd_disconnected(gfd);
+    int ready_bits = 0;
```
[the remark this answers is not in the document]
jotpalch: Right, and fixed. pselect_usb_writeback now reports back what the host's own select() already charged for each usbfs entry, and the caller trades that single count for the per-set bits it just wrote; pselect_usb_credit_disarmed is deleted outright, since keeping it alongside would double count a disarmed entry. With the fix reverted the new assertion reports ret=1 read=1 write=1; with it in place, ret=2.

src/syscall/usbdev.c
```diff
@@ -689,11 +965,679 @@ static int64_t usbdev_build_pipe_map(usbdev_iface_t *fi)
...
+ */
+static void urb_complete_locked(usbdev_t *u, usbdev_urb_t *rec, int32_t status)
+{
+    rec->status = status;
+    rec->state = URB_COMPLETED;
+    urb_list_append(&u->completed_head, &u->completed_tail, rec);
+    usbdev_ready_sync_locked(u);
+    if (u->pipe_wr >= 0) {
+        char b = 0;
+        (void) !write(u->pipe_wr, &b, 1);
```
jserv: The write is nonblocking and its result is discarded, so a completion whose byte does not fit is silently untracked while its record stays on `completed_head`. `usbdev_do_reap` consumes exactly one byte per reap, so once the guest has drained the stored bytes the surviving completions have no readable byte behind them, `usbdev_poll_guest_revents` returns 0 for an armed entry, and `poll`, `select` and `epoll` block on completions that are reapable. With the per-fd URB cap gone a zero-length URB costs only `sizeof(usbdev_urb_t)` against the 16 MB budget, which admits far more records than the pipe's 64 KiB of tokens. A level indicator (write on the transition to reapable, drain only when `completed_head` is empty and the fd is not disconnected) does not depend on every byte landing.

src/syscall/usbdev.c
```diff
@@ -689,11 +965,679 @@ static int64_t usbdev_build_pipe_map(usbdev_iface_t *fi)
...
+            r = u->pending_head;
+            while (r) {
+                usbdev_urb_t *nx2 = r->next;
+                if ((intf == NULL || r->intf == intf) &&
+                    r->state == URB_INFLIGHT) {
+                    urb_pending_unlink(u, r);
+                    u->inflight--;
+                    u->nurbs--;
+                    u->inflight_bytes -= r->charge;
+                    usbdev_memory_refund(r->charge);
```
jserv: `usbdev_memory_refund` hands the record's share of the process-wide budget back while `r->buf` is still allocated and still owned by an in-flight IOKit transfer, so the global usbfs accounting under-counts live memory until the late callback runs, and indefinitely if it never does. The reason given for settling here applies to `u->nurbs` and `u->inflight_bytes`, which are slot fields a reused slot must not see touched; the process-wide counter carries no such constraint. Move the refund to `urb_free_orphan_locked`, beside the `free(rec->buf)` that actually releases the memory.

src/syscall/usbdev.c
```diff
@@ -689,11 +965,679 @@ static int64_t usbdev_build_pipe_map(usbdev_iface_t *fi)
...
+                            u->ifaces[fi_i].orphans++;
+                            break;
+                        }
+                    }
+                    n++;
+                }
+                r = nx2;
+            }
+            log_warn("usbdev: %d in-flight URB(s) did not drain; orphaned", n);
+            u->draining = false;
```
jserv: This path clears `draining` and returns, while the success path below also kicks every endpoint key. The orphaned records' late callbacks take the `orphaned` early return in `usbdev_async_cb`, so nothing else restarts a FIFO: an endpoint whose leader completed during the drain, whose callback saw `draining` and did not start the follower, is left with a queued URB and no in-flight one, and a `REAPURB` for it never returns. Release interface 0 while interface 1 holds a leader and a follower to reach it.

```suggestion
            u->draining = false;
            for (unsigned k = 0; k < 256; k++)
                usbdev_kick_ep_locked(u, (uint8_t) k);
```

src/syscall/usbdev.c
```diff
@@ -1357,14 +2414,25 @@ int64_t usbdev_open_path(const char *path, int linux_flags)
...
     claimed_mask_reset(u);
     /* Nonzero while bound; equal for every fd open on the same device node. */
     devkey_publish(
         u, (1ull << 63) | ((uint64_t) (uint32_t) bus << 32) | (uint32_t) dev);
+    u->pending_head = u->pending_tail = NULL;
+    u->completed_head = u->completed_tail = NULL;
+    u->inflight = 0;
+    u->nurbs = 0;
+    u->inflight_bytes = 0;
+    u->disconnected = false;
```
jserv, line 2701: `disc_drained` is not reset with the rest of the per-open async state, and nothing else clears it: `usbdev_teardown_locked` resets `disconnected` but leaves this one, and `usbdev_init` zeroes the table once at startup. A slot whose previous open drained after a disconnect therefore starts its next open with the flag already set, so that open's own disconnect skips the drain in `usbdev_do_reap` and answers `ENODEV` with URBs still pending.

```suggestion
    u->disconnected = false;
    u->disc_drained = false;
```

src/syscall/usbdev.c
```diff
@@ -2098,6 +3208,505 @@ static int64_t usbdev_do_driver_ioctl(usbdev_t *u, guest_t *g, uint64_t arg)
...
+         * (measured: three URBs outstanding at terminate, zero callbacks in the
+         * next three seconds, and the same URBs recovered in 1 ms by close()'s
+         * AbortPipe), so the kill is issued here, once, the first time a reap
+         * finds the completion list empty on a disconnected fd. Raising the
+         * capability bit over the old behavior -- one URB of three returned and
+         * the other two lost -- was claiming a contract the code did not keep.
+         */
+        bool drain = disc && !u->disc_drained && u->pending_head != NULL;
+        if (drain)
+            u->disc_drained = true;
```
jserv: `disc_drained` is claimed on a pass that also has a completion to hand back, and that pass returns from the `if (rec)` block above the `if (drain)` block, so the kill never runs and every later pass sees the flag already true. A guest that leaves one completion unreaped and other URBs in flight when the device is unplugged gets that one completion and then `ENODEV`, with the in-flight URBs never handed back, which is what `CAP_REAP_AFTER_DISCONNECT` promises. Decide the drain only when there is no completion to return.

```suggestion
        bool drain =
            !rec && disc && !u->disc_drained && u->pending_head != NULL;
        if (drain)
            u->disc_drained = true;
```

src/syscall/usbdev.c
jotpalch: Confirmed, and reproduced. A backlog of 66001 completions built with the event thread idle (a never leader, 66000 refused followers, one DISCARDURB that drains the queue inline), then 65536 reaps to spend every byte the pipe could have held: poll answers 0x0 after 1002 ms, epoll_wait 0 after 1001 ms and select 0 after 1011 ms, while the REAPURBNDELAY issued after all three still hands a URB back. Your direction is what landed, with two departures I would rather state than leave silent. The level needs explicit per-slot state, a ready_token bool, because "write on the transition" has no memory of whether the token is already in the pipe, and a level with no state re-creates either the lost wake or an undrained spin. And the decision has to be taken and executed under async_lock: the old read sat outside it deliberately, but a decision taken there races a completion arming behind it and strands the token that completion just wrote, so the write and the nonblocking read now both sit inside, neither blocking and with nothing locked beneath (check-lock-order clean). The disconnect half is unchanged in behavior, since the token is taken and never given back, which is what a sticky unmaskable POLLERR|POLLHUP wants. This is the bill for removing the per-fd URB cap two rounds ago: the count was what made a byte per completion sound, and the pipe scheme did not go with it. One consequence is stated in internals.md rather than tested, because reaching it takes a guest that breaks the EPOLLET contract: a level is edged once per rise, not once per completion, so an EPOLLET consumer that reaps one URB per wake instead of draining to EAGAIN would wait here where Linux fires again. Restoring the byte per completion fails the three assertions above.

src/syscall/usbdev.c
jotpalch: Confirmed, and your patch is what landed: usbdev_memory_refund moved to urb_free_orphan_locked, beside the free(rec->buf) that releases the memory. The new scenario asks the budget a yes/no question rather than reading a counter, since 9 MiB fits in the 16 MiB allowance alone and does not fit beside 8 MiB. That probe is issued before the orphaning, while the record is orphaned and its 8 MiB buffer is still owned by an in-flight IOKit transfer, and again after the late abort frees it. With the refund back at the drain deadline the middle probe is accepted, which is the under-count you describe; that single assertion is what kills the mutation. Your reading of the reason given at that site is right, and the comment now says which of the three counters is which: nurbs and inflight_bytes are slot fields a reused slot must not see moved, and usbdev_memory_usage is neither, so a charge that never comes back is a transfer whose callback never arrived rather than a leak.

src/syscall/usbdev.c
jotpalch: Confirmed, driven exactly as you describe. ep0 belongs to no interface, so a SETINTERFACE kill for the claimed interface does not match its leader and only draining stops it: with that leader completing 400 ms into the drain and a wedged transfer on the interface missing the 2 s deadline, the ep0 follower is left queued with nothing in flight and its REAPURB never returns (d=(nil) after 2500 ms). Your suggestion is what landed, unchanged. All 256 keys is right rather than scoped, because the endpoints that need a restart are precisely the ones the kill did not match, and the success path below already kicks the same 256, so the two arms are now the same three lines. The kick cannot start something it should not: a per-interface kill that reaches this arm refuses the change, so the pipeRefs those followers reference stay valid, and a whole-device kill has already completed every queued record with -ENOENT at the top of the function, which leaves the loop nothing to start. Dropping the kick fails three assertions, the ep0 follower and the two disconnect rows it then spills into. For the record, this lands on usbdev_drain_for_change, which this series added last round in answer to cubic.

src/syscall/usbdev.c
jotpalch: Confirmed, and your patch is what landed, with the same pair added to usbdev_teardown_locked beside the disconnected reset it already had, since that is the other place the per-open async state is cleared. Reached through a slot reused across a close and a reopen, because the allocator takes the lowest free slot and two opens in a row land on the same one: with neither reset, the second open loses the URB only its own drain could return (2 back of 3, ep83's lost). I measured the redundancy rather than assuming it, and it cuts both ways: removing either reset alone leaves the lane green, and removing both is what fails it, so the assertion kills the bug but not each site on its own. That is stated in the commit message rather than left for a reader to discover. This one is inside CAP_REAP_AFTER_DISCONNECT, which this project's own review round had recorded as having no test at all.

src/syscall/usbdev.c
jotpalch: Confirmed, and your patch is what landed. Reproduced with the shape you name: one completion left unreaped at the unplug and two URBs still in flight, and the guest gets that completion and then -ENODEV, reaping 1 of 3, with the two in-flight URBs never handed back. The disconnect scenario now leaves that completion unreaped on purpose, since a pass with nothing to return never exercised the decision, which is why the old lane was green over this. The two disc_drained fixes are each killed by their own mutation and neither hides the other: reverting only the !rec && fails 4 assertions, both slot-reuse drains and both disconnect rows, and reverting only the two resets fails 1, the second open on the reused slot. Same provenance as the one above, the untested half of CAP_REAP_AFTER_DISCONNECT.

docs/internals.md
```diff
@@ -1005,27 +1012,247 @@ and a dispatcher restart would send it twice.
...
+### Deviations From Linux
+
+| usbfs behavior | elfuse behavior |
+|---|---|
+| `RESET` re-enumerates the device | clears the claimed pipes' stall state and returns 0; `USBDeviceReEnumerate` would tear down every open plugin handle |
+| isochronous URBs | `EINVAL` |
+| `DISCSIGNAL` delivers a signal on disconnect | signal and context stored, never delivered |
+| sync `BULK` on an interrupt endpoint works (converted to an interrupt URB) | `EINVAL`; the conversion exists only on the async URB path |
+| `BULK_CONTINUATION` unlinks the rest of the cascade on error | flag accepted, no cascade unlink |
+| `dup` of a usbfs fd works | `EBADF` |
```
jserv, line 1254: Address known limitations.

README.md
```diff
@@ -223,6 +223,13 @@ do.
   mask); the host scheduler picks the actual CPU.
 - `/proc`, `/dev`, and mount data are synthetic compatibility views,
   not host pass-throughs.
+- USB interfaces bound to an Apple class driver (CDC serial, HID, FTDI)
+  cannot be claimed; `CLAIMINTERFACE` reports `EBUSY`, and root-mode
+  device capture is not implemented. CDC serial devices are reachable by
+  opening the host's `/dev/cu.*` node instead.
+- USB mass storage will never be claimable, even with capture; macOS
+  does not release it.
+- Isochronous URBs are unimplemented; submitting one reports `EINVAL`.
```
jserv, line 232: Is it feasible to implement isochronous URB?

jotpalch:
The red check here is `Runtime (Release)` ([job 103707125751](https://github.com/sysprog21/elfuse/actions/runs/34750866978/job/103707125751)),
where `tests/test-shim-futex-toctou` printed `unimplemented syscall 2 (x0=0x200000000,
x1=0x80, ...)` and `FAIL: unexpected spin rc -38 (round 1655)`: the host dispatched the
guest's `FUTEX_WAIT` `svc` as syscall 2. It reproduces on `main` at 5b7741e with nothing
from this branch applied, at counts that came out equal to this branch at 2ee3279: 6/370
runs idle and 138/550 under eight concurrent compile jobs on each tree. The path is host
signal delivery and `rt_sigreturn`, which this branch does not touch -- its diff
is 26 files, none of them `src/syscall/signal.c`, `src/syscall/syscall.c` or
`src/runtime/futex.c`. Filed separately as #379.

Makefile
```diff
@@ -80,12 +80,40 @@ SRCS := \
     debug/log.c \
     debug/syscall-hist.c
 
+# The USB fixture seam (src/syscall/usbdev-fixture.h). usbdev.c calls it with no
+# conditional compilation of its own, so exactly one translation unit has to
+# define the entry points and the choice is made here: the stub in every build,
+# the loopback device model when USB_LOOPBACK_FIXTURE asks for it. Listing both
+# would be a duplicate-symbol link error, which is the property that keeps a
+# default build from quietly acquiring the model.
+ifeq ($(USB_LOOPBACK_FIXTURE),1)
```
jserv, line 89: `BUILD_FLAVOR` is `$(strip $(CFLAGS))` (mk/common.mk) and the fixture switch changes `SRCS`, not `CFLAGS`, so flipping it wipes nothing. Run the command mk/config.mk documents, `make USB_LOOPBACK_FIXTURE=1`, and the model is linked into `build/elfuse`; a later plain `make` finds `build/elfuse` newer than every object in the rebuilt `OBJS` and relinks nothing, so the default binary keeps the model until someone runs `make clean`. The `make check` path is safe because `elfuse-loopback` overrides `ELFUSE_BIN`, which is what the `nm` evidence measured. Either fold the fixture selection into the flavor stamp so the tree wipes on a flip, or make the documented manual command target `$(ELFUSE_LOOPBACK_BIN)` too.

src/syscall/poll.c
```diff
@@ -448,16 +483,74 @@ int64_t sys_ppoll(guest_t *g,
...
+                continue;
+            short before = host_fds[i].revents;
+
+            /* A disarmed entry's pipe interest was withdrawn (fd -1, revents
+             * 0), so probe with a hypothetical POLLIN and let the disconnect
+             * and reapable maps decide what the guest sees. Armed entries
+             * record whether their pipe actually fired BEFORE the writeback
+             * below overwrites revents; the disarm pass keys off that.
+             */
+            usb_pipe_fired[i] = !usb_disarmed[i] && (before & POLLIN) != 0;
```
jserv, line 574: `usb_pipe_fired` only recognises `POLLIN`, but a usbfs entry's readiness pipe reports `POLLHUP` once teardown closes its write end, which a concurrent `close()` of a polled fd does while `host_refs` still pins the read end. `POLLHUP` maps to no guest event (the fd is gone, so `fd_snapshot` fails and the maps are clear), `ret` drops back to 0, the entry is not disarmed because its pipe never "fired", and the `goto ppoll_retry` below re-polls the same standing `POLLHUP` immediately: an indefinite `ppoll` burns a core until a signal arrives. Treat `POLLHUP|POLLERR|POLLNVAL` as an invisible fire alongside `POLLIN` so the disarm pass withdraws the interest.

src/syscall/poll.c
```diff
@@ -1301,8 +1585,24 @@ void epoll_instance_free(void *inst)
...
 {
+    if (reg->usbdev) {
+        /* Completion-pipe readability alone does not mean "URBs reapable": the
+         * disconnect wake travels the same pipe, so EPOLLOUT needs the reapable
+         * map to agree (devio.c poll grants writability only while
+         * async_completed is non-empty). The usbfs fd never signals EPOLLIN.
+         * EOF/errors on the pipe only happen while the fd is being torn down --
+         * report the hangup pair Linux uses for a removed device.
+         */
+        if (kev->filter == EVFILT_READ && (reg->events & LINUX_EPOLLOUT) &&
```
jserv: The writable interest here is `reg->events & LINUX_EPOLLOUT`, while the poll and select half answers `LINUX_POLLOUT | LINUX_POLLWRNORM` and masks by the guest's events (`usbdev_poll_guest_revents`). A registration that asks for `EPOLLWRNORM` alone therefore never sees a completion here and would see it through `poll`, where Linux's `usbdev_poll` returns both bits and `ep_send_events` masks by what was requested. `epoll_mute_usb_wakes`'s `wantout` and the stamp path read the same single-bit mask. Use `EPOLLOUT|EPOLLWRNORM` as the writable mask in all three.

src/syscall/usbdev.c
```diff
@@ -673,190 +1134,1227 @@ static int64_t usbdev_build_pipe_map(usbdev_iface_t *fi)
...
+     * concurrent teardown + slot reuse (table -> async is the documented order,
+     * so the nested mark is safe).
+     */
+    unsigned idx;
+    uint64_t gen;
+    if (!usbdev_watch_unpack((uintptr_t) refcon, &idx, &gen))
+        return;
+    usbdev_t *u = &usbdev_fds[idx];
+    pthread_mutex_lock(&usbdev_table_lock);
+    if (u->used && (u->generation & USBDEV_WATCH_GEN_MASK) == gen) {
```
jserv: The peer walk applies `o->dead` to every slot it stamps (`usbdev_mark_key_disconnected_tlocked`); this test does not. A terminate notification IOKit had already dispatched can run after `usbdev_fd_cleanup` has torn the slot down and cleared the maps but before `usbdev_unref` clears `used` and `generation`, so `used` is still true and the packed generation still matches. It then stamps `discmap` for a guest fd number a sibling's `open()` may already have taken, leaving that fd permanently reported `POLLERR|POLLHUP` and answering `ENODEV`. Add `&& !u->dead`.

src/syscall/usbdev.c
```diff
@@ -673,190 +1134,1227 @@ static int64_t usbdev_build_pipe_map(usbdev_iface_t *fi)
...
+{
+    rec->next = NULL;
+    if (*tail)
+        (*tail)->next = rec;
+    else
+        *head = rec;
+    *tail = rec;
+}
+
+static void urb_pending_unlink(usbdev_t *u, usbdev_urb_t *rec)
```
jserv, line 1714: `urb_pending_unlink` and `usbdev_kick_ep_locked` both walk `pending_head` from the head, and the list is fd-wide rather than per endpoint. With the 256-record cap gone, a ring queued on one endpoint sits in front of every other endpoint's records, so each completion on a second endpoint walks the whole first queue twice while holding `async_lock` on the one event thread that carries every usbdevfs fd's completions. A guest can fill the 16 MB budget with about 87000 zero-length records and drive a second endpoint against them. A per-endpoint intrusive FIFO plus an in-flight pointer makes append, unlink and restart O(1); worth doing in this change, since removing the cap is what exposes it.

src/syscall/usbdev.c
```diff
@@ -673,190 +1134,1227 @@ static int64_t usbdev_build_pipe_map(usbdev_iface_t *fi)
...
+    usbdev_urb_t *r = u->pending_head;
+    while (r) {
+        usbdev_urb_t *nx = r->next;
+        if ((intf == NULL || r->intf == intf) && r->state == URB_INFLIGHT) {
+            urb_pending_unlink(u, r);
+            u->inflight--;
+            u->nurbs--;
+            u->inflight_bytes -= r->charge;
+            r->orphaned = true;
+            for (int fi_i = 0; r->intf && fi_i < USBDEV_MAX_IFACES; fi_i++) {
```
jserv: The orphan pin is guarded by `r->intf`, so an ep0 record (which carries `intf == NULL`) leaves this loop having incremented nothing. Nothing else records that an ep0 request is still outstanding: the record is already unlinked from `pending`, so a later `usbdev_kill_urbs_locked(u, NULL)` in `usbdev_teardown_locked` finds no in-flight URB, returns `drained`, and `CFRelease`s `dev_src` and `Release`s `u->dev` while `DeviceRequestAsyncTO` still owns the record. Reachable through `usbdev_do_reap`'s `overdue` arm, which orphans on one call and leaves the close to a later one. Give the device the same pin `usbdev_iface_t.orphans` gives an interface, and let `usbdev_teardown_locked` read it the way `usbdev_release_locked` reads the interface count.

src/syscall/usbdev.c
```diff
@@ -2098,6 +3792,661 @@ static int64_t usbdev_do_driver_ioctl(usbdev_t *u, guest_t *g, uint64_t arg)
...
+    uint64_t seq = rec->seq;
+    IOUSBInterfaceInterface800 **intf = rec->intf;
+    u->ep_aborting[key]++;
+    pthread_mutex_unlock(&u->async_lock);
+
+    /* The interface handle cannot be released concurrently: release paths need
+     * the entry lock this thread holds.
+     */
+    if (pipe == 0)
+        (void) (*u->dev)->USBDeviceAbortPipeZero(u->dev);
```
jserv: Both aborts here discard their `IOReturn`. `usbdev_arm_disconnect_watch` calls itself best effort and names "kIOReturnNoDevice detection on ops" as the fallback, and this is an op whose return says exactly that. With no interest notification armed, a `DISCARDURB` against a departed device aborts nothing, waits out the 2 s deadline, logs and returns 0 with the record still `URB_INFLIGHT` in `pending` and the fd unstamped; a following `REAPURB` then blocks forever, because the drain arms all require `disc`. Route both through `usbdev_ioret_op` (or test `usbdev_ioret_device_gone` and stamp) instead of casting to void.

src/syscall/usbdev.c
```diff
@@ -2098,6 +3792,661 @@ static int64_t usbdev_do_driver_ioctl(usbdev_t *u, guest_t *g, uint64_t arg)
...
+         * pending is the blocking arm's wait, so a wire that answers no abort
+         * left a REAPURBNDELAY loop returning -EAGAIN for ever while poll held
+         * POLLERR|POLLHUP -- measured against ep81:wedge(600000) as 1958386
+         * -EAGAIN and no -ENODEV in 4 s, where Linux's proc_reapurbnonblock
+         * answers connected(ps) ? -EAGAIN : -ENODEV and the guest tears down.
+         */
+        bool overdue = !rec && disc && pending && !block && u->disc_drained &&
+                       usbdev_now_ms() >= u->disc_orphan_at_ms;
+        if (overdue) {
+            usbdev_orphan_stalled_locked(u, NULL);
```
jserv, line 4574: `usbdev_orphan_stalled_locked` unlinks the in-flight record and marks it orphaned without appending it to `completed_head`, and the late callback frees it through the `rec->orphaned` arm rather than completing it, so the URB pointer is never handed back and the same pass then answers `-ENODEV` through the `disc && !pending` test below. That is the one path where `CAP_REAP_AFTER_DISCONNECT`, which `GET_CAPABILITIES` raises, does not hold. The 2 s bound is a deliberate trade, so this belongs in the XFAIL list beside `discard-latency` rather than being left to the capability bit, or the orphan should complete with the status `usb_kill_urb` leaves.

docs/testing.md
```diff
@@ -533,6 +535,97 @@ answers with its primary alone, because the backing half belongs to a stream
...
+`tests/usbdev-ioctl-departed.tbl`, one per usbdevfs ioctl, and a row that
+diverges from Linux carries both tuples: the lane fails if the elfuse answer
+stops matching the recorded one, and fails just as loudly if it starts matching
+Linux's, because an XFAIL nobody notices closing is a row that should have been
+retired. The generator refuses to emit at all unless every ioctl
+`usbdev_ioctl` dispatches has a row, so the recording cannot fall behind the
+surface it describes.
+
+USB-layer coverage is split by what it needs. `test-uevent-socket` needs no
+hardware and runs in the matrix like any other unit test, and so do the two
```
jserv: Neither usbdevfs lane is in the matrix: `grep usbdev tests/test-matrix.sh` is empty on both sides of this change, and `test-usbdev-urb-host` is a native macOS binary, so it could not run there in any case (the matrix runs aarch64 guest binaries under elfuse and qemu). `test-uevent-socket`, the lane this sentence compares them to, is registered. Say they run from `make check` instead, or register them.

jserv (changes requested):
Rebase latest `main` branch and resolve conflicts. Meanwhile, response to review messages.

README.md
jserv: ?

src/syscall/poll.c
jotpalch: I could not reproduce this one, so I left the code alone. Here is what I measured.

The fd this path sees is the readiness pipe, both ends owned by elfuse, and macOS reports its EOF as `POLLIN|POLLHUP` (`0x11`) rather than a bare `POLLHUP`. The `POLLIN` bit is set, so `usb_pipe_fired` is true and the disarm pass does withdraw the interest. I could not construct a bare `POLLHUP` on this fd.

**Validation** (3 runs each): a sibling thread closing the write end under a parked blocking `poll` returns `rc=1 revents=0x0011` at the close; a `poll` nothing ever wakes returns `rc=0 revents=0x0000` at its own timeout. Neither spins. A 3 s blocking poll costs under 1 ms of CPU.

If you know a path that produces `POLLHUP` without `POLLIN` here, I will take the `POLLHUP|POLLERR|POLLNVAL` treatment you suggested. I did not want to widen the test on a case I could not make happen. The measurement is recorded at the disarm site, `poll.c:605-625`.

src/syscall/poll.c
jotpalch: Confirmed and fixed as you described. The writable interest was a single-bit `LINUX_EPOLLOUT` mask while poll and select answered both bits and let `usbdev_poll_guest_revents` mask by request, so an `EPOLLWRNORM`-only registration saw nothing through epoll and would have seen it through `poll`.

`LINUX_EPOLL_WRITABLE` (`poll.c:1461`) is now `EPOLLOUT|EPOLLWRNORM` and backs all three sites you named, plus the generic-fd arming gate, which had the same single-bit problem for pipes and sockets: `poll.c:1804-1805, 1837, 2174, 2246, 2310, 2513, 2608`.

While there I checked the read half and it had the mirror of this bug, so `LINUX_EPOLL_READABLE` gets the same treatment for `EPOLLRDNORM`. `EPOLLPRI`, `EPOLLRDBAND` and `EPOLLMSG` stay unmodeled with the reason written beside them.

src/syscall/usbdev.c
jotpalch: Confirmed and fixed, though my first attempt overshot.

I put `!u->dead` on the peer walk itself, which also skipped `usbdev_mark_peers_disconnected_tlocked`. That walk is the last thing that can tell a device's *other* fds it is gone, so a sync-only peer on the same node stopped learning about the disconnect entirely: it polled `0x0000` where it had been getting `0x0018`.

The guard is now on the mark alone, which is what you pointed at:

```c
if (!u->dead)
    usbdev_mark_disconnected(u);
```

**Validation** (3 runs each, loopback fixture, terminate delivered inside the 2 s close drain): the recycled fd reads `0x0018` before and `0x0000` after; the sync-only peer reads `0x0018` in both. Both measurements are recorded at `usbdev.c:1568-1599`.

src/syscall/usbdev.c
jotpalch: Done, and you were right that removing the cap is what exposed it.

`pending_head`/`pending_tail` stays fd-wide for the callers that genuinely need every record, and each record now also carries `ep_next`/`ep_prev` (`usbdev.c:465`) into a per-endpoint FIFO, so append, unlink and restart no longer walk another endpoint's queue.

**Validation**, median submit-to-reap round trip on one endpoint against a second endpoint's pending depth, loopback fixture:

| depth | before | after |
|---|---|---|
| 0 | 0.017 ms | 0.016 ms |
| 20k | 0.356 ms | 0.014 ms |
| 100k | 2.688 ms | 0.021 ms |

The three link words grow the record from 136 to 160 bytes, which drops how many zero-length records fit the 16 MB budget from 123361 to 104857. Both numbers are recorded at `usbdev.c:440-461`.

src/syscall/usbdev.c
jotpalch: Confirmed and fixed the way you suggested: an ep0 record now pins the device the same way an interface record pins `usbdev_iface_t.orphans`, through `dev_orphans`/`dev_pin` (`usbdev.c:656, 2169, 1935-1936`), and `usbdev_teardown_locked` reads it the way `usbdev_release_locked` reads the interface count.

One thing I should be straight about: I reproduced the state defect, not the use-after-free. Under the loopback fixture, an ep0 wedge plus a terminate plus an overdue `REAPURBNDELAY` left `drained` true with the transfer still outstanding, and close printed no leaking-handle line where the same scenario on ep02 printed one. That is the `r->intf` guard doing exactly what you described. The actual `Release` racing a live `DeviceRequestAsyncTO` needs real IOKit to schedule the late callback, and the fixture frees a COM wrapper nothing dereferences afterwards, so ASan stayed clean there.

src/syscall/usbdev.c
jotpalch: Confirmed and fixed. Both aborts now route through the device-gone check instead of casting the `IOReturn` to void (`usbdev.c:4352-4353`), so a `DISCARDURB` against a departed device stamps the disconnect and completes the record.

**Validation**, fixture taught to answer `NoDevice` from `AbortPipe`: before, `DISCARDURB` returned 0 after 2003 ms with `revents=0x0` and no URB handed back, and the following `REAPURB` never returned. After, it returns 0 in under 500 ms with `revents=0x0018` and the URB comes back with `-ENOENT`.

You were right that `usbdev_arm_disconnect_watch`'s own comment already named this as the fallback. That is the part that bothered me most.

src/syscall/usbdev.c
jotpalch: Fixed rather than XFAILed. `usbdev_orphan_stalled_locked` (`usbdev.c:2175-2189`) now appends the orphan to `completed_head` carrying the status the abort path leaves, instead of handing it to the late callback's `orphaned` free arm, and the `overdue` branch re-reads `completed_head` on the same pass (`usbdev.c:4573-4590`) so the pointer is handed back before the `disc && !pending` test can answer `-ENODEV`.

I went this way rather than adding it to the XFAIL list because `destroy_all_async` is the upstream behavior and the record can outlive the hand-back safely; the two-second bound stays where it is.

**Validation**: with `ep02:wedge(600000)` and a terminate, `REAPURBNDELAY` returned `-ENODEV` at 2000 ms with 0 URBs handed back before, and now hands back one with `-ENOENT` and then answers `-ENODEV`. `CAP_REAP_AFTER_DISCONNECT` holds on this path now.

docs/testing.md
jotpalch: You are right, and the sentence was simply false. I retracted it rather than registering the lanes: `test-usbdev-ioctl` drives `ELFUSE_USB_FIXTURE` with no reference kernel to check it against, and `test-usbdev-urb-host` is a native macOS binary the aarch64-guest lanes cannot run at all. `docs/testing.md:548` now says they run from `make check` and says nothing about the matrix.

Makefile
jotpalch: Confirmed. `BUILD_FLAVOR` never moves when the flavor flips, so nothing invalidated the tree and a plain `make` after `make USB_LOOPBACK_FIXTURE=1` handed back a `build/elfuse` still linked against the model.

I took the second option you offered rather than folding the switch into the flavor stamp: `mk/config.mk` now points `ELFUSE_BIN` at `$(ELFUSE_LOOPBACK_BIN)` when `USB_LOOPBACK_FIXTURE=1`, so the two builds write different paths and neither can be a stale copy of the other. `.ci/check-usb-fixture-bin.sh` asks `make print-elfuse-bin` under both flavors and fails if they ever collapse back to one path.

**Validation**: the four-command sequence you gave (`make clean; make elfuse; make USB_LOOPBACK_FIXTURE=1 elfuse; make elfuse`) now leaves `build/elfuse` at 815984 bytes with no `_usbdev_fixture_lock` beside `build/elfuse-loopback` at 834304 with it. Before the change the same sequence left the model in `build/elfuse` until `make clean`.

README.md
jotpalch: Feasible, and I did not find a structural blocker. It is bounded work rather than a redesign.

Today's `-EINVAL` is deliberate: the URB-type switch hits `case LINUX_URB_TYPE_ISO:` and returns it with a `log_warn` (`usbdev.c:4089-4091`), and the synchronous paths reject iso endpoints the same way at `usbdev.c:3624` and `:4164`, citing `devio.c:1715-1717`.

The IOKit side is `ReadIsochPipeAsync`/`WriteIsochPipeAsync`: a buffer, a frame number from `GetBusFrameNumber`, a packet count, and an `IOUSBIsocFrame[]` (`frReqCount` in, `frActCount`/`frStatus` out per packet). No pre-registered buffers are needed; that is only the `LowLatency` variant, and libusb's own darwin backend skips it for isoch too. On the usbfs side the fields that have to map are `number_of_packets` (1-128), the `usbdevfs_iso_packet_desc[]` array, `URB_ISO_ASAP`, `start_frame`, and the per-packet `actual_length`/`status` that `compute_isochronous_actual_length` computes on completion.

Worth saying, since you asked before this existed: the async URB engine in this PR changes the calculus. `ReadPipeAsync`/`WritePipeAsync` completion plumbing, the per-endpoint FIFO with one URB in flight, and the orphan and late-callback handling under `async_lock` all run on the same completion-callback and run-loop model IOKit's isoch calls use, so that scaffolding carries over instead of having to be built first. What is left is the frame-list mapping both directions, `ISO_ASAP`/`start_frame` emulation via `GetBusFrameNumber` plus interval bookkeeping (libusb's `submit_iso_transfer` already does this, so the shape is proven), and per-packet status on reap.

I am not committing to it in this PR. If you want it scoped as a follow-up, say so and I will size it properly.

jotpalch:
Rebased onto `afdcfce9`; `pr-d-usb-async` is at `5e7bad47` and `pr-e-tty-alias` at `ca577212`. CI is 19 checks, 0 failures, with `Frama-C WP proofs` and `cubic` skipped by design.

I replied on each of the nine threads. Summary, in the order you filed them:

1. `Makefile:89` — took your second option; `ELFUSE_BIN` points at `elfuse-loopback` for that flavor, and a gate keeps the two paths from collapsing.
2. `poll.c:574` — I could not reproduce it. macOS reports this pipe's EOF as `POLLIN|POLLHUP`, so the existing test already disarms. Details and a question for you on the thread.
3. `poll.c:1599` — fixed, and the read half had the mirror of the same bug, so both masks are widened now.
4. `usbdev.c:1464` — fixed, but my first attempt overshot and broke peer visibility; the guard is on the mark alone now.
5. `usbdev.c:1714` — per-endpoint FIFO. 2.688 ms to 0.021 ms at depth 100k.
6. `usbdev.c:1951` — device-level pin added. I reproduced the state defect, not the use-after-free; said so on the thread.
7. `usbdev.c:4116` — both aborts honour the `IOReturn` now. Your point about our own comment naming the fallback was the sharpest one.
8. `usbdev.c:4574` — the orphan is handed back rather than XFAILed.
9. `docs/testing.md:549` — retracted; the claim was false.

One you did not ask about: the qemu reference lane caught an `EPOLLRDHUP`-only registration answering a spurious `EPOLLIN` where Linux answers 0 at the timeout. Fixed by arming the read filter with a low-water mark, so the divergence is gone rather than recorded as an XFAIL. That lane earning its keep is a direct consequence of #9.

The isochronous question is answered on the `README.md` thread.

jserv:
Thank @jotpalch for contributing!

PR #376, opened 2026-09-08
Return from wait as soon as the child exits

xalestar:
A guest that forks and waits pays about 100 ms per child even though the child is already gone. The exit reaches the parent, but nothing wakes the thread waiting for it, so every wait runs out its safety-net timeout.

`proc_process_exit()` appends the exit to the lifecycle registry and rings the cross-process doorbell, SIGUSR2. The parent's sigwait thread takes the signal, drains the record and pokes the vCPU's signal pipe, but nothing on that path touches `pid_cond`, which is where `sys_wait4` and `sys_waitid` sleep between re-checks. The comment at the first of those waits says the condvar handles normal exits and the timeout only catches edge cases. The condvar was never signaled: the only caller of `proc_mark_child_exited()` is the `CLONE_VFORK` branch of `sys_clone`, so an ordinary `SIGCHLD` fork never reached it.

Broadcasting there fixes `wait(2)` and leaves `waitpid(2)` on a single pid exactly as slow, because the two re-check different things. The `pid == -1` loop imports the registry after waking, and the registry already carries the exit. The single-pid loop polls the child's host process with `WNOHANG`, and that process outlives the guest exit by about 0.7 ms while it tears down its VM, so the poll still reports it running when the wakeup arrives and nothing rings a second time. The second commit imports the registry there too, then re-enters through the table lookup at the top of `sys_wait4`.

That loop polls with no lock held and takes `pid_lock` only to sleep, so a broadcast landing between the two finds nobody parked and evaporates. `wait4` cannot be the predicate that closes the window, because it answers about the host process and still reports "running" at the moment of the broadcast; the doorbell sets `entry->exited` under `pid_lock` before it broadcasts, so the loop reads that instead, inside the same critical section the sleep releases.

The third commit sweeps unreaped host children from the watchdog tick. `proc_deferred_reap_poll()` collects the host process of a child whose status the guest already took from the registry, and `sys_wait4`'s entry was its only caller, so a guest that takes its last child's status and never waits again left that child a zombie for the rest of the run. The SIGALRM branch of the sigwait thread already runs on its own, which bounds the strand to a period or two without a new thread or any cost on the wait path. Two paths then have to hold up under a second concurrent sweeper: the poll claims an entry under `pid_lock` before its unlocked `wait4`, so two callers cannot reap the same host pid and have the second take a freshly spawned child that inherited it, and `proc_wait_autoreap_children()` now accounts and consumes an entry the registry already marked exited instead of only deactivating it.

Reproduction, against the Alpine rootfs under `externals/test-fixtures`:

```
build/elfuse --sysroot $RF $RF/bin/busybox sh -c \
    'i=0; while [ $i -lt 20 ]; do (:) ; i=$((i+1)); done'
```

Delay from the child's own last timestamp to the parent's return from wait, median of 8 rounds, measured by `tests/test-fork-wait-latency.c`:

| | before | after | Linux 6.12 |
|---|---:|---:|---:|
| `wait(2)` | 98.7 ms | 0.20 ms | 0.02 ms |
| `waitpid(2)` on one pid | 99.5 ms | 0.20 ms | 0.02 ms |

Shell workloads, wall clock, median of three runs, same tree built with and without the change:

| | before | after |
|---|---:|---:|
| fork only x20 | 2300 ms | 160 ms |
| fork and exec x100 | 11390 ms | 1420 ms |
| md5sum over 34 files | 2620 ms | 390 ms |
| find piped to wc | 120 ms | 40 ms |
| command substitution x30 | 440 ms | 450 ms |

The last row is a control. `$(...)` learns the child is done by reading the pipe to EOF, so `wait4` finds a zombie and never reaches the condvar. It does not move, and it already sat at the ratio the other rows reach after the change.

The test measures the delay rather than the status, because a missed wakeup returns the right answer late and a status check cannot see it. The child is held at a pipe read until the parent is on its way into wait, so the exit cannot land before the parent reaches the syscall and answer from the table lookup at the top of `sys_wait4` without ever reaching the condvar. It asserts on the median for the same reason: one such iteration would carry a minimum. It is registered in `test-matrix.sh`, so the same assertion is checked against Linux 6.12.

`make check`, `make check-format` and `make lint` are clean. `make check-tsan` and `make check-asan` are clean. Rebased on 73d8246.

### DIFF
--- a/src/syscall/proc.c
+++ b/src/syscall/proc.c
@@ -2263,6 +2263,25 @@ static int64_t proc_wait_autoreap_children(int pid, int options)
     }
 }
 
+/* Has the local table already recorded this child's exit? The lifecycle
+ * registry carries the guest exit before the child's host process finishes
+ * tearing down, so this can be true while waitpid(2) on the host pid still
+ * reports the process as running.
+ */
+static bool proc_guest_child_exited_locked(int64_t guest_pid)
+{
+    proc_entry_t *entry = proc_find_guest_entry(guest_pid);
+    return entry && entry->exited;
+}
+
+static bool proc_guest_child_exited(int64_t guest_pid)
+{
+    pthread_mutex_lock(&pid_lock);
+    bool exited = proc_guest_child_exited_locked(guest_pid);
+    pthread_mutex_unlock(&pid_lock);
+    return exited;
+}
+
 /* sys_wait4. */
 
 /* Reap host children whose terminal status the guest already consumed from the
@@ -2563,9 +2582,41 @@ int64_t sys_wait4(guest_t *g,
                         return -LINUX_EINTR;
                     struct timespec ts;
                     timespec_deadline_in_ms(&ts, 100);
+
+                    /* Import before the sleep as well as after it. The poll
+                     * above runs unlocked, so a broadcast landing between it
+                     * and the lock below finds nobody parked and evaporates:
+                     * pid_cond keeps no count. The doorbell sets entry->exited
+                     * under pid_lock before it broadcasts, so re-reading the
+                     * table inside the same critical section the sleep releases
+                     * is what closes that window. The import itself stays
+                     * outside the lock, like the poll: it touches the registry
+                     * file, and pid_lock holds only bounded table walks.
+                     */
+                    lifecycle_import_children();
                     pthread_mutex_lock(&pid_lock);
+                    if (proc_guest_child_exited_locked(gpid)) {
+                        pthread_mutex_unlock(&pid_lock);
+                        return sys_wait4(g, pid, status_gva, options,
+                                         rusage_gva);
+                    }
                     pthread_cond_timedwait(&pid_cond, &pid_lock, &ts);
                     pthread_mutex_unlock(&pid_lock);
+
+                    /* The doorbell reports the guest exit about a millisecond
+                     * before the child's host process becomes reapable, so the
+                     * poll above still says "running" when the wakeup arrives.
+                     * The registry carries the exit by then, so importing it
+                     * copies the status across and flags the host reap for
+                     * proc_deferred_reap_poll(), after which the table lookup
+                     * at the top of sys_wait4 answers without waiting for the
+                     * host process at all. Without this the wakeup is wasted
+                     * and the wait costs the full 100 ms timeout.
+                     */
+                    lifecycle_import_children();
+                    if (proc_guest_child_exited(gpid))
+                        return sys_wait4(g, pid, status_gva, options,
+                                         rusage_gva);
                 }
             }
             if (ret > 0) {
--- a/src/syscall/proc.c
+++ b/src/syscall/proc.c
@@ -2953,6 +2953,26 @@ static void *preempt_thread_main(void *arg)
             atomic_store_explicit(&g_external_guest_signal, 1,
                                   memory_order_release);
             drain_external_guest_signal();
+
+            /* The doorbell is also how a fork child reports that it exited.
+             * sys_wait4 and sys_waitid sleep on pid_cond between re-checks, and
+             * nothing on this path signaled it, so a waiting parent only
+             * re-checked when its 100 ms safety-net timeout expired: every
+             * fork/wait pair paid that timeout in full even though the child
+             * was already gone.
+             *
+             * Broadcast under pid_lock, not beside it. A sleeper holds the lock
+             * across its predicate check and only releases it inside
+             * pthread_cond_timedwait, so taking the lock here means the
+             * broadcast cannot land in the window between the two and be lost.
+             * pid_lock is a leaf (see the lock-order block in
+             * syscall/internal.h) and every critical section under it is a
+             * bounded table walk, so this cannot invert an order or park the
+             * sigwait thread on someone else's blocking call.
+             */
+            pthread_mutex_lock(&pid_lock);
+            pthread_cond_broadcast(&pid_cond);
+            pthread_mutex_unlock(&pid_lock);
             wakeup_pipe_signal();
         } else if (sig == SIGALRM) {
             static uint64_t last_progress;
--- a/src/syscall/proc.c
+++ b/src/syscall/proc.c
@@ -2242,10 +2242,27 @@ static int64_t proc_wait_autoreap_children(int pid, int options)
             } else if (ret == 0) {
                 still_active = true;
             } else {
+                /* ret < 0, in practice ECHILD: the host child is gone. It can
+                 * be gone because proc_deferred_reap_poll() took it from the
+                 * watchdog tick, which reaches a single-threaded guest parked
+                 * here and not only one racing a second guest thread. When the
+                 * registry already carried the exit, the entry holds the status
+                 * and the rusage, so finish it the way the ret == host_pid
+                 * branch does: deactivating alone drops the child's time from
+                 * RUSAGE_CHILDREN and leaves the lifecycle row to be imported
+                 * again.
+                 */
                 proc_entry_t *entry =
                     proc_find_host_guest_entry(host_pid, guest_pid);
-                if (entry)
+                if (entry && entry->exited) {
+                    proc_account_entry_locked(entry);
                     entry->active = false;
+                    pthread_mutex_unlock(&pid_lock);
+                    lifecycle_consume(guest_pid);
+                    pthread_mutex_lock(&pid_lock);
+                } else if (entry) {
+                    entry->active = false;
+                }
             }
         }
         pthread_mutex_unlock(&pid_lock);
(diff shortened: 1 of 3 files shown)


src/syscall/proc.c
```diff
@@ -2263,6 +2263,20 @@ static int64_t proc_wait_autoreap_children(int pid, int options)
     }
 }
 
+/* Has the local table already recorded this child's exit? The lifecycle
+ * registry carries the guest exit before the child's host process finishes
+ * tearing down, so this can be true while waitpid(2) on the host pid still
+ * reports the process as running.
+ */
+static bool proc_guest_child_exited(int64_t guest_pid)
```
jserv, line 2294: Now that the tick calls `proc_deferred_reap_poll()`, it can reap a host zombie while a single-threaded guest sits in `proc_wait_autoreap_children()`, which until now needed a second guest thread in `sys_wait4` to reach. That function's `ret < 0` branch only clears `entry->active`: it skips the `proc_account_entry_locked()` and `lifecycle_consume()` that both sibling branches run, so the child's time is dropped from `RUSAGE_CHILDREN` and its lifecycle row survives to be re-imported.

The recheck this diff adds to the single-pid `sys_wait4` branch fits there too: look the entry up, and if it now reads `exited`, account and consume it rather than just deactivating.

src/syscall/proc.c
```diff
@@ -2953,8 +2999,37 @@ static void *preempt_thread_main(void *arg)
...
+             * Broadcast under pid_lock, not beside it. A sleeper holds the lock
+             * across its predicate check and only releases it inside
+             * pthread_cond_timedwait, so taking the lock here means the
+             * broadcast cannot land in the window between the two and be lost.
+             * pid_lock is a leaf (see the lock-order block in
+             * syscall/internal.h) and every critical section under it is a
+             * bounded table walk, so this cannot invert an order or park the
+             * sigwait thread on someone else's blocking call.
+             */
+            pthread_mutex_lock(&pid_lock);
```
jserv, line 3070: The invariant this comment states holds for the `pid == -1` loop and for `sys_waitid`, which both keep `pid_lock` across the scan and the `pthread_cond_timedwait`. It does not hold for the single-pid loop this PR's second commit targets: that one polls `wait4(host_pid, WNOHANG)` with no lock held and takes `pid_lock` only to sleep, so a broadcast landing between the poll and the lock is lost and the waiter still spends the full 100 ms.

The window is the table scan plus one `wait4`, and the exit has to land inside it, which is why the min-of-8 measurement does not show it. Checking `proc_guest_child_exited(gpid)` under `pid_lock` immediately before the `pthread_cond_timedwait`, or an exit generation counter read under the same lock, closes it.

src/syscall/proc.c
```diff
@@ -2953,8 +2999,37 @@ static void *preempt_thread_main(void *arg)
...
             wakeup_pipe_signal();
         } else if (sig == SIGALRM) {
+            /* sys_wait4's entry is the only other caller, so a guest that takes
+             * its last child's status and never waits again strands that
+             * child's host process as a zombie for the rest of the run. The
+             * tick already exists and already runs on this thread, so sweeping
+             * from it bounds the strand to a period or two without adding a
+             * thread, a wakeup, or a cost on the wait path.
+             */
+            proc_deferred_reap_poll();
```
jserv, line 3082: `proc_deferred_reap_poll()` releases `pid_lock` around its `wait4(host_pid, WNOHANG)`, and this makes the alarm thread a second concurrent caller for a guest that has only one thread. Two callers can read the same `host_reap_pending` slot, capture the same `host_pid`, and both call `wait4` on it; after the first reaps it macOS can hand that pid to a freshly spawned guest fork child, and the second call can reap that one instead and discard its status.

The paired-lookup guard already in that function protects the flag clearing, not the reap itself. Claiming the entry under `pid_lock` before dropping it, so exactly one caller owns the reap, covers both.

tests/manifest.txt
```diff
@@ -124,6 +124,7 @@ test-clone3                    # diff=skip
 test-clone-childtid
 test-fork-exec $TESTDIR/echo-test
 test-fork-lowbase
+test-fork-wait-latency
```
jserv: The scope note at the top of this file keeps a portable, cross-checkable test out of it, and this one is registered in `test-matrix.sh` already: the PR's own point is that the same assertion runs against Linux 6.12.

The entry also costs more than duplication. `Fork edge cases` matches `SANITIZER_SECTIONS` in `mk/tests.mk`, so a 20 ms wall-clock assertion now runs in the ASAN, UBSAN and TSAN lanes, and nothing on that path gives it any allowance: `driver.sh` re-runs only on a timeout, not on a failed assertion, and `test_host_is_busy` is consulted only by `test-bench-guardrail.sh`. Dropping this line leaves the coverage where the note puts it.

tests/test-fork-wait-latency.c
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * Both spellings are exercised because they take different paths inside elfuse.
+ * wait(2) settles from the lifecycle registry, waitpid(2) on one pid polls the
+ * child's host process, and the host process outlives the guest exit by about a
+ * millisecond -- so a wakeup that satisfies the first can still leave the
+ * second waiting out the whole timeout.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
```
jserv: `stdlib.h`, `string.h` and `sys/resource.h` are unused here: no `malloc`/`free`/`exit`, no `str*` or `mem*`, and no `rusage`. `stdio.h`, `time.h`, `unistd.h` and `sys/wait.h` cover everything the file calls.

tests/test-fork-wait-latency.c
```diff
@@ -0,0 +1,125 @@
...
+#define MAX_GAP_MS 20.0
+#define ITERS 8
+
+static double ms_between(const struct timespec *a, const struct timespec *b)
+{
+    return (double) (b->tv_sec - a->tv_sec) * 1000.0 +
+           (double) (b->tv_nsec - a->tv_nsec) / 1e6;
+}
+
+/* Fastest observed gap from the child's final timestamp to the parent's return
```
jserv: "a missed wakeup costs every iteration the full timeout" is the premise the minimum rests on, and `sys_wait4` disproves it: it opens with `lifecycle_import_children()`, and the single-pid scan answers from `proc_table[i].exited` before it ever reaches the poll loop. So on an unfixed tree any iteration where the parent reaches the syscall after the child's host process finished tearing down (the ~0.7 ms this PR measures) returns immediately, and one such iteration in eight makes the whole test pass.

Releasing the child only once the parent is parked in `wait`, and asserting on the median rather than the minimum, gives the regression nowhere to hide without making an unlucky trial fatal.

tests/test-fork-wait-latency.c
```diff
@@ -0,0 +1,125 @@
...
+            close(fd[0]);
+            clock_gettime(CLOCK_MONOTONIC, &c);
+            (void) !write(fd[1], &c, sizeof(c));
+            _exit(0);
+        }
+
+        close(fd[1]);
+        int status;
+        if (specific_pid)
+            waitpid(pid, &status, 0);
```
jserv: Neither wait checks its return value or `status`. A `waitpid` that fails with `EINTR` leaves the child unreaped and still records a fast gap, so an error reads as a pass, and the leftover child is then visible to the plain `wait()` in the other case. Retry on `EINTR`, require the returned pid to be the child and `WIFEXITED(status) && WEXITSTATUS(status) == 0`, and treat anything else as a failure.

src/syscall/proc.c
xalestar: Fixed in 3248719. The `ret < 0` branch now looks the entry up and, when it reads `exited`, runs `proc_account_entry_locked()` and `lifecycle_consume()` before deactivating, matching the `ret == host_pid` branch. Deactivation alone stays for an entry the registry never marked.

src/syscall/proc.c
xalestar: Fixed in b7d9177. The loop now imports outside the lock, then takes `pid_lock` and re-reads the table before parking, so the check and the sleep share one critical section. `wait4` could not serve as that predicate: it answers about the host process, which is still running at the moment of the broadcast, so it would report "not yet" and park anyway. `proc_guest_child_exited()` gained a `_locked` variant for the call under the lock; the import stays outside it because it touches the registry file.

src/syscall/proc.c
xalestar: Fixed in 3248719, though not with a separate claimed field. Clearing `host_reap_pending` before dropping the lock is itself the claim, since the scan is by that flag alone: a second sweeper sees the slot as not pending and skips it. The flag goes back only when `wait4` returned 0, under the same paired lookup, so a slot that got reused in the window is not asked to reap a live child.

tests/manifest.txt
xalestar: Dropped in b7d9177; the PR no longer touches `manifest.txt` at all. The binary still builds — `TEST_C_SRCS` is a `wildcard tests/*.c` — so `test-matrix.sh` runs it unchanged.

tests/test-fork-wait-latency.c
xalestar: Fixed in b7d9177. Down to `stdio.h`, `time.h`, `unistd.h` and `sys/wait.h`; `test-harness.h` already pulls `errno.h` for `FAIL`. The median needs a sort, but eight elements is an insertion sort in place rather than a reason to bring `stdlib.h` back.

tests/test-fork-wait-latency.c
xalestar: Fixed in b7d9177, both parts: the child now blocks on a pipe read until the parent has written the go byte and is entering `wait`, and the assertion is on the median of eight.

One measurement worth recording against your reasoning. The escape path is real in the code — the single-pid scan does answer from `proc_table[i].exited` before the poll loop — but I could not get it to fire: on 73d8246 the old test failed 5 runs out of 5, and the new one reports a 98.7 ms median for `wait` and 99.5 ms for `waitpid`. A parent that forks and immediately waits seems to reach the syscall well before the child's host process finishes tearing down. So the minimum was not hiding the regression on this machine, but it was resting on a premise the code does not guarantee, and nothing about a slower or busier host keeps it resting there.

tests/test-fork-wait-latency.c
xalestar: Fixed in b7d9177. `wait_for()` retries on `EINTR`, requires the returned pid to be the child, and requires `WIFEXITED(status) && WEXITSTATUS(status) == 0`; anything else fails the iteration instead of recording its gap. The child also reports failure through its own exit status now, so a short read or a failed write on its side cannot pass either.

xalestar:
All seven addressed, and the branch is squashed back to three commits — `manifest.txt` is untouched now, so the diff is `src/syscall/proc.c`, the new test, and its `test-matrix.sh` entry.

`make check`, `make check-format`, `make lint`, `make check-tsan` and `make check-asan` are clean. The new test passes under QEMU against Linux 6.12 and fails on 73d8246, which is the property the rewrite was meant to keep.

jserv:

PR #377, opened 2026-09-12
Return from a sleep as soon as a wake arrives

xalestar:
A guest thread parked in `nanosleep` does not notice a queued signal until its own 100 ms recheck quantum expires. `sys_nanosleep` splits the interval into 100 ms chunks and tests `signal_pending` between them, so the delay is uniform over 0 to 100 ms, about 50 ms on average, against 0.12 ms under Linux 6.12.

The wake those chunks stand in for already exists: queueing a guest signal, `exit_group` and `futex_interrupt` all reach `wakeup_pipe_signal`, and `io.c` and `poll.c` already join it. A pure timeout has no descriptor to watch alongside the pipe, so it parks on a condition variable added to that same call instead, which leaves all five existing callers of the signal covering it unchanged.

A signal, exit_group and an execve teardown end the sleep on the wake. An execve handed to the leader keeps the `SLEEP_CHUNK_NS` cadence: a handoff issues a stop request thousands of times a second, and answering every ring turns each into a guest-visible EINTR that the guest answers by reissuing the rest of its interval; the numbers that cost are below. One broadcast also wakes every sleeper, so the sleep claims a process-directed signal under `sig_lock` before reporting EINTR, reusing the pick `rt_sigsuspend` makes, moved into a helper in the first commit.

## Reproduction

`tests/test-nanosleep-signal-latency.c` holds one thread in a sleep and times the gap between `pthread_kill` and the handler running, for both `nanosleep` and `clock_nanosleep(TIMER_ABSTIME)`. The delay before the signal walks forward across iterations: a fixed delay phase-locks every iteration to one offset inside the quantum and reports that point as though it were the whole distribution.

On the parent commit the test fails with 60.6 ms and 28.9 ms medians against its 20 ms ceiling. With this change it passes, and it also passes under the QEMU reference lane.

`tests/test-nanosleep-process-signal.c` parks four threads in `nanosleep` and sends the group one `kill()`. Without the claim 2 to 4 sleepers return EINTR while the handler runs once; with it exactly one does, as under QEMU.

## Measurements

| | before | after | Linux 6.12 |
|---|---|---|---|
| signal to handler, median of 12 | 50 ms | 0.069 ms | 0.12 ms |
| 1 us sleep, min/median of 200 | 7 / 8 us | 4 / 5 us | |
| 10 us sleep | 18 / 20 us | 15 / 17 us | |
| 100 us sleep | 132 / 135 us | 130 / 133 us | |
| 1 ms sleep | 1051 / 1273 us | 1111 / 1274 us | |

`nanosleep`, `pselect` and `pthread_cond_timedwait_relative_np` are indistinguishable in resolution at 200 samples (3 / 2 / 3 us for a 1 us request), so the primitive swap costs nothing. Short sleeps come out slightly faster because the old path rebuilt a `timespec` and ran the saturating remainder arithmetic once per chunk.

The stop cadence is what the 400 ms guest sleep in `test-exec-handoff` measures, under handoff pressure. It only separates the two revisions inside the full sanitizer suite; run on its own the test sits at 1800 to 2400 ms either way.

Same tree built with and without the change, alternating.

| under `make check-tsan` | interruptions | sleep |
|---|---|---|
| before the change | 3 | 432 ms |
| handoff stop tested on every wake | 29390 | 3150 ms |
| this revision | 3 | 403 ms |

## Commands

- `make check`, `make check-tsan`, `make check-asan`, `make check-ubsan`: all exit 0, with `check-tsan` green over four consecutive runs
- `make check-format`, `make lint`: exit 0
- `make check-ascii`: clean across 427 source files
- `test-wakeup-pipe-host` goes from 5 to 9 checks, covering the counter advancing, a stale counter not parking, a wait holding until the wake, and a signal releasing a parked wait

Rebased on 4023647.

---
## Summary by cubic
Guest threads parked in `nanosleep` or `clock_nanosleep(TIMER_ABSTIME)` used to notice queued signals only when their 100 ms recheck quantum expired, adding about 50 ms average signal-to-handler latency. They now wake immediately via `wakeup_pipe_signal()`, cutting that to about 0.07 ms without regressing short sleeps.

**Bug Fixes**

- Pure timeout sleeps now park on a condition variable driven by `wakeup_pipe_signal()`, since they have no descriptor to watch; existing callers are unchanged.
- Adds an atomic wake counter so a wake that lands between a caller's predicate check and its park is not lost.
- Recomputes remaining sleep time from `CLOCK_MONOTONIC`, because the wait can return early on any process-wide wake.
- Keeps stop-request checks on the old recheck cadence; testing them on every wake would make exec handoffs produce a guest-visible EINTR storm.
- A process-directed signal is now claimed under `sig_lock` by one sleeper, so a single `kill()` interrupts one thread instead of every sleeper.
- Adds latency and process-signal tests and expands the host wakeup-pipe test to cover the condvar wait.

<sup>Written for commit 76ba1ad0c55c0241bf2fb2eda4a6bce0f34541b2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/377?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


xalestar:
The elapsed-time and test-robustness points were both real.

- `time.c`: elapsed now comes from the saturating `host_timespec_to_ns_sat` on both samples rather than a raw multiply.
- Latency test: `handler_ran`/`handler_ns` are `_Atomic` with explicit ordering, the `pthread_create` and `pthread_kill` return codes are reported separately from a handler that never ran, and the handshake loop has a deadline so a stranded sleeper fails the test instead of hanging the lane.
- `test-wakeup-pipe-host`: the generation snapshot now happens before the ringer thread exists. Taking it afterwards could hand the wait a generation that already matched, which would have parked the full 5 s on a loaded machine.

The `pthread_create` failure message in `test-wakeup-pipe-host` is left as it was, to match the reader thread a few lines above it in the same file.

Separately, the TSAN lane was red on the first push and that was mine. The counter read sat under `wake_lock`, so every sleep in the process serialized against an exec handoff's ringer. Making the counter atomic removes the read from the lock. The 400 ms guest sleep `test-exec-handoff` runs under handoff pressure now takes 1001 to 1028 ms against 986 to 1002 ms without the change, alternating the two builds back to back on an idle machine.

xalestar:
The TSAN red on the previous push came from moving both halves of the sleep's predicate onto the wake. They are not the same condition. A tagged counter over one run of `test-exec-handoff` under handoff pressure:

```
passes=45000 stop=22465 sig=20
```

Almost every early return was `thread_stop_requested`, not a queued guest signal. An exec handoff issues a stop request thousands of times a second, so testing it on every wake turns each ring into a guest-visible EINTR, and the guest answers each one by reissuing the rest of its interval.

This push tests `signal_pending` on every wake, which is what the change is for, and keeps `thread_stop_requested` on the SLEEP_CHUNK_NS cadence it had before, which is unchanged behaviour for that half.

The measurement only separates the two versions inside the full sanitizer suite. Run on its own the test sits at 1800 to 2400 ms either way; under `make check-tsan`:

| | interruptions | sleep |
|---|---|---|
| before the change | 3 | 432 ms |
| previous push | 29390 | 3150 ms |
| this push | 3 | 407 ms |

`make check-tsan` is green over four consecutive runs here, as are `check`, `check-asan`, `check-ubsan`, `check-format`, `lint` and `check-ascii`. Signal to handler is unchanged at 0.053 to 0.074 ms.

src/syscall/wakeup-pipe.h
```diff
@@ -30,3 +40,22 @@ int wakeup_pipe_read_fd(void);
...
 /* Consume every queued byte. The pipe is nonblocking, so this cannot park. */
 void wakeup_pipe_drain(void);
+
+/* The wake counter, read before testing whatever predicate the caller sleeps
+ * on. wakeup_wait_ns() takes the value back and returns at once when the
+ * counter has moved since, which is what closes the window between a caller's
+ * predicate check and its park: a wake landing there bumps the counter instead
+ * of being broadcast to nobody.
+ */
+uint64_t wakeup_generation(void);
```
jserv: Is it better to use `counter`?

xalestar: Renamed to `wakeup_counter()`, along with `wake_counter` and `seen_counter`, in the latest push.

alanhc (approved):

src/syscall/time.c
```diff
@@ -117,8 +118,40 @@ static int64_t interruptible_sleep_ns(guest_t *g,
...
+         * A signal is what the guest asked to be interrupted for. A stop
+         * request is this tree asking its own threads to leave the guest, and
+         * an exec handoff issues one thousands of times a second. Testing it on
+         * every wake turns each of those into a guest-visible EINTR, and the
+         * guest reissues the rest of its interval after every one:
+         * test-exec-handoff measured the 400 ms sleep it runs under handoff
+         * pressure taking 3150 ms across 29390 returns, against 432 ms across 3
+         * on this cadence.
+         */
+        bool stop_due = elapsed_ns >= stop_due_ns;
```
jserv, line 147: `thread_stop_requested()` is three conditions, and only one of them is the handoff the EINTR measurement above is about. `thread_stop_is_leader_work_only()` already separates them, so `exit_group` and an `execve` teardown could be answered on the wake that is now in hand while the handoff keeps its `SLEEP_CHUNK_NS` cadence: `stop_due || !thread_stop_is_leader_work_only()`. As written, a guest thread in a multi-second sleep still holds up teardown for up to 100 ms after the wake has already reached it.

src/syscall/time.c
```diff
@@ -117,8 +118,40 @@ static int64_t interruptible_sleep_ns(guest_t *g,
...
+        if (stop_due)
+            stop_due_ns = elapsed_ns + SLEEP_CHUNK_NS;
+
+        if (signal_pending() || (stop_due && thread_stop_requested())) {
```
jserv: `signal_pending()` reads the shared process-directed pending set, which `self_pending_locked` unions into every thread's view, and nothing reserves that bit for one thread. Before this change each sleeper polled at its own phase inside the quantum, so whichever thread reached `signal_deliver` usually consumed the bit before its siblings next looked. One broadcast now wakes every sleeper at the same instant, so with several threads in `nanosleep` a single `kill(pid, SIGUSR1)` returns `EINTR` from all of them while only one runs the handler; Linux interrupts only the chosen thread. Claim the shared bit for one eligible thread under `sig_lock` before reporting the interruption, and cover it with a multi-sleeper test.

tests/test-nanosleep-signal-latency.c
```diff
@@ -0,0 +1,207 @@
...
+        if (rc != 0) {
+            errno = rc;
+            return GAP_NO_THREAD;
+        }
+
+        /* Let the sleeper reach the host wait before the signal is sent. A
+         * signal that arrives first is answered by the check at the top of the
+         * sleep loop, which is fast whether or not the wait itself is joined.
+         */
+        struct timespec settle = {0, (150 + i * 11) * 1000000L};
```
jserv, line 148: The settle delay is a guess, not a handshake. If the sleeper has not reached the wait when `pthread_kill` fires, the signal is answered by the `signal_pending()` test at the top of the sleep loop and the iteration records a fast gap on the unfixed tree too. A slow `pthread_create` under the sanitizer lanes can put several iterations in that state, and once half of them land there the median stops measuring the thing this test exists for. Have the sleeper publish an atomic flag immediately before it calls `nanosleep` and spin on that flag before the settle delay.

tests/test-wakeup-pipe-host.c
```diff
@@ -81,5 +104,36 @@ int main(void)
...
+    pthread_t ringer;
+    if (pthread_create(&ringer, NULL, ringer_main, NULL) != 0) {
+        fprintf(stderr, "FAIL thread: cannot start the ringer thread\n");
+        return 1;
+    }
+    t0 = now_ms();
+    wakeup_wait_ns(5000000000LL, before_ring);
+    long long waited_ms = now_ms() - t0;
+    pthread_join(ringer, NULL);
+    host_check(waited_ms < 1000, "a signal releases a parked wait",
```
jserv, line 142: Both new waits assert only an upper bound, so a `wakeup_wait_ns` that returns at once and never parks passes them: it would report about 0 ms, well under the 1000 ms ceiling, and the stale-counter check above passes for the same reason. Add the lower bound that separates the two, something like `waited_ms >= 40`, since the ringer sleeps 50 ms before it signals.

src/syscall/time.c
xalestar: The sleep tests `thread_stop_requested() && (stop_due || !thread_stop_is_leader_work_only())`, so exit_group and an execve teardown end it on the wake while the handoff keeps the `SLEEP_CHUNK_NS` cadence. The stop test also runs before the signal test, because a thread an execve reaps would otherwise claim a process-directed signal into its private set and take it along. Under `make check-tsan` the handoff sleep in `test-exec-handoff` is 403 ms across 3 interruptions, the same as without the change.

src/syscall/time.c
xalestar: Confirmed, and this branch introduced it. `tests/test-nanosleep-process-signal.c` parks four threads in `nanosleep`, blocks SIGUSR1 in main and sends one `kill(getpid(), SIGUSR1)`. On 2d6482a, 4, 2 and 2 sleepers returned EINTR across three runs while the handler ran once; with the fix it is 1 and 1 in five runs, matching Linux 6.12 under QEMU.

The sleep calls `signal_claim_interruption()`, which takes the signal under `sig_lock` through the private-then-shared pick `rt_sigsuspend` already made, moved into `signal_claim_waking_locked()` in its own commit. The pick carries the disposition filter, so a SIG_IGN or default-ignored signal does not interrupt a sleep.

tests/test-nanosleep-signal-latency.c
xalestar: The sleeper sets `sleeper_armed` immediately before `nanosleep` or `clock_nanosleep`, and `median_gap` spins on it, bounded at `SLEEP_SEC`, before the settle delay.

tests/test-wakeup-pipe-host.c
xalestar: Added `waited_ms >= 40`. `t0` also moved ahead of `pthread_create`: taken after it, a ringer already partway through its 50 ms could put a correct wait under the bound on a slow start.

tests/test-nanosleep-process-signal.c
```diff
@@ -0,0 +1,125 @@
...
+    sigset_t set;
+    sigemptyset(&set);
+    sigaddset(&set, SIGUSR1);
+    pthread_sigmask(SIG_BLOCK, &set, NULL);
+
+    TEST("kill() interrupts exactly one of several sleepers");
+
+    pthread_t th[SLEEPERS];
+    for (long i = 0; i < SLEEPERS; i++) {
+        if (pthread_create(&th[i], NULL, sleeper, (void *) i) != 0) {
```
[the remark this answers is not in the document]
xalestar: `send_one_signal()` returns the `pthread_create` error and the caller sets `errno = rc` before `FAIL()`.

tests/test-nanosleep-process-signal.c
```diff
@@ -0,0 +1,125 @@
...
+            return 1;
+        }
+    }
+
+    long long give_up = now_ms() + 2000;
+    while (atomic_load_explicit(&armed, memory_order_acquire) < SLEEPERS &&
+           now_ms() < give_up)
+        ;
+    /* The flag is set just before nanosleep; this lets each reach the park. */
+    usleep(150000);
```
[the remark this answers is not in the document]
xalestar: A user-space flag cannot mark a thread as already inside `nanosleep`, so the settle stays and the outcome it cannot rule out is retried instead. The scenario runs up to three times, and only `runs == 1 && woke == 0` repeats it. Any result with more than one EINTR fails on the first attempt, and the test still fails on 2d6482a with 3 sleepers interrupted.

henrybear327:
This unblocks the LTP test! Thanks for this improvement! :)

Issue #379, opened 2026-09-13
`rt_sigreturn` overwrites the restored X8, so a signal taken on an `svc` re-enters it as syscall 2


jotpalch:
## Symptom

`tests/test-shim-futex-toctou` fails intermittently on `main`. The host prints:

```
WARN  src/syscall/syscall.c:2841: unimplemented syscall 2 (x0=0x200000000, x1=0x80, x2=0x0, x3=0x0, x4=0x0, x5=0x0)
FAIL: unexpected spin rc -38 (round 24992)
FAIL: 1 unexpected returns in the spin phase
```

`x1=0x80` is `FUTEX_WAIT|FUTEX_PRIVATE_FLAG` and `x0` is the futex word, so the guest
issued `futex` and the host dispatched syscall 2. The guest gets `-ENOSYS` out of a
call it made as `futex`. Nothing here is specific to that test's `mprotect` race; the
reproduction below has no `mprotect` in it.

## Mechanism

Measured by running the reproduction below under a print-only probe copy of `main`,
which records per thread the last signal delivery, the last `rt_sigreturn`, and the
state of the restart record. It caught ten bad-X8 events across nine failing runs; nine
of the ten are identical, e.g.:

```
INSTR bad x8=2 elr=0x400948 restarted=0 | mark site=2 seq=792 armed=0 elr=0x400944 rewound=0x0 nr=0 | cancel out=0 seq=789 elr=0x0 rewound=0x0 | arm seq=0 elr=0x0 nr=0 | seq=792
INSTR deliv el0=1 pc=0x400944 x8=98 seq=790 | sigret pc=0x400944 x8=98 seq=791
WARN  src/syscall/syscall.c:2931: unimplemented syscall 2 (x0=0x4a1aa8, x1=0x80, x2=0x0, x3=0x0, x4=0x0, x5=0x0)
FAIL: futex returned -38 at round 404994 (1 total)
```

`deliv` is the delivery that took the EL0-preemption branch, `sigret` the `rt_sigreturn`
that returned through its frame, `bad` the SVC that reached dispatch as syscall 2;
`arm seq=0` and `cancel out=0` say `syscall_restart_arm` never ran on that thread and
the cancel found nothing armed. The tenth is a different shape -- delivery on the EL1
branch (`el0=0`), the frame already carrying X8 = 2, the saved PC still on the `svc`.
Same symptom, nothing armed there either, and I did not chase how that PC came to be
[...]

Issue #378, opened 2026-09-13
A guest signal does not interrupt futex, epoll_wait or a finite select


Max042004:
A thread parked in these host waits notices a queued signal only when its
own polling quantum or timeout ends. Linux returns EINTR at once.

PR #377 fixed nanosleep not returning on a queued signal; futex,
epoll_wait and select have the same defect. Measured on Apple M4, 
against Linux 6.12 under qemu with HVF.

| signal to handler, median of 12 | elfuse | Linux 6.12 |
|---------------------------------|--------|------------|
| nanosleep (after #377) | 0.080 ms | 0.044 ms   |
| futex, finite timeout  | 40.7 ms  | 0.044 ms   |
| futex, indefinite      | 40.6 ms  | 0.044 ms   |
| epoll_wait, indefinite | 125 ms   | 0.047 ms   |
| select, 1 s timeout    | 779 ms   | 0.050 ms   |
| epoll_wait, 1 s timeout| 777 ms   | 0.047 ms   |

futex re-checks signal_pending on a 100 ms quantum
(FUTEX_OS_SYNC_POLL_CAP_NS) and an indefinite epoll_wait on a 200 ms kevent
slice. A finite select or epoll_wait hands the guest's whole timeout to
pselect or kevent with no wakeup pipe fd and no slice, so the signal waits
for the timeout.

Reproducer: a sibling thread parks in each wait, the main thread sends
SIGUSR1 with pthread_kill after a settle delay, and the handler stamps
CLOCK_MONOTONIC.

xalestar:
I'll start with select and epoll_wait, which can join the wakeup pipe the way poll already does, and take futex after #377 lands.

jserv:
Utilize [frama-c-mcp](https://github.com/sysprog21/frama-c-mcp) to introduce formal verification of guest signal handling.

PR #380, opened 2026-09-13
Report EINTR from one waiter per signal

xalestar:
`kill(2)` queues one signal for the whole thread group, and Linux `complete_signal()` wakes only the thread it picks to run the handler. On elfuse, several threads blocked in `read`, `ppoll`, `pselect6`, `epoll_pwait` or `flock` could each return EINTR for that one signal, while only one of them ran the handler.

These waits tested the shared pending set with `signal_pending_interruption()` and never took the signal. One byte on the wakeup pipe returns every thread parked in host `poll()` at the same time, so each saw the same bit and reported EINTR, and whichever reached `signal_deliver()` first ran the handler. The others handed their guests an EINTR that no handler explains. `nanosleep` had the same defect and claims the signal since #377; this moves the other waits onto that claim: `io_wait_fd_timed_or_interrupted`, `tty_drain_interruptible`, `io_retry_backoff`, `ppoll`, `pselect6` and `epoll_pwait`, and in the third commit `futex_os_sync_wait`, `futex_poll_signal_relock` and `rt_sigtimedwait`.

The second commit fixes an older defect cubic pointed out in review. `ppoll`, `pselect6` and `epoll_pwait` put the caller's mask back before the syscall epilogue delivers, and `signal_deliver()` tests that restored mask, so a signal that only the wait's sigmask unblocks ended the wait with EINTR and was never delivered: a guest that blocks the signal everywhere else and unblocks it only inside the wait got EINTR on every retry and no handler. Linux keeps the temporary mask until delivery and restores the original at sigreturn. These waits now do the same when they return EINTR for a signal they claimed, through the `saved_blocked` path `rt_sigsuspend` already uses, and the syscall epilogue puts back a saved mask that no handler frame took, as `restore_saved_sigmask()` does. `epoll_pwait` still claims only when kevent produced nothing, so ready events keep outranking the signal.

The third commit changes what a claim is, after review. It used to move the signal into the claimer's private set, which made it thread-directed for good: a mask restored over it, or the claimer exiting, stranded it where no other thread could take it. A claim now leaves the signal in the shared set and records the thread, and the other threads leave it alone only while that thread can still deliver it; any dequeue, `thread_deactivate()`, exec and a fork snapshot release it. With a wait's saved mask pending, delivery takes the claimed signal first, so the frame carrying that mask is its own. The fourth commit lets ready descriptors outrank a pending signal in `ppoll` and `pselect6`, as `epoll_pwait` already did, and has all three make one non-blocking pass when a signal is already pending, so a finite `pselect6` or `epoll_pwait` no longer sits out its whole timeout for it.

FUSE stays on `signal_pending_interruption()`, which it needs for `restart_out`. The comment in `io_retry_backoff()` said that function filtered `SA_RESTART`; it never did, and the rewritten comment no longer says so.

## Reproduction

`tests/test-wait-process-signal.c` parks four threads in each of `read`, `ppoll`, `pselect6`, `epoll_pwait`, `flock` and `FUTEX_WAIT`, blocks the signal in the main thread, and sends the group one `kill()`, up to four rounds per wait. It fails if more than one waiter returns EINTR, or if no round caught a waiter parked. The futex kind returned 3 EINTRs for one `kill()` on the second commit and passes with the third.

On 4872bb3 it failed 5 of 5 runs, with every wait failing at least once. With this change it passes 10 of 10, and 3 of 3 under the QEMU reference lane.

`tests/test-wait-sigmask-signal.c` waits in each of `ppoll`, `pselect6` and `epoll_pwait` under a sigmask that is the only one unblocking SIGUSR1, and checks that the handler runs, that it runs under the wait's mask, and that the original mask is back after both an EINTR and a ready return. Without the second commit 8 of those 12 checks fail. Two more checks come with the fourth commit: a signal already pending at entry ends a 5 s wait at once, and a ready descriptor beats a pending signal, which stays pending. On the second commit they fail for `pselect6` and `epoll_pwait` (5005 and 5001 ms spent) and for `ppoll` and `pselect6` (EINTR instead of the ready count). With all four commits the 18 checks pass, as they do under QEMU.

## Measurements

A separate program with the same shape, 30 rounds each, `SA_RESTART` off:

| EINTR per `kill()` at 4 waiters | before | after | Linux 6.12 |
|---|---|---|---|
| `read` | 1 to 3 (42 over 30) | 1 (30) | 1 (30) |
| `ppoll` | 1 to 4 (80 over 30) | 1 (30) | 1 (30) |
| `epoll_pwait` | 1 to 3 (48 over 30) | 1 (30) | 1 (30) |

The handler ran exactly once per round in every row.

The temporary-mask case: every thread blocks SIGUSR1, and the main thread loops for 1 s on each wait with an empty sigmask after one `kill()`. On 4872bb3 and on the first commit the handler never runs and the waits return EINTR 4 (`pselect6`), 7 (`ppoll`) and 4 (`epoll_pwait`) times. With the second commit, as under Linux 6.12, the handler runs once on the waiting thread and each wait returns EINTR once.

## Not in this change

- `SA_RESTART` is still not honored for a signal-interrupted wait, so the thread that does run the handler still sees EINTR from `read()`. `tests/test-poll.c` already documents that.
- A signal that arrives during the wait still reaches an indefinite `epoll_pwait` only on its 200 ms kevent slice (about 96 ms on average here), and a finite `pselect6` or `epoll_pwait` only at its deadline, since none of them joins the wakeup. That is #378, which will add the wake on top of this claim.
- With two signals pending that only a wait's mask unblocks, elfuse builds one frame per epilogue, so unless the first handler makes a syscall the second stays pending under the restored mask instead of nesting under the first as on Linux. The reply on the review thread at `signal.c` describes the two ways to close it; it is left for a follow-up.

## Commands

- `make check`: exit 0, all 99 unit tests pass. An earlier run failed only `test-shim-futex-toctou` with the -38 from #379; looped 300 times it fails 19 on 4872bb3 and 9 with this change, so it is not caused here
- `make check-tsan`: exit 0, all 72 tests pass, no ThreadSanitizer reports
- `test-nanosleep-process-signal`, `test-sigsuspend`, `test-sigtimedwait`, `test-signalfd`, `test-poll`, `test-epoll`, `test-epoll-mt`, `test-epoll-edge`, `test-signal`, `test-signal-thread`, `test-tgkill-directed`: pass with all four commits
- Each commit builds on its own
- `make check-format`, `make lint`: exit 0
- `make check-ascii`: clean across 438 source files

Rebased on 4872bb3.


src/syscall/signal.c
```diff
@@ -1658,9 +1663,16 @@ bool signal_claim_interruption(void)
...
     blocked = atomic_load_explicit(thread_blocked_ptr(), memory_order_acquire);
-    bool claimed = signal_claim_waking_locked(blocked) != 0;
+
+    /* Claim only what this thread can deliver once @saved_blocked is back. A
+     * signal only the temporary mask unblocks still ends the wait, but it stays
+     * shared: signal_deliver() tests the restored mask, and binding it here
+     * would take it away from a sibling that could run the handler.
+     */
+    bool woke = signal_claim_waking_locked(blocked | saved_blocked) != 0 ||
+                signal_set_would_wake_locked(self_pending_locked() & ~blocked);
```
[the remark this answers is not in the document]
xalestar: This was already the behavior before #380: on 4872bb3, a signal that only the wait's sigmask unblocks ends `pselect6`, `ppoll` or `epoll_pwait` with EINTR, the original mask is put back before the syscall epilogue delivers, and the handler never runs. With every thread blocking SIGUSR1 and one `kill()`, a 1 s loop on each wait ran the handler 0 times and got EINTR 4, 7 and 4 times, identical on 4872bb3 and on e4d1c02; Linux 6.12 runs the handler once and returns EINTR once.

Fixed in 3c9c38c by following Linux. When one of these waits returns EINTR for a signal it claimed, it leaves the temporary mask installed and hands the original to the handler frame as `uc_sigmask`, through the `saved_blocked` path `rt_sigsuspend` already uses, so the handler runs under the wait's mask and `rt_sigreturn` restores the original. Any other return restores the mask immediately, and the syscall epilogue puts back a saved mask that no frame took, as `restore_saved_sigmask()` does. With the mask still in force at delivery the plain claim is correct, so `signal_claim_interruption_masked()` is gone.

The same probe now runs the handler once, on the waiting thread, with one EINTR for all three waits. `tests/test-wait-sigmask-signal.c` covers it: the handler runs, it runs under the wait's mask, and the original mask is back after both an EINTR and a ready return. It fails 8 of 12 checks without the change and passes with it and under QEMU.

tests/test-wait-process-signal.c
```diff
@@ -0,0 +1,247 @@
...
+            epfds[i] = epoll_create1(EPOLL_CLOEXEC);
+            if (epfds[i] < 0 ||
+                epoll_ctl(epfds[i], EPOLL_CTL_ADD, pipes[i][0], &ev) < 0)
+                return -1;
+        }
+    }
+
+    pthread_t th[WAITERS];
+    for (long i = 0; i < WAITERS; i++) {
+        if (pthread_create(&th[i], NULL, waiter, (void *) i) != 0)
```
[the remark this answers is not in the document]
xalestar: Fixed in 6e53c10: `send_one_signal()` returns the `pthread_create` error number, and a partial create releases and joins the threads it already started before returning.

src/syscall/io.c
```diff
@@ -235,11 +235,12 @@ int64_t io_retry_backoff(unsigned *backoff_us)
...
-     * were reachable by neither. signal_pending_interruption already filters
-     * SIG_IGN, default-ignore, and SA_RESTART, so it cannot manufacture an
-     * EINTR the guest would not have seen.
+     * were reachable by neither. SIG_IGN and default-ignore signals do not
+     * count. Teardown is tested first so a thread an execve reaps never claims
+     * a process-directed signal, and the claim keeps a sibling contending for
+     * the same lock from also reporting EINTR for it.
      */
-    if (thread_stop_requested() || signal_pending_interruption(NULL))
+    if (thread_stop_requested() || signal_claim_interruption())
```
jserv, line 243: The claim makes the converted waits agree with each other, but the waits that were not converted still test the shared set non-destructively, so one `kill(2)` can still hand out two EINTRs. `futex_os_sync_wait` returns `-LINUX_EINTR` on a bare `signal_pending()`, `futex_poll_signal_relock` wakes a bucket waiter the same way, and `rt_sigtimedwait` decides on `signal_set_would_wake_locked()`. A guest with one thread in `futex_wait` and one in `ppoll` still sees exactly the defect this fixes. Either move those onto the claim as a follow-up, or say in the commit that the invariant holds only among the waits listed here.

src/syscall/poll.c
```diff
@@ -383,9 +384,13 @@ int64_t sys_ppoll(guest_t *g,
...
 
-        /* Check for process/thread interrupts after waking. */
-        if (thread_stop_requested() || futex_interrupt_consume() ||
-            signal_pending_interruption(NULL)) {
+        /* Check for process/thread interrupts after waking. The signal is
+         * claimed, not just seen: every thread in poll() returns on the same
+         * wakeup byte, and only one of them may report EINTR for it.
+         */
+        bool stopped = thread_stop_requested() || futex_interrupt_consume();
+        signal_interrupted = !stopped && signal_claim_interruption();
```
jserv: The claim runs whatever `poll()` returned, so a `ppoll` that already has ready `revents` still spends the process's one signal, reports EINTR, and drops the events. `sys_epoll_pwait` guards this with `nready <= 0` for the reason its own comment gives, and Linux `do_poll()` consults `signal_pending()` only under `if (!count)`. Gating the claim on a result with no guest-visible events, here and in `sys_pselect6`, would make the three agree.

src/syscall/signal.c
```diff
@@ -1122,6 +1122,34 @@ void signal_restore_blocked(uint64_t saved)
     pthread_mutex_unlock(&sig_lock);
 }
 
+void signal_defer_restore_blocked(uint64_t saved)
```
jserv, line 1187: `deliver_signal_locked` hands `saved` to whichever signal it delivers first, not to the one the wait claimed, and a claimed process-directed signal is now indistinguishable from a thread-directed one in `tpending`. Thread A blocks SIGUSR2 outside the wait, unblocks it in the wait's mask, and claims a process-directed SIGUSR2; a SIGUSR1 aimed at A before the epilogue is lower-numbered, so `signal_deliver_one` picks it, stamps `uc_sigmask` with `saved`, and `rt_sigreturn` restores the mask that blocks SIGUSR2. That SIGUSR2 is left in A's private set where no other thread can take it, and it dies with A. A reservation on the shared set, released on mask restore or teardown, would keep process-directed signals out of `tpending`.

tests/test-wait-process-signal.c
```diff
@@ -0,0 +1,258 @@
...
+        if (rc != 0) {
+            errno = rc;
+            FAIL("setup failed");
+            return;
+        }
+        if (runs != 1 || woke > 1)
+            break;
+    }
+
+    if (runs == 1 && woke <= 1) {
```
jserv: `woke == 0` passes, and that is a round in which nothing was observed: the signal ran its handler before any waiter parked, so the round proves nothing about how many waiters report EINTR. Every round can come out that way and the lane stays green with the defect present. Track whether any round saw `runs == 1 && woke == 1` and fail when none did, so a vacuous run is inconclusive rather than a pass.

src/syscall/poll.c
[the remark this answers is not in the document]
xalestar: A signal already pending when the wait starts now ends it after one non-blocking pass, in `ppoll`, `pselect6` and `epoll_pwait` (9f2a5c7). The pass still runs, so ready descriptors win. `pselect6: pending at entry` and `epoll_pwait: pending at entry` in `tests/test-wait-sigmask-signal.c` took 5005 and 5001 ms against their 5 s timeout on 3c9c38c and finish under 1 s now, as under QEMU. A signal that arrives partway through a finite `pselect6` or `epoll_pwait` is still noticed only at the deadline, because those waits carry no wakeup descriptor; that part is #378.

tests/test-wait-sigmask-signal.c
```diff
@@ -0,0 +1,193 @@
...
+    atomic_store_explicit(&handler_runs, 0, memory_order_relaxed);
+    atomic_store_explicit(&handler_saw_usr2_blocked, 0, memory_order_relaxed);
+    pthread_t th;
+    int rc = pthread_create(&th, NULL, killer, NULL);
+    if (rc != 0) {
+        errno = rc;
+        FAIL("pthread_create failed");
+        return;
+    }
+    int ret = wait_masked(kind, p[0], epfd, &wait_mask);
```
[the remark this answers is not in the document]
xalestar: Fixed in 9f2a5c7: the thread that sends SIGUSR1 writes a byte to the pipe if the wait has not returned within 5 s, so a signal that never reaches the waiter fails the check with a ready return instead of hanging the lane.

tests/test-wait-sigmask-signal.c
```diff
@@ -0,0 +1,193 @@
...
+    sigemptyset(&wait_mask);
+    sigaddset(&wait_mask, SIGUSR2);
+
+    snprintf(label, sizeof(label), "%s: handler runs", name);
+    TEST(label);
+    atomic_store_explicit(&handler_runs, 0, memory_order_relaxed);
+    atomic_store_explicit(&handler_saw_usr2_blocked, 0, memory_order_relaxed);
+    pthread_t th;
+    int rc = pthread_create(&th, NULL, killer, NULL);
+    if (rc != 0) {
```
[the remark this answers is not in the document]
xalestar: Fixed in 9f2a5c7: both failure branches close `epfd` and the pipe before returning.

src/syscall/io.c
xalestar: Moved them onto the claim in 8dee3fb. `futex_os_sync_wait`, `futex_poll_signal_relock` and the `rt_sigtimedwait` interrupt test, which claims outside its own wait set, now take the signal the way the other waits do; each site tests teardown before it claims. `tests/test-wait-process-signal.c` gains a futex kind: four threads in `FUTEX_WAIT` returned 3 EINTRs for one `kill()` on 3c9c38c, and exactly one now, as under QEMU.

src/syscall/poll.c
xalestar: Gated in 9f2a5c7. `ppoll` and `pselect6` consume the futex interrupt and claim only for a pass that produced no guest-visible events: the wakeup descriptor does not count, and POLLNVAL entries and the entries evaluated through `select()` do, so the three waits agree. `ready beats signal` in `tests/test-wait-sigmask-signal.c` makes a descriptor ready, sends the signal, and expects the ready count with the signal still pending under the restored mask. On 3c9c38c `ppoll` and `pselect6` returned EINTR and ran the handler; they pass now, as under QEMU.

src/syscall/signal.c
xalestar: Done as you suggest, in 8dee3fb. A claim no longer moves the signal into `tpending`: it stays in the shared set with the claiming thread recorded, and every other thread's view of the shared set (delivery, `rt_sigtimedwait`, signalfd, the claim itself) leaves it out for as long as the claimer does not block it. Once a mask is restored over it, or the claimer exits (`thread_deactivate` releases its claims and raises attention), it is an ordinary shared signal again. Any dequeue from the shared set releases the claim, and exec and a fork snapshot start with none. When a wait has left its saved mask, delivery takes the signal this thread claimed before a thread-directed one, so the frame that carries `saved` belongs to the claimed signal.

I could not make the interleaving you describe, a SIGUSR1 landing between the claim and the epilogue, deterministic from a guest, so no test pins that order; the wait, signal, sigtimedwait and signalfd tests pass with the change, and `make check-tsan` is clean.

One difference from Linux is left for a follow-up.

The case: two signals are pending that only the wait's mask unblocks.
- Linux delivers both before returning to user space, nesting a frame for each.
- elfuse builds one frame per return. The first handler's frame carries the saved mask, so its `rt_sigreturn` restores the original mask over the second signal. Unless that handler makes a syscall, the second stays pending: a process-directed one in the shared set, where another thread can still take it, a thread-directed one in `tpending`.

The one-frame limit is `syscall_return_epilogue`'s. Its comment calls `signal_deliver` a once-per-epilogue call by contract, since a second call "would stack a frame on the handler the first one just installed". That rule came in with the ptrace-stop work in 348e0e5, and its message gives no correctness reason against nesting.

Two ways to close it:
1. Nest frames as Linux does. This changes the one-frame rule, and has to get the X8 frame-drop marker (#379), where a ptrace stop is taken, and the altstack check right.
2. Chain at `rt_sigreturn`: when the frame being restored carries a wait's saved mask, keep the handler's mask for one more delivery and pass the saved mask to that frame through the same `saved_blocked` path. Both handlers run, but in the opposite order from Linux.

I'd rather do 2 as a separate PR after this one than grow this one further.

tests/test-wait-process-signal.c
xalestar: Fixed in 8dee3fb: `check_kind()` fails a kind whose rounds never caught a waiter parked, meaning no round with one handler run and one EINTR, so a run that tested nothing no longer passes.

src/syscall/signal.c
```diff
@@ -770,6 +811,19 @@ bool signal_attention_needed(void)
...
 bool signal_pending_interruption(bool *restart_out)
 {
+    /* Same lock-free fast path as signal_pending(): the fd waits ask this
+     * before every host wait.
+     */
+    uint64_t hint =
+        atomic_load_explicit(&sig_pending_hint, memory_order_acquire);
+    uint64_t hint_blocked =
+        atomic_load_explicit(thread_blocked_ptr(), memory_order_acquire);
+    if ((hint & ~hint_blocked) == 0) {
```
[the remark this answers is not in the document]
xalestar: The lock-free check is only the first gate. When it does not rule a signal out, `signal_pending_interruption()` takes `sig_lock` and returns false unless `self_pending_locked() & ~blocked` is non-zero, and `self_pending_locked()` reads the shared set through `shared_visible_locked()`, which leaves out a signal another thread has claimed while that thread can deliver it. A waiter that lost the claim therefore gets false, keeps its normal slice or timeout, and does not reach the non-blocking pass; the hint only costs it one locked recheck per pass, which is the same false-positive cost `signal_pending()` already documents for its fast path.

tests/test-wait-sigmask-signal.c
```diff
@@ -0,0 +1,289 @@
...
+{
+    int wfd = *(int *) arg;
+    usleep(100000);
+    kill(getpid(), SIGUSR1);
+    for (int i = 0; i < 500; i++) {
+        if (atomic_load_explicit(&wait_returned, memory_order_acquire))
+            return NULL;
+        usleep(10000);
+    }
+    ssize_t n = write(wfd, "x", 1);
```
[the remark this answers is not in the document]
xalestar: Fixed in 2ee7170: after joining the sending thread, `check_kind()` drains a byte the watchdog left in the pipe, so a failed first wait no longer turns the later checks into ready returns.

jserv:

PR #383, opened 2026-09-14
Restore the guest X8 when the shim drops its frame

### LINKED ISSUE #379: `rt_sigreturn` overwrites the restored X8, so a signal taken on an `svc` re-enters it as syscall 2

jotpalch:
Fixes #379.

A signal delivered while the vCPU is at EL0 with PC on an `svc` that has not
executed yet is saved with that PC, correctly and without the drop-frame marker.
The handler returns, `signal_rt_sigreturn` restores all 31 GPRs from the frame
and then writes the marker `2` over `X8`, and `exec_drop_frame` pops and `ERET`s
without restoring a register -- so the `svc` resumes with `X8 = 2` and runs as
syscall 2. Dispatch finds nothing and answers `-ENOSYS` to a call the guest made
as something else.

The value goes in the frame's own `X8` slot, and `exec_drop_frame` reloads `X8`
from it before the pop. That slot is EL1-only memory on this vCPU's exception
stack, unshared, holding nothing live once the frame is being dropped, and
retired by the pop, so it cannot collide the way a register does. No second
register can carry it instead: every GPR on that tail is meant to be final EL0
state, which is why `X7`'s ptrace request is already left unwritten there.

The slot alone does not close the window, which was measured rather than argued.
A signal delivered after the `rt_sigreturn` but before the vCPU resumes
snapshots the live registers, where `X8` is still the marker and `ELR_EL1` is
the unexecuted `svc`. `rt_sigreturn` therefore also parks the value it published
and `deliver_signal_locked` takes it in place of the register, bounded by a
forget at the single `hv_vcpu_run` call site. With only the slot in place the 32
nested rounds fail and nothing else does; with the slot and the park but no
forget, the rounds pass and the stale-park phase fails.

`tests/test-shim-sigreturn-x8.c` is deterministic and needs no race: 64 rounds
on the syscall route, a stale-park phase, and two rounds that reach the same
drop tail from HVC #9 instead. All 64 rounds and both HVC #9 rounds fail on
`afdcfce`, entering with `0x2` where the guest held `0xa5`; all four phases pass
here. Every assertion is plain Linux signal ABI, so it is registered in
`tests/test-matrix.sh` rather than exempted -- the qemu lane needs a reference
kernel and was not run here, so neither `EXPECTED_BASELINES` floor moves.

The reproduction the issue carries, a raw `FUTEX_WAIT` loop with a signal
landing on it, was run interleaved so both binaries met the same machine: 5
failures in 400 runs on `afdcfce` against 0 in 400 here, and 161 in 400 against
0 in 400 under eight concurrent compile jobs. That machine is shared, so the
unloaded column is a floor rather than a rate -- the same pair read 53 in 400
while another build was running. The loaded column is the one to read.

`scripts/check-svc-tails.py` gains the two structural rules: the drop tail must
load `X8` from the frame before it pops, at the offset the gate reads out of
`signal.c` rather than trusting the two sides to match by eye, and every
`hv_vcpu_run` call site must forget the parked value first. It cannot gate the
substitution itself, and the commit says so -- that half rests on the test,
which fails deterministically on it.

Two pre-existing neighbours are recorded rather than fixed: the same tail hands
`X9` through `X11` out to EL0 after a page-table syscall in the same epilogue,
and a ptrace stop taken inside the `rt_sigreturn` epilogue already shows the
tracer the marker through `PTRACE_GETREGSET`.

---
## Summary by cubic
Fixes #379: a signal delivered while the vCPU has PC on an unexecuted `svc` no longer resumes that `svc` with the drop-frame marker (`2`) in `X8`, so the guest's syscall runs as its own number instead of as syscall 2 and returning `-ENOSYS`.

**Bug Fixes**
- `exec_drop_frame` now reloads `X8` from the frame's own `X8` slot before popping; `signal_rt_sigreturn` publishes the restored value there.
- `signal_rt_sigreturn` also parks the restored `X8`; a delivery in the same epilogue takes the parked value, the run loop forgets it before every `hv_vcpu_run()`, and `signal_repark_sigreturn_x8()` re-keys it after a `PTRACE_INTERRUPT` stop changes the resume PC.
- `handle_brk` now keeps its frame live across the `HVC`, reloads `X8` from the slot, then pops, so a `SIGTRAP` handler enters with the guest's `X8` instead of the marker.
- Both `HVC #9` W^X tails branch to the same drop tail, so their `SIGSEGV` handlers now enter with the guest's `X8` rather than the marker.
- Pre-existing shadowing on the syscall-return path is left as-is: delivery after a page-table syscall still hands `X9` through `X11` to EL0, and ptrace stops on this tail still show the marker via `PTRACE_GETREGSET`.

**Validation**
- Adds deterministic `tests/test-shim-sigreturn-x8.c` covering plain and nested returns, the stale-park window, both W^X tails, and the BRK tail.
- `scripts/check-svc-tails.py` now checks the `X8` reload offset against `signal.c`, requires `signal_forget_sigreturn_x8()` above every `hv_vcpu_run()`, and holds both `exec_drop_frame` and `handle_brk` to the reload.
- The reproduced `FUTEX_WAIT` loop failed 5/400 on base vs 0/400 here, and 161/400 vs 0/400 under concurrent build load.
- The qemu matrix lane wasn't run, so `EXPECTED_BASELINES` is unchanged.

<sup>Written for commit a83f704ab98960a5a519faf0b216141379bf2919. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/383?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src/syscall/signal.c
```diff
@@ -2304,6 +2411,20 @@ static int deliver_signal_locked(hv_vcpu_t vcpu,
...
+    /* X8 is not the guest's while the drop-frame marker occupies it, and an
+     * rt_sigreturn earlier in this same epilogue leaves the marker live with
+     * ELR_EL1 on the instruction the guest is about to resume, which can be an
+     * SVC it has not executed yet. Snapshotting the register there would record
+     * the marker as the guest's X8, and the handler's own rt_sigreturn would
+     * hand it back, so that SVC would run as syscall 2. Take what rt_sigreturn
+     * parked instead. A record exists only inside the epilogue that wrote it,
+     * so there is no later delivery for this to reach; see sigreturn_x8.
+     */
+    if (!el0_preempt && sigreturn_x8.valid && sigreturn_x8.elr == saved_pc) {
```
jserv, line 2423: The park closes the `rt_sigreturn` route into the drop tail, but the ordinary syscall-return route into it is left open and reaches EL0 the same way. `deliver_signal_locked` snapshots the live GPRs, and on a page-table syscall `tlbi_request_emit_to_vcpu` has already put the wire values in `X8` through `X11`, so the frame records `X8` as 0/1/3/4 and the handler's own `rt_sigreturn` hands that back; the note on `exec_drop_frame` measures all 707 deliveries running the following `svc` as the wrong syscall. The shim frame is live at `SP_EL1` for that delivery too, so the slot this change reads could supply `X8` through `X11` there as well, with the park winning when it is set. Worth its own issue rather than living only in a source comment, since #379 closes here and the same corruption survives.

docs/internals.md
```diff
@@ -317,25 +319,109 @@ separate broadcast after the split lands.
...
+`src/core/shim.S`), because the host answers a flip request with a
+`SIGSEGV` delivery when the region never had the permission asked for.
+Nothing publishes an `X8` on that route, so the reload hands EL0 the
+value the exception was taken with, which is the guest's own. Measured
+with a guest holding `0xa5` in `X8` across a branch to a page with no
+`PROT_EXEC` and across a store to a page with no `PROT_WRITE`: the
+handler enters with `X8 = 0x2` on the tree before the reload and with
+`0xa5` after it, on both tails. `tests/test-shim-sigreturn-x8` covers
+the pair. The same reload is what a `BRK`-delivered `SIGTRAP` returns
+through. Its handler still enters with the marker, since the frame
```
jserv, line 347: This one is cheap to close alongside the others rather than leave standing. `handle_brk` is the only marker consumer that never looks at it: `RESTORE_GPRS` pops the frame before `hvc #10` and the tail is a bare `eret`, so `deliver_signal_locked`'s `X8 = 2` reaches the `SIGTRAP` handler verbatim. Giving it the `handle_el0_fault` shape (`LOAD_GPRS`, then `ldr x8, [sp, #64]` and `add sp, sp, #256` after the HVC) costs the same two instructions this change spends on the other three tails, and JIT translators using `BRK` as a trampoline are exactly the callers that read `X8`.

src/syscall/signal.c
```diff
@@ -2203,6 +2206,110 @@ static void build_sigcontext_reserved(uint8_t *reserved,
...
+ */
+static void shim_publish_frame_x8(hv_vcpu_t vcpu,
+                                  const guest_t *g,
+                                  uint64_t x8_for_el0)
+{
+    uint64_t sp_el1 = vcpu_get_sysreg(vcpu, HV_SYS_REG_SP_EL1);
+    uint64_t lo = g->shim_data_base;
+    uint64_t hi = lo + BLOCK_2MIB;
+
+    if (sp_el1 < lo || sp_el1 > hi - SHIM_FRAME_BYTES) {
```
jserv, line 2299: The accepted window is the whole 2 MiB shim data block, but only the top `MAX_THREADS` 4 KiB slots of it are EL1 stack; everything below is the shim-globals cache (identity slots, urandom ring, attention bitmask). An `SP_EL1` that is wrong but still inside the block passes this check, and the `memcpy` then rewrites eight bytes of that cache instead of taking the `log_error` path. Bound it at the slot region `thread_alloc_sp_el1` carves from the top of the block so a bad `SP_EL1` reports rather than corrupts.

tests/test-matrix.sh
```diff
@@ -728,6 +728,8 @@ run_unit_tests()
         "$bindir/test-wait-process-signal"
     test_check "$runner" "test-wait-sigmask-signal" " - PASS" \
         "$bindir/test-wait-sigmask-signal"
+    test_check "$runner" "test-shim-sigreturn-x8" "0 failed" \
```
jserv, line 731: The block above `run_unit_tests` still reads that `test-shim-futex-fast` is the exception to `test-shim-*` staying out of the qemu lane; this line makes a second one. Record it there with its reason (every assertion is plain Linux signal ABI), and note that the lane has not actually been run against it yet, so the next reader does not take the enumeration as complete or the registration as verified.

src/syscall/signal.c
jotpalch: Confirmed and fixed. You are right that this is the reverse of what the check is
for, and it is the one finding in this round that was a memory-safety hole rather
than a gap in coverage. The branch introduced it.

The bound was the whole 2 MiB block, `[shim_data_base, shim_data_base +
BLOCK_2MIB - 256)`. Only the top `MAX_THREADS` slots are EL1 stack, so any
`SP_EL1` landing in the shim-globals cache below them passed and the `memcpy`
rewrote eight bytes of live shim state.

It is now the slot region, derived rather than written out:

```c
void thread_sp_el1_region(const guest_t *g, uint64_t *lo, uint64_t *hi)
{
    uint64_t top = sp_el1_top(g);

    *hi = top;
    *lo = top - (uint64_t) MAX_THREADS * SP_EL1_SLOT_BYTES;
}
```

`sp_el1_top`, `MAX_THREADS` and `SP_EL1_SLOT_BYTES` are the three constants
`thread_alloc_sp_el1` carves the slots with, so the check and the allocator
cannot drift apart. `shim_publish_frame_x8` calls it and keeps the same
`log_error` and return.

**Validation.** Temporary instrumentation forced `SP_EL1` to `block + 0x100`,
which is inside the urandom ring, and was reverted afterwards.

```
with the block bound:  ring word 0xbd0f39f71fe88c59 -> 0xac, no log line
with the slot bound:   rt_sigreturn: SP_EL1 0xeffe00100 is outside the EL1
                       stack region [0xefffc0000, 0xf00000000)
                       ring word unchanged
```

Boundary probed from both sides: `block + 0x1bff00` is rejected and
`block + 0x1c0000` is the first address accepted, which is the region floor
exactly.

**What the new bound rejects that the old one accepted: the low 1835008 bytes of
the block** (2097152 minus 64 slots of 4096). Every byte of that window is
shim-globals cache, so there was no address in it the old check should have been
letting through.

docs/internals.md
jotpalch: Confirmed and fixed. The branch had documented this tail as a known gap instead
of closing it, which was the wrong call for two instructions, and your point
about JIT translators is the one that settles it: a `BRK` trampoline is
precisely the caller that reads `X8` in the handler.

`handle_brk` now has the `handle_el0_fault` shape:

```asm
handle_brk:
    LOAD_GPRS          /* was RESTORE_GPRS, which popped the frame here */

    hvc #10

    ldr x8, [sp, #64]
    add sp, sp, #256
    eret
```

**Validation**, guest holding `0xc3` in `X8` across a `BRK #0` with a `SIGTRAP`
handler installed, three runs each. This machine is shared, so both halves were
measured on the same head back to back.

```
                        handler entry X8    X8 after the handler's rt_sigreturn
without fix (old tail)        0x2                        0xc3
with fix (new tail)           0xc3                       0xc3
```

The right-hand column was already correct on both, since that half returns
through `HVC #5` like any other syscall. The left-hand column is what changes.

`tests/test-shim-sigreturn-x8.c` gains a fifth phase asserting both halves; it
fails against the old tail with `entered with X8 = 0x2, want 0xc3` and passes
now. The gate follows: `check_drop_tail` is generalized to `check_reload_tail`
and applied to `exec_drop_frame` and `handle_brk` alike, so deleting the `ldr`
from either is rc=1. Self-test is 28 cases, all passing.

tests/test-matrix.sh
jotpalch: Confirmed and fixed. The comment did read as though `test-shim-futex-fast` were
the only one, and a single recorded exception is exactly what gets the next
registration held out by mistake. Both are named now, each with its reason, and
the text says why they are named together.

On the second half of your request I did the opposite of what you asked, and it
is worth being precise about what that means. Rather than record that the lane
had not been run, **I ran it**: fixtures fetched, Alpine `6.18.52-0-virt` booted,
`test-shim-sigreturn-x8` reports `0 failed` there. So the comment records the run
instead of the caveat.

What that does not cover, and what the comment also says: this was one binary run
by hand, not a full lane. No `EXPECTED_BASELINES` floor moves, because an
unobserved floor would assert a run that did not happen.

src/syscall/signal.c
jotpalch: Confirmed, and I agree it belongs in its own issue rather than in a source
comment. Filed as #384. The comment here and the passage in `docs/internals.md`
now say it is tracked separately, with the number going in once the issue exists.

Nothing on this branch changes for it, as you asked.

Re-measured on the current head for the issue, and the fresh numbers sharpen your
point rather than soften it. Three interleaved rounds per tree, 60000 `mprotect`
rounds each under a 200 us itimer, two adjacent `svc`s with the first the
`mprotect`. Shared machine, 1-minute load average 3.24 to 3.73 throughout,
so the delivery counts are scheduling and only the proportions inside a run
mean anything. In round 1 with the fix the last delivery landed between the two
counters being printed, which is the 760 against 759 there.

```
                    deliveries   X8 the second svc ran as   ucontext X8 wrong
without fix         710/705/715  2 x710 / 2 x705 / 2 x715   100% (all rounds)
(afdcfce)
with fix            760/704/715  0 x462,3 x297 / 0 x410,3   100% (all rounds)
(this branch)                    x294 / 0 x440,3 x275
```

The guest held 226 at every `svc`. Two things follow that I had not stated
plainly enough before:

- The corruption rate is 100 percent of deliveries on **both** trees. This branch
  changes which wrong number arrives, from the marker to the TLBI wire value, and
  nothing else. `afdcfce` is not better here and this branch is not worse.
- `X9` through `X11` are wrong on exactly the deliveries where `X8` is 3, which
  is `TLBI_RANGE`, the only kind that writes them: `X9 = 0x200000000`,
  `X10 = 0x1`, `X11 = 0x0` against sentinels `0xa9`, `0xaa`, `0xab`. So the same
  issue covers the `X9` through `X11` shadowing this PR declines, since it is one
  epilogue and one snapshot.

**On the direction.** Your observation that the shim frame is live at `SP_EL1`
for this delivery too is carried into the issue as the suggested starting point
and credited to you, deliberately without being adopted: the issue records that
the same slot could supply `X8` through `X11` with the park winning when it is
set, and records equally that it has not been implemented or measured. The one
caveat I attached is that `shim_publish_frame_x8` is written to be callable only
where the frame is provably live, so extending the reader to a second route means
extending that argument rather than just the call site.

jserv (changes requested):
Shorten changes against docs/internals.md and response to cubic AI reviews.

PR #386, opened 2026-09-16
Wake select and epoll_wait on a queued signal

xalestar:
Part of #378; futex is left for after #383.

A thread parked in a host wait learns of a guest signal only through the wakeup pipe. pselect6 joined the pipe for an indefinite wait only, and epoll_pwait never did, so the handler ran only once the current 200 ms slice ended.

pselect6 now joins the pipe for every wait except a zero timeout, which does not park and would only drain a wake meant for a sibling. epoll_pwait parks in select() on the kqueue fd plus the pipe, then collects with a zero-timeout kevent(), so the pipe is never registered on the guest's epoll instance. select() rather than poll(), because poll() answers POLLNVAL for a kqueue fd and would never park. A guest with a full fd table puts the kqueue past FD_SETSIZE, so `poll.c` builds with `_DARWIN_UNLIMITED_SELECT` and that case parks on a heap bitmap.

Reproduction: `tests/test-wait-signal-latency.c` parks a sibling thread in each wait on an empty pipe, sends SIGUSR1 with pthread_kill, and takes the median signal-to-handler gap over 8 rounds against a 20 ms bound. It also reads the waiter's `CLOCK_THREAD_CPUTIME_ID` share over the wait against a 5% bound, since a spinning wait answers the signal at once too. A fourth case fills the fd table before epoll_wait. On `0a63368` every case fails at about 42 ms. With this change all four pass under 1% CPU. A build with the select() park forced to a zero timeout reads 10% and fails; one that skips the park past FD_SETSIZE fails the full-table case at 41 ms.

Validation, rebased on `b05f639`:
- `make check`: exit 0. One earlier run failed `test-shim-futex-toctou` with -ENOSYS in its spin phase, the #379 symptom that #383 fixes; it passed 5/5 on rerun.
- `test-wait-signal-latency`, `test-wait-process-signal`, `test-wait-sigmask-signal`, `test-poll`, and all nine `test-epoll*`: pass
- `make test-matrix-elfuse-aarch64` (on the first revision): 291 passed, 1 failed, 10 skipped. The failure is `test-sigio` (no SIGURG on a TCP OOB byte). It also fails 3/3 on `0a63368` and does not touch select or epoll.
- The qemu lane was not run.


tests/test-wait-signal-latency.c
```diff
@@ -0,0 +1,242 @@
...
+             */
+            pthread_join(th, NULL);
+            errno = rc;
+            return GAP_NO_THREAD;
+        }
+
+        /* Bounded, so a regression that strands the waiter fails the test
+         * rather than hanging the lane it runs in.
+         */
+        long long give_up_ns = sent_ns + (long long) SLEEP_SEC * 2000000000LL;
```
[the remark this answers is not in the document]
xalestar: Every failure path in median_gap now detaches the waiter instead of joining it, so a waiter stuck in epoll_wait(-1) is torn down by the process exit after SUMMARY. Checked with a variant whose indefinite waiter blocks SIGUSR1: it reports "the handler never ran" and exits 1 instead of hanging.

src/syscall/poll.c
```diff
@@ -2612,6 +2616,31 @@ static int epoll_mute_usb_wakes(int kq,
...
+                            int cap,
+                            int slice_ms)
+{
+    int wake_fd = wakeup_pipe_read_fd();
+    if (slice_ms > 0 && wake_fd >= 0) {
+        struct pollfd pfds[2] = {
+            {.fd = kq, .events = POLLIN},
+            {.fd = wake_fd, .events = POLLIN},
+        };
+        if (poll(pfds, 2, slice_ms) > 0 && (pfds[1].revents & POLLIN))
```
jserv: macOS `poll()` answers `POLLNVAL` for a kqueue descriptor (the comment above `poll_eval_unpollable` records this). This `poll()` therefore returns at once on every slice, leaves `pfds[1]` unset, and falls through to a zero-timeout `kevent()`. The loop in `sys_epoll_pwait` never parks: a finite `epoll_wait` spins until its deadline, and an idle `epoll_wait(-1)` holds a core at 100% forever. The latency test still passes because a spinning loop also rechecks signals immediately. `select()` accepts kqueue descriptors, so park there instead, and keep the old plain `kevent()` slice when either descriptor is at or above `FD_SETSIZE`:
```c
if (slice_ms > 0 && RANGE_CHECK(wake_fd, 0, FD_SETSIZE) &&
    RANGE_CHECK(kq, 0, FD_SETSIZE)) {
    fd_set rset;
    FD_ZERO(&rset);
    FD_SET(kq, &rset);
    FD_SET(wake_fd, &rset);
    struct timeval tv = {slice_ms / 1000, (slice_ms % 1000) * 1000};
    if (select((kq > wake_fd ? kq : wake_fd) + 1, &rset, NULL, NULL, &tv) > 0 &&
        FD_ISSET(wake_fd, &rset))
        wakeup_pipe_drain();
    slice_ms = 0;
}
```

tests/test-wait-signal-latency.c
```diff
@@ -0,0 +1,243 @@
...
+                      sent_ns) /
+            1e6;
+        pthread_join(th, NULL);
+    }
+
+    sort_gaps(gaps, ITERS);
+    return (gaps[ITERS / 2 - 1] + gaps[ITERS / 2]) / 2.0;
+}
+
+static void check(const char *name, void *(*body)(void *) )
```
jserv, line 237: The test only checks how fast the handler runs, and a busy loop wins that check. A wait that never parks passes it. Reading the waiter's CPU time around the wait (`clock_gettime(CLOCK_THREAD_CPUTIME_ID)` inside the waiter body) and failing when it approaches the wall-clock settle time would catch the spin above. It would also catch the same regression if it comes back later.

jserv:
`epoll_wait` no longer sleeps: macOS `poll()` answers `POLLNVAL` for a kqueue descriptor, so the park added in `epoll_wait_slice` returns at once and an idle `epoll_wait(-1)` holds a full core. The latency test cannot see this, because a spinning loop also rechecks signals immediately. Details and a `select()`-based fix are on `epoll_wait_slice`.

src/syscall/poll.c
xalestar: The park now uses select() on the kqueue fd and the wakeup pipe, and falls back to the plain kevent() slice when either fd is at or above FD_SETSIZE. The poll() version parked on my local host, which is why the spin did not show there. On the same host, forcing the select() park to a zero timeout reproduces it: an idle epoll_wait spends 10% of the wait on CPU. The parked version stays under 1%.

tests/test-wait-signal-latency.c
xalestar: The waiter now reads CLOCK_THREAD_CPUTIME_ID around the wait, and the test fails above a 5% CPU share. Through that clock a parked waiter reads under 1% and a spinning one 10 to 13%, not 100%, so the bound sits well below the spin. The zero-timeout select() build fails both epoll_wait cases on it.

src/syscall/poll.c
```diff
@@ -2612,6 +2616,39 @@ static int epoll_mute_usb_wakes(int kq,
...
+ * slice.
+ */
+static int epoll_wait_slice(int kq,
+                            struct kevent *kevents,
+                            int cap,
+                            int slice_ms)
+{
+    int wake_fd = wakeup_pipe_read_fd();
+    if (slice_ms > 0 && RANGE_CHECK(wake_fd, 0, FD_SETSIZE) &&
+        RANGE_CHECK(kq, 0, FD_SETSIZE)) {
```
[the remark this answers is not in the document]
xalestar: The park no longer skips high descriptors. poll.c builds with _DARWIN_UNLIMITED_SELECT, and when the kqueue or pipe fd is past FD_SETSIZE the select() runs on a heap bitmap sized to it. Only a failed allocation still takes the plain kevent() slice. A new test case fills the guest fd table first, which put the host kqueue at fd 1026. It took 41 ms before this change and now passes under the 20 ms and 5% CPU bounds.

tests/test-wait-signal-latency.c
```diff
@@ -0,0 +1,307 @@
...
+    check("select with a timeout wakes on a queued signal", select_finite);
+    check("epoll_wait with a timeout wakes on a queued signal", epoll_finite);
+    check("epoll_wait without a timeout wakes on a queued signal",
+          epoll_indefinite);
+
+    int full[2];
+    while (pipe(full) == 0)
+        ;
+    close(full[0]);
+    close(full[1]);
```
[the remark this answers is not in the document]
xalestar: The case now saves errno right after the pipe() loop, before the two close() calls can overwrite it, and exits with "fd table not filled" unless it is EMFILE.

src/syscall/poll.c
```diff
@@ -2612,6 +2622,46 @@ static int epoll_mute_usb_wakes(int kq,
...
+        fd_set *rset = &stack_set;
+        if (nfds > FD_SETSIZE)
+            rset = calloc(howmany(nfds, NFDBITS), sizeof(int32_t));
+        else
+            FD_ZERO(rset);
+        if (rset) {
+            FD_SET(kq, rset);
+            FD_SET(wake_fd, rset);
+            struct timeval tv = {.tv_sec = slice_ms / 1000,
+                                 .tv_usec = (slice_ms % 1000) * 1000};
```
jserv: Append a comma (`,`) at the end of `* 1000`, so that clang-format can indent better.

src/syscall/poll.c
```diff
@@ -2612,6 +2622,46 @@ static int epoll_mute_usb_wakes(int kq,
...
+            if (select(nfds, rset, NULL, NULL, &tv) > 0 &&
+                FD_ISSET(wake_fd, rset))
+                wakeup_pipe_drain();
+            slice_ms = 0;
+            if (rset != &stack_set)
+                free(rset);
+        }
+    }
+    struct timespec slice_ts = {.tv_sec = slice_ms / 1000,
+                                .tv_nsec = (slice_ms % 1000) * 1000000L};
```
jserv: Ditto. Append a comma.

src/syscall/poll.c
```diff
@@ -9,6 +9,12 @@
  * called from syscall_dispatch() in syscall/syscall.c.
  */
 
+/*
+ * Lets select() take descriptors at or above FD_SETSIZE in a caller-sized
+ * bitmap (select(2)); epoll_wait_slice parks on a kqueue fd that can be one.
+ */
+#define _DARWIN_UNLIMITED_SELECT
```
jserv, line 17: Avoid redefining this macro if any.

src/syscall/poll.c
xalestar: Added.

src/syscall/poll.c
xalestar: Added.

src/syscall/poll.c
xalestar: Guarded with ifndef.

jserv:

PR #388, opened 2026-09-18
Let kill reach a relative that is not a child

xalestar:
A forked guest process gets ESRCH from `kill(getppid(), sig)`, where Linux delivers the signal. `sc_kill` resolves a single pid only through the child table, which holds the caller's descendants, so the parent is never found. The group and broadcast forms of `kill` already read the fork-family registry through `proc_get_namespace_targets`; the first commit falls back to the same registry when the child table has no answer.

Reading the registry from a pid-directed `kill` exposed three weaknesses that the group forms already had, and the rest of the series closes them.

A record outlives the member that wrote it until the next publish compacts the file, and a reader kept any record whose host pid was alive. Once macOS reused that pid for another elfuse process, `kill` on the exited guest pid returned 0 instead of ESRCH. Each record now carries the host process start time from `proc_pidinfo`, and a record survives only while the process at that pid has the same start time. The reverse lookup gets the same check, so `F_GETLK` stops naming an exited member as the holder of a conflicting lock.

The family id was the root's host pid, which macOS recycles too. Once a family outlived its root, a later elfuse root landing on that pid passed the owner test, deleted the orphan family's registry and published into the same path. The orphans then resolved guest pid 1 to that unrelated root, and `proc_send_guest_signal` let the signal through, because the tag the receiver checks was the same colliding value, so an orphan's `kill(1, SIGKILL)` reached a stranger. The root now mints a random id, children receive it over the fork IPC field that already carries it, and the owner tests read a flag set by whoever minted.

Reproduction: `tests/test-kill-parent.c` has the child call `kill(getppid(), 0)` and then `kill(getppid(), SIGUSR1)`; on `main` both return ESRCH (`FAIL: 2 failed`), with this series the parent's handler runs (`PASS: 0 failed`). `make test-registry-stale-pid` plants a record carrying a live elfuse host pid and a start time that process does not have, and expects ESRCH; it also pins that the registry file is not named after the root's pid. Both assertions fail without the commits they cover.

Rebased on `main` at c0458b4. `make check` passes. `make test-matrix` reports two failures, both of which reproduce on `main` with the same binary: the Rosetta `audit-known-limitations` case (a Rosetta assertion in `rt_sigreturn`, rc=133) and `test-sigio`, which is flaky at 7 failures in 10 runs on `main` against 3 in 10 here.


src/syscall/proc.c
```diff
@@ -1760,6 +1767,21 @@ int proc_get_namespace_targets(proc_signal_target_t *out,
...
 }
 
+int proc_get_namespace_targets(proc_signal_target_t *out,
+                               int max,
+                               int64_t pgid_filter)
+{
+    return registry_collect(out, max, pgid_filter, 0);
+}
+
+pid_t proc_namespace_host_pid(int64_t guest_pid)
```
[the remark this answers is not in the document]
xalestar: Valid, but it predates this PR. `kill(-1)`, `kill(0)`, and `kill(-pgid)` already read the same registry records through `registry_collect`, with the same `kill(pid, 0)` and `proc_pidpath` checks; this PR added single-pid `kill` as one more reader. The family tag in `proc_send_guest_signal` already stops delivery to the wrong process, so the visible bug was `kill` returning 0 instead of ESRCH.

Fixed in e0012bb: each record now stores the host process start time (`proc_pidinfo`), and `registry_parse_cb` keeps a record only while the process at that pid has the same start time. This covers every reader, including members that died without cleaning up. `test-registry-stale-pid` plants a stale record and expects ESRCH.

mk/tests.mk
```diff
@@ -726,6 +727,39 @@ test-absock-cleanup: $(ELFUSE_BIN) $(BUILD_DIR)/test-absock-cleanup
...
+		grep -q READY "$$tmp/out" && break; \
+		sleep 0.1; \
+	done; \
+	grep -q READY "$$tmp/out" || fail "family never reported READY"; \
+	reg="$$(getconf DARWIN_USER_TEMP_DIR)elfuse-procs-$$fpid"; \
+	[ -f "$$reg" ] || fail "no registry at $$reg"; \
+	printf '%s 99 1 1\n' "$$xpid" >> "$$reg" || fail "cannot append to $$reg"; \
+	echo go >&4; \
+	exec 4>&-; \
+	wait $$fpid; \
```
jserv, line 757: `wait $$fpid` has no bound, so a regression that leaves the guest parked (the `read(0, ...)` on the fifo, or the forked child's `read(release[0])`) hangs this recipe instead of failing it. The `EXIT` trap never runs, so `make check` stops here with both elfuse roots still alive and the scratch directory left behind. Every other lane in this file reaches the guest through a runner that wraps the invocation in `timeout`; this recipe launches `$(ELFUSE_BIN)` directly. Wrapping both launches in `timeout $(TEST_TIMEOUT)` and failing when `fpid` outlives it keeps a hang reportable.

src/syscall/proc.c
```diff
@@ -1217,19 +1220,37 @@ typedef struct {
     bool truncated;
 } registry_parse_ctx_t;
 
-/* Upsert one "hostpid guestpid pgid" record, keeping the latest guest_pid/pgid
- * per LIVE host pid. Dead, malformed, and out-of-range records are dropped.
+/* Start time of host process @pid in microseconds.
+ *
+ * Returns false once @pid has exited.
+ */
+static bool host_start_us(pid_t pid, uint64_t *out)
```
jserv, line 1227: Nothing on the reverse lookup uses this. `registry_find_by_host_cb` still parses three fields, and `proc_host_to_guest_pid` still accepts a hit on host pid plus a matching `proc_pidpath`, which is the exact identity test this commit demonstrates is not sufficient. Once macOS hands an exited member's host pid to another elfuse process, `F_GETLK` reports the exited member's guest pid as the conflicting lock holder, the same stale-record class the second commit closes for `kill`. Parsing the fourth field in `registry_find_by_host_cb` and comparing it against `host_start_us` closes it on that path too.

src/syscall/proc.c
```diff
@@ -1760,6 +1792,21 @@ int proc_get_namespace_targets(proc_signal_target_t *out,
...
+int proc_get_namespace_targets(proc_signal_target_t *out,
+                               int max,
+                               int64_t pgid_filter)
+{
+    return registry_collect(out, max, pgid_filter, 0);
+}
+
+pid_t proc_namespace_host_pid(int64_t guest_pid)
+{
+    proc_signal_target_t target;
```
jserv, line 1804: Two inputs break the contract `proc.h` states for this function. A `guest_pid` of 0 or negative leaves `guest_filter > 0` false inside `registry_collect`, which disables the filter entirely and returns the first non-self member, so the caller gets an arbitrary host pid where the header promises -1. The caller's own guest pid returns -1 even though it is a live member, because `registry_collect` skips `host_pid == self`. `sc_kill` screens both cases before it calls, so nothing is wrong today, but this is exported as a general resolver and the next caller has no reason to know it must screen.

```suggestion
pid_t proc_namespace_host_pid(int64_t guest_pid)
{
    if (guest_pid <= 0)
        return -1;
    proc_signal_target_t target;
```

src/syscall/proc.h
```diff
@@ -495,6 +495,16 @@ int proc_get_namespace_targets(proc_signal_target_t *out,
...
 
+/* Resolve one guest pid to its host pid through the same fork-family registry
+ * proc_get_namespace_targets reads. The child table only holds descendants, so
+ * this is what lets a process signal a relative that is not its own child (its
+ * parent, most commonly).
+ *
+ * Returns the host pid, or -1 when the registry holds no live member with that
+ * guest pid.
+ */
+pid_t proc_namespace_host_pid(int64_t guest_pid);
```
jserv, line 506: `sys_pidfd_open` and `sys_pidfd_send_signal` still resolve through `proc_guest_to_host_pid` alone, so after this change `kill(getppid(), 0)` succeeds while `pidfd_open(getppid())` still returns ESRCH, and a pidfd on any relative that is not a descendant stays unreachable. Linux draws no such line. Worth a follow-up routing the three `proc_guest_to_host_pid` call sites in `src/syscall/proc-pidfd.c` through the same resolver.

src/syscall/syscall.c
```diff
@@ -1034,6 +1034,17 @@ static int kill_deliver_targets(const proc_signal_target_t *targets,
     return delivered;
 }
 
+/* Resolve a guest pid for kill(2). The child table answers for descendants.
+ * Every other member of the fork family, the caller's own parent above all,
+ * exists only in the namespace registry, the same source the group and
+ * broadcast forms already read.
+ */
+static pid_t kill_resolve_host_pid(int64_t gpid)
```
jserv, line 1042: This puts pid-directed `kill` on the registry, where before only the group and broadcast forms read it, and the registry's file name is keyed on `absock_get_namespace_id()`, which falls back to the root's host pid. macOS recycles that pid like any other: once a family's root exits with a child still running, a new elfuse root that lands on the reused pid passes the owner test in `proc_registry_reset_if_owner`, unlinks the orphan family's registry, and then publishes into the same path. The start time added by the second commit proves the target is live and `proc_pidpath` proves it runs the same binary, but neither proves it is in the same family, so the orphan's `kill(1, SIGKILL)` reaches the unrelated root, and `proc_send_guest_signal` passes it through because the namespace tag it checks is that same colliding value. Keying the registry on something that is not a recycled pid, a nonce carried through the existing fork IPC state for instance, would close it for the group forms as well.

mk/tests.mk
xalestar: Fixed in 6ff3f96. Both launches go through timeout, and the lane fails on a nonzero exit from the family, so a parked guest reports rc=124 instead of hanging. Wrapping the launches moves `$!` to timeout, so the recipe no longer derives anything from it: it finds the registry by diffing the temp dir and takes the holder's pid from `pgrep`. Checked by parking the family in `pause()`: the lane fails in 3s with `TEST_TIMEOUT=3`.

src/syscall/proc.c
xalestar: Fixed in 6e21ed8. `registry_find_by_host_cb` now parses the fourth field and compares it against `host_start_us`, so `proc_host_to_guest_pid` stops trusting pid plus `proc_pidpath` alone and `F_GETLK` no longer names an exited member.

No test. Reaching that path needs a stale record whose host pid belongs to a live process holding a conflicting lock, and I could not stage that from a guest test without a second elfuse family taking the lock. Open to a suggestion on where that lane would fit.

src/syscall/proc.c
xalestar: Applied in 594f4b9.

src/syscall/proc.h
xalestar: Agreed, and out of scope here. The three call sites in `proc-pidfd.c` follow in a separate PR once this one lands.

src/syscall/syscall.c
xalestar: Confirmed, and it reaches further than the resolver: with the id colliding, `proc_send_guest_signal` delivers, since the tag the receiver checks is that same value. An orphan's `kill(1, SIGKILL)` lands on the unrelated root.

Fixed in a323958. The root mints a random id instead of taking `getpid()`, children receive it through the `absock_namespace_id` field fork IPC already carries, and the four owner tests read a flag set by whoever minted rather than comparing against `getpid()`. The stale-registry lane pins that the registry file is not named after the root's pid; that assertion fails on the old id.

Not covered by a test: the collision itself needs a recycled pid, which the minted id now makes unreachable by construction.

src/syscall/proc.c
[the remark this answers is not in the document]
xalestar: Correction to the reply above: the family tag does not stop delivery in every case. The tag is the family id, and that id was the root's host pid, so a later root landing on a recycled pid carried the same value and accepted an orphan family's signals. a323958 makes the root mint a random id instead, which is what the tag now relies on.

jserv:

PR #392, opened 2026-09-21
Let a pidfd name any relative kill can name

xalestar:
## Summary

Follow-up to #388, which routed pid-directed `kill` through the fork-family registry so a guest can signal its parent. The three pidfd call sites still resolved through `proc_guest_to_host_pid` alone, so after #388 `kill(getppid(), 0)` succeeded while `pidfd_open(getppid())` returned ESRCH, and a pidfd on any relative that is not a descendant stayed unreachable. Linux draws no such line. This routes all three through the same resolver.

## What

`proc_resolve_guest_pid()` (src/syscall/proc.c) holds the two-step lookup that was previously private to `sc_kill`: the child table answers for descendants, the fork-family registry for every other relative. `sc_kill` and the three sites in `src/syscall/proc-pidfd.c` — `pidfd_create`, `sys_pidfd_open`, `sys_pidfd_send_signal` — now share it, so every caller that turns a guest-named pid into a host pid reaches the same set of processes.

Routing `pidfd_create` matters beyond the open. That resolve is what arms the exit monitor, so with only the open fixed, a pidfd on a sibling would open and then never become readable — the process could exit and `poll()` would wait forever. The test below covers that separately.

## Why

Reported by @jserv on #388 (src/syscall/proc.h:506).

## Test plan

`tests/test-pidfd-parent.c` covers both halves, registered in `tests/test-matrix.sh` next to `test-kill-parent`:

| Scenario | What it proves |
|---|---|
| A child opens a pidfd on its parent and sends sig 0 and SIGUSR1 through it | `sys_pidfd_open` and `sys_pidfd_send_signal` resolve an ancestor |
| A second child opens a pidfd on its sibling and polls it for the sibling's exit | `pidfd_create` arms the exit monitor for a process that is neither ancestor nor descendant |

The sibling half is ordered by two pipe handshakes rather than by sleeps: the sibling blocks on a pipe and exits only after the watcher reports its pidfd open, so a slow fork cannot turn the test red. An earlier sleep-ordered draft was observably flaky; this one passed 10/10 consecutive runs under elfuse.

Mutation check — reverting any one of the three call sites on its own turns the test red, each with a distinct message:

| Reverted site | Failure |
|---|---|
| `pidfd_create` | sibling pidfd never reported the exit |
| `sys_pidfd_open` | `pidfd_open` fails for both the parent and the sibling |
| `sys_pidfd_send_signal` | signal through the parent pidfd fails |

Commands run, all clean:

```
make check                       # exit 0
make check-format
make lint                        # no new warnings on the touched lines
.ci/check-newline.sh .ci/check-security.sh .ci/check-matrix-lists.sh .ci/check-cppcheck.sh
```

The new test fails on the merge commit of #388 and passes here. Cross-checked against a real kernel with `tests/qemu-runner.sh` (qemu-aarch64), where it passes 3/3.

## Backward compatibility

No ABI or flag changes. `proc_guest_to_host_pid` keeps its meaning and its other callers; the registry fallback only adds pids that previously resolved to ESRCH, so nothing that worked before resolves differently now.

## Related issue

No separate issue — follow-up to review feedback on #388. Rebased on main at 669ea25.

---
## Summary by cubic
Lets a pidfd name any relative `kill` can name: the three pidfd call sites now resolve guest pids through the same fork-family registry as `kill`, so pidfds on a parent or sibling work as on Linux. A pidfd whose target exits during setup now reads as exited instead of polling forever, and a pidfd on the caller itself stays unreadable.

**Details**
- Routes `pidfd_create`, `sys_pidfd_open`, and `sys_pidfd_send_signal` through `proc_resolve_guest_pid`, the two-step lookup `sc_kill` already used.
- `pidfd_create` marks its fd exited whenever no exit monitor is armed, covering a target that vanished between the caller's resolve and the open's re-resolve, plus a monitor thread that failed to start.
- A self-pidfd returns before the resolve, since a guest cannot observe its own exit.
- Adds `tests/test-pidfd-parent.c` covering pidfds on a parent, a sibling, and the caller itself, with pipe handshakes that order the open before the exit so the test is not flaky.

<sup>Written for commit aff1cc533b9443f71e140b7a72457280ece3ffa1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/elfuse/pull/392?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src/syscall/proc-pidfd.c
```diff
@@ -141,7 +141,7 @@ int pidfd_create(guest_t *g, int64_t target_pid)
     entry->write_end = pfd[1];
     pthread_mutex_unlock(&pidfd_lock);
 
-    pid_t host_pid = proc_guest_to_host_pid(target_pid);
+    pid_t host_pid = proc_resolve_guest_pid(target_pid);
```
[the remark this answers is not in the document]
xalestar: Real, and fixed in `aff1cc5`. `pidfd_create` now arms the monitor or marks the fd exited, never neither: `monitor_ok` moved out of the `host_pid > 0` branch, so a target lost between the two lookups completes the fd instead of leaving it unreadable forever.

Carrying the resolved pid down from `sys_pidfd_open` would have narrowed the window without closing it, since `pidfd_create` has two other callers (the CLONE_PIDFD path in forkipc, and the self case), so completing on a lost target covers all three.

One wrinkle worth flagging: completing unconditionally would have been wrong. `pidfd_open(self)` never resolves, because `registry_collect` skips the caller's own record, so a blanket "unresolved means exited" made a self pidfd poll readable while the process was still running. That case now returns before the resolve, and test-pidfd-parent pins it — dropping the early return turns the test red.

src/syscall/proc.c
```diff
@@ -1814,6 +1814,12 @@ pid_t proc_namespace_host_pid(int64_t guest_pid)
                : -1;
 }
 
+pid_t proc_resolve_guest_pid(int64_t guest_pid)
```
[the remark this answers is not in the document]
xalestar: This validation is already in place, so there is nothing to change here.

`registry_collect` is not the only filter. It reads through `registry_read_locked` → `registry_parse_cb`, which parses each record as `hostpid guestpid pgid startus` and drops it unless the live process's start time matches the recorded one:

```c
uint64_t live_us;
if (!host_start_us((pid_t) hp, &live_us) || live_us != (uint64_t) st)
    return;
```

A record naming a reused host pid therefore never reaches the `proc_pidpath` comparison; that check is a second filter layered on top, not the only one. This landed in #388 as 6ff3f96 ("Tie a registry record to its process start time") and 6e21ed8 ("Check the start time on the reverse lookup too"). `proc_resolve_guest_pid` inherits it unchanged, since the registry half is `proc_namespace_host_pid` and the child-table half holds live descendants.

src/syscall/proc-pidfd.c
```diff
@@ -164,10 +170,16 @@ int pidfd_create(guest_t *g, int64_t target_pid)
...
-        if (!monitor_ok)
-            proc_pidfd_notify_exit(target_pid);
     }
 
+    /* No monitor means no one will ever mark this fd readable. A target that no
+     * longer resolves exited between the caller's lookup and this one, and
+     * Linux hands out a pidfd that reads as exited rather than one that waits
+     * forever, so complete it here.
+     */
+    if (!monitor_ok)
```
jserv, line 181: This completion now runs for a target that never resolved, and `clone3(CLONE_PIDFD | CLONE_VM)` reaches `pidfd_create` with exactly that: `sys_clone_vm` allocates the child tid in the thread table and never calls `proc_register_child` or publishes to the registry, so `proc_resolve_guest_pid` returns -1 for a child that is running. The guest then gets a pidfd that reads as exited from the moment `clone3` returns. Gate the completion on the target having resolved, and handle the unresolvable non-self case separately.

src/syscall/proc-pidfd.c
```diff
@@ -164,10 +170,16 @@ int pidfd_create(guest_t *g, int64_t target_pid)
...
+        proc_pidfd_notify_exit(target_pid);
```
jserv, line 182: `proc_pidfd_notify_exit` matches by guest pid rather than by fd, so a `malloc` or `pthread_create` failure here also completes every pidfd already watching that same live target, not just the one being created. The premise in the comment above is only half true today: `pidfd_monitor_thread` returns without notifying when `kqueue()` fails, which leaves exactly the forever-unreadable fd this block exists to prevent. Worth a follow-up that completes only this entry and closes the monitor-side gap.

src/syscall/proc-pidfd.c
```diff
@@ -207,7 +219,7 @@ int64_t sys_pidfd_open(guest_t *g, int64_t pid, unsigned int flags)
     if (pid == proc_get_pid())
         return pidfd_create(g, pid);
 
-    if (proc_guest_to_host_pid(pid) > 0)
+    if (proc_resolve_guest_pid(pid) > 0)
```
jserv: `pidfd_open` with a non-positive pid lands on `-LINUX_ESRCH` here, but the kernel rejects it earlier: `pidfd_open` returns `EINVAL` for `pid <= 0`. `proc_resolve_guest_pid` already refuses those, so only the errno the guest sees differs.

```suggestion
    if (pid <= 0)
        return -LINUX_EINVAL;

    if (proc_resolve_guest_pid(pid) > 0)
```

src/syscall/proc-pidfd.c
xalestar: Confirmed and fixed in `655d6a0`. Reproduced first: `clone3(CLONE_PIDFD | CLONE_VM)` then `ppoll` on the pidfd while the child slept gave `poll=1 revents=0x1` — the fd read as exited immediately, exactly as described.

Rather than gate the completion, `pidfd_create` no longer resolves at all: the caller passes the host pid to watch, and 0 means the target lives in this host process and has nothing to watch, so the fd is left unreadable as before. `sys_pidfd_open` passes what it already resolved, and the clone3 path passes `proc_resolve_guest_pid(ret)`, which is the child's host pid for a fork child and -1 for a CLONE_VM child.

That also removes the second lookup cubic's earlier finding was about, so the window it described no longer exists. A target that died in the meantime is handled where it belongs: the monitor thread's `kill(hpid, 0)` ESRCH check and the `kevent` failure path both complete the fd.

Covered by a new case in test-pidfd, which asserts the CLONE_VM child's pidfd is not readable while it runs; restoring the old behavior turns it red. It went to test-pidfd rather than the cross-checked test because a CLONE_VM child resumes on a fresh stack, where the compiler's copy of the clone3 return value is undefined — the `if (ret == 0)` branch is only reliable under elfuse, and against qemu the child took the parent branch and died. The same reason your existing CLONE_VM tests sit in QEMU_SKIP, I think.

Unrelated observation from writing it: a VM-clone child calling `exit` (93) directly takes the whole host process down with it, before stdio is flushed — the parent's output disappears entirely and the process still exits 0. I parked the child in a nanosleep loop instead. Happy to file that separately if it is not already known.

src/syscall/proc-pidfd.c
xalestar: Both halves are right, and a follow-up suits them better than this PR — they are pre-existing, and fixing them properly means changing what `proc_pidfd_notify_exit` keys on, which touches the reap path and the CLONE_PIDFD path as well as this one.

What I will do there: complete a single entry by fd for the creation-failure case, and make `pidfd_monitor_thread` notify when `kqueue()` fails rather than returning silently, so the block's premise stops being half true.

`655d6a0` does narrow it slightly in passing — the completion now runs only for a target that was actually given a host pid to watch, so the self and CLONE_VM cases no longer reach it at all. The over-broad match on a `malloc`/`pthread_create` failure stands as you describe.

I will open it once this merges, unless you would rather have it here.

src/syscall/proc-pidfd.c
xalestar: Applied in `655d6a0`, with the guard first so nothing is looked up before it.

test-pidfd-targets now asserts `pidfd_open(0)` and `pidfd_open(-1)` both return EINVAL; dropping the guard turns it red. Both also pass against qemu-aarch64, so the errno matches what the kernel gives rather than only what I read.

tests/test-pidfd.c
```diff
@@ -148,6 +150,54 @@ int main(void)
...
+                uint64_t ts[2] = {5, 0};
+                raw_syscall2(__NR_nanosleep, (long) ts, 0);
+            }
+        }
+        if (child < 0) {
+            FAIL("clone3 CLONE_VM failed");
+            goto done;
+        }
+
+        struct linux_pollfd pf = {.fd = vm_pidfd, .events = 1, .revents = 0};
```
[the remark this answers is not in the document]
xalestar: Correct, and the test was passing vacuously. Fixed in `dbb1f57`, which asserts `vm_pidfd >= 0` before polling, matching the CLONE_PIDFD case above.

Verified rather than assumed, by forcing `pidfd_create` to fail on exactly the path this PR touches so clone3 writes no pidfd:

- old test: `clone3 CLONE_PIDFD|CLONE_VM pidfd not readable while child runs OK`
- with the assertion: `FAIL: clone3 wrote no pidfd for the CLONE_VM child`

So a regression that stopped creating the pidfd entirely was invisible to the check meant to catch it.

jserv:

